PR #2, opened 2019-03-05
Add ping/pong to the protocol

coderasher:


packages/protocol/src/browser/client.ts
```diff
@@ -22,6 +22,8 @@ export class Client {
 	private readonly sharedProcessActiveEmitter = new Emitter<SharedProcessData>();
 	public readonly onSharedProcessActive = this.sharedProcessActiveEmitter.event;
 
+	private readonly pingTimeout = 10000;
```
coderasher: I'm not sure if 10s is reasonable or not.

kylecarbs: Seems fine to me.

Issue #4, opened 2019-03-05
Fix password feature


icebob:
First of all, it's an awesome project. Thank you! By the way, do you plan to add fix password feature? Maybe it can be set as CLI argument like `./code-server --password 123456`.

Or can I create a PR with this feature?

Keep up the good work!

tvarwig:
I agree

nol166:
I like that Ide @icebob. I will bring it up to the front end peeps.

ddon:
Wow, this code-server is amazing, thank you for such a great new tool! we will start using it in our agency... Password parameter for sure is important, otherwise user needs to login to ssh first, run the code-editor, copy the password, and only then be able to use it in the browser.

kylecarbs:
@icebob feel free to create a PR for it! We'll also put it on the high-priority list :)

icebob:
Great! I will send it when I'm done.

Pitu:
I'm confused since according [to the docs](https://github.com/codercom/code-server/blob/master/doc/self-hosted/index.md) `--password` is a valid flag.

ddon:
$ ./code-server --help
Start your own self-hosted browser-accessible VS Code

USAGE
  $ server [WORKDIR]

ARGUMENTS
  WORKDIR  [default: /home/ddon/code-server-1.31.0-x86_64-linux] Specify working dir

OPTIONS
  -d, --data-dir=data-dir
  -h, --host=host          [default: 0.0.0.0]
  -o, --open               Open in browser on startup
  -p, --port=port          [default: 8443] Port to bind on
  -v, --version            show CLI version
  --allow-http
  --cert=cert
  --cert-key=cert-key
  --help                   show CLI help
  --no-auth

Pitu:
What I meant with my message is that if the flag is in the docs maybe it's a feature not pushed yet but already implemented. I don't know the roadmap, was just pointing that out.

icebob:
Docs updated with #5 https://github.com/icebob/code-server/commit/e1099f0fb4240ac39f9fe788d86a2e1527724b17#diff-3e3c890e38761b2c49f651e11b921f9d
But it's not released yet.

NGTmeaty:
While regular Math.Random is probably fine, we may want to look into using a CSPRNG.

sr229:
why not just use `/dev/urandom`?

Nothing could possibly go wrong using the OS's Random device.

NGTmeaty:
[...]

PR #5, opened 2019-03-05
Implement #4 - fix password via CLI

icebob:
Add `--password=123456` CLI argument to use fix password instead of random.

@kylecarbs Currently I'm on Windows so I can't test it but I hope the code is correct. Maybe could you try it?

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -26,6 +26,7 @@ export class Entry extends Command {
 		version: flags.version({ char: "v" }),
 		"no-auth": flags.boolean({ default: false }),
 		"allow-http": flags.boolean({ default: false }),
+		password: flags.string(),
 
 		// Dev flags
 		"bootstrap-fork": flags.string({ hidden: true }),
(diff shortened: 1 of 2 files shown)


packages/server/src/cli.ts
```diff
@@ -26,6 +26,7 @@ export class Entry extends Command {
 		version: flags.version({ char: "v" }),
 		"no-auth": flags.boolean({ default: false }),
 		"allow-http": flags.boolean({ default: false }),
+		password: flags.string(),
```
kylecarbs: We could generate the password here and add it as a default. What do you think?

andreimc: @kylecarbs for non interactive env's a flag should be provided I think

packages/server/src/cli.ts
```diff
@@ -132,13 +133,17 @@ export class Entry extends Command {
...
 		});
 
-		const passwordLength = 12;
-		const possible = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
-		const chars = [];
-		for (let i = 0; i < passwordLength; i++) {
-			chars.push(possible[Math.floor(Math.random() * possible.length)]);
+		let password = flags["password"];
+		if (!password) {
+			// Generate a random password
```
nhooyr: Unrelated issue but this needs to be changed. The password should generated from a CPRNG.

packages/server/src/cli.ts
icebob: @kylecarbs Maybe, but in this case, `code-server --help` will print always a random password as a default value for `--password` argument, won't it?

packages/server/src/cli.ts
icebob: This PR is only related to adding a static password feature, not changing original password generation method.

packages/server/src/cli.ts
kylecarbs: Ah thats correct. I agree.

packages/server/src/cli.ts
Multishifties: For those reading this thread, aforementioned issue is located at https://github.com/codercom/code-server/issues/26

PR #6, opened 2019-03-05
Add task for packaging release

coderasher:


build/tasks.ts
```diff
@@ -260,4 +260,43 @@ const ensurePatched = register("vscode:patch", async (runner) => {
...
+	const releasePath = path.resolve(__dirname, `../release/${releaseTag}`);
+	fse.removeSync(releasePath);
+	fse.mkdirpSync(releasePath);
+
+	await Promise.all([{
+		bin: "cli-linux",
+		name: "linux",
+	}, {
+		bin: "cli-osx",
+		name: "apple-darwin",
```
kylecarbs: Wouldn't we only want to do one of these on the specific platform?

coderasher: Ah, good point, currently this is set up so you download the two built binaries then package them up into archives. Would definitely be better to do this as part of the build process.

kylecarbs:
Optimally we'd fix the `v1.0.0` output when launching the binary as well. It'd be helpful to have the version once we add automatic-updates too.

build/tasks.ts
```diff
@@ -260,4 +260,30 @@ const ensurePatched = register("vscode:patch", async (runner) => {
...
+	fse.copySync(binaryPath, binaryDestination);
+	fs.chmodSync(binaryDestination, "755");
+	["README.md", "LICENSE"].forEach((fileName) => {
+		fse.copySync(path.resolve(__dirname, `../${fileName}`), path.join(archiveDir, fileName));
+	});
+
+	runner.cwd = releasePath;
+	await os.platform() === "linux"
+		? runner.execute("tar", ["-cvzf", `${archiveName}.tar.gz`, `${archiveName}`])
+		: runner.execute("zip", ["-r", `${archiveName}.zip`, `${archiveName}`]);
```
Multishifties: For our use case, is tar preferable to zip? And if so, is there any way for us to set a conditional based on whether or not said commands are installed rather than OS platform?

coderasher: The goal is to provide tarballs for Linux and zips for Mac & Windows. This runs in the CI and is what will be available for download on the "releases" page.

PR #10, opened 2019-03-05
Fixes calculate num of cpu threads for HappyPack

thekiba:
HappyPack requires a positive integer for threads. And when the number of cpu is 1, it is calculated 1 - 1 = 0. Now a minimal value will be 1.


kylecarbs:
CI is broken for PRs, but this LGTM.

PR #29, opened 2019-03-06
Fix yarn install missing vscode lib dependency

### LINKED ISSUE #22: Dependency `spdlog` missing
- `code-server` version: 1d8da2161f4330e2cf26fd28005729e3b6e21286
- OS Version: Darwin Kernel Version 17.7.0

#### Steps to Reproduce

1. `yarn; yarn start;`, then throw TS Compile Error
2. Missing dependency `spdlog`: https://github.com/codercom/code-server/blob/master/packages/protocol/src/common/helpers.ts#L1

barretlee:
close #22

### DIFF
--- a/build/tasks.ts
+++ b/build/tasks.ts
@@ -217,16 +217,11 @@ const ensureCloned = register("vscode:clone", async (runner) => {
 	} else {
 		fse.mkdirpSync(libPath);
 		runner.cwd = libPath;
-		const clone = await runner.execute("git", ["clone", "https://github.com/microsoft/vscode"]);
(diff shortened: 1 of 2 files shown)


kylecarbs:
In order for everything to run properly, you must run `yarn task build:server:binary`. We should probably change this to throw an error in `cli.ts` if required files don't exist.

kylecarbs:
I don't believe we should add this. I'd prefer to add a message after `postinstall` in the root that `yarn task server:package:binary is the correct route for development.

kylecarbs:
Closing as this seems to have been fixed.

PR #32, opened 2019-03-06
WIP: Prettify readmes and files

grikomsn:
Main purpose was to organize readmes, but I also prettified other files. Still a work in progress, will push other updates soon.


grikomsn:
Going to add Docker usage based on #57

PR #49, opened 2019-03-06
Remove unused markup from password page

Multishifties:
screen shot 2019-03-06 at 1 56 20 pm

Currently this checkbox does nothing. Neither does "helper text" provide any useful context at this time. We can remove them for now.


Multishifties:
I've compiled and tested locally. My environment is as follows:

PR #54, opened 2019-03-06
Revise welcome page

nol166:
Add Coder specific links, remove non-working links.

### DIFF
--- a/packages/web/webpack.config.js
+++ b/packages/web/webpack.config.js
@@ -65,7 +65,7 @@ module.exports = merge(
 			"spdlog": path.join(vsFills, "spdlog.ts"),
 			"native-watchdog": path.join(vsFills, "native-watchdog.ts"),
 			"iconv-lite": path.join(vsFills, "iconv-lite.ts"),
-			"welcome": path.join(vsFills, "welcomePage.ts"),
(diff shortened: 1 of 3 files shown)


packages/web/webpack.config.js
```diff
@@ -65,6 +65,7 @@ module.exports = merge(
 			"spdlog": path.join(vsFills, "spdlog.ts"),
 			"native-watchdog": path.join(vsFills, "native-watchdog.ts"),
 			"iconv-lite": path.join(vsFills, "iconv-lite.ts"),
+			"welcome": path.join(vsFills, "welcomePage.ts"),
```
coderasher: The alias will need to match the import, so in this case `vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page`.

kylecarbs:
We should be careful about this. Wouldn't want to miss an update on their welcome page. Not sure if theres really a way around that.

coderasher:
Ahh, true, maybe we should patch it after all? I suggested this but the updates are definitely a good reason to go with patching.

Multishifties:
Should we consider using the update modal that we currently have on our hosted version and stick to using that for all coder-related stuff?

nol166:
Closing this one. Going to make a patch.

PR #42, opened 2019-03-06
add dockerfile

tacyarg:
Very basic implementation of a docker file, would need some optional arguments to support the certs

PR #57, opened 2019-03-06
Add Dockerfile and some cleanup

### LINKED ISSUE #9: Docker container to try out Coder.
I think you should add a docker container deployment one liner to your landing page similar to https://www.theia-ide.org/

Should be simple enough to create the image and add to your CI and keep an official one with Docker Hub...with docker for windows this also *sorta* meets your windows goals

nhooyr:
Closes #9 #42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM node:8.15.0
+
+# Install VS Code's deps. These are the only two it seems we need.
+RUN apt-get update
+RUN apt-get install -y libxkbfile-dev libsecret-1-dev
+
+# Ensure latest yarn.
+RUN npm install -g yarn
+
+# In the future, we can use https://github.com/yarnpkg/rfcs/pull/53 to make it use the node_modules
+# directly which should be faster.
+WORKDIR /src
+COPY . .
+RUN yarn
+RUN yarn task build:server:binary
+
+# We deploy with ubuntu so that devs have a familiar environemnt.
+FROM ubuntu:18.10
+RUN apt-get update
+RUN apt-get install -y openssl
+RUN apt-get install -y net-tools
+WORKDIR /root/project
+COPY --from=0 /src/packages/server/cli-linux /usr/local/bin/code-server
+EXPOSE 8443
+# Unfortunately `.` does not work with code-server.
+CMD code-server $PWD
--- a/packages/runner/src/runner.ts
+++ b/packages/runner/src/runner.ts
@@ -1,5 +1,5 @@
 import * as cp from "child_process";
-import { logger, Logger, field, time } from "@coder/logger";
+import {field, Logger, logger, time} from "@coder/logger";
 
 export interface CommandResult {
 	readonly exitCode: number;
@@ -9,7 +9,9 @@ export interface CommandResult {
 
 const execute = (command: string, args: string[] = [], options: cp.SpawnOptions, logger: Logger): Promise<CommandResult> => {
 	let resolve: (result: CommandResult) => void;
-	const prom = new Promise<CommandResult>(res => resolve = res);
+	const prom = new Promise<CommandResult>((res): void => {
+		resolve = res;
+	});
 
 	const stdout: string[] = [];
 	const stderr: string[] = [];
@@ -44,6 +46,7 @@ export type TaskFunction = (runner: Runner) => void | Promise<void>;
 
 export interface Runner {
 	cwd: string;
+
 	execute(command: string, args?: string[], env?: object): Promise<CommandResult>;
 }
 
@@ -90,10 +93,22 @@ export const run = (name: string = process.argv[2]): void | Promise<void> => {
 			cwd = path;
 		},
 		execute(command: string, args: string[] = [], env?: object): Promise<CommandResult> {
-			return execute(command, args, {
+			const prom = execute(command, args, {
 				cwd,
 				env: env as NodeJS.ProcessEnv,
 			}, log);
+
+			return prom.then((result: CommandResult) => {
+				if (result.exitCode != 0) {
+					log.error("failed",
+						field("exitCode", result.exitCode),
+						field("stdout", result.stdout),
+						field("stderr", result.stderr)
+					);
+				}
+
+				return result;
+			});
 		},
 	});
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
 set -e
 
-npm install -g cross-env
 yarn task build:server:binary
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile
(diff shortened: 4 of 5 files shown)


scripts/build.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
 set -e
 
-npm install -g cross-env
```
kylecarbs: Why was this removed?

packages/runner/src/runner.ts
```diff
@@ -90,10 +93,21 @@ export const run = (name: string = process.argv[2]): void | Promise<void> => {
 			cwd = path;
 		},
 		execute(command: string, args: string[] = [], env?: object): Promise<CommandResult> {
-			return execute(command, args, {
+			const prom = execute(command, args, {
 				cwd,
 				env: env as NodeJS.ProcessEnv,
 			}, log);
+			prom.then((result: CommandResult) => {
```
kylecarbs: Should return this instead of prom below

packages/runner/src/runner.ts
```diff
@@ -44,6 +46,7 @@ export type TaskFunction = (runner: Runner) => void | Promise<void>;
 
 export interface Runner {
 	cwd: string;
+
```
kylecarbs: Unneeded space, but not a biggie

scripts/build.sh
nhooyr: Wasn't necessary to install globally according to @coderasher as its a library.

packages/runner/src/runner.ts
nhooyr: Interesting. It seems to work both ways. Is it a convenience thing?

packages/runner/src/runner.ts
nhooyr: Goland formatted automatically, will fix.

packages/runner/src/runner.ts
nhooyr: screen shot 2019-03-06 at 6 12 36 pm

Can't.

scripts/build.sh
coderasher: Yup it's part of the root `package.json`.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN yarn task build:server:binary
+
+# We deploy with ubuntu so that devs have a familiar environemnt.
+FROM ubuntu
+RUN apt-get update
+RUN apt-get install -y openssl
+WORKDIR /root/project
+COPY --from=0 /src/packages/server/cli-linux /usr/local/bin/code-server
+EXPOSE 8443
+# Unfortunately `.` does not work with code-servr
```
nol166: small typo

nhooyr: Fixed.

packages/runner/src/runner.ts
kylecarbs: You can add
```
return result;
```
to the end of that

nhooyr: Not sure what you mean. To the end of what?

kylecarbs: Inside the `prom.then` block

nhooyr: ah very fancy.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# In the future, we can use https://github.com/yarnpkg/rfcs/pull/53 to make it use the node_modules
+# directly which should be faster.
+WORKDIR /src
+COPY . .
+RUN yarn
+RUN yarn task build:server:binary
+
+# We deploy with ubuntu so that devs have a familiar environemnt.
+FROM ubuntu
```
sr229: Pin the version - since this will always resolve to `18.10` (current `latest`).

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+RUN apt-get update
+RUN apt-get install -y openssl
+WORKDIR /root/project
```
sr229: I do not recommend using `root`. Consider using a actual user or a pseudo-user like what we do in Theia.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+# In the future, we can use https://github.com/yarnpkg/rfcs/pull/53 to make it use the node_modules
+# directly which should be faster.
+WORKDIR /src
+COPY . .
+RUN yarn
+RUN yarn task build:server:binary
+
+# We deploy with ubuntu so that devs have a familiar environemnt.
+FROM ubuntu
+RUN apt-get update
```
sr229: Compress these two `RUN`s as one instruction.

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM node:8.15.0
+
+# Install VS Code's deps. These are the only two it seems we need.
+RUN apt-get update
+RUN apt-get install -y libxkbfile-dev libsecret-1-dev
```
sr229: Compress these two `RUN`s as one instruction as well.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Ensure latest yarn.
+RUN npm install -g yarn
+
+# In the future, we can use https://github.com/yarnpkg/rfcs/pull/53 to make it use the node_modules
+# directly which should be faster.
+WORKDIR /src
+COPY . .
+RUN yarn
```
sr229: compress this as one instruction as well.

Dockerfile
nhooyr: Its a small thing but I like em separate to make better use of docker's layer cache.

Dockerfile
nhooyr: I personally don't really see much value in that. Especially for a dev setup, you pretty much need sudo 100% of the time at which point it just becomes annoying and repetitive.

cc @kylecarbs

kylecarbs: I'm not certain what purpose there'd be in limiting the user when running inside the docker container. What do you think @sr229 ?

sr229: @kylecarbs @nhooyr we're excercising a good practice that the user should be aware that some administrative actions might need `sudo` or a manual switch to the root. Not all Cloud IDE users are proficient in Linux and know the security implications. 

This is also how Theia and Cloud9 does their images.

Dockerfile
sr229: Optimizing layers no matter how small should be done.

Dockerfile
nhooyr: Lets get this PR in for now and discuss this in #65.

Dockerfile
frol: @nhooyr it is fine to separate them while you are getting the image done, but once it is there, it is [strongly recommended (see “apt-get” section)](
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/) to combine update + install + remove caches

nhooyr: That is interesting. Does seem reasonable to me to bust the cache every time the package list is modified.

PR #71, opened 2019-03-07
fix Dockerfile

yush1ga:
Thanks for making this awesome soft open source.

Unfortunately, `Dockerfile` doesn't work in my environment.
It may depend on os arch environment.

I tried to fix it by tweaking `Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,8 @@ RUN apt-get update
 RUN apt-get install -y openssl
 RUN apt-get install -y net-tools
 WORKDIR /root/project
-COPY --from=0 /src/packages/server/cli-linux /usr/local/bin/code-server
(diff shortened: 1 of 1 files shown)


nhooyr:
Could you elaborate on the error/issue you were having? I don't see what this fixes.

coderasher:
Ah, probably because I added the architecture to the binary name here: https://github.com/codercom/code-server/commit/994531d8bbfa62d849b65a5828f87f3b4c8983fa#diff-09fded6ea343c3bb08cac161ed7a3cbaR17

nhooyr:
Dammit Asher. I'll push a fix for this. Thanks for pointing this out @yush1ga

nhooyr:
See #109

yush1ga:
@nhooyr @coderasher Thanks!

PR #72, opened 2019-03-07
make docker build working with after a fresh checkout

zerdos:


build/tasks.ts
```diff
@@ -224,7 +224,7 @@ const ensureCloned = register("vscode:clone", async (runner) => {
 	}
 
 	runner.cwd = vscodePath;
-	const checkout = await runner.execute("git", ["checkout", "tags/1.31.1"]);
+	const checkout = await runner.execute("git", ["checkout", "tags/1.31.1"]); //TODO: this tag should come from a parameter
```
kylecarbs: What do you mean by a parameter?

nhooyr: I guess he means it shouldn't be hard coded.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.15.0
+FROM node:8.9.3
```
nhooyr: Why the change?

Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN yarn && yarn task build:server:binary
 # We deploy with ubuntu so that devs have a familiar environemnt.
 FROM ubuntu:18.10
 WORKDIR /root/project
-COPY --from=0 /src/packages/server/cli-linux /usr/local/bin/code-server
+COPY --from=0 /src/packages/server/cli-linux-x64   /usr/local/bin/code-server
```
nhooyr: afaik, there shouldn't be a `-x64` suffix?

@kylecarbs

nhooyr:
I don't understand what this PR fixes. How does `docker build .` not work after a fresh checkout?

Dockerfile
zerdos: There should be - that is actually the only output

nhooyr:
See #72

zerdos:
This PR makes the Dockerfile actually build without error :)

Dockerfile
nhooyr: Yup. This was recently changed without me knowing. It's fixed in #109.

PR #75, opened 2019-03-07
version from 1.31.0->1.31.1

sujunmin:
#45 missing this to docker user.


kylecarbs:
This is just an arbitrary version in code for the apps which will be coming out soon. This version isn't actually used anywhere atm!

PR #84, opened 2019-03-07
Generate the en_US.UTF-8 locale in the docker image

### LINKED ISSUE #83: Locales are not available in local docker image
- `code-server` version: master
- OS Version:  4.20.8-arch1-1-ARCH

not having locales can cause unexpected behaviour when working with text that is utf-8 encoded

#### Steps to Reproduce

1. docker run -p 8443:8443 -v "${PWD}:/root/project" codercom/code-server code-server --allow-http --no-auth
2. open terminal and type `locale -a` see `en_US.UTF-8 is` missing

xMrWhite:
Fixes issues when working with utf-8 text from the command line

close #83

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,8 @@ FROM ubuntu:18.10
 RUN apt-get update
 RUN apt-get install -y openssl
 RUN apt-get install -y net-tools
+RUN apt-get install -y locales
+RUN locale-gen en_US.UTF-8
 WORKDIR /root/project
 COPY --from=0 /src/packages/server/cli-linux /usr/local/bin/code-server
 EXPOSE 8443


Dockerfile
```diff
@@ -19,6 +19,8 @@ FROM ubuntu:18.10
 RUN apt-get update
 RUN apt-get install -y openssl
 RUN apt-get install -y net-tools
+RUN apt-get install -y locales
+RUN locale-gen en_US.UTF-8
```
nhooyr: You'll also need

```
# We unfortunately cannot use update-locale because docker will not use the env variables
ENV LANG=en_US.UTF-8
```

nhooyr:
If you don't mind, I'll take this over to get this out quicker. Will credit you in my PR.

PR #88, opened 2019-03-07
update example docker command port bind options

Omeryl:
The existing command errors out with 

```
docker: invalid publish opts format (should be name=value but got 'localhost:8443:8443').
```

it needs to explicitly use an IP and not a hostname.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 Try it out:
 ```bash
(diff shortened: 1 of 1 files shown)


Omeryl:
CI failing is unrelated, given it's only updating a markdown file.

coadler:
@kylecarbs this also needs to be fixed on the homepage

PR #95, opened 2019-03-07
Patch welcome page

nol166:
Patch welcome page for non-working links

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -864,6 +864,31 @@ index d8de423529..5953e920c3 100644
 +		Object.defineProperty(this._options, 'useSameOriginForRoot', {
 +			value: true,
 +		});
+diff --git a/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts b/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
+index d4cd525812..6904db5518 100644
+--- a/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
++++ b/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
+@@ -30,8 +29,0 @@ export default () => `
+-				<div class="section recent">
+-					<h2 class="caption">${escape(localize('welcomePage.recent', "Recent"))}</h2>
+-					<ul class="list">
+-						<!-- Filled programmatically -->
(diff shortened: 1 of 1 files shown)


scripts/vscode.patch
```diff
@@ -864,6 +864,31 @@ index d8de423529..5953e920c3 100644
...
+-				<div class="section recent">
+-					<h2 class="caption">${escape(localize('welcomePage.recent', "Recent"))}</h2>
+-					<ul class="list">
+-						<!-- Filled programmatically -->
+-						<li class="moreRecent"><a href="command:workbench.action.openRecent">${escape(localize('welcomePage.moreRecent', "More..."))}</a><span class="path detail if_shortcut" data-command="workbench.action.openRecent">(<span class="shortcut" data-command="workbench.action.openRecent"></span>)</span></li>
+-					</ul>
+-					<p class="none detail">${escape(localize('welcomePage.noRecentFolders', "No recent folders"))}</p>
+-				</div>
+@@ -41,6 +33,5 @@ export default () => `
+-						<li class="keybindingsReferenceLink"><a href="command:workbench.action.keybindingsReference">${escape(localize('welcomePage.keybindingsCheatsheet', "Printable keyboard cheatsheet"))}</a></li>
```
coderasher: Just leaving this here as a note not to merge yet; we're gonna provide the URLs (via `product.ts`) instead of replacing the `href`s.

scripts/vscode.patch
```diff
@@ -864,6 +864,31 @@ index d8de423529..5953e920c3 100644
 +		Object.defineProperty(this._options, 'useSameOriginForRoot', {
 +			value: true,
 +		});
+diff --git a/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts b/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
+index d4cd525812..6904db5518 100644
+--- a/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
++++ b/src/vs/workbench/parts/welcome/page/electron-browser/vs_code_welcome_page.ts
+@@ -30,8 +29,0 @@ export default () => `
+-				<div class="section recent">
```
coderasher: I would prefer commenting this out `<!-- -->` so the patch can be as minimal and additive as possible, should help with potential merge conflicts when we update VS Code.

coderasher: Also, it might be a good idea for us to actually implement the open recent command at some point.

kylecarbs:
Not sure we should do this. I feel like it could be rough to maintain.

Issue #100, opened 2019-03-07
remove install packages script


nhooyr:
See https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/

nhooyr:
@kylecarbs marking this as urgent as its leaving a bad impression on contributors

mko-x:
Same issue here - pretty annoying. Wanted to help with some stuff, but if the 'stable' is so unstable at local build it's getting to a problem.

See the logs at my [hastebin](https://paste.m-ko.de/ocobiwihag.sql)

Any insights what to do?

lherman-cs:
same issue too 
![Screenshot_2019-03-09_14-32-55](https://user-images.githubusercontent.com/15654932/54076290-10fc7500-4278-11e9-862a-4b0ada8058b7.png)

kylecarbs:
Will be fixing asap. In the meantime, running yarn multiple times should result in success.

PR #104, opened 2019-03-07
Docker issues

zerdos:
- add .dockerignore files
- instead of parallel execution, use sequential one (to avoid async errors)
- use locked yarn files (otherwise, we can't be sure which version will be installed)
- use the same node version what we are using on Travis


zerdos:
I uploaded an image to dockerhub, which was built from my branch, by dockerhub:
https://hub.docker.com/r/zerdos/vscode-server/tags

nhooyr:
> use the same node version what we are using on Travis

Lets upgrade the travis version instead @kylecarbs 

> instead of parallel execution, use sequential one (to avoid async errors)

Lets just do #100 

> use locked yarn files (otherwise, we can't be sure which version will be installed)

This is a solid change.

> add .dockerignore files

This is a good change too.

.dockerignore
```diff
@@ -1 +1,12 @@
+.git
+.gitignore
 Dockerfile
+lib
+node_modules
```
nhooyr: Why do we want to not put node_modules into the container?

Right now yarn doesn't use it as a cache but it could in the future. See the comment in the Dockerfile.

build/tasks.ts
```diff
@@ -235,6 +235,11 @@ const ensureClean = register("vscode:clean", async (runner) => {
 
 	const status = await runner.execute("git", ["status", "--porcelain"]);
 	if (status.stdout.trim() !== "") {
+
+		// this is an uggly workaround - inside docker git clean throws an error message because of this directory
+		// related to git modules.
+		await runner.execute("rm", ["-rf", "node_modules/spdlog"]);
```
nhooyr: I've never ran into an error with this. Can you describe in more detail what you did and what happened?

yarn.lock
```diff
@@ -27,172 +27,175 @@
   integrity sha1-7ihweulOEdK4J7y+UnC86n8+ce4=
```
nhooyr: Why does this file have so many changes?

scripts/install-packages.ts
```diff
@@ -41,7 +41,7 @@ const handlePackages = async (dir: string): Promise<void> => {
 		const pkgDir = join(dir, pkg);
 		const pkgJsonPath = join(pkgDir, "package.json");
 		if (existsSync(pkgJsonPath)) {
-			const ip = doInstall(pkg, pkgDir);
+			const ip = await doInstall(pkg, pkgDir);
```
nhooyr: @kylecarbs you fine with this for now?

nhooyr: (below we should remove the `await` if this is good to go.

packages/vscode/package.json
```diff
@@ -14,6 +14,7 @@
   },
   "devDependencies": {
     "@types/tar-stream": "^1.6.0",
+    "cross-env": "^5.2.0",
```
nhooyr: @kylecarbs why wasn't this here already?

build/tasks.ts
zerdos: the git clean command throws an error, that a submodule is not git repository.
I didn't get the error message on my mac - but inside docker yes.

Give me a minute and I try to reproduce the issue.

nhooyr:
@zerdos instead of opening multiple PRs, in the future please edit the existing PR.

.dockerignore
nhooyr: I guess we can add it back later.

Dockerfile
```diff
@@ -1,32 +1,47 @@
-FROM node:8.9.3
+FROM node:8.15.1
```
nhooyr: 8.15.1 will not build as nexe does not support it.

https://github.com/nexe/nexe/releases

.dockerignore
zerdos: https://medium.com/@rdsubhas/docker-for-development-common-problems-and-solutions-95b25cae41eb

In linux we have linux binaries, on mac we have mac binaries, on windows we have exe files.

NPM is a hot mess, so node_modules should be git + docker ignored.

build/tasks.ts
zerdos: reverted. Maybe it was an earlier version of node - or the yarn - installed by npm (yakk). 
Now its working without the workaround

scripts/install-packages.ts
zerdos: I try to remove it and see what happens in docker.

zerdos:
```ERROR Failed to install "common" dependencies {"error":{"killed":false,"code":1,"signal":null,"cmd":"yarn --frozen-lockfile --network-concurrency 1"},"stdout":"[1/4] Resolving packages...\n[2/4] Fetching packages...\ninfo If you think this is a bug, please open a bug report with the information provided in \"/src/packages/app/common/yarn-error.log\".\ninfo Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.\n","stderr":"warning package.json: No license field\nwarning ../../package.json: No license field\nwarning @coder/app-common: No license field\nerror An unexpected error occurred: \"https://registry.yarnpkg.com/@material/layout-grid/-/layout-grid-0.41.0.tgz: ENOENT: no such file or directory, open '/usr/local/share/.cache/yarn/v4/npm-@material-layout-grid-0.41.0-2e7d3be76313e0684d573b10c2c6a88b3230d251/node_modules/@material/layout-grid/.yarn-tarball.tgz'\".\n"}
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
ERROR: "packages:install" exited with 1.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command '/bin/sh -c yarn --frozen-lockfile' returned a non-zero code: 1```

Comment (nhooyr) [reply scripts/install-packages.ts]: That's half the error msg.

File: packages/vscode/package.json (hunk 5 above)
Comment (zerdos) [reply packages/vscode/package.json]: I've seen a commit yesterday, which removed the global instal of this packet from travis.
I think you guys have this package installed globally :)

File: scripts/install-packages.ts (hunk 4 above)
Comment (zerdos) [reply scripts/install-packages.ts]: Nah, its the full of it. 
Try to have a clean checkup (without the node modules).

Full log:
https://gist.github.com/zerdos/e94dd4867f3b2866b87ca002d218ff82

**Comment by zerdos** [comment]
Removed the yarn.lock file changes, also the other non version bumps.

...and it seems the docker build is still working :)

File: Dockerfile [hunk 7]
```diff
@@ -11,7 +11,7 @@ RUN npm install -g yarn
 # directly which should be faster.
 WORKDIR /src
 COPY . .
-RUN yarn
+RUN yarn --forzen-lockfile
```
Omeryl: This should be `--frozen-lockfile`

scripts/install-packages.ts
kylecarbs: Once we convert to yarn workspaces this should be a non-issue. I'm good w/ it for now.

Dockerfile
```diff
@@ -1,32 +1,45 @@
...
+RUN yarn --frozen-lockfile 
+
+RUN yarn task vscode:install
+RUN yarn task build:copy-vscode
+RUN yarn task build:web
+RUN yarn task build:bootstrap-fork
+RUN yarn task build:default-extensions
+RUN yarn task build:server:bundle
+RUN yarn task build:app:browser
+RUN yarn task build:server:binary:package
```
kylecarbs: Tasks depend on prior tasks. Running in this way will likely increase total build time significantly.

packages/vscode/package.json
coderasher: This should be fixed now; the scripts directly reference the `cross-env` binary installed in the root `node_modules` directory.

Dockerfile
mko-x: Yes @kylecarbs I think so too
Changing this to 'supposed' serial execution ruins the idea of tasks.ts file which clearly respects the dependencies.

The only benefit - that is to admit - there would be a Docker build layer for each step, allowing to cache single yarn steps

.dockerignore
nhooyr: Are you sure about this? I thought `node_modules` only contained source code? Not actual binaries.

lsmoura: @zerdos is right. Check the folder `node_modules/.bin/`. There are modules that build executables in there (the `node-sass` package is one that comes to mind. I don't think code-server uses it, but there are many other examples).

Dockerfile
kylecarbs: That's a fair point. We can also optimize to cache previous tasks.

nhooyr: We can't cache single yarn steps with this because the input to each step is the same source. So every time the source changes, everything will still be rebuilt, it won't get reused.

sr229: I would like to also add this is very inefficient and doesn't really do what your PR is supposed to do.

package.json
```diff
@@ -6,8 +6,8 @@
   "description": "Run VS Code remotely.",
   "scripts": {
     "build:rules": "cd ./rules && tsc -p .",
-    "packages:install": "cd ./packages && yarn",
-    "postinstall": "npm-run-all --parallel packages:install build:rules",
+    "packages:install": "cd ./packages && yarn --frozen-lockfile",
+    "postinstall": "npm-run-all packages:install build:rules",
```
sr229: I want a rationale why the parallel flag have to be removed.

zerdos:
I wasn't able to run a successful build without this. 
Yes, its slower and less efficient, but at least running, after a fresh checkout.

sr229:
Unfortunately this is a cherry pick not a merge, I don't think that Dockerfile is a pass even for me. It's way too many layers for a build step whoch won't be used anyways. Practice Layer optimization please.

zerdos:
The end result will be small (2 stages build).

Those yarn commands could be on one layer for sure, but the build should be sequential - its not running with the parallel execution task unfortunately after a fresh checkout.

The biggest difference in the PR, that I put the node_modules directories to the .dockerignore.

Btw, on my laptop (mac environment) I never managed to run a successful build from the master branch.

PR #120, opened 2019-03-08
[README] Remove --allow-http from docker option

ymmt2005:
Chrome disables clipboard API when connecting to HTTP sites.
As a consequence, if a user connects to `code-server` with HTTP, they cannot cut or copy texts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 Try it out:
 ```bash
-docker run -p 127.0.0.1:8443:8443 -v "${PWD}:/root/project" codercom/code-server code-server --allow-http --no-auth
+docker run -p 127.0.0.1:8443:8443 -v "${PWD}:/root/project" codercom/code-server code-server --no-auth
 ```
 
 - Code on your Chromebook, tablet, and laptop with a consistent dev environment.


kylecarbs:
Hm, this makes sense.

What are your thoughts? @nhooyr

nhooyr:
Makes me really sad. Really don't want people seeing that insecure stuff.

nhooyr:
I do have a plan for this though. Will need to be discussed in private.

korzq:
I think we should keep --allow-http since some people use nginx to reverse proxy.

nhooyr:
Yea not sure this is worth it. I think its best we keep `--allow-http` as the default for the docker container one-liner.

Or we could distribute a script instead that uses `mkcert` to register a self signed certificate and start things up using that.

GottZ:
my reverse proxy does the https part. i prefer running my containers without tls in between them and my nginx. (just a reminder that there are people who know what they are doing)

example: https://rutorrent.home.gottz.de

so.. tls should always be optional and never be forced. at least something that can be turned off or is turned off by default and requires additional configuration otherwise.
(literally as it is right now)

codehz:
But chrome allow use clipboard API in localhost without ssl, it is ok for testing

ymmt2005:
I'm closing this.  Thank you all for comments.

PR #130, opened 2019-03-08
Fixed documentation download links based on latest version

MoienTajik:
The specified download links in the Docs is no longer exist and gives 404 not found: https://github.com/codercom/code-server/releases/download/0.1.4/code-server-linux

Fixed by replacing them with the latest release [links](https://github.com/codercom/code-server/releases/tag/1.31.1-100).

No


kylecarbs:
Might be better to generalize or remove these commands so we don't need to update the README on each release.

MoienTajik:
It could be better if the latest version always exists on the Coder.com website with a general URL which always points to the latest version. So this URL could always be referenced in the documentation, like: https://coder.com/releases/download/latest/code-server-linux

Another option could be [`/releases/latest`](https://help.github.com/en/articles/linking-to-releases#linking-to-the-latest-release) in Github.

kylecarbs:
I'm good with using `/releases/latest`.

MoienTajik:
Updated based on `/releases/latest`.

MoienTajik:
**Ping !**

doc/admin/install/google_cloud.md
```diff
@@ -24,23 +24,38 @@ If you're just starting out, we recommend [installing code-server locally](../..
...
+tar -xvzf code-server-1.31.1-100-linux-x64.tar.gz
 ```
-3. Make the binary executable if you run into any errors regarding permission:
-  ```
-  chmod +x code-server-linux
-  ```
+
+4. Make the binary executable if you run into any errors regarding permission:
+```
+chmod +x code-server-linux
```
coderasher: This has changed a bit; the packaged binary name will always be `code-server` now.

doc/admin/install/digitalocean.md
```diff
@@ -16,9 +16,13 @@ If you're just starting out, we recommend [installing code-server locally](../..
...
 - Once in the SSH session, visit code-server [releases page](https://github.com/codercom/code-server/releases/) and copy the link to the download for the latest linux release
-- In the shell run the below command with the URL from your clipboard
+- Find and download the latest Linux release from this URL:
   ```
-  wget https://github.com/codercom/code-server/releases/download/0.1.4/code-server-linux
+  https://github.com/codercom/code-server/releases/latest
+  ```
+- Extract the downloaded tar.gz file with this command, for example:
+  ```
+  tar -xvzf code-server-1.31.1-100-linux-x64.tar.gz
```
coderasher: Since this creates a directory we might want to add a step to run `cd code-server-1.31.1-100-linux-x64`.

doc/admin/install/aws.md
```diff
@@ -32,10 +32,14 @@ If you're just starting out, we recommend [installing code-server locally](../..
...
-- With the URL in the clipboard, run:
+- At this point it is time to download the `code-server` binary. We will of course want the linux version.
+- Find and download the latest Linux release from this URL:
   ```
-  wget https://github.com/codercom/code-server/releases/download/0.1.4/code-server-linux
+  https://github.com/codercom/code-server/releases/latest
+  ```
+- Extract the downloaded tar.gz file with this command, for example:
+  ```
+  tar -xvzf code-server-1.31.1-100-linux-x64.tar.gz
```
coderasher: Would it make sense to leave the `wget` command? We could say something like:

1. Find the latest version number on the [releases page](https://github.com/codercom/code-server/releases/latest).
2. Replace `{version}` in the following command with the version found on the releases page and run it (or just copy the download URL from the releases page):
```
wget https://github.com/codercom/code-server/releases/download/{version}/code-server-{version}-linux-x64.tar.gz
```
This would also let us use `{version}` as a placeholder in all the other commands.

MoienTajik: I've made it like this now.

doc/admin/install/google_cloud.md
```diff
@@ -24,23 +24,48 @@ If you're just starting out, we recommend [installing code-server locally](../..
...
+
+7. Start the code-server
 ```
 sudo ./code-server-linux -p 80
 ```
+
 > For instructions on how to keep the server running after you end your SSH session please checkout [how to use systemd](https://www.linode.com/docs/quick-answers/linux/start-service-at-boot/) to start linux based services if they are killed
-5. Access code-server from the public IP of your Google Cloud instance we noted earlier in your browser. 
+
+8. Access code-server from the public IP of your Google Cloud instance we noted earlier in your browser.
```
avelino: Use markdown number list (generate auto sequence)

```
1. one
1. two
1. ...
```

Render:
1. one
1. two
1. ...

MoienTajik: I've changed that to an unordered list now to be same as other docs.

PR #146, opened 2019-03-08
Add support for Kubernetes 

### LINKED ISSUE #60: Kubernetes Support
I wanted to see if this project would be interested in Kubernetes as a deployment option. I worked on a  proof of concept and have things mostly working. I'm happy to share this with the repo but wanted to check in and see if there was interest before cleaning things up with a PR.

stevesloka:

Fixes #60 by adding support for Kubernetes. This deploys code-server, Contour for ingress controller, and Cert Manager for Let's Encrypt certificates, with persistent storage with AWS EBS volumes. Also, I left the Dockerfile I used for golang development. If there's a public push of this somewhere it would be great to document how you can create base images with the right tooling up front. 

This is also somewhat opiononated in that it's using AWS EBS volumes, but can be adapted for many use cases. I wasn't sure how that might get organized within the repo. 

Signed-off-by: Steve Sloka <slokas@vmware.com>

#60 

// cc @jbeda

### DIFF
--- a/deployment/deployment.yaml
+++ b/deployment/deployment.yaml
@@ -1,0 +1,43 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: code-server
+---
+apiVersion: v1
+kind: Service
+metadata:
+ name: code-server
+ namespace: code-server
+spec:
+ ports:
+ - port: 8443
+   name: https
+   protocol: TCP
+ selector:
+   app: code-server
+ type: ClusterIP
+---
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  labels:
+    app: code-server
+  name: code-server
+  namespace: code-server
+spec:
+  selector:
+    matchLabels:
+      app: code-server
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: code-server
+    spec:
+      containers:
+      - image: codercom/code-server
+        imagePullPolicy: Always
+        name: code-server
+        ports:
+        - containerPort: 8443
+          name: https
--- a/deployment/aws/deployment.yaml
+++ b/deployment/aws/deployment.yaml
@@ -1,0 +1,74 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: code-server
+---
+apiVersion: v1
+kind: Service
+metadata:
+ name: code-server
+ namespace: code-server
+spec:
+ ports:
+ - port: 8443
+   name: https
+   protocol: TCP
+ selector:
+   app: code-server
+ type: ClusterIP
+---
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: gp2
+  annotations:
+    storageclass.kubernetes.io/is-default-class: "true"
+provisioner: kubernetes.io/aws-ebs
+parameters:
+  type: gp2
+  fsType: ext4 
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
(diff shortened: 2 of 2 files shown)


NGTmeaty:
This looks awesome! I wonder if it could be improved by possibly integrating or using parts of our current [Dockerfile](https://github.com/codercom/code-server/blob/master/Dockerfile) as we currently do some building?

stevesloka:
@NGTmeaty Sounds good! I just updated to use the example Dockerfile. I should probably add some docs to this as well. Let me see if I can pull those together so it's more clear to folks.

NGTmeaty:
Awesome! +1

bhack:
It could be nice to have an official helm chart.

sr229:
Please do not rely on EBS only, make it vendor agnostic

stevesloka:
It doesn't, just an example. You could reference any storage class and it would work.

stevesloka:
I could integrate kustomize to allow for customization. Could pick some clouds and show examples (e.g. AWS, GCP, Azure). What would you suggest?

sr229:
@stevesloka since its an example, might as well use default storage class for PVC or `hostMount`, then have providers documented if they want to

deployment/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN npm install -g yarn@1.13
+
+WORKDIR /src
+COPY . .
+
+# In the future, we can use https://github.com/yarnpkg/rfcs/pull/53 to make yarn use the node_modules
+# directly which should be fast as it is slow because it populates its own cache every time.
+RUN yarn && yarn task build:server:binary
+
+# Use golang base dev for go development tooling
```
kylecarbs: Not sure we should change the base image

deployment/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+    matchLabels:
+      app: code-server
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: code-server
+    spec:
+      containers:
+      - image: stevesloka/code-server
```
kylecarbs: I believe this can now be `codercom/code-server`

deployment/Dockerfile
stevesloka: I'll leave it generic. My intent was to show how to have bootstrapped images. Having to install deps is not great, but doesn't make sense at this point in time.

deployment/deployment.yaml
stevesloka: I'll change it back. 👍

deployment/Dockerfile
kylecarbs: I like the idea of extending it.

We could include an example of something like:

```dockerfile
FROM codercom/code-server

RUN apt-get install mysql -y

...
```

PhilsLab:
I think the ingress controller and TLS should not be part of this. Depending on the cluster setup, those are unwanted extras in most cases.

deployment/Dockerfile
Zate: Any desire to move it to 1804. LTS for the base instead of 18.10 ?  Generally prefer to build docker images on things that have longer support.

kylecarbs:
Is this ready to go? Happy to merge this in.

stevesloka:
@nhooyr I can add some docs on a follow-up PR.

nhooyr:
Awesome, sounds good man.

nhooyr:
@kylecarbs can you disable the signed commits requirement?

PR #149, opened 2019-03-08
open websocket on same path as page

dbushong:
This is the critical piece to let you serve code-server proxied under a
path.  Otherwise if you proxy e.g. `/editor/` thru to
`http://localhost:8000`, everything works fine except the websocket
connection is still opened to `/`

### DIFF
--- a/packages/ide/src/fill/client.ts
+++ b/packages/ide/src/fill/client.ts
@@ -91,8 +91,9 @@ class WebsocketConnection implements ReadWriteConnection {
 	 */
 	private async openSocket(): Promise<WebSocket> {
 		this.dispose();
+		const wsProto = location.protocol === "https:" ? "wss" : "ws";
 		const socket = new WebSocket(
-			`${location.protocol === "https:" ? "wss" : "ws"}://${location.host}`,
+			`${wsProto}://${location.host}${location.pathname}`,
 		);
 		socket.binaryType = "arraybuffer";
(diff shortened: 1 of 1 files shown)


dbushong:
The app with this works minimally when proxied.  There are some 404ing assets, which I'm not immediately sure how to fix:

* [ ] `/lib/vscode/src/vs/workbench/parts/webview/electron-browser/webview.html`
* [ ] `/resource/home/travis/build/codercom/code-server/packages/server/build/extensions/theme-seti/icons/seti.woff`
* [ ] `/__webpack_hmr`
* [ ] `/lib/vscode/src/vs/workbench/browser/parts/editor/media/split-editor-horizontal-inverse.svg`
* [ ] `/lib/vscode/src/vs/workbench/parts/webview/electron-browser/webview-pre.js`
* [ ] `/lib/vscode/src/vs/platform/extensionManagement/node/media/defaultIcon.png`

Basically all of those should be `/whatever/path/page/is/on/lib/vscode/....`, I assume the fix is similar to this websocket url fix.

kylecarbs:
Did these assets 404 prior to the WebSocket fix?

dbushong:
The only one I can confirm is `__webpack_hmr` - the others don't even load without the websocket connecting.

avelino:
This issue fixed: 
https://github.com/codercom/code-server/issues/141

PR #155, opened 2019-03-09
Adding ENTRYPOINT to omit `code-server` at the CLI. 

ay-b:
Having `code-server` at the ENTRYPOINT will free the CMD line, so launching that container will no longer require writing the binary name after image name and will not confuse people anymore.


nhooyr:
Very nice change. Could you rebase off dev?

nhooyr:
See #170

Please don't open additional PRs for the same, you can use `git push --force` to rewrite a PR.

PR #167, opened 2019-03-09
Clone exact vscode release branch when build task

jeasonstudio:

Build toooo slow.

### DIFF
--- a/build/tasks.ts
+++ b/build/tasks.ts
@@ -219,17 +219,12 @@ const ensureCloned = register("vscode:clone", async (runner) => {
 	} else {
 		fse.mkdirpSync(libPath);
 		runner.cwd = libPath;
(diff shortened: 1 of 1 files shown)


kylecarbs:
I believe the release is 1.32.0

nhooyr:
This broke local builds.

nhooyr:
Because the checkout was removed.

kylecarbs:
Ah. We can revert this.

kylecarbs:
A bit confused how the CI passed

nhooyr:
CI clones every time the exact release after this PR. When we try and reuse the existing lib/vscode directory, this PR makes it not checkout the exact release.e

nhooyr:
Will include fix in my upcoming PR.

PR #170, opened 2019-03-09
Added ENTRYPOINT to omit full command at launch.

ay-b:
Having `code-server` at the ENTRYPOINT will free the CMD line, so launching that container will no longer require writing the binary name after image name and will not confuse people anymore.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,4 +27,5 @@ RUN apt-get install -y locales &&
 # configured in /etc/default/locale so we need to set it manually.
 ENV LANG=en_US.UTF-8
 # Unfortunately `.` does not work with code-server.
-CMD code-server $PWD
+ENTRYPOINT ["code-server"]
+CMD ["$PWD"]


ay-b:
I don't know where the agrument logic lives, but at least as a workaround in it to replace " . " with " $PWD " before actual submitting to binary should work. Isn't it?

nhooyr:
Sorry, I don't understand. Could you elaborate? Are you asking why `.` doesn't work?

ay-b:
> Sorry, I don't understand. Could you elaborate? Are you asking why `.` doesn't work?

I don't know why it doesn't work — didn't see the code, sorry. But replacing the " . " argument with the " $PWD " prior the arguments get worked out should do the trick.

I'm sorry if I'm speaking nonsense.

nhooyr:
Ah looks like this was fixed to work now: https://github.com/codercom/code-server/issues/19

I will update. Thanks.

PR #185, opened 2019-03-10
Ansible Deployment Support

### LINKED ISSUE #182: Ansible Deployments
Because not everyone has the expense of doing the same thing per machine.

I made [one](https://github.com/sr229/code-server-ansible) already, I just need feedback before I upstream this.

sr229:

Adds Ansible Playbook deployments for repetitive deployments.

Closes #182 .


nhooyr:
I would be against putting this stuff in the repo as we don't want to maintain it.

You can put this in your own repo and we'd be happy to link to it.

nhooyr:
See #182

PR #192, opened 2019-03-11
feat: user-mode docker

### LINKED ISSUE #65: sudo in Dockerfile?
See https://github.com/codercom/code-server/pull/57#discussion_r263202554

sr229:

Prior to discussion to #65, this runs all file operations in the user `coder` in container along with some refactoring of the Dockerfile to make it easier to read.

This resolves #65. This will be blocked until general consensus that this merit as a merge.


kylecarbs:
Wonderful change.

ping @nhooyr

Dockerfile
```diff
@@ -17,15 +17,27 @@ RUN yarn && yarn task build:server:binary
...
-	locale-gen en_US.UTF-8
+
 # We unfortunately cannot use update-locale because docker will not use the env variables
 # configured in /etc/default/locale so we need to set it manually.
 ENV LANG=en_US.UTF-8
-ENTRYPOINT ["code-server"]
+
+COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
+RUN apt-get update && apt-get install -y \
+    sudo \
```
nhooyr: Please don't combine this. Minimizing layers in favour of readability is premature optimization.

https://github.com/codercom/code-server/issues/183#issuecomment-472776658

nhooyr: By combine I mean the `echo` and `adduser` stuff.

Its fine to install all deps in a single RUN.

sr229: I don't think it merits its own `RUN`? It's better to have it on one build step already than to force it in another layer when they do the exact same thing. 

On a novice Linux user's perspective, which I also asked from a couple of Linux beginners, this is still readable, since you can tell after `apt` the next step is adding user and changing modes, which most are acquainted at already. Which defeats your "Minimzing layers in favor of readability" point. If you really want unreadable Docker images, look further more into the official library images.

This chain is actually more innocent than the rest of how the Docker library images are done (where compliancy is a must). And to validate my claim this is readable in most beginners' perspective, I asked people to read it and ask if they can read and describe me how everything works (which 94% of the people agreeing this is readable).

kylecarbs:
Once the conflicts are resolved I'm good to merge @sr229

sr229:
merge issues fixed, CI seems stuck but PR build reports fine

davefinster:
I feel like this is worth mentioning, but I made my own changes to shift over to a non-root user that mimic the changes in this PR. The Dockerfile in question is here: https://gist.github.com/davefinster/39825c0fa7cf168b8114d24bd3b3df53

When I first deployed the container and everything worked except Workspace creation. The container was setup via Kube using these params:
```yaml
command: ["code-server"]
args:
  - --allow-http
  - --no-auth
  - --port=8440
  - --data-dir=/home/coder/.code-server
  - /home/coder
workingDir: /home/coder
```
Despite that, I was getting file write errors whenever VS Code tried to anything Workspace related since it was always attempting to write into /root/.code-server/Workspaces/<number>/workspace.json regardless of what I did. I ended up having to add these lines to get it to work:
```bash
mkdir -p /root/.code-server/Workspaces && \
chown -R 3000 /root/.code-server && \
chmod -R 777 /root/.code-server && \
chmod 755 /root
```

kylecarbs:
@davefinster interesting. @sr229 have you experienced that with this dockerfile?

Dockerfile
sabrehagen: > It's better to have it on one build step already than to force it in another layer when they do the exact same thing.

What does "the exact same thing"? If it did the "exact same thing", why would you do it twice?

This command has three purposes: installing dependencies, adding a user, and configuring application state. None of those three should be in the same step; they are all distinct, unrelated operations.

sr229:
@davefinster @kylecarbs This is a mirror of my [Dockerfile](https://github.com/sr229/code-server-openshift), particularly the Vanilla dockerfile. Every operation works, and you shouldn't be mounting on `/root`, rather in `/home/coder` on my end.

davefinster:
Ignore my comment - turns out it was due to a dirty workspace key in localstorage.

sr229:
Request for another review for merge @coadler @kylecarbs

lsmoura:
@sr229 you need to merge master into your branch again. 😓

sr229:
Should be ready to merge now.

Issue #55, opened 2019-03-06
No error/warning for incorrect password.


Multishifties:
- `code-server` version: 1.31.0
- OS Version: Mac OS 10.13.6
#### Steps to Reproduce

1. Build binary
2. Run binary
3. Enter incorrect password

Expected: 
Some kind of message saying "password is incorrect"

Actual:
¯\\_(ツ)_/¯

Multishifties:
I can take this one. I've already made mention of it in https://github.com/codercom/code-server/pull/51

Multishifties:
UPDATE: PR is here https://github.com/codercom/code-server/pull/201

PR #201, opened 2019-03-11
Add error message for incorrect password. Fix issue #55

Multishifties:

Issue:
On initial welcome/login screen of binary, no error message is displayed in the event of an incorrect password. (See https://github.com/codercom/code-server/issues/55)

Changes:
Adds simple logic and styling for error display. Currently is of identical styling/design to the `coder.com` login page.

Test Environment:
(Result: Pass! Compile was successful and everything functions as expected)

Caveats:
None that I can think of at the moment. (I mean, the error message is hard coded. But this page has only one input and the page only has one purpose and does exactly one thing. So I don't know that this is necessarily a bad thing)

### DIFF
--- a/packages/app/browser/src/app.html
+++ b/packages/app/browser/src/app.html
@@ -8,21 +8,20 @@
 
 <body>
 	<div class="login">
-		<div class="back">
-			<- Back </div>
-			<h4 class="title">code-server</h4>
-				<h2 class="subtitle">
-					Enter server password
-				</h2>
-				<div class="mdc-text-field">
-					<input type="password" id="password" class="mdc-text-field__input" required>
-					<label class="mdc-floating-label" for="password">Password</label>
-					<div class="mdc-line-ripple"></div>
-				</div>
-				<button id="submit" class="mdc-button mdc-button--unelevated">
-					<span class="mdc-button__label">Enter IDE</span>
-				</button>
+		<div class="back"> <- Back </div>
+		<h4 class="title">code-server</h4>
+		<h2 class="subtitle">
+			Enter server password
+		</h2>
+		<div class="mdc-text-field">
+			<input type="password" id="password" class="mdc-text-field__input" required>
+			<label class="mdc-floating-label" for="password">Password</label>
+			<div class="mdc-line-ripple"></div>
 		</div>
+		<button id="submit" class="mdc-button mdc-button--unelevated">
+			<span class="mdc-button__label">Enter IDE</span>
+		</button>
+		<div id="error-display"></div>
 	</div>
 </body>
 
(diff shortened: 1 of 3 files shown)


Multishifties:
@nhooyr Can you review this PR? I can't request you as a reviewer and Asher and Kyle probably have a backlog that will last for a while.

nhooyr:
@kylecarbs why can't he request me as a reviewer?

packages/app/browser/src/app.html
```diff
@@ -8,21 +8,20 @@
...
-				</h2>
-				<div class="mdc-text-field">
-					<input type="password" id="password" class="mdc-text-field__input" required>
-					<label class="mdc-floating-label" for="password">Password</label>
-					<div class="mdc-line-ripple"></div>
-				</div>
-				<button id="submit" class="mdc-button mdc-button--unelevated">
-					<span class="mdc-button__label">Enter IDE</span>
-				</button>
+		<div class="back"> <- Back </div>
```
nhooyr: Glad you fixed the indentation here, but it did make the diff a little harder to read. The only thing you changed here is to add the error display div right?

packages/app/browser/src/app.ts
```diff
@@ -28,3 +28,14 @@ submit.addEventListener("click", () => {
 	document.cookie = `password=${password.value}`;
 	location.reload();
 });
+
+/**
+ * Notify user on load of page if previous password was unsuccessful
+ */
+const reg = new RegExp(`password=(\\w+);?`);
```
nhooyr: Could you describe how this works for me? Like why `document.cookie.match`? How are you telling if the previous password was unsuccessful?

packages/app/browser/src/app.html
Multishifties: Yep, just added the error display element.

packages/app/browser/src/app.ts
Multishifties: Certainly!

This page is only ever loaded if a valid password cookie does not exist. You can see that on line 29 the original code calls `location.reload()` which would just load into the IDE because the key is set (assuming it's valid). 

Document cookies are weird. Rather than doing something logical like returning a map or object. The returned value from `document.cookie` is a single string value with all the cookies concatenated together and separated by semi-colons. So I'm using `string.prototype.match()`to apply a regex and search to see if a cookie named "password" exists in the cookie string.

So with the exception of the first load we can determine that if this page ever loads and "password" cookie exists than it must not be correct. As for the handling of first load, I'm using `document.referrer` (line 39) which returns null if there was/is no previous page load.

nhooyr: Isn't it insecure to keep the password in the cookie like this?

nhooyr: Oh nvm, not coder.com, I guess this is fine for self hosted.

PR #221, opened 2019-03-13
Combine apt-get in final docker image and remove cache

SuperSandro2000:

I had no problem but this reduces the amount of layers needed to download and removes the apt-get cache to reduce the image size

nop

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,9 +22,10 @@ COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
 EXPOSE 8443
 RUN apt-get update && apt-get install -y \
 	openssl \
-	net-tools
-RUN apt-get install -y locales && \
-	locale-gen en_US.UTF-8
+	net-tools \
+	locales && \
+	locale-gen en_US.UTF-8 && \
+	rm -rf /var/lib/apt/lists/*
 # We unfortunately cannot use update-locale because docker will not use the env variables
 # configured in /etc/default/locale so we need to set it manually.
 ENV LANG=en_US.UTF-8


Dockerfile
```diff
@@ -22,9 +22,10 @@ COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
 EXPOSE 8443
 RUN apt-get update && apt-get install -y \
 	openssl \
-	net-tools
-RUN apt-get install -y locales && \
-	locale-gen en_US.UTF-8
+	net-tools \
+	locales && \
+	locale-gen en_US.UTF-8 && \
+	rm -rf /var/lib/apt/lists/*
```
sr229: you forgot to run `apt-get clean` first.

SuperSandro2000: ``Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.``
Source: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

nhooyr:
The two `RUN`s are intentionally separate for easier readability.

The apt-get cache removal looks reasonable to me.

SuperSandro2000:
Combined it is readable too and there is already a comment. Don't think we need to keep it seperated.

If we don't combine it the cache removal is totally useless as we download all the data with the first layer and then remove it in the next layer. This means if we don't combine the apts we would make the image overall worse.

And personally I use Alpine for every image because it is so damn small and keeps my hard drive empty. I am currently doing them myself and maybe going to contribute them if they work good.

nhooyr:
> Combined it is readable too and there is already a comment. Don't think we need to keep it separated.

Readability is subjective. I understand you think its readable but imo it is harder. Also, there is no comment.

> If we don't combine it the cache removal is totally useless as we download all the data with the first layer and then remove it in the next layer. This means if we don't combine the apts we would make the image overall worse.

That is true, but I think that's ok. The package cache is very small. I favour the readability of separate runs.

Thank you for the contribution but for the above reasons, I cannot merge it.

SuperSandro2000:
Well then I do my own like always. People just can't make nice small docker images.

coderasher:
@nhooyr Isn't it recommended to combine `update` and `install` commands? https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

ammario:
Yes, this PR is good at least because we want to couple the `apt-get update` layer with the `apt-get install` layer. It ensures the package cache is updated whenever we change which packages are installed.

SuperSandro2000:
@coderasher oh thanks man! ❤️ I should get a printer and hang that on my wall.

nhooyr:
> @nhooyr Isn't it recommended to combine update and install commands? https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

Yes, but that would mean we add a `apt-get update` to the second `RUN`, not combine them.

> Yes, this PR is good at least because we want to couple the apt-get update layer with the apt-get install layer. It ensures the package cache is updated whenever we change which packages are installed.

We don't have to combine the RUNs to do that.

SuperSandro2000:
> Yes, but that would mean we add a apt-get update to the second RUN, not combine them.

I am glad you didn't do that cause that would be kinda stupid.

nhooyr:
@SuperSandro2000 Look man, we're all just trying to make code-server better. Can you please be a little more respectful?

I don't see why it would have been stupid. It would have been fine either way.

SuperSandro2000:
English is not my first language so it is a bit hard for me. I didn't mean it in an offensive way or so.

We would basically run the same command twice. I don't understand why you don't like the idea of just merging the two lines. Readability is just fine.

nhooyr:
> English is not my first language so it is a bit hard for me. I didn't mean it in an offensive way or so.

Understood. It's all good.

Issue #203, opened 2019-03-12
Git in docker container. 


greenfox1505:
I believe Git just be installed by default in the docker container. Git integration is a core feature of VSCode. It's easy to add `apt-get install git` without significant complication or inflation of the docker image.

jeasonstudio:
sgtm

sr229:
actually there's some inflation but its not detriminal as git itself is very lightweight.

greenfox1505:
I'M HOME, it's not *significant* inflation.

SuperSandro2000:
Maybe it would be a good idea for future additions to add a bare bone image for people like me that wan't to save drive space and a more featured image.
If we base the more featured image on the barer one we could make the building process really easy.

nhooyr:
Opened #234

PR #230, opened 2019-03-13
Add git to docker container

pokemonlover1234:
Install git in the docker container, proposed by #203. While it is possible that issue requires further discussion, this is just a proposal pull request. If that proposal is accepted, here's the pull request.

  This is a proposal pull request in response to an open issue
#203


nhooyr:
You want to put `git` after the second `FROM` where we install `openssl` for example.

pokemonlover1234:
Closing in preference to #231

PR #231, opened 2019-03-13
Dockerfile: include git

### LINKED ISSUE #203: Git in docker container.

nhooyr:
Also improves installation of runtime deps.

Closes #221
Closes #230
Closes #203

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,10 @@ COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
 EXPOSE 8443
 RUN apt-get update && apt-get install -y \
 	openssl \
-	net-tools
-RUN apt-get install -y locales && \
-	locale-gen en_US.UTF-8
+	net-tools \
+	locales && \
+	locale-gen en_US.UTF-8 && \
+	rm -rf /var/lib/apt/lists/*
 # We unfortunately cannot use update-locale because docker will not use the env variables
 # configured in /etc/default/locale so we need to set it manually.
 ENV LANG=en_US.UTF-8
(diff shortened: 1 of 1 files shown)


ammario:
I think we should merge in everyone else's PRs instead so they get attribution.

pokemonlover1234:
Mine (#230) does the same thing, except I put the git install in the wrong spot.

nhooyr:
@ammario the changes in this PR are different from the other PRs. Regardless, I've included all their commits here so they still get attribution and then included my own commit at the end to fix things up.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM node:8.15.0
 
-# Install VS Code's deps. These are the only two it seems we need.
+# Install VS Code's deps. These are the only two it seems we need. Install git for source control.
```
ammario: Second sentence is in the wrong place (and extraneous imo)

nhooyr: I agree, from old a commit I just added. Will remove.

Dockerfile
```diff
@@ -20,11 +20,13 @@ FROM ubuntu:18.10
...
+
 RUN apt-get update && apt-get install -y \
 	openssl \
-	net-tools
-RUN apt-get install -y locales && \
-	locale-gen en_US.UTF-8
+	net-tools \
+	git \
+	locales
+RUN locale-gen en_US.UTF-8
```
SuperSandro2000: We could squash that into the other run.

PR #245, opened 2019-03-14
Fix install from VSIX for TAR and ZIP formats

### LINKED ISSUE #92: Detect archive type of extension to decompress
We've filled the client zip implementation with tar so we can eventually load extensions on the client. Installing from a VSIX no longer works because the standard format is a zip.

We should detect the archive type then use the appropriate fill.

foresthoffman:
Fixes installing extensions from VSIX files formatted as either TAR or ZIP.

Fixes #92


packages/vscode/src/fill/zip.ts
```diff
@@ -58,81 +58,38 @@ export function zip(tarPath: string, files: IFile[]): Promise<string> {
...
-	const sourcePathRegex = new RegExp(options.sourcePath ? `^${options.sourcePath}` : '');
-
+export async function extract(archivePath: string, extractPath: string, options: IExtractOptions = {}, token: CancellationToken): Promise<void> {
 	return new Promise<void>(async (c, e) => {
-		const buffer = await promisify(fs.readFile)(tarPath);
-		const extractor = tarStream.extract();
-		extractor.once('error', e);
-		extractor.on('entry', (header, stream, next) => {
-			const rawName = header.name;
+		extractZip(archivePath, extractPath, options, token).then(c).catch((ex) => {
```
kylecarbs: Should we try extracting a tar first? We expect _most_ extensions to be in the tar format.

foresthoffman: That makes sense, extraction only occurs for installing from the marketplace.

Issue #35, opened 2019-03-06
ARM binary release


frozenmaple:
I have a RPi 3B+ module，and i hope code-server will be running within this platform, thank u

sr229:
@frozenmaple due to the resource usage of VSCode, no, its not possible unless you use a Memory-optimized version of Electron for Raspberry Pi.

kylecarbs:
We'll be reducing the amount of memory `code-server` uses. It's a decently-high priority for us.

Would running this on a PI just require us to have an ARM target?

sr229:
@kylecarbs yep, `armv7l` and `arm64`.

kylecarbs:
Will take a bit as nexe 2.x.x doesn't support arm as a target.

icebob:
@kylecarbs do you insist on using nexe? [pkg](https://github.com/zeit/pkg) is a similar project to create executable from Node project with [arm support](https://github.com/zeit/pkg#targets). It's just my two cents.

kylecarbs:
Hm that's fair. We've considered using `pkg`, will consider it again.

Jakst3r:
It would be amazing to be able to run this in docker on my [Odroid XU4Q ](https://www.odroid.co.uk/index.php?route=product/product&product_id=813) &[ ODROID-N2](https://www.hardkernel.com/blog-2/odroid-n2/) when it's released.

SuperSandro2000:
@Jakst3r are they arm64? If not you can forget it cause build time on qemu arm32 is like 10x and more.

Jakst3r:
@SuperSandro2000 

Looks like the Odroid XU4Q is out of the question, what do you think about the ODROID-N2?

[Odroid XU4Q ](https://www.odroid.co.uk/index.php?route=product/product&product_id=813) = 32-bit
Samsung Exynos-5422 : Cortex™-A15 and Cortex™-A7 big.LITTLE processor with 2GByte LPDDR3 RAM


[ ODROID-N2](https://www.hardkernel.com/blog-2/odroid-n2/) = 64-bit
quad-core ARM Cortex-A73 CPU cluster and a dual core Cortex-A53 cluster with a new generation Mali-G52 GPU. 4GByte DDR4 RAM running at 1320Mhz

sr229:
@kylecarbs I would like to also note we need to tell V8 explicitly to use Ignition (JITLess) to lower memory requirements for ARM boards (but I believe we should also apply this to x86 as well).

[...]

PR #246, opened 2019-03-14
Add basic cross-compile code

SuperSandro2000:
Build script now should be able to cross compile. I most likely missed some places.
The binary builds but I did not yet test functionality.
#35 

BTW this is my first real js contribution.

### DIFF
--- a/build/tasks.ts
+++ b/build/tasks.ts
@@ -6,13 +6,19 @@ import * as path from "path";
 import * as zlib from "zlib";
 
 const isWin = os.platform() === "win32";
+const arg = process.argv.slice(3);
+const availableArches = ["ia32", "x64", "arm", "arm64"];
+const linuxArch = arg.length === 0 ? "x64" : arg;
 const libPath = path.join(__dirname, "../lib");
 const vscodePath = path.join(libPath, "vscode");
 const pkgsPath = path.join(__dirname, "../packages");
-const defaultExtensionsPath = path.join(libPath, "VSCode-linux-x64/resources/app/extensions");
+const defaultExtensionsPath = path.join(libPath, `VSCode-linux-${linuxArch}/resources/app/extensions`);
 const vscodeVersion = "1.32.0";
 
 const buildServerBinary = register("build:server:binary", async (runner) => {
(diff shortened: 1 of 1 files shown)


kylecarbs:
The `arm` target still won't work with this, correct?

SuperSandro2000:
IIRC the pulled nexe binary is wrong.

build/tasks.ts
```diff
@@ -197,7 +215,7 @@ const buildDefaultExtensions = register("build:default-extensions", async (runne
 	if (!fs.existsSync(defaultExtensionsPath)) {
 		await copyForDefaultExtensions();
 		runner.cwd = extDirPath;
-		const resp = await runner.execute(isWin ? "npx.cmd" : "npx", [isWin ? "gulp.cmd" : "gulp", "vscode-linux-x64", "--max-old-space-size=32384"]);
+		const resp = await runner.execute(isWin ? "npx.cmd" : "npx", [isWin ? "gulp.cmd" : "gulp", `vscode-linux-${linuxArch}`, "--max-old-space-size=32384"]);
```
kylecarbs: We _should_ be able to target linux from any OS since it normally builds with electron. This might not be necessary.

build/tasks.ts
```diff
@@ -64,7 +70,19 @@ const dependencyNexeBinary = register("dependency:nexe", async (runner) => {
 			if (!fs.existsSync(upxBinary)) {
 				fse.mkdirpSync(upxFolder);
 				runner.cwd = upxFolder;
-				const upxExtract = await runner.execute("bash", ["-c", "curl -L https://github.com/upx/upx/releases/download/v3.95/upx-3.95-amd64_linux.tar.xz | tar xJ --strip-components=1"]);
+				var upxArch = "";
+				if ("ia32" === linuxArch) {
```
kylecarbs:
```ts
switch (linuxArch) {
 ...
}
```
Might work better here.

NGTmeaty:
This may have been solved by moving to nbin. If not I imagine this will need a few changes(or moved entirely to nbin.)

kylecarbs:
Fixed by moving to `nbin` :confetti_ball:

PR #269, opened 2019-03-17
Use CMD to allow running a shell inside the container

SuperSandro2000:

I can't start the container with /bin/sh to debug it.

No but I hope this https://stackoverflow.com/a/21564990/4446318 helps explain the issue better than I can.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,4 +30,4 @@ RUN locale-gen en_US.UTF-8
 # We unfortunately cannot use update-locale because docker will not use the env variables
 # configured in /etc/default/locale so we need to set it manually.
 ENV LANG=en_US.UTF-8
-ENTRYPOINT ["code-server"]
+CMD ["code-server"]


SuperSandro2000:
No you can't just if no container is running.
``docker run image bash`` will fail cause bash is no argument. Additionally if ever the code-server binary will fail to start on startup you have no easy way to debug this. You would need to build a docker image based on this on to get into that container. Entrypoint is meant to be used to change the shell or to add helper scripts at startup.

sr229:
I don't think its limited JUST for helper scripts. ENTRYPOINT was designed for applications to take arguments during run. Which is basically a glorified `exec` passing.

If you really want to run bash in it, don't do it via `docker run`. Run the image first then `docker exec`.

It's easier to debug it while running rather than slapping `CMD` which does not allow us to pass more args if needed.

SuperSandro2000:
Yes, it is.

I mean I could just build my own image, too. Doesn't solve the problem.

If it is not running you can't debug it and my container wasn't running.
CMD only needs the executable in it otherwise it does the same here.

So the only change would be that all docker-compose users with a command entry need to append code-server at the beginning.

nhooyr:
@SuperSandro2000 I appreciate your concern but the issue for us is we want to keep the docker one-liner as short as possible.

You have two options for debugging.

1. Start the container and then use `docker exec`.
2. Start a new Dockerfiles and use the `COPY` instruction to copy the binary from our image into your own and set it up however you want.

nhooyr:
Ah, you actually want `docker run  --entrypoint=/bin/bash`. That should work.

SuperSandro2000:
I literally listed your debug ideas already and said why they are not ideal.

``docker run --entrypoint=/bin/bash``
That is more useful. Thanks

PR #272, opened 2019-03-17
Add .node-version file

kemuridama:

Many developers are using **env tools such as [nodenv](https://github.com/nodenv/nodenv) to treat various version environments in each project.
nodenv is specifying node version for projects using by .node-version file.
So, this PR adds `.node-version` file for using nodenv and sets node version v8.15.0 according to Dockerfile.

None.


kylecarbs:
Do lots of people want this addition? It's very small, but I could see this getting out of hand with version files.

kemuridama:
Maybe yes. Some contributors or developers who try to build code-server locally fail to build code-server due to Node.js version does not match its version requiring to build code-server. Because 
latest Node.js version is v11, they are often used it mainly and switch other version using nodenv.

PR #284, opened 2019-03-18
Fix vscode-textmate dependency

foresthoffman:

There was a 'vscode-textmate' dependency for the `CoenraadS.bracket-pair-colorizer-2` extension, which included 'oniguruma' as a dependency. This PR adds a section to the bootstrapFork module requirement which will allow the inclusion of dependencies that should exist with the unpacked version of VS Code.

https://github.com/codercom/code-server/issues/200#issuecomment-473182369


foresthoffman:
Closing in favor of https://github.com/codercom/code-server/pull/299. Making the PR from branch rather than a fork.

PR #285, opened 2019-03-19
Refactor evaluations

code-asher:
This refactors evaluations to send messages. With this method all we send are the arguments along with the method name. This is accomplished using proxies and interfaces for those proxies to tie everything together nicely.

Proxies are advantageous mainly they allow us to ~share interfaces between the client and server~ (it's even better now because it just uses the actual class so we don't even need to maintain any interfaces), and they also allow code like:

```
this.proxy.createWriteStream(...);
```

instead of something like:

```
this.client.call("createWriteStream", ...);
```

TODO:
- [X] fs
- [X] fs tests
- [x] child_process
- [x] child_process tests
- [x] net
- [x] net tests
- [x] spdlog
- [x] spdlog tests
- [x] node-pty
- [x] node-pty tests
- [x] trash
- [x] benchmarks (speed and CPU usage)

Some concerns:
- [x] ~Right now there is no way to ensure that all the events are properly hooked up. It would be nice if there was a way to enforce that events from the proxy on the client are all handled and that events from the actual instance on the server are all handled.~ <-- I think this is solved now in a fairly reasonable way.
- [ ] Making sure everything is disposed (no memory leaks). <-- Pretty sure this is OK but would be advantageous to run more tests.


code-asher:
I'm not sure the best way to benchmark this, but I created and wrote to 10k files and here are the results (mean taken over 10 runs):

Native: 383ms
Evaluations (old method): 10568ms (~27.6x slower than native)
Messages (new method): 2022 (~5.3x slower than native)

code-asher:
Same benchmark except with raw node:

Native: 300.8ms
Evaluations (old method): 8648.2ms (~28.75x slower than native)
Messages (new method): 1408.5ms (~4.68x slower than native)

code-asher:
Same benchmark except with raw node:

Native: 300.8ms
Evaluations (old method): 8648.2ms (~28.75x slower than native)
Messages (new method): 1408.5ms (~4.68x slower than native)

code-asher:
Uh...whoops.

code-asher:
I also profiled them; the old version spent most of its time in JavaScript looping across event emitters callbacks apparently. In the new version that's all gone and `parse` has become the biggest spender (but it's not taking much longer than it did before, 20 ticks now vs the 15 from before---for reference the old `evalDoneEmitter.event` was taking 510).

The old version spent most of its C++ time in `runInContext` and evaluating the code, which we already knew. In the new version that's completely gone. The biggest spenders in the new version are the `map` lookups.

andreimc:
i ran this branch and my extensions wouldn't activate

kylecarbs:
This also adds [`nbin`](https://github.com/codercom/nbin), our custom binary packer. Will reduce a lot of tech debt in the future.

PR #303, opened 2019-03-20
Add descriptions for flags in the CLI

nol166:
Fills in the descriptions for flags that are missing one ( --help)

n/a

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -235,7 +235,7 @@ export class Entry extends Command {
 			logger.warn("Launched without authentication.");
 		}
 
-		const url = `https://localhost:${flags.port}/`;
+		const url = flags["allow-http"] ? `http://localhost:${flags.port}/` : `https://localhost:${flags.port}/`;
 		logger.info(" ");
 		logger.info("Started (click the link below to open):");
 		logger.info(url);
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -235,7 +235,7 @@ export class Entry extends Command {
 			logger.warn("Launched without authentication.");
 		}
 
-		const url = `http://localhost:${flags.port}/`;
+		const url = `https://localhost:${flags.port}/`;
 		logger.info(" ");
 		logger.info("Started (click the link below to open):");
 		logger.info(url);
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -15,21 +15,22 @@ import { createApp } from "./server";
 import { forkModule, requireFork, requireModule } from "./vscode/bootstrapFork";
 import { SharedProcess, SharedProcessState } from "./vscode/sharedProcess";
 import opn = require("opn");
+import { args } from "@oclif/parser";
 
 export class Entry extends Command {
 	public static description = "Start your own self-hosted browser-accessible VS Code";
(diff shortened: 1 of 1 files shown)


packages/server/src/cli.ts
```diff
@@ -15,21 +15,22 @@ import { createApp } from "./server";
...
 import opn = require("opn");
+import { args } from "@oclif/parser";
 
 export class Entry extends Command {
 	public static description = "Start your own self-hosted browser-accessible VS Code";
 	public static flags = {
-		cert: flags.string(),
-		"cert-key": flags.string(),
-		"data-dir": flags.string({ char: "d" }),
+		cert: flags.string({char: "c", description: "Point to an SSL certificate .crt file"}),
```
creativeguy2013: Should be TLS not SSL

TLS certificates don't have to end in .crt

Maybe change to "Point to a TLS certificate file"

packages/server/src/cli.ts
```diff
@@ -15,21 +15,22 @@ import { createApp } from "./server";
...
+		"cert-key": flags.string({ char: "k", description: "Point to an SSL certificate .key file"}),
```
creativeguy2013: Should be TLS not SSL

TLS certificate key files don't have to end in .key

Maybe change to "Point to a TLS certificate private key file"

packages/server/src/cli.ts
```diff
@@ -15,21 +15,22 @@ import { createApp } from "./server";
...
+		"data-dir": flags.string({ char: "d", default: "Current directory", description: "Specify the working directory for the IDE"  }),
 		help: flags.help(),
 		host: flags.string({ char: "h", default: "0.0.0.0" }),
 		open: flags.boolean({ char: "o", description: "Open in browser on startup" }),
 		port: flags.integer({ char: "p", default: 8443, description: "Port to bind on" }),
 		version: flags.version({ char: "v" }),
-		"no-auth": flags.boolean({ default: false }),
-		"allow-http": flags.boolean({ default: false }),
-		password: flags.string(),
+		"no-auth": flags.boolean({ char: "n", default: false, description: "Start code-server without password" }),
```
creativeguy2013: Change to "Start code-server without a password"

packages/server/src/cli.ts
```diff
@@ -15,21 +15,22 @@ import { createApp } from "./server";
...
+		"allow-http": flags.boolean({ char: "a", default: false, description: "Allow use of HTTP protocol"  }),
+		password: flags.string( {description: "Specify the IDE password inline"}),
```
creativeguy2013: Is the _inline_ necessary? Is there an other way of specifying the password?

packages/server/src/cli.ts
```diff
@@ -234,7 +234,7 @@ export class Entry extends Command {
 			logger.warn("Launched without authentication.");
 		}
 
-		const url = `http://localhost:${flags.port}/`;
+		const url = `http${flags["allow-http"] ? "s" : ""}://localhost:${flags.port}/`;
```
kylecarbs: I believe this is inverted.

nol166: corrected

kylecarbs:
I referenced these titles for descriptions in the new CLI structure. If any need to be adjusted, feel free to reopen.

Issue #241, opened 2019-03-14
base url support


yamajik:
I wanne run behind proxy

Multishifties:
@yamajik I think there are already docs on how to run it behind a proxy https://github.com/codercom/code-server/blob/master/doc/self-hosted/index.md#user-content-nginx-reverse-proxy

yamajik:
I am managing many code servers behind a proxy.
Each server serves with baseurl like “/code-server/:server_id/“ .
But, some requests still do from root.

yamajik:
I wanne config for global baseurl setting

tiffany352:
I just found code-server, and I want to give it a try, but without base url support it looks like I'll have to setup a separate subdomain for it on my website, which is pretty inconvenient.

sr229:
@tiffany352 @yamajik have you tried using Docker with different IPs to see if it works as a temporary workaround? This is possible by creating a new network.

mgcaret:
I also request this.  It's a non-starter if I can't use it at another HTTP path, instead of /.

@sr229 that won't help.  The issue is that you cannot set it up to be at http://myserver.foo/vscode/, rather it must be at http//myserver.foo/

sgama:
I also would like this functionality to be added in, preferably as an environment variable such as `BASE_URL`

LetsEncrypt limits the number of SSL certs for subdomains on one host, thus taking up another subdomain is not preferable.

tiffany352:
@sgama Off topic, but I'm kind of confused about your LE comment, I don't see any kind of limit other than (large) rate limits on certificate issuing? There's also wildcard certificates available now.

I ended up just making a new subdomain to run code-server on, and it's working fine.

sgama:
@tiffany352, oh right. Wildcards are available now. Thanks for reminding me.

kylecarbs:
This _should_ be fixed. The socket should now be created at the current page path.

yamajik:
Not sure.

resourse urls
webview urls
...

Both incorrect.

I have fixed some in my fork.

yamajik:
reopen this, plz.

yamajik:
I will pr my fix later which may not fix all.

chylvina:
+1

kylecarbs:
We'll have to remove the `/resource` and replace it with `resource/` to use subpaths.

patrickstjohn:
+1 this will be a big help for us

patrickstjohn:
[...]

PR #307, opened 2019-03-21
Feature/relative url

yamajik:
Issue: #241


yamajik:
Problems still exist while using relative url ("resource").
location.pathname + relative_url will fix it.

umdstu:
@kylecarbs - I know this is only a week old, but do you have an estimate for when you'll be able to review this PR?

Thanks!

au5ton:
polite ping

packages/ide/src/fill/electron.ts
```diff
@@ -47,7 +47,8 @@ const newCreateElement = <K extends keyof HTMLElementTagNameMap>(tagName: K): HT
 			},
 			set: (value: string): void => {
 				if (value) {
-					value = value.replace(/file:\/\//g, "/resource");
+					const resourceBaseUrl = location.pathname.replace(/\/$/, "") + "/resource";
```
kylecarbs: Wouldn't this only catch URLs that end with `/`?

Something like `https://ide.kwc.io/banana` wouldn't be caught with this.

yamajik: This makes sure the url is correct.

Both 
`https://ide.kwc.io/banana`
`https://ide.kwc.io/banana/`
use
`https://ide.kwc.io/banana/resource`
to request the resources

ShannonMasters:
any chance of getting this merged in for the next release?  this is a pretty big issue for how my environment is atm (of which I have limited control)

PR #317, opened 2019-03-22
Ensure workspace configPath is a valid URI object

foresthoffman:
Adding a folder to a workspace results in a failure to reload the workspace. This was due to the workspace URI being invalid. The scheme (https://tools.ietf.org/html/rfc3986#section-3.1) for the URI was actually the path to the workspace configuration file, so the URI would always fail to parse upon workspace initialization.

This PR will ensure that the configuration path is actually defined as a valid `file://` URI, instead of simply casting it.

Fixes https://github.com/codercom/code-server/issues/309

### DIFF
--- a/packages/vscode/src/workbench.ts
+++ b/packages/vscode/src/workbench.ts
@@ -185,7 +185,13 @@ export class Workbench {
 			_: [],
 		};
 		if ((workspace as IWorkspaceIdentifier).configPath) {
-			config.workspace = workspace as IWorkspaceIdentifier;
+			// tslint:disable-next-line:no-any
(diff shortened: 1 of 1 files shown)


broady:
Can we get a version of this released?

(or is there some easy way for me to easily run a copy of code-server at a particular commit, in a similar fashion to the single binary distro? sorry, a bit of a Node.js noob here)

nol166:
You can clone the repo and then:

1. Cd into `code-server`
2. Run `yarn`
3. Run `yarn task build:binary:linux` or darwin depending on what OS you are using

It should output the CLI to `packages/server/cli-(os)`

PR #342, opened 2019-03-26
Remove local machine reference

Multishifties:
There's a reference to Kyle's local machine. I've replaced this.

Nope!


kylecarbs:
Fixed in another branch!

PR #344, opened 2019-03-27
Chrome e2e tests

foresthoffman:
- [x] `--port` flag
- [x] `--password` flag
- [x] `--allow-http` flag
- [x] `--no-auth` flag
- [x] Can load in successfully.
- [x] Can open file
- [x] Can create file
- [x] Can delete file
- [x] Can install extension
- [x] Can uninstall extension
- [x] Can debug JavaScript file

Fixes https://github.com/codercom/code-server/issues/339


packages/app/chrome/test/chrome.test.ts
```diff
@@ -0,0 +1,34 @@
+import { resolve } from "path";
```
kylecarbs: `packages/app/chrome` will be disappearing soon. Might be nice to have this in `test/*`

packages/app/chrome/test/chrome.test.ts
```diff
@@ -0,0 +1,34 @@
...
+	beforeEach(async () => {
+		page = await browser.newPage();
+	});
+	afterAll(async () => {
+		await browser.close();
+		server.dispose();
+		await server.killProcesses();
+	});
+
+	server.test("should open IDE", async () => {
```
kylecarbs: I'd use normal `it` here if possible. Will assist in readability.

packages/app/chrome/test/chrome.test.ts
```diff
@@ -0,0 +1,34 @@
...
+		// page.once("load", async () => {
+		// 	await page.goto(server.url);
+		// 	await page.screenshot({ path: resolve(__dirname, "./screenshot-test.jpg"), fullPage: true });
+		// });
+
+		await page.waitFor(4000);
```
kylecarbs: I'd create a general function for loading the IDE page.

packages/app/chrome/test/chrome.test.ts
foresthoffman: `packages/app/test`?

kylecarbs: Nah just `test/`

kylecarbs: Like at the root of the repo. These are tests for the overall build of the repo afterall.

packages/app/chrome/test/chrome.test.ts
foresthoffman: My thought process is that:
```
server.test("should do a thing", async () => {
    await doThing();
}, 1000);
```
Is prettier than:
```
it("should do a thing", async () => {
    try {
        await doThing();
    } catch (e) {
        // make sure the server closes itself, if the thing doesn't work
        await this.dispose();
        throw e;
    }
}, 1000);
```

code-asher: Could you dispose the server in `afterEach`?

foresthoffman: Yes, that would. I've made changes so that's no longer necessary. The tests are using `it` now.

code-asher:
Just had a thought that it might be good to run the tests as Mac, Windows, and Linux clients. Those values are provided by the user agent, could possibly use that? Or import and override the values provided here: https://github.com/microsoft/vscode/blob//src/vs/base/browser/browser.ts#L126

foresthoffman:
@code-asher It is possible to set the user agent for a page with puppeteer: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetuseragentuseragent. I think it uses the base platform to determine the user agent.

PR #362, opened 2019-03-27
Check version

### LINKED ISSUE #213: Add release number in CLI, tell you when to update.
It would be neat if the release number was added to the CLI and a way to check for the latest release and tell you if you should update.

nol166:
Checks to see if there is a newer release available when running the CLI

Fixes #213.

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -166,10 +166,21 @@ const bold = (text: string | number): string | number => {
 		}
 	}
 
-	if (process.env.VERSION !== "development") {
-		await getRecentRelease().then((release) => console.log(release));
-	}
+	let newestVersion = "";
+	// TODO: remove bang
+	if (!process.env.VERSION) {
+		let currentVersion = process.env.VERSION;
+		const recentRelease = await getRecentRelease();
 
+		// TODO: replace string with currentVersion
+		if (compareVersions("1.31.0-20", recentRelease)! <= 1) {
+			newestVersion = recentRelease;
+		} else {
+			newestVersion = "1.31.0-20";
+		}
+
+	}
+	logger.info(`Latest version is \u001B[1m${newestVersion}`);
 	logger.info(`\u001B[1mcode-server ${process.env.VERSION ? `v${process.env.VERSION}` : "development"}`);
 
 	if (options.dataDir) {
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -177,10 +177,11 @@ const bold = (text: string | number): string | number => {
 			newestVersion = currentVersion;
 		}
 	}
-	if (newestVersion === process.env.VERSION){
+	if (newestVersion === process.env.VERSION) {
 		logger.info(`You are using the latest version: ${newestVersion}`);
+	} else {
+		logger.info(`New version (\u001B[1m${newestVersion}) available at https://github.com/codercom/code-server/releases`);
 	}
-	logger.info(`Latest version is \u001B[1m${newestVersion}`);
 	logger.info(`\u001B[1mcode-server ${process.env.VERSION ? `v${process.env.VERSION}` : "development"}`);
 
 	if (options.dataDir) {
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -180,7 +180,7 @@ const bold = (text: string | number): string | number => {
 	if (newestVersion === process.env.VERSION || newestVersion === "") {
 		logger.info("You are using the latest version");
 	} else {
-		logger.info(`New version (\u001B[1m${newestVersion}) available at https://github.com/codercom/code-server/releases`);
+		logger.info(`New version (\u001B[1m${newestVersion}) available https://github.com/codercom/code-server/releases`);
 	}
 	logger.info(`\u001B[1mcode-server ${process.env.VERSION ? `v${process.env.VERSION}` : "development"}`);
 
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -167,7 +167,7 @@ const bold = (text: string | number): string | number => {
 	}
 
 	if (isCli) {
-		let currentVersion = process.env.VERSION;
+		const currentVersion = process.env.VERSION;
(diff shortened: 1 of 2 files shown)


packages/server/src/updater.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const getRecentRelease = (): Promise<any> => {
+	return new Promise<string>((resolve, rej) => {
+		const options = {
+			host: "api.github.com",
+			path: "/repos/codercom/code-server/releases",
+			method: "GET",
+			agent: false,
+			headers: {
+				"User-Agent": "test",
```
NGTmeaty: We'll want to follow [Github's user-agent requirements](https://developer.github.com/v3/#user-agent-required) and change it to code-server.

packages/server/src/updater.ts
```diff
@@ -0,0 +1,56 @@
...
+			return -1;
+		}
+	}
+};
+
+export const getRecentRelease = (): Promise<any> => {
+	return new Promise<string>((resolve, rej) => {
+		const options = {
+			host: "api.github.com",
+			path: "/repos/codercom/code-server/releases",
```
NGTmeaty: Looking at the API again, we could possibly [call /latest](https://developer.github.com/v3/repos/releases/#get-the-latest-release) and get just compare those names?

NGTmeaty:
The Docker version we compile doesn't get a version number. Might be a different issue, but it'd be nice to get version checking in there too :)

nol166:
> The Docker version we compile doesn't get a version number. Might be a different issue, but it'd be nice to get version checking in there too :)
> 

The docker image eventually inherits the source code from the binary so as soon as the CI catches up (if and when) this is merged it should work for the docker image as well.

packages/server/src/updater.ts
```diff
@@ -0,0 +1,50 @@
...
+			method: "GET",
+			agent: false,
+			headers: {
+				"User-Agent": "code-server",
+			},
+		};
+
+		https.get(options, function (res) {
+			if (res.statusCode !== 200) {
+				throw new Error("Failed to acquire release information");
```
lsmoura: You're inside a Promise. Maybe you should let the caller knows by rejecting it?

```suggestion
				rej(new Error("Failed to acquire release information"));
```

packages/server/src/cli.ts
```diff
@@ -139,6 +140,22 @@ if (isCli) {
 		}
 	}
 
+	let newestVersion = "";
+	if (process.env.VERSION) {
+		let currentVersion = process.env.VERSION;
+		const recentRelease = await getRecentRelease();
+
+		if (compareVersions(currentVersion, recentRelease)! <= 1) {
```
kylecarbs: No need to compare versions. If the latest version !== the current version then an update is available.

packages/server/src/cli.ts
```diff
@@ -139,6 +140,22 @@ if (isCli) {
 		}
 	}
 
+	let newestVersion = "";
```
kylecarbs: Would be cleaner to do something like:
```
if (isCli) {
    const currentVersion = process.env.VERSION;
    if (!currentVersion) {
        return;
    }
    const recentRelease = await getRecentRelease();
    ...
}
```

packages/server/src/cli.ts
```diff
@@ -139,6 +140,22 @@ if (isCli) {
...
+		const recentRelease = await getRecentRelease();
+
+		if (compareVersions(currentVersion, recentRelease)! <= 1) {
+			newestVersion = recentRelease;
+		} else {
+			newestVersion = currentVersion;
+		}
+	}
+	if (newestVersion === process.env.VERSION || newestVersion === "") {
+		logger.info("You are using the latest version");
```
kylecarbs: Shouldn't output anything if latest imo

packages/server/src/updater.ts
```diff
@@ -0,0 +1,50 @@
...
+			return 1;
+		}
+		if (aVal < bVal || isNaN(aVal)) {
+			return -1;
+		}
+	}
+};
+
+export const getRecentRelease = (): Promise<any> => {
+	return new Promise<string>((resolve, rej):void  => {
```
kylecarbs: Should spell out `reject` fully since we spelt out `resolve` fully

packages/server/src/updater.ts
```diff
@@ -0,0 +1,50 @@
...
+		const options = {
```
kylecarbs: Should throw this directly into the `https.get`

packages/server/src/cli.ts
```diff
@@ -139,6 +140,22 @@ if (isCli) {
 		}
 	}
 
+	let newestVersion = "";
+	if (process.env.VERSION) {
+		let currentVersion = process.env.VERSION;
+		const recentRelease = await getRecentRelease();
```
kylecarbs: Needs to handle if this is rejected

packages/server/src/cli.ts
```diff
@@ -139,8 +140,19 @@ if (isCli) {
 		}
 	}
 
-	logger.info(`\u001B[1mcode-server ${process.env.VERSION ? `v${process.env.VERSION}` : "development"}`);
+	if (isCli) {
+		let currentVersion = process.env.VERSION;
```
lsmoura: This value is never reassigned. Use `const` instead?

```suggestion
		const currentVersion = process.env.VERSION;
```

packages/server/src/updater.ts
```diff
@@ -0,0 +1,28 @@
...
+			if (res.statusCode !== 200) {
+				reject(Error("Failed to acquire release information"));
+			}
+			let body = "";
+			res.on("data", (chunk) => {
+				body += chunk;
+			});
+
+			res.on("end", () => {
+				let release = JSON.parse(body);
```
lsmoura: This value is never reassigned. use `const` instead?

```suggestion
				const release = JSON.parse(body);
```

packages/server/src/cli.ts
nol166: made all these changes

nol166:
resolved conflicts

PR #363, opened 2019-03-27
Handle disconnects

code-asher:
We already reconnect the web socket when it drops but we also need to restart/reconnect services that started before the disconnection.

This PR will do that for the following:
- [X] Terminal instances (just spawn a new instance for now)
- [x] Extension host (already supported, just make it automatic)
- [x] Watcher (already supported, but need to use our retry for it)
- [x] Searcher
- [x] spdlog instances

Won't be in this PR:
- Shared process

Other:
- [x] Show message when the shared process dies explaining what won't work while it's dead and to restart/refresh the page for now if you need any of that.

### DIFF
--- a/packages/protocol/src/browser/client.ts
+++ b/packages/protocol/src/browser/client.ts
@@ -143,6 +143,7 @@ export class Client {
 			clearTimeout(this.pingTimeout as any);
 			this.pingTimeout = undefined;
 			handleDisconnect();
+			this.dispose();
 		});
 		connection.onUp(() => this.disconnected = false);
 
@@ -158,6 +159,12 @@ export class Client {
 	 */
 	public dispose(): void {
 		this.connection.close();
+		this.proxies.clear();
+		this.successEmitter.dispose();
+		this.failEmitter.dispose();
+		this.eventEmitter.dispose();
+		this.initDataEmitter.dispose();
+		this.sharedProcessActiveEmitter.dispose();
 	}
 
 	public get initData(): Promise<InitData> {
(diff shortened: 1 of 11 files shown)


packages/protocol/src/browser/client.ts
```diff
@@ -173,9 +174,18 @@ export class Client {
 	/**
 	 * Make a remote call for a proxy's method using proto.
 	 */
-	private remoteCall(proxyId: number | Module, method: string, args: any[]): Promise<any> {
-		if (this.disconnected) {
-			return Promise.reject(new Error("disconnected"));
+	private async remoteCall(proxyId: number | Module, method: string, args: any[]): Promise<any> {
```
lsmoura: There's no `await` inside the function, and every return in this function is already a promise. It's better for either this function *not* to be `async` or return/throw regular values.

You don't need to return promises from `async` functions just for the sake of the caller getting a promise since every return will eventually be converted to a promise.

```suggestion
	private remoteCall(proxyId: number | Module, method: string, args: any[]): Promise<any> {
```

Let me know if you need more clarification on the subject. I'll be happy to expand.

packages/protocol/src/node/server.ts
```diff
@@ -331,9 +331,9 @@ export class Server {
 		return stringify(value, undefined, (p) => this.storeProxy(p));
 	}
 
-	private getProxy(proxyId: number | Module): ProxyData {
+	private getProxy(proxyId: number | Module, method: string): ProxyData {
```
lsmoura: Should we really pass a value for this function, even if it's only going to be used as a message on the error?

packages/protocol/src/browser/client.ts
code-asher: Nice catch. :+1: My best guess is that I added an `await` for something then removed it.

I also managed to forget to send the promise through to `this.parse`. Must have looked strange to see the promise put into a variable only to immediately return that variable.

packages/protocol/src/node/server.ts
code-asher: Yeah, it would have been better to just use `trace` logging to figure out why that error was occurring. Will revert.

Issue #286, opened 2019-03-19
Setting password with command line parameter makes it visible in ps aux


kykc:
- `code-server` version: 1.32.0-275-linux-x64
- OS Version: Ubuntu 18.04
## Description

Setting password with command line parameter makes it visible in `ps aux`

## Steps to Reproduce
```
./code-server --password=supersecret $CODE_SERVER_ROOT
ps aux
```

## Possible solution

Seems that there's some possible ways to hide it mentioned [here](http://broken-by.me/node-js-change-visible-process-name/)

coadler:
cc @kylecarbs @code-asher

kykc:
Maybe it's also a good idea not to output password into code-server stdout when it's been explicitly set by startup parameter. Should I create a separate issue for that?

NGTmeaty:
We can probably keep it in here because it's related.

Multishifties:
UPDATE: I have a PR for this https://github.com/codercom/code-server/pull/365

code-asher:
The recommended way now to set a password is by using the `PASSWORD` environment variable, which will prevent the password from displaying in the process listing.

https://github.com/codercom/code-server/commit/a3ac4567e39133e736e7e74365f10e9b97861621 also makes it so the password is only logged when it was generated.

PR #365, opened 2019-03-27
Fix/issue #286

Multishifties:
Problem: CLI Password argument is visible in `ps -ax`

Issue: https://github.com/codercom/code-server/issues/286

Changes: Adds script to change the process display name.

Test Environment:
(Result: Pass! Compile was successful and everything functions as expected)


packages/server/src/cli.ts
```diff
@@ -166,6 +166,20 @@ if (isCli) {
...
+		const parts = [process.title];
+		parts.push(path.relative(process.cwd(), process.argv[1]));
+		for (let i = 2; i < process.argv.length; i++) {
+			if (process.argv[i].includes("--password")) {
+				parts.push(process.argv[i].replace(/=(.*)/, "=****"));
+			} else {
+				parts.push(process.argv[i]);
+			}
+		}
+		process.title = parts.join().replace(/\,/g, " ");
```
lsmoura: This will look bad if any of the parameters have a comma in it. It's also sub-optimal. Why not provide a parameter for the `join()` function?

```suggestion
		process.title = parts.join(" ");
```

Multishifties: Good call!

code-asher:
We should handle `code-server --password secretpassword` (no `=`) as well. `code-server -P` is also supposed to work I believe.

lsmoura:
@code-asher is right. According to `code-server --help`, this is the format:

```
-P, --password <value>   Specify a password for authentication.
```

Multishifties:
Suggested changes have been made! Now works with or without `=`;

I did notice that `-P` as a CLI argument does not appear to be working correctly. But that can be handled in another issue and another PR.

packages/server/src/cli.ts
```diff
@@ -171,8 +171,14 @@ if (isCli) {
 		const parts = [process.title];
 		parts.push(path.relative(process.cwd(), process.argv[1]));
 		for (let i = 2; i < process.argv.length; i++) {
-			if (process.argv[i].includes("--password")) {
-				parts.push(process.argv[i].replace(/=(.*)/, "=****"));
+			if (process.argv[i].includes("--password") || process.argv[i].includes("-P")) {
```
lsmoura: I would change this... remove the `includes` and use something for equality. (`===`, maybe?).

If we somehow code-server is invoked by something like `code-server --foo "i-am-a-silly--password-usecase" --password "secret"`, we are going to have problems.

Multishifties: True, that could happen. If we were to go the `===` route, we'd need to handle 4 different cases.

`--password`
`--password=`
`-P`
`-P=`

I figured the code looked and read cleaner in it's current state. And I figured the edge case of someone using `--password` in their filepath or other CLI args would be pretty rare and unlikely. And on that topic both approaches would break if a user wanted to run code-server in a directory named `--password`

lsmoura: I agree. Let's leave the final word with the project owners then! 😉

Multishifties: @code-asher 🙂

code-asher: Personally I'd prefer we properly handle all cases. It wouldn't even really be much more code, would it? 

Omitting the `-P` check for brevity:

```
if (process.argv[i].startsWith("--password=")) {
  parts.push(process.argv[i].replace(/=.*/, "=****");
} else if (process.argv[i] === "--password")) {
  parts.push(process.argv[i++], "****")
} else if (process.argv[i] === "--") {
  parts.push(...process.argv.slice(i));
  break;
} else {
  parts.push(process.argv[i]);
}
```

I think that might handle everything:
```
code-server --password=password            # =
code-server --password password            # space
code-server --password password My-Project # directory that contains the flag
code-server -- --password=                 # directory that is the same as or starts with the flag
code-server --password-max-tries           # a new flag we create that includes part of the password flag
```

Multishifties: Ah! `--password-max-tries` is an example use case I hadn't even considered. Strict equality checks in this context would indeed make much more sense!

packages/server/src/cli.ts
```diff
@@ -166,6 +166,26 @@ if (isCli) {
 		password = buffer.toString("hex");
 	}
 
+	// If CLI password was provided, obsfucate password from process title
+	if (options.password) {
+		const parts = [process.title];
+		parts.push(path.relative(process.cwd(), process.argv[1]));
```
code-asher: When running a script with `node` this is correct but when running the binary this won't normally be present in the process listing, right? I think you'll get something like `./code-server out/cli.js --password ****` instead of just `./code-server --password ****`. (It'll also be truncated since the title cannot increase in length.)

packages/server/src/cli.ts
code-asher: Another note, as far as I remember `commander` requires the arguments to be exact, including capitalization, so checking against that isn't necessary (or desired).

packages/server/src/cli.ts
```diff
@@ -166,6 +166,24 @@ if (isCli) {
 		password = buffer.toString("hex");
 	}
 
+	// If CLI password was provided, obsfucate password from process title
+	if (options.password) {
+		const parts = [process.title];
+		for (let i = 2; i < process.argv.length; i++) {
+			if (process.argv[i].startsWith("--password=")) {
+				parts.push(process.argv[i].replace(/=.*/, "=****"));
+			} else if (process.argv[i] === "--password") {
```
code-asher: My bad, I didn't include `-P` in my example but we'll need that as well. No need to check for `-P=` though since `=` isn't used with short options.

packages/server/src/cli.ts
```diff
@@ -166,6 +166,24 @@ if (isCli) {
...
+				parts.push(process.argv[i++], "****")
+			} else if (process.argv[i] === "--") {
```
code-asher: Another unfortunate edge case comes to mind:

```
code-server --data-dir -- password pass
```

That is, `--` is only interpreted as the end of options if it isn't the argument for another option.

code-asher: It's possible we should use an environment variable and/or configuration file instead of a command line argument for the password. Or possibly a prompt, although I'm not really keen on that solution.

kylecarbs:
We should probably use an environment variable and deprecate the usage via CLI.

Multishifties:
Using an ENV password will make this PR obsolete. I'll go ahead and close this PR.

PR #379, opened 2019-03-29
Add description for using code-server as a systemd unit

nwtnsqrd:

I needed this extra piece of information and had to get it from another source. Might be convenient to include it here.

No.

### DIFF
--- a/doc/self-hosted/index.md
+++ b/doc/self-hosted/index.md
@@ -115,5 +115,29 @@ OPTIONS
   ```
   *Important:* For more details about Apache reverse proxy configuration checkout the [documentation](https://httpd.apache.org/docs/current/mod/mod_proxy.html) - especially the [Securing your Server](https://httpd.apache.org/docs/current/mod/mod_proxy.html#access) section
   
+  ### Start code-server as systemd unit
+  Create `/lib/systemd/system/code-server.service` with the following contents:
+  ```
+  [Unit]
+  Description=Code Server IDE
+  After=network.target
+
+  [Service]
+  Type=simple
+  User=<USER>
+  Restart=on-failure
+  RestartSec=10
+
+  ExecStart=<PATH TO BINARY> --allow-http
+
+  StandardOutput=file:/var/log/code-server-output.log
+  StandardError=file:/var/log/code-server-error.log
+
+  [Install]
+  WantedBy=multi-user.target
+
+  ```
+  Enable code-server with `systemctl enable code-server.service`. Start with `systemctl start code-server.service`.
+  
   ### Help
   Use `code-server -h` or `code-server --help` to view the usage for the cli. This is also shown at the beginning of this section. 
--- a/doc/self-hosted/index.md
+++ b/doc/self-hosted/index.md
@@ -125,6 +125,7 @@ OPTIONS
   [Service]
   Type=simple
   User=<USER>
+  EnvironmentFile=$HOME/.profile
   Restart=on-failure
   RestartSec=10
 


doc/self-hosted/index.md
```diff
@@ -115,5 +115,29 @@ OPTIONS
...
+  ### Start code-server as systemd unit
+  Create `/lib/systemd/system/code-server.service` with the following contents:
+  ```
+  [Unit]
+  Description=Code Server IDE
+  After=network.target
+
+  [Service]
+  Type=simple
+  User=<USER>
```
sr229: Please also set environment file. There is a instance the User environment is not being used by `systemd`.

nwtnsqrd: Please suggest changes as I am not aware of the issue.

sr229:
```suggestion
  User=<USER>
  EnvironmentFile=$HOME/.profile
```

sr229: For reference, see my [systemd Unit file](https://gist.github.com/sr229/9600b26d9eada8133a8587ce05d846a9)

kylecarbs:
Ping @nwtnsqrd

nwtnsqrd:
I have updated the PR

doc/self-hosted/index.md
```diff
@@ -115,5 +115,30 @@ OPTIONS
...
+  EnvironmentFile=$HOME/.profile
+  Restart=on-failure
+  RestartSec=10
+
+  ExecStart=<PATH TO BINARY> --allow-http
+
+  StandardOutput=file:/var/log/code-server-output.log
```
fbartels: On Ubuntu 16.04 the Standard* options were not recognized. I ended up with the following systemd unit (I anyways prefer logging to go to journald):

```
[Unit]
Description=VSCode in a browser
After=network.target

[Service]
Type=simple
User=myuser
EnvironmentFile=$HOME/.profile
Restart=on-failure
RestartSec=10

ExecStart=/usr/local/bin/code-server --allow-http --port 1234 --no-auth --disable-telemetry $HOME/dev-workspace

ExecStop=/sbin/start-stop-daemon --stop -x /usr/local/bin/code-server

[Install]
WantedBy=multi-user.target
```

sr229: Have you tested this on other distros? This should work on all that uses systemd not just one

sr229: I think a more viable alternative is to log this to journald if needed.

nodesocket:
This is a very useful doc PR. Having a peer reviewed systemd config file is 💰.

My only comment is to add `--disable-telemetry` to `ExecStart`.

doc/self-hosted/index.md
frank-dspeed: @sr229 The Environment file can't work it has wrong formart systemd expects a other formart it will not execute that file with if else statements.

```
VAR=VALUE
```

is the only thing that is accepted in it!

doc/self-hosted/index.md
nhooyr: We should always be logging to journald.

nhooyr:
@sr229 Can you take this over with the one modification to log to journald? Just removing the log file line should fix it. Make sure you include @nwtnsqrd's commit in your PR so we can give him attribution.

sr229:
@nhooyr We already have it on `quickstart.md`, I'll Co-Author them while adding their changes.

nhooyr:
Sounds good.

PR #380, opened 2019-03-29
Removed files that are in .gitignore but still in the tree

MOZGIII:
I'm doing the packing for debian, and I found out there are somes that are in the `.gitignore` (`git clean -Xdf` removes them) but they're in the tree. It cases some problems for the way I'm trying to package things.
Those files were probably added by mistake, so this PR removes them.

Not that I'm aware of.

### DIFF
--- a/packages/requirefs/test/lib/chained-1.js
+++ b/packages/requirefs/test/lib/chained-1.js
@@ -1,1 +1,0 @@
-exports = require("./chained-2");
--- a/packages/requirefs/test/lib/chained-2.js
+++ b/packages/requirefs/test/lib/chained-2.js
@@ -1,1 +1,0 @@
-exports = require("./chained-3");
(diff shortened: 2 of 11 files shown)


code-asher:
These are part of the `requirefs` tests and shouldn't be removed. We should probably un-ignore the `lib` folder there. We do that for `lib/node_modules` but not `lib`: https://github.com/codercom/code-server/blob/master/packages/requirefs/test/.gitignore#L1

We could also rename `lib` to something else.

MOZGIII:
I'm closing this PR then. At which level do you prefer them ignored?

code-asher:
I think it'd actually be better to rename the directory.

MOZGIII:
Is the pattern in this repo to ignore all the `lib` dirs? Or maybe it should be `/lib` instead of `lib` at the top level?

code-asher:
Ah, that's a good point, we could just ignore `/lib` at the top level. That sounds good to me.

PR #393, opened 2019-04-01
Add getting started for contributing to README.md

andreimc:

Just some docs for getting started

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,25 @@ How to [secure your setup](/doc/security/ssl.md).
 
 ## Contributing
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,7 +69,25 @@ How to [secure your setup](/doc/security/ssl.md).
 
 ## Contributing
 
-Development guides are coming soon.
+### Getting started
```
kylecarbs: We already have a `Getting Started` guide, so this could be confusing. Where do you think this should be placed @nol166 ?

PR #402, opened 2019-04-02
Convert fully to protobuf (was partially JSON)

### LINKED ISSUE #390: Improve protocol serialization
Currently transmitted data is sent/received as JSON.

Converting most of the logic to use protobufs directly should result in a large performance bump.

code-asher:
This removes all the stringifying and parsing of JSON and uses protobuf directly for everything.

Still need to run some benchmarks.

Closes #390.


code-asher:
For a buffer with 10 million bytes, full-proto serializes/deserializes in 396ms. Using partial JSON as we had before takes 3021.3ms. 

Got similar results writing ten 10MB files (421ms vs 3442ms).

For writing 10k tiny 1k files full-proto was about 50% faster, which makes sense since this change mostly benefits serializing and deserializing buffers.

code-asher:
https://plot.ly/~code-asher/6

The takeaway:
- Using 100% proto is much improved over using JSON inside of proto.
- JSON-only is actually a bit faster for buffers < 6k in size.
- Above 6k proto beats out JSON.
- Base64 over JSON is actually the fastest.

Notes:
- Each result is an average over 100 runs.
- Tests were all ran in Node, browser results could be different.
- This only tests the time it takes to serialize and then immediately deserialize (or stringify/parse/encode/decode as the case may be).

Issue #383, opened 2019-03-30
Use a tinyinit in the dockerfile


nhooyr:
Will make ctrl+c work.

sr229:
How exactly does this affect us in the container-level?

Issue #361, opened 2019-03-27
docker one-liner should use the -i flag


nhooyr:
Otherwise signals aren't passed correctly and the container keeps running in the bg.

Also, we need to make code-server handle signals correctly.

PR #407, opened 2019-04-03
Use an init system

RichardMcSorley:

While using docker container the user can not send POSIX commands like SIGKILL (CTR+C) or SIGTERM to code-server, causing the server to continue running in the background. Best practice in this case is to run an init system.

If a users wants to run the application in the background they should use docker run's `-d` flag
```bash
docker run -d codercom/code-server code-server --allow-http --no-auth
```
[dumb-init documentation](https://github.com/Yelp/dumb-init)

#361 
#383

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,10 +25,11 @@ RUN apt-get update && apt-get install -y \
 	openssl \
 	net-tools \
 	git \
-	locales
+	locales \
+	dumb-init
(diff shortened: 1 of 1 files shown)


RichardMcSorley:
@nhooyr Great work guys. I love this :heart:

nhooyr:
I tested this change locally, it's good @kylecarbs

PR #411, opened 2019-04-03
Add serviceworker + Monaco Service Worker

andreimc:
https://github.com/codercom/code-server/pull/154 original PR, rebased and fixed conflicts

Also added monaco webpack plugin which enables monaco to operate via the service worker.

https://github.com/codercom/code-server/issues/375


creativeguy2013:
Does this add web workers or does it add service workers? AFAIK Monaco uses web workers rather than service workers, because Monaco doesn't have to do background processing, rather it does parallel processing. If this is the case this should not be merged together with the "Service worker and PWA" PR as it is not related.

You might be confused by Monaco's "language server service" workers, which are not service workers but rather just regular web workers which are running a "language server service". (https://www.quora.com/Whats-the-difference-between-service-workers-and-web-workers-in-JavaScript)

andreimc:
@creativeguy2013 you are  right 👍

PR #416, opened 2019-04-03
Improve CI caching

kylecarbs:
Reduces build time from ~20m to ~5m. 

Stores vscode source and built default-extensions within object-storage.


packages/server/src/cli.ts
```diff
@@ -47,6 +49,8 @@ if (isCli) {
 		readonly host: string;
 		readonly port: number;
 
+		readonly extensionsDir?: string;
+		readonly userDataDir?: string;
```
code-asher: `extensionsDir` and `userDataDir` are unused?

build/tasks.ts
```diff
@@ -255,4 +202,24 @@ register("package", async (runner, releaseTag) => {
 		: runner.execute("zip", ["-r", `${archiveName}.zip`, `${archiveName}`]);
 });
 
+/**
+ * If we're in the CI and the VS Code version is the same,
+ * then we don't need to rebuild.
+ */
+const ifCiAndVsc = (vscName: "vscode" | "vscode-default-extensions"): boolean => {
```
code-asher: Should this be used somewhere?

build/tasks.ts
```diff
@@ -129,97 +130,43 @@ const buildWeb = register("build:web", async (runner) => {
...
+		});
+	});
 });
 
 const ensurePatched = register("vscode:patch", async (runner) => {
 	if (!fs.existsSync(vscodePath)) {
 		throw new Error("vscode must be cloned to patch");
 	}
-	await ensureClean();
+	await ensureInstalled();
```
code-asher: Should we still keep the `git reset` that was inside `ensureClean`? Otherwise trying to build the binary again fails because the patch won't apply. Or maybe if the patch fails we log a notice and keep going anyway. That might be better so we can build the binary locally for testing with custom changes that we haven't put into the patch yet.

packages/server/src/cli.ts
kylecarbs: Oh right. Going to add this in another PR.

build/tasks.ts
kylecarbs: Meant for the old CI stuff. Will remove!

build/tasks.ts
kylecarbs: Ah I agree. Will add.

Issue #408, opened 2019-04-03
Error events.js:183


juanmaioli:
- `code-server` version: 1.408-vsc1.32.0
- OS Version:  Linux server.com.ar 4.9.0-8-amd64 #1 SMP Debian 4.9.144-3.1 (2019-02-19) x86_64 GNU/Linux

## Description
On start fail with this error 

juan@server:~/code-server1.408-vsc1.32.0-linux-x64 $ ./code-server 
INFO  code-server v1.408-vsc1.32.0
INFO  Additional documentation: http://github.com/codercom/code-server
INFO  Initializing {"data-dir":"/home/juan/.local/share/code-server","working-dir":"/home/juan/code-server1.408-vsc1.32.0-linux-x64","log-dir":"/home/juan/.cache/code-server/logs/20190403015055217"}
INFO  Starting webserver... {"host":"0.0.0.0","port":8443}
WARN  No certificate specified. This could be insecure.
WARN  Documentation on securing your setup: https://coder.com/docs
INFO   
INFO  Password: da72d89827c89498bc27a815
INFO   
INFO  Started (click the link below to open):
INFO  http://localhost:8443/
INFO   
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE 0.0.0.0:8443
    at u.setupListenHandle [as _listen2] (net.js:1360:14)
    at listenInCluster (net.js:1401:12)
    at doListen (net.js:1510:7)
    at _combinedTickCallback (internal/process/next_tick.js:142:11)
    at process._tickCallback (internal/process/next_tick.js:181:9)

## Steps to Reproduce

only execute code-server on /bin directory

nol166:
@juanmaioli Looks like your port is in use.

To kill the process that is using it: 
`kill $(lsof -t -i :8443)`

kylecarbs:
Seems to be fixed!

PR #418, opened 2019-04-03
Add port in use message

nol166:
Add better error handling and informative error message for when a port is in use

#408


packages/server/src/cli.ts
```diff
@@ -234,7 +238,12 @@ if (isCli) {
 			logger.info(`WebSocket closed \u001B[0m${req.url}`, field("client", id), field("code", code));
 		});
 	});
-
+	app.wss.on("error", (err: {code: string}) => {
```
code-asher: It might be better to use the `NodeJS.ErrnoException` type.

PR #420, opened 2019-04-03
Add flags for customizing user data dir and extensions dir

kylecarbs:
Mirrors vscode CLI options `--user-data-dir` and `--extensions-dir`.

Allows inheriting local VS Code configuration (eg. `--user-data-dir ~/.config/User --extensions-dir ~/.vscode/extensions`).


kylecarbs:
@ammario Going to add a section about compatibility with VS Code in another PR

Issue #440, opened 2019-04-05
127.0.0.1 by default vs 0.0.0.0


nol166:
- `code-server` version: v1.604-vsc1.32.0
- OS Version: macOS 10.14.4 (18E226)

## Description
It has been discussed that using 127.0.0.1 might be better than using 0.0.0.0 for the default host for security reasons when running code-server. A user should have to explicitly set 0.0.0.0 as the host if they want to use it with the `-h` flag

## Steps to Reproduce
1. Start code server `INFO  Starting webserver... {"host":"0.0.0.0","port":8443}`

Multishifties:
@nol166 is there any record of this discussion that you can link to this issue?

nol166:
@nhooyr Discussed it this morning and asked me to make this issue

coadler:
I think the overwhelming use case of `code-server` would be to access it over the internet, not from the same machine. If this is a security issue why not just remove the `--no-auth` flag from the docker one-liner?

coadler:
I think a better solution would be to just add a warning. Changing the default behavior would most likely break a lot of scripts/deployments

nol166:
@coadler That might be a better solution if it will screw up deployments. Thoughts @nhooyr?

nhooyr:
Yea I agree it would be a bad idea to change the behaviour. I guess what we really is to listen on localhost if the `--no-auth` or `--allow-http` flags are passed as at that point listening on 0.0.0.0 is potentially insecure.

coadler:
That seems sensible to me. I could see this affecting people behind reverse proxies that handle auth/tls upstream though, so we should be sure to document the change on the next release

nol166:
updated the pr

Multishifties:
PR is here for people from the future following this thread https://github.com/codercom/code-server/pull/441

PR #441, opened 2019-04-05
Add warning about 0.0.0.0

nol166:
If 0.0.0.0 is a bad idea, this sets the default to 127.0.0.1

#440


packages/server/src/cli.ts
```diff
@@ -235,8 +235,13 @@ const bold = (text: string | number): string | number => {
 		} : undefined,
 	});
 
-	logger.info("Starting webserver...", field("host", options.host), field("port", options.port));
-	app.server.listen(options.port, options.host);
+	if (options.noAuth || options.allowHttp) {
+		logger.info("Starting webserver...", field("host", "127.0.0.1"), field("port", options.port));
+		app.server.listen(options.port, "127.0.0.1");
```
coadler: Won't this override if the user has specifically set `0.0.0.0`?

coadler: We should also let the user know we've switched to localhost for them

packages/server/src/cli.ts
```diff
@@ -235,8 +235,13 @@ const bold = (text: string | number): string | number => {
 		} : undefined,
 	});
 
-	logger.info("Starting webserver...", field("host", options.host), field("port", options.port));
-	app.server.listen(options.port, options.host);
+	if (options.noAuth || options.allowHttp) {
+		logger.info("Starting webserver...", field("host", "127.0.0.1"), field("port", options.port));
```
kylecarbs: We shouldn't adjust the host automatically I feel. Maybe we just output a warning?

Multishifties: A warning would also be the easiest option to implement, no?

nol166: changed it to just a warning

packages/server/src/cli.ts
```diff
@@ -236,12 +236,10 @@ const bold = (text: string | number): string | number => {
 	});
 
 	if (options.noAuth || options.allowHttp) {
-		logger.info("Starting webserver...", field("host", "127.0.0.1"), field("port", options.port));
-		app.server.listen(options.port, "127.0.0.1");
-	} else {
-		logger.info("Starting webserver...", field("host", options.host), field("port", options.port));
-		app.server.listen(options.port, options.host);
+		logger.warn("0.0.0.0 is accessible to any device on your network");
```
cy91244: I think we should not generate this message when the string `host` is not `0.0.0.0`.

nol166: updated

sr229:
LGTM

packages/server/src/cli.ts
```diff
@@ -235,6 +235,9 @@ const bold = (text: string | number): string | number => {
 		} : undefined,
 	});
 
+	if (options.noAuth && options.host === "0.0.0.0"|| options.allowHttp && options.host === "0.0.0.0") {
+		logger.warn("0.0.0.0 is accessible to any device on your network");
```
sr229: Nit: maybe we can elaborate this more. Here's a suggestion
```suggestion
		logger.warn("0.0.0.0 is accessible to every device on the network your connected to. Make sure you set up authentication to prevent unauthorized access.");
```

packages/server/src/cli.ts
```diff
@@ -235,6 +235,9 @@ const bold = (text: string | number): string | number => {
 		} : undefined,
 	});
 
+	if (options.noAuth && options.host === "0.0.0.0"|| options.allowHttp && options.host === "0.0.0.0") {
```
kylecarbs: Add space between between the string and the pipes `0.0.0.0"||`

nol166: added

code-asher:
Closed in favor of https://github.com/cdr/code-server/pull/857.

Issue #94, opened 2019-03-07
add install script to automatically download binary into /usr/local/bin/code-server


nhooyr:

robertpeteuil:
You might be able to leverage this [installation script](https://github.com/robertpeteuil/terraform-installer) that I created for Terraform.

It can provide a base as it has a lot of error checks, parameter options, and logic for:
- finds the latest release from the GitHub API 
- determines OS and architecture and creates the download URL
- downloading, extracting and copying to binary directory

kylecarbs:
Fixed with [sshcode](https://github.com/codercom/sshcode) :confetti_ball:

PR #450, opened 2019-04-06
Update QuickStart guides to use copy/paste links to lastest version

illiapoplawski:
No problem, just a feature update to follow up on comment submitted in PR [255](https://github.com/codercom/code-server/pull/255).
It uses the api to find the latest release version, downloads it, and extracts it over the old version to update code server to the latest  version of code-server with no changes to the commands required by the user.

No

### DIFF
--- a/doc/admin/install/aws.md
+++ b/doc/admin/install/aws.md
@@ -33,21 +33,17 @@ If you're just starting out, we recommend [installing code-server locally](../..
   >example: `ssh -i "/Users/John/Downloads/TestInstance.pem" ubuntu@ec2-3-45-678-910.compute-1.amazonaws.co`
 - You should see a prompt for your EC2 instance like so<img src="../../assets/aws_ubuntu.png">
 - At this point it is time to download the `code-server` binary. We will of course want the linux version.
-- Find the latest Linux release from this URL:
+- Download the latest Linux release with this command:
   ```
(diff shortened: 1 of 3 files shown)


illiapoplawski:
to test the script you can use the following command as I set the link in the guides to point to codercom/code-server/master rather than to my forked repo:
```
 bash <(curl -s https://raw.githubusercontent.com/illiapoplawski/code-server/quickstart-copy-paste-links/scripts/update.sh)
```
If there's a better place to store the script feel free to store it elsewhere and then update the link to point to that more permanent location.

coadler:
cc @nhooyr

coadler:
In the future it would be nice for the script to be OS aware (darwin vs linux), but since this is only mentioned from the cloud guides we can work on that later.

illiapoplawski:
I can easily update the script to automatically find the current running os darwin or linux then download and install the correct file.  The rest of the guide would need to be updated as well.  Should I do it in this PR or in a new one?

illiapoplawski:
I've updated to be in-line with master and am not planning on adding anything else to this PR unless anyone else has anything to change.

nhooyr:
See #94 

We can remove almost all of these docs now.

illiapoplawski:
Are you recommending closing because sshcode automatically downloads the latest version of code-server?

nhooyr:
Yes.

Issue #53, opened 2019-03-06
prevent ctrl+w from closing tab


0xfd000000:
I guess it's not possible to catch the ctrl+w shortcut before the browser interprets it, but would it be possible to show a "leave this page?" confirmation dialog to prevent the good old muscle memory from screwing up ones work?

kylecarbs:
Will definitely add this!

odiak:
Can you use Ctrl + W to close editor tabs or some other VSCode functions?
And I think it's better to close browser tab (maybe with confirmation) if any editor tabs are opened.

jeasonstudio:
Hotkey should be consistent with `VSCode` to get an immersive experience.

avelino:
> Will definitely add this!

Important to have a page in the documentation listing all the default code-server shortcuts

gijo-varghese:
Right now when I press `Ctrl + W` it closes the browser tab, instead of closing the file. Since I used to close files using this in real VS Code, we need to find a solution for this

gijo-varghese:
we could use something like 
```javascript
window.onbeforeunload = function (e) {
    // Cancel the event
    e.preventDefault();

    // Chrome requires returnValue to be set
    e.returnValue = 'Really want to quit the game?';
};
```

It doesn't prevent from closing, but at least show you an additional confirmation

code-asher:
Fixed with https://github.com/codercom/code-server/pull/463.

MicroDroid:
It totally closes the tab for me.

Chrome: Version 80.0.3962.2 (Official Build) dev (64-bit)
Code Server: 2.1692-vsc1.39.2

MicroDroid:
It totally closes the tab for me.

Chrome: Version 80.0.3962.2 (Official Build) dev (64-bit)
Code Server: 2.1692-vsc1.39.2

MicroDroid:
I didn't unassign anyone :thinking:

sr229:
@MicroDroid unfortunately we can't solve this - I believe this is also persistent in upstream VSCode so this is more of a by-design since you can't override Ctrl-W as far as I investigated.

MicroDroid:
Perhaps we can at least show a confirmation dialog?

sr229:
I think we did that over in Theia, not sure how we want it on VSCode though.

Feel free to bring this up to Microsoft as well.

CC @code-asher what do you think?

code-asher:
[...]

PR #463, opened 2019-04-08
Add confirmation for closing window

nol166:
Control W closes the browser window instantly without respect for unsaved work. This shows a prompt asking if you want to leave.

#53

### DIFF
--- a/packages/ide/src/client.ts
+++ b/packages/ide/src/client.ts
@@ -35,9 +35,12 @@ export abstract class IdeClient {
 
 		let appWindow: Window | undefined;
 
-		window.onbeforeunload = () => {
-			return "Are you sure you want to navigate away?";
-		}
+		window.addEventListener("beforeunload", (e) => {
+			let confirmationMessage = "Are you sure you want to navigate away?";
+			(e || window.event).returnValue = confirmationMessage; //IE
+
+			return confirmationMessage; // webkit, safari, chrome
+		});
 
 		window.addEventListener("message", (event) => {
 			if (event.data === "app") {
--- a/packages/ide/src/client.ts
+++ b/packages/ide/src/client.ts
@@ -34,6 +34,11 @@ export abstract class IdeClient {
 		this.loadTime = time(2500);
 
 		let appWindow: Window | undefined;
+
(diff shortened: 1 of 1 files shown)


packages/ide/src/client.ts
```diff
@@ -34,6 +34,11 @@ export abstract class IdeClient {
 		this.loadTime = time(2500);
 
 		let appWindow: Window | undefined;
+
+		window.onbeforeunload = () => {
```
code-asher: It might make sense to use `addEventListener` to stay consistent with the other event handlers.

packages/ide/src/client.ts
```diff
@@ -34,6 +34,14 @@ export abstract class IdeClient {
 		this.loadTime = time(2500);
 
 		let appWindow: Window | undefined;
+
+		window.addEventListener("beforeunload", (e) => {
+			let confirmationMessage = "Are you sure you want to navigate away?";
```
kylecarbs: Should be a const

packages/ide/src/client.ts
```diff
@@ -34,6 +34,14 @@ export abstract class IdeClient {
 		this.loadTime = time(2500);
 
 		let appWindow: Window | undefined;
+
+		window.addEventListener("beforeunload", (e) => {
```
sr229: This should only fire if there is unsaved changes that needs to be pushed to the origin.

code-asher: The goal is to prevent accidentally closing the tab for those who have `ctrl+w` ingrained in muscle memory. Unfortunately I think this is the only way we can do it. :disappointed:

packages/ide/src/client.ts
```diff
@@ -34,6 +34,14 @@ export abstract class IdeClient {
...
+			const confirmationMessage = "Are you sure you want to navigate away?";
+			(e || window.event).returnValue = confirmationMessage; //IE
+
+			return confirmationMessage; // webkit, safari, chrome
```
code-asher: I don't think the confirmation message is actually used, is it? https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#Examples

PR #471, opened 2019-04-09
[GH-459] explicitly create a volume and set permission for project dir

### LINKED ISSUE #459: Docker issue
os Debian 9.0

```docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "chdir to cwd (\"/home/coder/project\") set in config.json failed: permission denied": unknown.```

### PR DESCRIPTION
**Pull request by sr229**

This ensures a emptydir volume if the user forgot to mount. This allows OpenShift compatibility and solves
Docker issues.

This fixes #459 and #469 partially

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,8 +36,14 @@ RUN adduser --gecos '' --disabled-password coder && \
 
 USER coder
 # We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
+RUN mkdir -p /home/coder/project && \
+    chmod g+rw /home/coder/project;
+
 WORKDIR /home/coder/project
+
+# This assures we have a volume mounted even if the user forgot to do bind mount.
+# XXX: Workaround for GH-459 and for OpenShift compatibility.
+VOLUME [ "/home/coder/project" ]
 
 COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
 EXPOSE 8443

### CODE REVIEW

File: Dockerfile [hunk 1]
```diff
@@ -36,9 +36,15 @@ RUN adduser --gecos '' --disabled-password coder && \
 
 USER coder
 # We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
+RUN mkdir -p /home/coder/project && \
+    chmod g+rw /home/coder/project;
```
nhooyr: Why is this necessary?

sr229: I need to set the R/W permissions explicitly because I found out on OpenShift hosts, if this is not set, the container would outright crash.

sr229:
CC @nhooyr, what do you think

kylecarbs:
LGTM

Dockerfile
nhooyr: Still not clear why this is necessary. You shouldn't be able to write if you don't own it, that is correct behaviour. I'm going to revert this @kylecarbs

Multishifties: `container would outright crash` <-- still a little concerned about the possibility of reintroducing this possibility with a revert.

nhooyr: > container would outright crash <-- still a little concerned about the possibility of reintroducing this possibility with a revert.
 
Not a concern, don't worry. Its definitely not outright crashing, there is a permission related issue.

sr229: @Multishifties I explained it already becuase of how OpenShift handles UIDs and GIDs, which would make `/home/coder/project` inaccessible practically on the official image. So I guess this would make sense if they revert it, but the official image will never work on OpenShift due to missing permissions

nhooyr: > becuase of how OpenShift handles UIDs and GIDs, 

What does that mean? How does it handle it?

You're not being clear with your concerns, I can't give you a good response.

PR #475, opened 2019-04-10
change path for /home/coder/project

mkenigs:
The first time I ran coder I copied and pasted the command from the README, changed the IP, and immediately exposed my home directory to the network. Since it's the first example most people will run and it's encouraged to "try it out," I think mounting PWD to /home/coder/project is a bad recommendation. I was certainly surprised when I realized what I'd done, and I think it would make more sense for the README to make a better recommendation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 Try it out:
 ```bash
-docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}:/home/coder/project" codercom/code-server:1.621 --allow-http --no-auth
+docker run -it -p 127.0.0.1:8443:8443 -v ~/coder:/home/coder/project codercom/code-server:1.621 --allow-http --no-auth
 ```
 
 - Code on your Chromebook, tablet, and laptop with a consistent dev environment.


cy91244:
I think it's fine as long as it would generate warning message 
when the IP is set to `0.0.0.0` and with `allow-http` or `no-auth` like what #441 does.

Furthermore, the danger can not be resolved by simply changing the volume.
e.g. the terminal is accessible for anyone.

mkenigs:
True it's not bad default behavior so much as a bad configuration decision. I just don't see why the example in the README should be an example of a poor decision when it would be simple to make it a little better.

sr229:
You don’t need to follow the README 100%. We have official docs for this.

sr229:
Recommending a closure since this isn't what this PR is supposed to do as described.

mkenigs:
Sorry if I'm misunderstanding, but it seems like if following the README leads to what you said was me not configuring something correctly, that should be changed. I cannot imagine when mounting home to /home/coder/project would be a good idea, and for a lot of users home will be PWD when they run this command. Also, if I wanted to mount home, I would much rather specify that explicitly (ie ~) so I could run the same command every time. I changed my initial description to make what I mean more clear. Maybe I'm missing something, but is there any situation under which you would actually want to mount whatever random directory you currently happen to be in?

kylecarbs:
I see what you're saying. Not sure _exactly_ sure what a solution would be. Defaulting to `~/coder` seems kinda strange imo.

nhooyr:
This doesn't make things any more secure because the container terminal is still exposed on your network. Closing as we can't really help here, the command explicitly uses localhost so if you change it to be public, you need to understand the ramifications and #441 will add a warning.

mkenigs:
I couldn't cd to any other directory so I assumed there wasn't some other way to do it - I guess there is? @nhooyr

nhooyr:
There is no legitimate way to escape a container but it still happens occasionally due to vulns. Without a tuned docker setup, it's not a good idea to rely on containers for security. E.g. see https://access.redhat.com/security/vulnerabilities/runcescape

I think the convenience is worth the marginal loss in security by mounting in $PWD.

mkenigs:
Ok I would never want to mount PWD but I see what you're saying. Thanks for the explanation!

Issue #478, opened 2019-04-10
Allow press enter on login screen to process


RealSlimMahdi:
At the moment, I have to click the enter button to proceed.
Would be good to be able to just press enter, once the password is entered.

thanks

sr229:
I think this is a bug and not a FR

creativeguy2013:
This is merged, can be closed.

coadler:
Resolved by #479

PR #479, opened 2019-04-10
Allow use of the enter key for password input for code-server

nol166:
Accept enter in addition to clicking the submit button on the code-server login page.

#478


code-asher:
Could we use a regular form instead? Then enter would just work as expected. It feels slightly odd to me to re-implement form logic in JavaScript.

packages/app/browser/src/app.ts
```diff
@@ -21,10 +21,14 @@ window.addEventListener("message", (event) => {
 
 const password = document.getElementById("password") as HTMLInputElement;
 const submit = document.getElementById("submit") as HTMLButtonElement;
```
code-asher: Can get rid of the button.

packages/app/browser/src/app.ts
```diff
@@ -21,10 +21,14 @@ window.addEventListener("message", (event) => {
...
+const form = document.getElementById("login-form") as HTMLFormElement;
+
 if (!submit) {
```
code-asher: Might be good to check if the form exists instead of the button.

PR #502, opened 2019-04-15
Use port from PORT env variable if available

BurntCaramel:
Some services such as Google Cloud Run specify the port via the `PORT` env variable. If we allow reading from that (if it doesn’t exist we fallback to 8443) then they will be more likely to work out of the box.


packages/server/src/cli.ts
```diff
@@ -26,7 +26,7 @@ commander.version(process.env.VERSION || "development")
 	.option("--data-dir <value>", "DEPRECATED: Use '--user-data-dir' instead. Customize where user-data is stored.")
 	.option("-h, --host <value>", "Customize the hostname.", "0.0.0.0")
 	.option("-o, --open", "Open in the browser on startup.", false)
-	.option("-p, --port <number>", "Port to bind on.", 8443)
+	.option("-p, --port <number>", "Port to bind on.", parseInt(process.env.PORT, 10) || 8443)
```
kylecarbs: If port is set to a non-integer it'd throw. Maybe we should parse below to offer a nice error?

BurntCaramel: If it’s a non integer, it’d be `NaN`, and due to the `||` would fallback to `8443`

nathang21:
Anyone have luck getting this to work on Cloud Run?
I've tried both the authenticatedUsers & allUsers options. 

authenticatedUsers users: 403 error "Your client does not have permission to get URL / from this server."
allUsers: I keep getting a 301 redirect loop when viewing the public URL until Chrome says "ERR_TOO_MANY_REDIRECTS" 

Maybe I need to figure out the certificates since Cloud Run forces HTTPS.

PR #503, opened 2019-04-15
Add GCP logging extender

kylecarbs:
Adds `logger.extend(<Extender>)` API along with GCP stackdriver extender.

### DIFF
--- a/packages/logger/src/logger.ts
+++ b/packages/logger/src/logger.ts
@@ -59,6 +59,8 @@ export const field = <T>(name: string, value: T): Field<T> => {
 	return new Field(name, value);
 };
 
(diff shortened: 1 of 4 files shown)


packages/logger/src/logger.ts
```diff
@@ -59,6 +59,8 @@ export const field = <T>(name: string, value: T): Field<T> => {
 	return new Field(name, value);
 };
 
+export type Extender = (msg: { section?: string, type: "trace" | "info" | "warn" | "debug" | "error", message: string, fields?: FieldArray, level: Level }) => void;
```
foresthoffman:
```suggestion
export type Extender = (msg: {
	message: string,
	level: Level,
	type: "trace" | "info" | "warn" | "debug" | "error",
	fields?: FieldArray,
	section?: string,
}) => void;
```

I think adding a little indentation here would make this easier to read.

PR #504, opened 2019-04-16
Add flag to install an extension from the command line

code-asher:
Opted to just run VS Code's CLI with a small modification to ensure it always only does extension management. Not sure if this is the best way. We could directly import the file but then there were Typescript errors that I wasn't sure how to solve cleanly.

Also wasn't sure how to make Commander accept variadic options for an option. Maybe it's not possible?


packages/server/src/cli.ts
```diff
@@ -162,6 +165,26 @@ const bold = (text: string | number): string | number => {
 		logger.warn('"--data-dir" is deprecated. Use "--user-data-dir" instead.');
 	}
 
+	if (options.installExtension) {
+		const fork = forkModule("vs/code/node/cli", [
```
kylecarbs: Super cool that this just works :confetti_ball:

PR #518, opened 2019-04-17
Make assets unique

code-asher:
- [x] Make all CSS and JavaScript assets uniquely named.
- [x] Move the login page to `/login`.
- [x] Manually include all desired assets. This prevents caching `/` and `/login` itself which made it so you had to refresh or log in twice in some cases.

### DIFF
--- a/scripts/webpack.client.config.js
+++ b/scripts/webpack.client.config.js
@@ -7,102 +7,81 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
 const WebpackPwaManifest = require("webpack-pwa-manifest");
 const { GenerateSW } = require("workbox-webpack-plugin");
 
-// const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer");
-
 const root = path.join(__dirname, "..");
 const prod = process.env.NODE_ENV === "production" || process.env.CI === "true";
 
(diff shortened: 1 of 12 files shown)


kylecarbs:
Pretty neat. I like it :confetti_ball:

scripts/webpack.client.config.js
```diff
@@ -7,102 +7,83 @@ const HtmlWebpackPlugin = require("html-webpack-plugin");
...
+			icons: [{
+				src: path.join(root, "packages/web/assets/logo.png"),
+				sizes: [96, 128, 192, 256, 384],
+			}],
+		})
+	].concat(prod ? [
+		// Exclude everything from the precache and add everything we want to be
+		// cached at runtime.
+		new GenerateSW({
+			exclude: [/.*/],
```
lucacasonato: We don't precache the bundle anymore now... before it could cache this and other assets while still on the login screen. This made the loading after the login significantly faster. Why exactly do you exclude all now? Maybe explicitly set it to cache the same files as in the runtimeCaching pattern.

code-asher: Yeah, I was deliberating on this but I think you're right and it's probably better to load them all up at once at the beginning.

PR #541, opened 2019-04-18
Fix multiple save-as prompts

### LINKED ISSUE #536: Save shortcut triggers even when save dialog is open
- `code-server` version: 1.868-vsc1.33.1
- OS Version: Ubuntu 18.04

## Description

When you press ctrl-S to save, it opens a dialog and puts a dark black rect in front of the rest of the UI. When you press ctrl-S again, it does it again! If you hold ctrl-S down it does it many times.

## Steps to Reproduce

1. Open a file.
2. Hold ctrl-S.

You might need to use [this workaround](https://github.com/codercom/code-server/issues/112#issuecomment-470757090) for the shortcut to work.

nol166:
The save as overlays open on top of each other if invoked before dismissing. This checks to see if there is an overlay present before spawning another one.

fixes #536

### DIFF
--- a/packages/ide/src/fill/dialog.ts
+++ b/packages/ide/src/fill/dialog.ts
@@ -131,7 +131,7 @@ export class Dialog {
 	 */
 	public show(): void {
 		if (document.querySelector(`.${this.overlay.className}`)) {
-			throw new Error ("Save prompt is already open");
+			throw new Error ("Overlay is already open");
 		}
 		if (!this.cachedActiveElement) {
 			this.cachedActiveElement = document.activeElement as HTMLElement;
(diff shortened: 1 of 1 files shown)


kylecarbs:
How is a dialog displayed twice? This feels like more of a patch and less of a fix.

nol166:
@kylecarbs If the prompt is already open and the user presses the keybind again, it opens a second prompt

code-asher:
One solution that might work (this is untested) is something like:
```
overlayDiv.addEventListener("keydown", (event) => {
  event.preventDefault();
});
```

Could also bind `escape` in there while we're at it to close the dialog.

PR #546, opened 2019-04-19
Fix markdown preview focus

### LINKED ISSUE #543: Markdown preview steals focus
- `code-server` version: 1.868-vsc1.33.1
- OS Version: Ubuntu 18.04

## Description

Preview window of markdown steals focus while editing the corresponding markdown file.

## Steps to Reproduce

1. Open preview window to the side.
1. Edit the markdown file.
    For example, try to add following texts:
    ```markdown
    ### abc
    ```
    As the preview window steals focus, you will not be able to write `abc`.

nol166:
Fixes markdown preview from stealing focus away from the editor while typing

fixes #543

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -827,7 +827,7 @@ index 780147c..2e8c9af 100644
 -					if (platform.isMacintosh) {
 +					if (browser.isMacintosh) {
 diff --git a/src/vs/workbench/contrib/webview/electron-browser/webview-pre.js b/src/vs/workbench/contrib/webview/electron-browser/webview-pre.js
-index 74fc798..03728d0 100644
+index 74fc798..0196be0 100644
 --- a/src/vs/workbench/contrib/webview/electron-browser/webview-pre.js
 +++ b/src/vs/workbench/contrib/webview/electron-browser/webview-pre.js
 @@ -10 +10,19 @@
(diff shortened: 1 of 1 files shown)


scripts/vscode.patch
```diff
@@ -824,6 +824,15 @@ index 74fc798..03728d0 100644
 +			// 	supportFetchAPI: true,
 +			// 	corsEnabled: true
 +			// });
+@@ -310 +328 @@
+-
++			
+@@ -328 +346,4 @@
+-					newFrame.contentWindow.focus();
++					/* Comment out to avoid VS Code from focusing on markdown preview on reload */
```
code-asher: This comment doesn't match the behavior; maybe something like "Only focus the updated frame if it already had focus"?

scripts/vscode.patch
```diff
@@ -885,7 +894,7 @@ index 48ef482..dc47f81 100644
 -			placeHolder: isMacintosh ? nls.localize('openRecentPlaceHolderMac', "Select to open (hold Cmd-key to open in new window)") : nls.localize('openRecentPlaceHolder', "Select to open (hold Ctrl-key to open in new window)"),
 +			placeHolder: browser.isMacintosh ? nls.localize('openRecentPlaceHolderMac', "Select to open (hold Cmd-key to open in new window)") : nls.localize('openRecentPlaceHolder', "Select to open (hold Ctrl-key to open in new window)"),
 diff --git a/src/vs/workbench/electron-browser/main.contribution.ts b/src/vs/workbench/electron-browser/main.contribution.ts
-index 71bc992..111ce51 100644
+index 71bc992..2692f39 100644
```
code-asher: Looks like there's a conflict here with master that needs to be resolved. You could just revert this line, or merge and regenerate the patch. I'm not sure why it generated a different hash. :confused:

scripts/vscode.patch
nol166: fixed

scripts/vscode.patch
nol166: fixed

scripts/vscode.patch
```diff
@@ -824,6 +824,16 @@ index 74fc798..03728d0 100644
 +			// 	supportFetchAPI: true,
 +			// 	corsEnabled: true
 +			// });
+@@ -310 +328 @@
+-
```
code-asher: We might want to remove all the extra whitespace here so the patch is as small as possible.

PR #557, opened 2019-04-21
Add sample deployment for Google Kubernetes Engine

tadashi0713:
Added sample code for Google Kubernetes Engine(GKE)
Mainly differs PersistentVolume compared to AWS

No

### DIFF
--- a/deployment/google-cloud/deployment.yaml
+++ b/deployment/google-cloud/deployment.yaml
@@ -1,0 +1,77 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: code-server
+---
+apiVersion: v1
+kind: Service
+metadata:
+ name: code-server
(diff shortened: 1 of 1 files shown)


nhooyr:
We already have a k8s example at https://github.com/cdr/code-server/blob/master/doc/examples/kubernetes.yaml

That should suffice imo.

nhooyr:
Likewise for AWS at https://github.com/cdr/code-server/blob/master/doc/examples/kubernetes.aws.yaml

I guess we should include this GKE example there as `kubernetes.gke.yaml`.

sr229:
Based on @nhooyr's feedback I think we can close this one. We'll revert to a vendor-neutral config instead.

nhooyr:
@sr229 Please open a new issue for that so we don't forget.

PR #564, opened 2019-04-22
Add --socket flag

kylecarbs:
Allows code-server to bind on a unix socket. Fixes https://github.com/codercom/code-server/issues/561


RyanEwen:
Could someone please post an example apache conf that reverse proxies the unix socket instead of tcp port?

Also I assume we need to make a volume mount to expose the sock file when running in docker

PR #572, opened 2019-04-23
Offer https/http url based on schema

nol166:
Shows link for HTTPS or HTTP based on flags used on running. Let users know if telemetry is disabled.

n/a

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -314,8 +314,8 @@ const bold = (text: string | number): string | number => {
 		logger.info("Telemetry is disabled");
 	}
 
-	let schema = options.allowHttp ? "http" : "https";
-	const url = `${schema}://localhost:${options.port}/`;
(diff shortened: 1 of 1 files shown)


packages/server/src/cli.ts
```diff
@@ -310,8 +310,12 @@ const bold = (text: string | number): string | number => {
 	} else {
 		logger.warn("Launched without authentication.");
 	}
+	if (options.disableTelemetry) {
+		logger.info("Telemetry is disabled");
+	}
 
-	const url = `http://localhost:${options.port}/`;
+	let schema = options.allowHttp ? "http" : "https";
```
code-asher: This can be a `const`.

code-asher: Also I think it might be called `scheme` rather than `schema`?

nol166: I had it as protocol to begin with. I will change it backsies.

PR #573, opened 2019-04-23
Fix no-auth disabling https

### LINKED ISSUE #300: --no-auth flag also disables SSL
Not sure if this was intentional, but when you disable auth with the `--no-auth` flag on the CLI, [SSL is also disabled](https://github.com/codercom/code-server/blob/21690453775629a97627d37ee8703ab5718ebe35/packages/server/src/server.ts#L135).

It's a bit confusing. In order to get SSL working, you need to use a password. When you don't have a password, you have to use the site/app via plain ol' HTTP.

Maybe there's some logic that I'm missing, but this had me stumped until I dug into the code.

nol166:
`--no-auth` disallows password requirement and SSL, it should instead only disallow the password requirement. Changes the default value from undefined to false for the ```--no-auth``` flag

fixes #300


RyanEwen:
Either this issue still exists or it hasn't made it into the docker version as of yet

packages/server/src/server.ts
```diff
@@ -133,7 +133,7 @@ export const createApp = async (options: CreateAppOptions): Promise<{
 		});
 	});
 
-	const server = httpolyglot.createServer(options.bypassAuth ? {} : options.httpsOptions || certs, app) as http.Server;
+	const server = httpolyglot.createServer(options.allowHttp ? {} : options.httpsOptions || certs, app) as http.Server;
```
nwtnsqrd: This change causes `(SSL: error:14094410:SSL routines:ssl3_read_bytes:sslv3 alert handshake failure:SSL alert number 40)` when running code-server behind a reverse proxy which provides SSL. Although this is likely due to faulty configuration of the reverse proxy. In nginx, I had to change
```
location / {
       proxy_pass https://localhost:8443;
       [...]
```
to
```
location / {
       proxy_pass http://localhost:8443;
       [...]
```

Just posting this here in case anyone mis-configured their reverse proxy like I did ;)

PR #584, opened 2019-04-24
Disable beforeUnload during app mode

### LINKED ISSUE #574: window.beforeclose prompt should not alert if running in app mode in chrome
- `code-server` version: v1.939-vsc1.33.1
- OS Version: linux server, osx client

## Description

When running code-server in chrome app mode (ex sshcode) the window close prompt should not show up when doing ctrl+w, since running in app mode, it's able to overload the ctrl/cmd+w shortcut to actually close editor tabs

## Steps to Reproduce

1. Start code-server
1. Open app with
```
/Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome --app=https://server-ip:8443
```
1. Open several editor tabs
1. cmd+w to close 1 tab

nol166:
When code-server is running in app mode, the close window confirmation is annoying and unnecessary. This change prevents that behavior if in app mode.

fixes #574

### DIFF
--- a/packages/ide/src/client.ts
+++ b/packages/ide/src/client.ts
@@ -35,10 +35,14 @@ export abstract class IdeClient {
 
 		let appWindow: Window | undefined;
 
-		window.addEventListener("beforeunload", (e) => {
-			e.preventDefault(); // FireFox
-			e.returnValue = ""; // Chrome
-		});
+		if (!window.matchMedia("(display-mode: standalone)").matches) {
+			console.log("Not in app mode");
+			window.addEventListener("beforeunload", (e) => {
+				e.preventDefault(); // FireFox
+				e.returnValue = ""; // Chrome
+			});
+		}
+
 
 		window.addEventListener("message", (event) => {
 			if (event.data === "app") {


packages/ide/src/client.ts
```diff
@@ -35,10 +35,13 @@ export abstract class IdeClient {
 
 		let appWindow: Window | undefined;
 
-		window.addEventListener("beforeunload", (e) => {
-			e.preventDefault(); // FireFox
-			e.returnValue = ""; // Chrome
-		});
+		if (!window.matchMedia("(display-mode: standalone)").matches) {
```
kylecarbs: Won't match fullscreen. Is there a cleaner way to check if we're in a minified window mode?

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode#Syntax

nol166: The fullscreen mode would still have the problem where `control + w` closing the browser window, correct?

nol166: `standalone` should cover it afaik

PR #586, opened 2019-04-24
Remove reveal in finder

### LINKED ISSUE #576: Right click on tab -> open in finder on Mac has unclear functionality and can break the instance
- `code-server` version: v1.696-vsc1.33.0
- OS Version: Mac Mojave locally on latest version of chrome, Ubuntu 18.04 remotely

## Description

When you right click on a tab, "Reveal in Finder" appears. I've clicked on it in a previous version of code server and it opened the file in its own folder without allowing me to get out of the folder. I just tested it, and it appears to just refresh the window and do nothing else. This isn't super crucial, but it doesn't seem like it should exist on a remote browser since finder should not be able to open the file at all.

I personally think that option should just be removed since it is possible for users to accidentally click it, and it is confusing.

## Steps to Reproduce

1. Open any file in Code Server
2. Right click the tab and click on the "Open in finder" dialog

nol166:
The "reveal in finder/explorer" option has no utility in the browser and causes a page to reload to invoke. This PR simply removes the option from the context menu to avoid mistakenly causing a refresh/confusion.

fixes #576


kasplat:
Thanks guys!

PR #629, opened 2019-05-01
Fix/redesign dialogs

nol166:
Redesign of the "move to trash" and "open" dialogs. Adds manual path entry.

n/a


jeasonstudio:
@nol166 Whether to consider refactoring `showOpenDialog`, use the `quickPick` provided by vscode to select a file or dir. Just like the picture below(vscode-insiders + remote development ext):

image

code-asher:
Closed in favor of https://github.com/cdr/code-server/pull/857 which uses the same dialog used by the VS Code's remote extensions.

PR #633, opened 2019-05-02
expose terminal service in IDE API

nhooyr:


nhooyr:
@code-asher what do you think of just returning the service collection from the IDE api instead?

Issue #439, opened 2019-04-05
Support docker run with  --user $(id -u):$(id -g) cli arg (shared process error)


satlus:
### Test environment
Host OS: OSX 10.14.4
Docker: 2.0.0.3 (31259)
image: codercom/code-server:1.621

### Request
Hi,

We run docker in a locked down environment where we alias the `docker run` cli command to prepend --user $(id -u):$(id -g) on all host operating systems. This is a security constraint we can't remove. When i try to execute code server simulating this environment it throws a shared process error. Any ideas on an approach to resolve?

### Repro
```
docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}/codeserver:/home/coder/project" -u $(id -u):$(id -g) -v "${PWD}/codeserver/.cache:/.cache" -v "${PWD}/codeserver/.local:/.local" codercom/code-server:1.621 --allow-http --no-auth
INFO  code-server development
INFO  Additional documentation: http://github.com/codercom/code-server
INFO  Initializing {"data-dir":"/.local/share/code-server","extensions-dir":"/.local/share/code-server/extensions","working-dir":"/home/coder/project","log-dir":"/.cache/code-server/logs/20190405152816797"}
INFO  Starting webserver... {"host":"0.0.0.0","port":8443}
WARN  No certificate specified. This could be insecure.
WARN  Documentation on securing your setup: https://github.com/codercom/code-server/blob/master/doc/security/ssl.md
WARN  Launched without authentication.
INFO   
INFO  Started (click the link below to open):
INFO  http://localhost:8443/
INFO   
INFO  Starting shared process [1/5]...
INFO  Starting shared process [2/5]... {"error":"Exited with 1"}
INFO  Starting shared process [3/5]... {"error":"Exited with 1"}
/src/packages/server/out/cli.js:325
[...]

PR #640, opened 2019-05-03
Support non-root users (runtime UID remapping)

satlus:
Our environment restricts docker execution to images that can only with a non root user `docker run -u $UID...`. This PR adds boxboat/fixuid to allow for runtime mapping of host->container UID

https://github.com/cdr/code-server/issues/439


DockerLaunchCommands.MD
```diff
@@ -0,0 +1,14 @@
+# Docker launch commands for coder
+
+## default coder uid:guid
+docker run -it -p 127.0.0.1:8443:8443 \
+ -v "${PWD}/code-server:/home/coder/project" \
+ satlus-code-server:latest --allow-http --no-auth
```
ibnesayeed: The image name here is of yours and not the official one.

DockerLaunchCommands.MD
```diff
@@ -0,0 +1,14 @@
...
+
+ ## passing `id(-g|-u)` through the system
+docker run -it -p 127.0.0.1:8443:8443 \
+-v "${PWD}/code-server:/home/coder/project" \
+-v "${PWD}/code-server/.cache:/home/coder/.cache" \
+-v "${PWD}/code-server/.local:/home/coder/.local" \
+-u $(id -u):$(id -g) \
+satlus-code-server:latest --allow-http --no-auth
```
ibnesayeed: The image name here is of yours and not the official one.

Dockerfile
```diff
@@ -34,14 +34,24 @@ RUN locale-gen en_US.UTF-8
...
+	chown root:root /usr/local/bin/fixuid && \
+	chmod 4755 /usr/local/bin/fixuid && \
+	mkdir -p /etc/fixuid && \
+	printf "user: $USER\ngroup: $GROUP\n" > /etc/fixuid/config.yml
+
 USER coder
+
 # We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
+RUN mkdir -p /home/coder/workdir
```
ibnesayeed: Is renaming the directory necessary here? I don't see any problem, but there might be documentation elsewhere which might go out of sync due to this change.

Dockerfile
```diff
@@ -34,14 +34,24 @@ RUN locale-gen en_US.UTF-8
...
 
-WORKDIR /home/coder/project
+WORKDIR /home/coder/workdir
```
ibnesayeed: Is renaming the directory necessary here? I don't see any problem, but there might be documentation elsewhere which might go out of sync due to this change.

DockerLaunchCommands.MD
```diff
@@ -0,0 +1,14 @@
+# Docker launch commands for coder
```
ibnesayeed: You might want to fix the title cases of headings and subheads to match with the style of the README file. Also, I personally do no like upper-case file extensions (as ignore patterns are easier if extensions are kept lower-cased) and CamelCased file names, but that could just me.

nhooyr:
Not sure if this is a good idea, the reamde for fixuid says:

```
fixuid should only be used in development Docker containers. DO NOT INCLUDE in a production container image
```

Not sure why.

ibnesayeed:
> Not sure if this is a good idea, the reamde for fixuid says:
> ```
> fixuid should only be used in development Docker containers. DO NOT INCLUDE in a production container image
> ```
> Not sure why.

@nhooyr editors are generally used in development environment, unless someone deploys it on a server and makes it available to a larger user-base.

ibnesayeed:
> @nhooyr: Not sure why.

https://github.com/boxboat/fixuid/issues/1

DockerLaunchCommands.MD
satlus: Great catch - I'm going to remove this file, and just add a section to the readme for non-root usage. It will avoid confusion

Dockerfile
satlus: I think it's a good idea, especially if we want to restrict container FS writes to the coder user. We're making the work directory a sub-path of the $HOME of the coder directory, as all docker instructions will be invoked as the coder user at this point. It also means that we can volume mount the /home/coder/project directory without risking masking/overwriting the entrypoint.sh script.

satlus:
I'm cleaning this up to make fixuid logging configurable (and enabled by default if the user runs as host UID).

Dockerfile
ibnesayeed: If you are creating a yet another directory, shouldn't one already existing be left alone as it is? If you are simply renaming it to something else in the same location then I am not sure what are you gaining from it, but I might be missing something.

satlus: check my latest commit - the workdir is now used only to hold the entrypoint.sh script

Dockerfile
```diff
@@ -34,14 +34,27 @@ RUN locale-gen en_US.UTF-8
...
+	chmod 4755 /usr/local/bin/fixuid && \
+	mkdir -p /etc/fixuid && \
+	printf "user: $USER\ngroup: $GROUP\n" > /etc/fixuid/config.yml
+
 USER coder
+
 # We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
+RUN mkdir -p /home/coder/workdir
+copy entrypoint.sh /home/coder/workdir/
```
ibnesayeed:
```suggestion
COPY entrypoint.sh /home/coder/workdir/
```

Dockerfile
ibnesayeed: I think you are missing my point here. There was a `/home/coder/project` directory before where users were supposed to mount their code from the host machine. You should have left it as it was and created `/home/coder/workdir` (or something more appropriately named) directory next to it where you can place your entrypoint script so it does not get overwritten. In my opinion, a better option would be to leave the directory as they were before your PR and place your entrypoint script in something like `/usr/local/bin/` so you can refer to it without an absolute path from anywhere and have the least chance of being overwritten due to bind mounts.

satlus:
@ibnesayeed - appreciate the feedback, but if you check my latest commit in the Dockerfile, the `VOLUME` bind mount is pointed to /home/coder/project, as per the original in master. The `WORKDIR` is a different directory, where the entrypoint.sh is copied (and only used for that purpose). What's the issue?

ibnesayeed:
> @satlus: @ibnesayeed - appreciate the feedback, but if you check my latest commit in the Dockerfile, the `VOLUME` bind mount is pointed to /home/coder/project, as per the original in master. The `WORKDIR` is a different directory, where the entrypoint.sh is copied (and only used for that purpose). What's the issue?

Removing the `mkdir` command and only relying on `WORKDIR` has the following side effect as per an existing comment:
```
# We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
```

Dockerfile
ibnesayeed: Docker instructions are written in upper case.

satlus:
I added it back in, but per the commit history the only reason this was needed in the past was because the `WORKDIR` and `VOLUME` paths resolved to the same location - /home/coder/project. This is no longer the case (as it should be), and workdir is only used to house the entrypoint. There really is no need for this going forward IMO.

Funnily, in testing, without the mkdir, the WORKDIR was still created as the coder user. Here's an example from a first start with fixuid disabled:

```bash
coder@d13691a6f654:~$ cd /home/coder
coder@d13691a6f654:~$ ls -la
total 36
drwxr-xr-x 1 coder coder 4096 May  4 10:45 .
drwxr-xr-x 1 root  root  4096 May  3 15:17 ..
-rw-r--r-- 1 coder coder  220 May  3 15:17 .bash_logout
-rw-r--r-- 1 coder coder 3771 May  3 15:17 .bashrc
drwxr-xr-x 3 coder coder 4096 May  4 10:45 .cache
drwxr-xr-x 3 coder coder 4096 May  4 10:45 .local
-rw-r--r-- 1 coder coder  807 May  3 15:17 .profile
drwxr-xr-x 4 coder coder  128 Apr 25 16:47 project
-rw------- 1 coder coder 1024 May  4 10:45 .rnd
drwxr-xr-x 1 coder coder 4096 May  3 21:08 workdir <== THIS
coder@d13691a6f654:~$ 
```

When fixuid is enabled it will recursively chown the home directory of the user, including workdir, making this even less important.

ibnesayeed:
@satlus in that case I would get rid of that `mkdir` command and the comment that says why it was important. Personally, I know `WORKDIR` and some other Docker commands run in the scope of the user if a `USER` is declared in a prior instruction, but the comment was confusing me.

That said, I would perhaps `COPY` the `entrypoint.sh` file into `/usr/local/bin/`. This would:

* No need to create yet another directory specifically for this
* Minimize the chances being overwritten if the user chooses the mount the whole HOME directory inside
* The `ENTRYPOINT` instruction can be changed to `["dumb-init", "entrypoint.sh", "code-server"]` without the need of absolute or relative path

satlus:
> @satlus in that case I would get rid of that `mkdir` command and the comment that says why it was important. Personally, I know `WORKDIR` and some other Docker commands run in the scope of the user if a `USER` is declared in a prior instruction, but the comment was confusing me.
> 
> That said, I would perhaps `COPY` the `entrypoint.sh` file into `/usr/local/bin/`. This would:

Good suggestions, implemented.

In testing I noticed that we'd have to provide explicit volume mounts for /home/coder/.cache and /home/coder/.local when enabling non-root use. This seems less than ideal from a usability perspective, and as it has the effect of writing transient data inside of the container. To mitigate this I went ahead and changed the `VOLUME` mount to export the entire /home/coder directory (and updated docker CLI run examples in the README). This seems like smarter way to ensure config and runtime data are separated from the container image appropriately. 

Would definitely appreciate the committers commenting here - it seems this is getting a bit broader in scope than originally anticipated.

Dockerfile
```diff
@@ -34,20 +34,29 @@ RUN locale-gen en_US.UTF-8
...
+	printf "user: $USER\ngroup: $GROUP\n" > /etc/fixuid/config.yml
+
 USER coder
-# We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
 
-WORKDIR /home/coder/project
+# Setup our entrypoint
+COPY entrypoint.sh /usr/local/bin/
+RUN sudo chmod +x /usr/local/bin/entrypoint.sh
```
ibnesayeed: These two `COPY` and `RUN` commands can be moved above the `USER` command so that the `entrypoint.sh` file is copied as root under the `/usr/local/bin/` directory and `chmod` needs no `sudo`.

README.md
```diff
@@ -9,7 +9,7 @@
 
 Try it out:
 ```bash
-docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}:/home/coder/project" codercom/code-server --allow-http --no-auth
+docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}/code-server:/home/coder" codercom/code-server --allow-http --no-auth
```
ibnesayeed: I am not too sure about this change. Are we assuming that there will be `code-server` available under `${PWD}/`?

Dockerfile
```diff
@@ -34,20 +34,29 @@ RUN locale-gen en_US.UTF-8
...
 
 # This assures we have a volume mounted even if the user forgot to do bind mount.
 # So that they do not lose their data if they delete the container.
-VOLUME [ "/home/coder/project" ]
+VOLUME [ "/home/coder" ]
```
ibnesayeed: Can we change this to:

```suggestion
VOLUME [ "/home/coder/project", "/home/coder/.cache", "/home/coder/.local" ]
```

Instead of creating a volume for the whole home directory?

README.md
satlus: It will be created if it doesn't exist - I think it's better than polluting the $PWD with the rnd, .local and cache as it does now.

Dockerfile
satlus: Yes - will test and update tomorrow

Dockerfile
satlus: Good call, will be in the next commit

Dockerfile
satlus: Saw some odd things after testing this change and i'm not sure we should go this route. The first is that if we decide to bind mount /home/coder, to have the full home directory on the local filesystem, duplicate folders are created:

```bash
code-server$ !find
find ./code-server/ -maxdepth 1 -type d
./code-server/
./code-server//project,
./code-server//.local
./code-server//.local,
./code-server//.cache,
./code-server//.cache
```

The comma suffixed directories are dupes and don't have data. I'm wondering if this is a side-effect of us bind mounting a parent path of previously defined volume containers. Taking a step back, do we really want to create three anonymous volume containers here by default? I'd rather just have the home directory be portable as an anonymous volume container,  and recommend a compose file or docker run commands where the volumes are explicitly created for this purpose.

Dockerfile
```diff
@@ -34,20 +34,29 @@ RUN locale-gen en_US.UTF-8
 # configured in /etc/default/locale so we need to set it manually.
 ENV LC_ALL=en_US.UTF-8
 
-RUN adduser --gecos '' --disabled-password coder && \
+RUN addgroup --gid 1000 coder && \
```
sr229: by default the first user to be made in Docker build is 1000 already, I don't see the need for extra steps.

Dockerfile
```diff
@@ -34,20 +34,29 @@ RUN locale-gen en_US.UTF-8
...
+    adduser --uid 1000 --ingroup coder --home /home/coder --shell /bin/sh --disabled-password --gecos "" coder && \
 	echo "coder ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
 
+RUN USER=coder && \
```
sr229: use `su -c` instead or `gosu`.

nhooyr:
Why can't you run the container as UID 1000?

ptoulouse:
As an alternative solution to fixuid, you can differ the creation of the coder user and group to the entrypoint.sh script.

- Create the user based on a user id and group id environment variable (often seen: PUID, PGID) with default values of 1000 and 1000: PUID=${PUID:-1000}, PGID=${PGID:-1000}
- chown -R coder:coder /whatever/directory
- runuser -u coder -- _code-server-command_

satlus:
> As an alternative solution to fixuid, you can differ the creation of the coder user and group to the entrypoint.sh script.
> 
> * Create the user based on a user id and group id environment variable (often seen: PUID, PGID) with default values of 1000 and 1000: PUID=${PUID:-1000}, PGID=${PGID:-1000}
> * chown -R coder:coder /whatever/directory
> * runuser -u coder -- _code-server-command_

We looked at a similar strategy but it won't work if we want init.d in the container to be non-root (the entire point of passing -u to docker at runtime). Unless I'm missing something, this won't satisfy the requirement.

carlos-sarmiento:
Any chance of this PR landing anytime soon? I'm facing this issue and would love to have a solution

sr229:
@carlos-sarmiento from what info I can gather, PRs are on hold until we can merge v2. If you really want to test, `chinodesuuu/coder:satlus` is a feature parity replica of this PR.

carlos-sarmiento:
@sr229 is there any idea on when v2 might land?

sr229:
> @sr229 is there any idea on when v2 might land?

v2 is quite in early stages, and barely feature-parity with v1, mainly because of the new VSCode Web architecture (I believe which is VS Online), you can check #857.

nhooyr:
> Our environment restricts docker execution to images that can only with a non root user docker run -u $UID.... This PR adds boxboat/fixuid to allow for runtime mapping of host->container UID

If you run with UID 1000, things should always just work as the image itself always sets itself to run with uid 1000 by default. You can always access root with `sudo`.

https://github.com/cdr/code-server/blob/22058c5f861d6f39bc42a768d5cb65ffdf696ed1/Dockerfile#L46

Is that not enough for your use case?

nhooyr:
I guess what I'm really suggesting is ensuring your files/folders outside the container are also UID 1000 as that's the only real way to do this without jank with Docker.

nhooyr:
The other being using user namespacing and explicitly mapping some external UID to UID 1000 inside the container. See https://docs.docker.com/engine/security/userns-remap/

ptoulouse:
> I guess what I'm really suggesting is ensuring your files/folders outside the container are also UID 1000 as that's the only real way to do this without jank with Docker.

What if I am not user 1000 on the base system? I won't be able to read the files. What about systems with multiple users? Linuxserver,io has solved the issue using s6_overlay.

nhooyr:
> What if I am not user 1000 on the base system? I won't be able to read the files.

You're using docker, so you should be accessing the files inside the container. If you need to still access them outside the container, use sudo or start the container with the entrypoint as bash. It's the way docker works best right now.

Dockerfile
SuperSandro2000: gosu is actually preferred  as it mirrors ``--user`` flag exactly and fixes weird tty and signal forwarding issues. See https://github.com/tianon/gosu. FYI Tianon creates a lot of official Docker Images on hub.docker.com.

polarathene: @sr229 @SuperSandro2000 this is setting temporary env vars for `USER` and `GROUP` as per [instructions of how to use `fixuid`](https://github.com/boxboat/fixuid#install-fixuid-in-dockerfile), which afaik is only intended for the last line in the command:

```
printf "user: $USER\ngroup: $GROUP\n" > /etc/fixuid/config.yml
```

Rather than add an additional config file externally. Has nothing to do with `gosu`, which you say to use "instead of", but there is no mention of `gosu` in the Dockerfile you're reviewing...?(unless there was in prior commit history, doesn't seem like it since the review from @sr229 is *June 18th* and last commit was *May 6th*)

Just in case there is any confusion, `fixuid` is for remapping existing file ownership of the config user:group to the given uid:gid from `--user` flag in `docker run` command. `gosu` is intended for dropping privileges from root to another user to run a command as instead:

> The core use case for `gosu` is to step *down* from `root` to a non-privileged user during container startup (specifically in the `ENTRYPOINT`, usually). - [Source](https://github.com/tianon/gosu#warning)

Dockerfile
polarathene: > by default the first user to be made in Docker build is 1000 already

This is false. Docker does not enforce a default uid of 1000. The Node images that are being used as a base image however [do create a `node` user of 1000:1000](https://github.com/nodejs/docker-node/blob/f5875531604b4b3b9fbc36437182781c3655c8ae/10/buster/Dockerfile#L3).

That said, this doesn't need to create a new user for 1000:1000, I'm not sure what is up with all the other options in addition to that. The intention appears to be replacing `node` with `coder` which is used later on.  Looks like they're referencing this addition from [`fixuid` installation docs](https://github.com/boxboat/fixuid#install-fixuid-in-dockerfile) without much thought..

Dockerfile
SuperSandro2000: @polarathene I was referring to sr229.

sr229:
@code-asher Let's cherry-pick this PR then pull a new PR that adds this PR's changes.

nhooyr:
I'm gonna take care of this myself in the coming days.

Thank you @satlus and @ibnesayeed for your work.

Issue #673, opened 2019-05-10
No Permissions running in Docker


SlateDave:
- `code-server` version: Latest from Docker (codercom/code-server)
- OS Version: Ubuntu 18.04

## Description

When I create any file through the web, I get the error `Permission denied writing to file (file:///home/coder/project/test.txt)`

I've also tried appending `:z` to the Volume and even tried using the unsafe paramenter `--privileged`

## Steps to Reproduce

1. Run `docker run --name Coder -it -p 8008:8443 -v ${PWD}:/home/coder/project codercom/code-server --allow-http --no-auth`

2. Go to 127.0.0.1:8008

3. Create a file

sr229:
@SlateDave I already slated a fix for this, because said mountable path seems to not allow writes, but @nhooyr didn't like the mode change, so this bug persisted.

liuyanhang:
use `--user root`

sr229:
@liuyanhang that is not recommended. I'm working on a fix right now

liguobao:
@liuyanhang --user root is docker command

nhooyr:
@sr229 There is no indication that the cause of the issue seen by @SlateDave is the same as the one you are fixing in #679 

@SlateDave What are the permissions on the mounted in folder and what user are you inside the container?

nhooyr:
Duplicate of https://github.com/cdr/code-server/issues/1307

Issue #676, opened 2019-05-11
Run code-server with non root user


maxorlovsky:
## Description

I've set up code-server to run through services and I given all of my files and code-server files permission to run as non-root user.
I run non-docker version.

But when I enter URL to enter online-vscode and open terminal, by default it set user as root@* and not user@*.
Is there a way to achieve that? I didn't found answer to this problem in other Issues.

maxorlovsky:
Ok, found an option. It's not actually related to code-server itself, but to ubuntu, that I'm running code-server on. Just before running code server, you can specify it from which user to initiate it. with `su -c`.
So closing of this question.

Issue #454, opened 2019-04-07
"Insufficient permission" error on save


yusukeoshiro:
- `code-server` version: 1.621
- OS Version: Ubuntu 18.04.2 LTS (on GCP)

## Description

Files cannot be saved nor folders cannot be created from the editor. 
This problem only started to occur on the latest release. (I think)

## Steps to Reproduce

1. Start the server with `docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}:/home/coder/project" codercom/code-server:1.621 --allow-http --no-auth`
1. Open the editor on another page and try to make change to any file or folder. You will get this error `Failed to save 'README.md': Insufficient permissions. Select 'Retry as Sudo' to retry as superuser.`.

![image](https://user-images.githubusercontent.com/11641437/55679156-609a9300-5941-11e9-841a-faefa887d236.png)

evemorgen:
Setting wider access settings on those files/directories worked for me - from `664` to `666`. 
I believe it's due to trying to edit files by user 'docker' or something similar (since we're operating from inside of the container) and it is not issue with code-server itself.

michaelgao8:
Reverting back to the `1.604-vsc1.32.0` Dockerfile resolves the issue for me, which means it was potentially introduced in the new release.

coadler:
Yeah, we recently changed the dockerfile to run as a regular user rather than root, so any previous files created as root will have to fixed. You should be able to just `chown -R` the projects folder to the user inside the docker container

sr229:
@yusukeoshiro due to recent changes I proposed that are merged, you will need to fix the permissions. Your files are owned by root so you need to recursively own all files to the user.

DavisAre:
> Your files are owned by root so you need to recursively own all files to the user.

To which user? What user/group is the coder-server running?

DavisAre:
Ok, I circumvented this by passing `-u` variables and running as current user. In my case - root. You can also specify other user's/group's IDs like `-u 1000:1000`
`docker run -it -p 127.0.0.1:8443:8443 -u $(id -u):$(id -g) -v "${PWD}:/home/coder/project" codercom/code-server:1.621 --allow-http` 

Might not be the best approach but may be helpful for someone.

sr229:
@DavisAre that’s redundant. UID/GID is already 1000 when you exec the container

cstoneham:
[...]

PR #679, opened 2019-05-13
GH-673: Revert "Remove chmod on project dir"

sr229:

Due to the decision to remove the mode change, this inherently
reintroduced a bug that prevents the system to write to the bind
mounted folder. the chmod is intended to *fix* it.

this fixes GH-673

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -45,7 +45,7 @@ RUN mkdir -p /home/coder/project && \
 WORKDIR /home/coder/project
 
 # This assures we have a volume mounted even if the user forgot to do bind mount.
-# XXX: Workaround for GH-459 and for OpenShift compatibility.
+# So that they do not lose their data if they delete the container.
 VOLUME [ "/home/coder/project" ]
 
 COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,12 +39,13 @@ RUN adduser --gecos '' --disabled-password coder && \
 
 USER coder
 # We create first instead of just using WORKDIR as when WORKDIR creates, the user is root.
-RUN mkdir -p /home/coder/project
+RUN mkdir -p /home/coder/project && \
+    chmod g+rw /home/coder/project;
 
 WORKDIR /home/coder/project
 
 # This assures we have a volume mounted even if the user forgot to do bind mount.
-# So that they do not lose their data if they delete the container.
+# XXX: Workaround for GH-459 and for OpenShift compatibility.
 VOLUME [ "/home/coder/project" ]
 
 COPY --from=0 /src/packages/server/cli-linux-x64 /usr/local/bin/code-server


nhooyr:
How does this fix #676? In fact, if you don't have read perms for the folder, how would the chmod even succeed?

sr229:
@nhooyr if I got the writing part for this right, you still have ownership in the folder because you're inside `coder`. I've used this before in my Coder images and they seem to work. We can still change the modes since we have ownership on it.

jmiller-to:
@nhooyr, for what it is worth I would like to see this approved or denied as I am experiencing the same issue:

https://github.com/cdr/code-server/issues/454#issuecomment-502461873

nhooyr:
> you still have ownership in the folder because you're inside coder.

Then why do you need to make it group editable if you have ownership?

sr229:
Apparently this does not mandate the ownership well for bind mounts for some reason so explicitly setting them as RW is the only way to make sure you can write changes from container to host @nhooyr.

nhooyr:
We need to know what the reason is before merging this. The only reason I see is if the directory outside of the container has a different owner than uid 1000 at which point you should be changing the directory outside of the container or run code-server as a different user instead of uid 1000. See #640

liguobao:
> We need to know what the reason is before merging this. The only reason I see is if the directory outside of the container has a different owner than uid 1000 at which point you should be changing the directory outside of the container or run code-server as a different user instead of uid 1000. See #640

it will "Permission denied" on docker

liguobao:
> We need to know what the reason is before merging this. The only reason I see is if the directory outside of the container has a different owner than uid 1000 at which point you should be changing the directory outside of the container or run code-server as a different user instead of uid 1000. See #640

it will "Permission denied" on docker

sr229:
Even if we set it to UID1000, the permissions in the bind mounted folder is not the same, therefore, we should force writes and reads on the bind folder so we can write to the folder.

sr229:
There is a inherent reason we do this as well, permissions is irrelevant to the host because its a container-specific permission. There will be no impact on security since it only applies to the container.

nhooyr:
> There is a inherent reason we do this as well, permissions is irrelevant to the host because its a container-specific permission. There will be no impact on security since it only applies to the container.

I don't think this is true, if its mounted in, then any changes to it will affect the folder outside the container. Docker doesn't have user namespacing enabled by default, so the UIDs/GIDs inside the container are the same as the UIDs/GIDs outside.

Can you show the perms on the folder you're having trouble with and then show the UID the container is running as? Do a `ls -la` and post it here.

nhooyr:
In fact can you open a new issue? A issue is better suited for debugging than this PR.

sr229:
Revisiting this PR due to GH-992. Impact is considered major.

sr229:
> > There is a inherent reason we do this as well, permissions is irrelevant to the host because its a container-specific permission. There will be no impact on security since it only applies to the container.
> 
> I don't think this is true, if its mounted in, then any changes to it will affect the folder outside the container. Docker doesn't have user namespacing enabled by default, so the UIDs/GIDs inside the container are the same as the UIDs/GIDs outside.
> 
> Can you show the perms on the folder you're having trouble with and then show the UID the container is running as? Do a `ls -la` and post it here.

@marcdumais-work Provides a explanation over at our friends at Theia, which you can find [here](https://github.com/eclipse-theia/theia/issues/5842#issuecomment-519916201).

> The permission issues you have are expected, and a generic Docker issue when sharing files between a host and container using a volume. What happens is that inside the container, the theia user is different than your user on the host. You can use the id -u and id -g commands to confirm (respectively user id and current main group id) - the user name/group names do not matter, just the ids, when it comes to permissions on Linux.

> If you are looking for a solution specifically for yourself, one option would be to modify the image you use and have the theia user created in the image using your host uid/gid, and then you could access the same files from both without these issues.The permission issues you have are expected, and a generic Docker issue when sharing files between a host and container using a volume. What happens is that inside the container, the theia user is different than your user on the host. You can use the id -u and id -g commands to confirm (respectively user id and current main group id) - the user name/group names do not matter, just the ids, when it comes to permissions on Linux.

So essentially speaking in the terms of Unix permission model we do not have the ability to write to that folder even if we own `/projects` because the file contents isn't ours. Overriding permissions in the container isn't a real security issue since we're only mounting a specific folder. Giving write access on that folder explicitly is the only safe workaround. Another workaround is to have the user write it under the `root` group. We shouldn't worry about the container writing where its not supposed to because we *are* supposed to write there, but the workaround is intended so they can write normally to a mounted folder or a PV.

nhooyr:
This wouldn't work regardless. If we do something like this, it has to be done in the `entrypoint.sh`, not just for the top level project directory which will have the real project directory mounted in as even if the user owns the top level project directory, they still won't be able to access files in a subdirectory.

Issue #648, opened 2019-05-06
docker-compose examples


ageisen2000:
Please add some documentation for how to load this up with a docker compose file.

I believe I got it all running using the commands: directive in the yaml, but i'm having permissions issues.

The224:
@ageisen2000 I created a pull request `Add docker-compose file as example` #680

PR #680, opened 2019-05-13
Add docker-compose file as example

The224:

There wasn't an example of how to use `docker-compose` with code-server.

Yes. docker-compose examples #648

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,13 @@
+version: "3"
+
+services:
+  code-server:
+    container_name: code-server
+    image: codercom/code-server
+    ports:
+      - "8443:8443"
+    volumes:
+      - "${PWD}:/home/coder/project"
+    environment:
+      PASSWORD: ${PASSWORD}
+    command: --allow-http --no-auth


ntindle:
With docker-compose, would it be trivial to add https with certbot for  example purposes ?

The224:
> With docker-compose, would it be trivial to add https with certbot for example purposes?

It could be easy, but I don't think it is in the scope of the example. The goal was to reproduce the docker command example:

```
docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}:/home/coder/project" codercom/code-server --allow-http --no-auth
```

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+  code-server:
+    container_name: code-server
+    image: codercom/code-server
+    ports:
+      - "8443:8443"
+    volumes:
+      - "${PWD}:/home/coder/project"
+    environment:
+      PASSWORD: ${PASSWORD}
+    command: --allow-http --no-auth
```
visualfanatic: You have `PASSWORD` env var, but it will not work, because the `--no-auth` flag is set.

nhooyr: Should add a comment here indicating that you need to remove --no-auth with a docker-compose override file or edit this one if you want to use $PASSWORD.

nhooyr:
@sr229 why is this obsolete? There is no merge conflict.

deansheather:
@nhooyr @sr229 it's not obsoleted, it just uses old flags that changed since v2

sr229:
I manually changed it to v2 flags. I'll work on this so it can be merged.

sr229:
CC @code-asher feel free to merge this in or cherrypick.

PR #698, opened 2019-05-20
Fix formatting issue

8sd:


code-asher:
Funnily enough, the doc is actually correct as it currently stands. We
need to add the comma here as well: https://github.com/cdr/code-server/blob/db39eacfa1fa15b3f4aaa44e6469f71d58b4d07b/packages/server/src/cli.ts#L31

doc/self-hosted/index.md
```diff
@@ -43,7 +43,7 @@ Options:
   --cert <value>
   --cert-key <value>
   -e, --extensions-dir <dir>  Set the root path for extensions.
-  -d --user-data-dir <dir>    Specifies the directory that user data is kept in, useful when running as root.
+  -d, --user-data-dir <dir>    Specifies the directory that user data is kept in, useful when running as root.
```
lsmoura: You have one extra space
```suggestion
  -d, --user-data-dir <dir>   Specifies the directory that user data is kept in, useful when running as root.
```

8sd:
@lsmoura thanks for the hint :)
@code-asher Ithanks, I didn't know there was this second help

code-asher:
@8sd our final check requires the commits to be signed, once that's done we'll merge this in!

8sd:
The commits are now verified

PR #701, opened 2019-05-20
Expose actions registry in IDE API

nhooyr:


code-asher:
We need `SyncActionDescriptor` exposed as well right?

nhooyr:
@code-asher yea my bad, exposed.

packages/ide-api/api.d.ts
```diff
@@ -1,6 +1,9 @@
 // tslint:disable no-any
 
 import { ITerminalService } from "vs/workbench/contrib/terminal/common/terminal";
+import { IWorkbenchActionRegistry } from 'vs/workbench/common/actions';
```
kylecarbs: Importing from here won't work right?

It will within the project, but importing inside another would leave these unresolved.

nhooyr: Not sure what you mean.

code-asher: Those files aren't included in the `ide-api` package so anything that imports it won't be able to resolve those types (only applies if you're using TypeScript of course). We'll want to fully encapsulate the API within the typings file before publishing a new version of the API.

nhooyr: Why not just use the vscode typings? Why are we wrapping?

code-asher: We'd have to include VS Code's code with the `ide-api` package. Since the files we import may import other files, it'd probably be easiest to just import all of VS Code, but it feels odd to include the entirety of VS Code just for the types.

The way VS Code does it is to have an entirely separate typings file like we're doing here and redefine all the types. Actually maybe we can just import their typings file and re-export some things from it?

code-asher: https://github.com/microsoft/vscode/blob/master/src/vs/vscode.d.ts

code-asher: Actually that wouldn't work because that doesn't include types for their internal API, just what they expose to extensions.

PR #719, opened 2019-05-27
Initialize Sail for the repository

### LINKED ISSUE #713: How to build latest releases
It seems that the build processes has since changed, and I can no longer simple `npm install` and `scripts/build.sh` anymore. How do I build the latest releases now?

roberthmiller:
Added [sail](https://github.com/cdr/sail)

The reasoning for using `codercom/ubuntu-dev` instead of `codercom/ubuntu-dev-node12` is because yarn was producing an install error because of some deprecated `.node` packages.

This currently does not resolve #713 but aims to eventually solve that by providing a consistent environment.

I would like some help since I am getting this error:
```
- scripts/build.sh 
+ [[ linux-gnu == \d\a\r\w\i\n* ]]
scripts/build.sh: line 26: TARGET: unbound variable
```

when trying to build code-server through Sail. Is there a dependency that I need to add to the sail Dockerfile?

### DIFF
--- a/.sail/Dockerfile
+++ b/.sail/Dockerfile
@@ -1,0 +1,11 @@
+FROM codercom/ubuntu-dev
+
+RUN curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+
+RUN sudo apt-get update && \
+  sudo apt-get install -y yarn && \
+  sudo apt install build-essential
+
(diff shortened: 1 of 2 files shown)


nhooyr:
@roberthmiller We really appreciate your work here but for now we've determined its best not to add Sail to this repository.

Issue #737, opened 2019-06-02
Digital ocean guide: ERR_CONNECTION_REFUSED


jorgt:
After following the DigitalOcean guide, I cannot seem to be connect to the code instance: ERR_CONNECTION_REFUSED in ubuntu droplet on DigitalOcean when entering the public IP address of the droplet. 

Running as root. Instance is started with a self signed certificate as per the other guide:

```
INFO  code-server v1.1119-vsc1.33.1
INFO  Additional documentation: http://github.com/cdr/code-server
INFO  Initializing {"data-dir":"/root/.local/share/code-server","extensions-dir":"/root/.local/share/code-server/extensions","working-dir":"/root/code-server1.1119-vsc1.33.1-linux-x64","log-dir":"/root/.cache/code-server/logs/20190602040823676"}
INFO  Starting webserver... {"host":"0.0.0.0","port":"80"}
INFO
INFO  Password: nnnnnnnnn
INFO
INFO  Started (click the link below to open):
INFO  https://localhost:80/
INFO
INFO  Starting shared process [1/5]...
WARN  stderr {"data":"(node:1491) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.\n"}
INFO  Connected to shared process
```

`wget` says:

```
# wget --verbose localhost:80
--2019-06-02 04:34:26--  http://localhost/
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://localhost/ [following]
--2019-06-02 04:34:26--  https://localhost/
Connecting to localhost (localhost)|127.0.0.1|:443... failed: Connection refused.
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:443... failed: Connection refused.
```

firewall seems inactive:

```
# sudo ufw status
Status: inactive
```

starting `code-server` on port 443 works in the sense that it connects, but then the websocket bombs out with status code 400.

deansheather:
Are you by any chance using a reverse proxy or maybe another http server (nginx, apache) on the machine?

jorgt:
No, straight up Ubuntu plus the code-server download and ssl key 

—
Jorg

erwinverdonk:
[...]

PR #740, opened 2019-06-02
Change documentation to use port 8443 instead of port 80

deansheather:

The current documentation asks users to run `sudo ./code-server -p 80` instead of just `./code-server`. This PR changes the documentation (including all firewall stuff) to use `./code-server` and therefore the default port of 8443.

I don't think we should encourage users to run this as sudo just to get it working on a root port., especially since the target audience of these guides are people with little server admin experience.

Additionally, this PR makes the insecure certificate guide apply to Firefox in addition to just Chrome, changes the AWS AMI image to 18.04 to match the DO guide (which doesn't specify an Ubuntu version, DO defaults to 18.04), and removes the AWS AMI ID as I think it's different for every AWS region.

#737


code-asher:
@deansheather we can merge this right in once the commit is signed. :key:

deansheather:
Yeah I can sign it, sure.

PR #741, opened 2019-06-02
Add port to redirect URL when server port is 80

deansheather:

This fixes HTTP to HTTPS redirects when people are running code-server on port 80 (default HTTP port).

Essentially, if code-server's port is 80 and is accessed over HTTP, code-server would redirect to HTTPS without adding `:80` to the end of the URL, resulting in the browser accessing over port 443.

I'm hesitant about this PR as it could cause problems with reverse proxies but unable to try multiple reverse proxy scenarios. **I don't think this should be merged until reverse proxies are tested and everything works as expected.**

#737

I've made another PR for changing the documentation that lead the user in the above issue to using port 80: #740

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -273,6 +273,7 @@ const bold = (text: string | number): string | number => {
 			},
 		},
 		password,
+		port,
 		httpsOptions: hasCustomHttps ? {
 			key: certKeyData,
 			cert: certData,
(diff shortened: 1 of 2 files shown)


code-asher:
I think what we might want to do here is if the port is 80, also listen to 443.

code-asher:
Or allow the user to specify multiple ports and just have an expectation that the user will add all the necessary ports for their setup.

deansheather:
I think the second choice is a better solution. We could also add a warning during startup that says "you might want to add port 443 to your server ports" if port 80 is supplied but not 443 and `--allow-http` is false.

code-asher:
Yeah, that sounds great to me.

deansheather:
Closed in favor of #819 which allows for listening on multiple ports.

PR #767, opened 2019-06-13
Improve Dockerfile

gengjiawen:
Use node:lts version and it already contains the latest yarn (1.16 for now)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,9 @@
-FROM node:10.15.1
+FROM node:lts
 
 # Install VS Code's deps. These are the only two it seems we need.
 RUN apt-get update && apt-get install -y \
 	libxkbfile-dev \
 	libsecret-1-dev
-
-# Ensure latest yarn.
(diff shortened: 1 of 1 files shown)


deansheather:
We could try bumping the Node version to the current LTS verson (`10.16` I think) and see how that goes, but I don't think it's a good idea to use `node:lts` as it changes whenever the LTS version changes.

I don't think code-server works with newer versions major versions of Node, so if `lts` became `12.x` our builds would break.

Additionally, since we use [nbin](https://github.com/cdr/nbin) to make code-server binaries we would also have to ensure that the same version was being compiled over there.

Dockerfile
```diff
@@ -1,13 +1,10 @@
-FROM node:10.15.1
+FROM node:lts
```
sr229: Do not use this kinds of tags, we want to pin versions as per best practices.

deansheather:
I think this should be closed unless there's a good reason for us to switch to `10.16`.

PR #794, opened 2019-06-19
Ensure directories are create on upload before searching files

HenningCash:
Uploading many files and directories to slow filesystems often caused a race condition where the `find` was executed before `mkdir` has finished. This two shouldn't be executed in parallel.  
This PR simply waits to execute `find` until `mkdir` succeeds.


packages/ide/src/upload.ts
```diff
@@ -133,20 +133,19 @@ export class Upload {
...
 	private async prepareDirectory(path: string, dir: IUploadableDirectory): Promise<void> {
-		await Promise.all([
-			promisify(exec)(`mkdir -p ${escapePath(path)}`).catch((error) => {
-				const message = error.message.toLowerCase();
-				if (message.includes("file exists")) {
-					throw new Error(`Unable to create directory at ${path} because a file exists there`);
-				}
-				throw new Error(error.message || `Unable to upload ${path}`);
-			}),
+		await promisify(exec)(`mkdir -p ${escapePath(path)}`).catch((error) => {
```
deansheather: I think this first catch block should be merged into the one below so we don't have two catches in our code.

HenningCash: Done

HenningCash:
Will do this as soon as I have access to my PGP key (not on my daily machine currently).

HenningCash:
Signed. Sorry it took so long. I hope the bug is still present 😅

PR #795, opened 2019-06-19
Revamp loading

code-asher:
#### Server
- [x] Remove Webpack on the server. Instead we just require using native Node plus `tsconfig-paths` which hooks in and tells Node how to resolve our aliases. Pretty much as clean as it gets, no loader or bundler required.
- [x] For forking VS Code processes, we just use their loader as-is without any changes. This means we don't have to patch anything, and things like language support should just work. It also means we should be able to plug in any version of VS Code at will.
- [x] Fix "undefined symbol" error when importing `spdlog` and other native modules.
- [ ] Might have some problems with dependencies nested in the `packages` directory since files get moved into an `out` directory.
- [ ] There might be a few things to check out due to no longer patching that we should work around in other ways (for example the extension host has some exit condition we had to comment out).

#### Client
- [ ] Greatly simplify client-side Webpack configuration (or use another bundler). It shouldn't touch anything from VS Code. Eventually we will be able to get rid of it entirely, but not in this PR. These changes should at least make it very simple to remove as well as make it build much, much faster.
- [ ] For loading VS Code, we just use their loader to run the main workbench file and let their loader do all the work, so hopefully no more patching and working language support. We will still have injected overrides.

#### General
- [x] If a pre-built VS Code isn't available, it'll just clone and build it which should make testing new versions locally much easier.
- [x] Optimized version of VS Code is bundled, so we should now have a minimal bundle size.
- [x] Build times are far faster now (around 7 seconds to build and nearly instant changes while watching).
- [x] Debugging will probably be much easier now.
- [ ] Bundling should exclude source maps (we could host them, that's what VS Code does).
- [ ] Update/write docs on developing. Should be much, much easier now. Essentially there will be just `yarn` plus one command to run for either building (`yarn build`) or jumping into development (`yarn watch`).


sr229:
My main concern is still performance and this is mostly by the GC of Node. By default GC is very lazy so we can force GC intervals to have a more optimized code base so "hopefully" we can finally run this on Raspberry systems with 512MB memory.

so by default I want to nominate the following flags to be automatically appended to the node binary invocation:

- ``--max-executable-size=192`` 
 - ``--max-old-space-size=256``
 - `` --max-semi-space-size=2 ``

code-asher:
Made obsolete by https://github.com/cdr/code-server/pull/857.

PR #799, opened 2019-06-21
Expose part of the VS Code API on the client

code-asher:
- [ ] The views aren't quite working; I'm not sure why. The container works fine but the individual views inside the container don't show.
- [ ] Also need to implement menu items for views (for providing actions on individual items).
- [ ] File system provider doesn't seem to work either although it looks correct to me.


code-asher:
I'm moving this into the new web branch.

PR #722, opened 2019-05-28
Fix 670

geiseri:
This addresses #670 by making the service worker register at the base of the mount.  I also changed it so that workbox is hosted from local and not a CDN.  This helps simplify CSP issues that pop up when mounting as part of a larger site.   There are still quite a few other ones once we start talking plugins but this is a start.

PR #834, opened 2019-07-07
Fix service worker path

### LINKED ISSUE #670: service-worker script wont load from non-root path
- `code-server` version: 1.939-vsc1.33.1
- OS Version: irrelevant

## Description
We are running `code-server` behind a reverse proxy that uses paths instead of distinct domains. e.g. `domain.com/ide/123456`. The service-worker script introduced in #154 will try to load the script from `domain.com/service-worker.js` instead of `domain.com/ide/123456/service-worker.js`.

~It should be sufficient~ (it's not) to remove the leading slash in `navigator.serviceWorker.register("/service-worker.js");`.  So either `navigator.serviceWorker.register("./service-worker.js");` or `navigator.serviceWorker.register("service-worker.js");`. But I haven't tested all scenarios, yet.

deansheather:

This addresses #670 by making the service worker register at the base of the mount. I also changed it so that workbox is hosted from local and not a CDN. This helps simplify CSP issues that pop up when mounting as part of a larger site. There are still quite a few other ones once we start talking plugins but this is a start.

Closes #670.

Note: this was originally opened in #722 by @geiseri but blocked by inactive user and no PGP signature.

### DIFF
--- a/packages/web/src/index.html
+++ b/packages/web/src/index.html
@@ -29,7 +29,7 @@
 		if ("serviceWorker" in navigator) {
 			// Use the window load event to keep the page load performant
 			window.addEventListener("load", () => {
-				navigator.serviceWorker.register("/service-worker.js");
+				navigator.serviceWorker.register('service-worker.js', {scope: './'});
 			});
 		}
 	</script>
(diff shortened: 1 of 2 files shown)


sr229:
Just read everything and it seems fine.

I do recommend adding Co-Authored-By so the OP of the PR can still get credited.

PR #835, opened 2019-07-07
Add failed authentication attempt logger

### LINKED ISSUE #214: Log failed login attempts
Would be cool if `code-server` was able log failed login attempts.

Maybe include ip/user-agent?

Originally suggested by @quietsy in the Discord

deansheather:
When `isAuthed()` is called and the password cookie is not what we expected, the failed login attempt is logged with the provided password, remote address, user agent, and timestamp.

To allow for logging failed attempts with a reverse proxy, the `--trust-proxy` argument has been added to trust the `X-Forwarded-For` header. This implementation of an `X-Forwarded-For` parser uses the last value in the list, therefore only trusting the nearest proxy.

This allows people to know if their code-server instance is getting attacked, and allows for writing a fail2ban filter for failed login attempts.

The log output looks like the following:
```
INFO  Failed login attempt {"password":"password","remote_address":"127.0.0.1","user_agent":"curl/7.65.1","timestamp":1562634192}
```

All fields are optional in the log line except the password and timestamp.

Closes #214.


deansheather:
Added a Fail2Ban filter conf file and a documentation page under `doc/security/` documenting it.

roflcoopter:
@deansheather I would love to use fail2Ban for code-server, but i cant figure out where the log-file is stored so ican mount it to my letsencrypt container which runs fail2ban. It doesnt seem to appear in the ```/root/.local/share/code-server``` folder

code-asher:
The failed login attempts are logged to stdout.

Issue #845, opened 2019-07-11
fetching .wasm does not work behind proxy


recro-lataille:
Im using HA proxy for ssl termination with client certificate validation. the fetch for f8f821a66b20e5801e35df5ec4565b30.wasm  (onigasm I believe) fails this was reproduced in the firefox browser by using the console fetch('f8f821a66b20e5801e35df5ec4565b30.wasm') the solution was tested by doing fetch('f8f821a66b20e5801e35df5ec4565b30.wasm', {credentials: 'same-origin') and the fetch is successful.  

## Steps to Reproduce

1. ssl terminating load balance validating client certs 
1. load the codeserver on the browser

code-asher:
Closing per comments in https://github.com/cdr/code-server/pull/846.

PR #846, opened 2019-07-11
Change Fetch for onigasm to use credentials

recro-lataille:
will cause credentials to be used to fetch onigasm so that it will work behind a proxy that is validating client certificates. 

Issue:
fetching .wasm does not work behind proxy #845

### DIFF
--- a/packages/vscode/src/fill/vscodeTextmate.ts
+++ b/packages/vscode/src/fill/vscodeTextmate.ts
@@ -22,7 +22,7 @@ target.Registry = class Registry extends vscodeTextmate.Registry {
 					const onigasm = require("onigasm");
 					const wasmUrl = require("!!file-loader!onigasm/lib/onigasm.wasm");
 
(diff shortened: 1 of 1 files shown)


recro-lataille:
can this be added to the next release that is currently in preview

deansheather:
Feel free to reopen this PR against the web branch and close this one. Since it's a very small change it should be OK.

nhooyr:
@code-asher cc

code-asher:
I tested this behind a proxy but it seems to work on Chrome and Firefox as-is on v2. Ah it looks like Firefox was changed in 61.0b13 (somewhere around June) to use `same-origin` by default to match the spec https://github.com/whatwg/fetch/pull/585.

nhooyr:
Can this be closed then?

code-asher:
I think so, unless we need to support older Firefox versions.

Issue #802, opened 2019-06-22
Add apple-touch-icon for ipad webapp


el:
After the release of the new iPadOS, code-server works pretty well on the iPad.

Configuring the web app for the ipad makes sense now:

https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

Here is how it looks right now:
![ECA17C4A-CC37-4E4B-9EAB-1E8F32601A10](https://user-images.githubusercontent.com/1488164/59969866-498f3880-9561-11e9-9e84-d64c79d0012b.jpeg)

Revengervn:
Also, another important point: 
Coder's copy,cut and paste doesn't work on the ipadOS safari which is weird. 
Theia and other VS code-based IDE work perfectly. 
I spent the entire day trying to figure out why but couldn't.

code-asher:
Was added by https://github.com/cdr/code-server/pull/1060.

PR #848, opened 2019-07-12
Modify webpack-pwa-manifest config to support iOS icon

admvx:
The auto-generated icon for Code Server on iPad and other iOS devices was partially broken and unreadable. This change makes use of the existing support in webpack-pwa-manifest to assign the real app icon to the appropriate meta tag in the app HTML file.

https://github.com/cdr/code-server/issues/802


nhooyr:
Thank you for your contribution @admvx.

However, WebpackPwaManifest isn't used in v2 so this PR is not necessary anymore. @code-asher should it be added back somewhere or is there a new mechanism?

admvx:
No sweat, thanks for following up! This would still be useful to me, so I have an update that doesn't use WebpackPwaManifest - PR on the way :)

PR #857, opened 2019-07-17
Rewrite code-server to use new web stuff

code-asher:
This uses the new web functionality VS Code has implemented which has made it possible to shrink the code base by nearly a factor of ten. It should be much easier to upgrade VS Code versions now and a lot of issues will be fixed. As the web portion of VS Code matures we'll be able to shrink out patch (and maybe even remove it entirely or almost entirely).

I think v2 is on par feature-wise with v1 except that you can't upload vsix files anymore. I'm not sure if we want to implement this right now or not. I also had to disable a few things like profiling extensions and extension tips, but we can probably get those working as well.

The next big thing will be to get language support working which I think should be pretty easy now (that's another big benefit to this rewrite).


sr229:
Can someone share the commit hash for VSCode, I suspect that might be for VS Online.

Dockerfile
```diff
@@ -1,4 +1,6 @@
-FROM node:10.15.1
+FROM node:10.16.0
+ARG codeServerVersion=docker
```
kylecarbs: We used to pull the binary from the docker image that gets built. Is that still done? If so the distributed binary would have `codeServerVersion=docker` as well.

src/channel.ts
```diff
@@ -0,0 +1,268 @@
...
+			extensionHostLogsPath: URI.file(path.join(this.environment.logsPath, "extension-host")),
+			globalStorageHome: URI.file(this.environment.globalStorageHome),
+			userHome: URI.file(this.environment.userHome),
+			extensions: await this.scanExtensions(locale),
+			os: OS,
+		};
+	}
+
+	private async scanExtensions(locale: string): Promise<IExtensionDescription[]> {
+		const translations = {}; // TODO: translations
```
kylecarbs: Will not having this break language packs?

src/cli.ts
```diff
@@ -0,0 +1,181 @@
...
+
+import { validatePaths } from "vs/code/node/paths";
+import { parseMainProcessArgv } from "vs/platform/environment/node/argvHelper";
+import { ParsedArgs } from "vs/platform/environment/common/environment";
+import { buildHelpMessage, buildVersionMessage, options } from "vs/platform/environment/node/argv";
+import pkg from "vs/platform/product/node/package";
+import product from "vs/platform/product/node/product";
+
+import { MainServer } from "vs/server/src/server";
+import "vs/server/src/tar";
```
kylecarbs: Might be helpful to add a comment as to why this is necessary.

src/cli.ts
```diff
@@ -0,0 +1,181 @@
...
+		case "wait":
+		case "disable-gpu":
+		// TODO: pretty sure these don't work but not 100%.
+		case "max-memory":
+		case "prof-startup":
+		case "inspect-extensions":
+		case "inspect-brk-extensions":
+			options.splice(i, 1);
+			break;
+	}
```
kylecarbs: Should we add a `default` here that throws? Otherwise we'd loop forever if they add a new option.

src/cli.ts
```diff
@@ -0,0 +1,181 @@
...
+
+	const version = `${(pkg as any).codeServerVersion || "development"}-vsc${pkg.version}`;
+	if (args.help) {
+		const executable = `${product.applicationName}${os.platform() === "win32" ? ".exe" : ""}`;
+		return console.log(buildHelpMessage(product.nameLong, executable, version, undefined, false));
+	}
+
+	if (args.version) {
+		return console.log(buildVersionMessage(version, product.commit));
```
kylecarbs: We should probably fix the logger so we can have pretty output

src/client.ts
```diff
@@ -0,0 +1 @@
+import 'vs/css!./media/firefox';
```
kylecarbs: It'd be helpful to document when `client.ts` is imported

src/insights.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as appInsights from "applicationinsights";
+
+export class TelemetryClient implements appInsights.TelemetryClient {
+	public config: any = {};
+
+	public channel = {
+		setUseDiskRetryCaching: (): void => undefined,
+	};
+
+	public trackEvent(options: appInsights.EventTelemetry): void {
```
kylecarbs: It'd be helpful to add a track source env var.

Then for images deployed on clouds like DigitalOcean we can set something like `TELEMETRY_SOURCE=digitalocean` then analyze it

kylecarbs:
Might be helpful to include some sort of auto-update feature in this version of code-server. :thinking:

src/api.ts
```diff
@@ -0,0 +1,284 @@
...
+			showErrorMessage: (message: string): void => {
+				notificationService.error(message);
+			},
+		},
+		workspace: {
+			registerFileSystemProvider: (scheme: string, provider: vscode.FileSystemProvider): IDisposable => {
+				return fileService.registerProvider(scheme, new FileSystemProvider(provider));
+			},
+		},
+	} as any;
```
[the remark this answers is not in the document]
code-asher: Originally it was because it only implements a tiny part of the API (so it would get "X property is missing" errors) but I think what I'll do instead is return `Partial<typeof vscode>`.

Dockerfile
code-asher: For the CI right? The CI doesn't use this Dockerfile at all; it builds using `codercom/nbin-centos` or `codercom/nbin-alpine` and the binary builds with the version specified in the Travis config.

For the `codercom/code-server` image we've been using this Dockerfile and building it manually then pushing it and it would just always say "development" for the version. Now we can specify the version when building it (`docker` is just a default value).

The ideal would probably be to have Travis push `codercom/code-server` so we can just use the version specified by Travis and not have to manage it manually. I'm not sure exactly how we want to orchestrate that so for this PR I just left the overarching build process the same.

src/channel.ts
code-asher: Yeah, the language packs don't work at all yet I believe. I haven't messed with any of that yet. Do you think we should try getting that in for this PR?

src/cli.ts
code-asher: This shouldn't loop forever because of the `(i--)` right? If we `throw` on `default` then the app would crash on the first non-matching option (there will be a bunch since this is removing a subset of options that we don't or can't support).

scripts/nbin-shim.js
```diff
@@ -0,0 +1,14 @@
+if (!global.NBIN_LOADED) {
```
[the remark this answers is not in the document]
code-asher: Yup, it's replaced when the shim is prepended to VS Code's bootstrap fork and to our main entry point. One alternative I thought of would be to use `__dirname` and then travel up until we find `out`.

Dockerfile
kylecarbs: Ah right. Makes sense

sr229:
Building [weekly snapshots](https://github.com/sr229/cdr-web-lkgr/releases) of this PR for everyone, so if you're that discerning someone who can't wait for a build out of this PR, I took initiative to make weeklies out of the change list.

sr229:
Initial testing of the new PR, looks like everything works as expected but I noticed some things:

- Seems that VSCode Web has the ability to walk through directories programmatically by using URL Parameters `?folder=/home/user/project`. However, this seems very broken right now.

- Extensions were properly being downloaded but they're not being loaded into the IDE. Even putting it on my own `.vscode` data dir does not see the extensions somewhat.

- Login via password is broken as well.

sr229:
Done more testing, looks like the binary bundle fails to resolve some VS Code files for the recent revision. Not sure if its nbin or me but getting the build artifacts might give an idea.

sshantt:
The --base-url flag works as expected and routes to <url>/base-url/login where a login page successfully shows up. 

Once the password is entered, the page routes to <url>/login - an invalid page - instead of <url>/base-url.

As with the previous stable build, base url works fine when authentication is turned off.

Let me know if you guys need any more details

sr229:
Looks like the Alpine target is overwriting the regular x86_64. Is this intended @code-asher?

code-asher:
@sr229 Nope, that's my bad. Will fix.

sshantt:
I re-downloaded https://github.com/sr229/cdr-web-lkgr/releases. Not sure what I did differently but `./Downloads/code-server/code-server --port 1999 --base-path '/vscode' ~/ --auth 'password'` is working perfectly fine now. Redirects to VSCode page when login is successful and indicates incorrect password otherwise.

Current Nginx Setup

```        
location /vscode/ {
            proxy_buffers 8 32k;
            proxy_buffer_size 64k;
            proxy_pass http://localhost:1999/;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            proxy_http_version 1.1;
            proxy_set_header Accept-Encoding gzip;
            proxy_set_header Upgrade $http_upgrade;
            proxy_set_header Connection "upgrade";
            proxy_set_header Host $host;
            proxy_read_timeout 20d;
            keepalive_timeout 6000;
        }
```

code-asher:
@sr229 Any info on what files are failing to resolve? I fixed an issue with the favicon not being included in the final build but there could be others.

@sshantt That's odd, the login page was hard-coded to make a request against `/login` so it shouldn't be working unless that build includes https://github.com/cdr/code-server/pull/857/commits/dde683d911ff2a0e0f7569aff7370a6e2d561b3b.

sr229:
@code-asher Looks like some workbench scripts are not loading properly - returning 404 for some reason. Previous builds were fine until 950bfce4200088fd5f8289ce8fc3b3b3216965ea

sshantt:
@code-asher Is there a way to find out what commits were included in that build?

code-asher:
I took a look but I didn't see any obvious way to see what commit the build is based on.
:thinking: In any case, if it's working then it seems likely the fix was included. :tada:

sr229:
Extensions now work properly on latest LKGR build (dc333d4)

visualfanatic:
Am I the only one having issues when trying to start a container using `Dockerfile` from this PR?

sr229:
@visualfanatic I don't think the Dockerfile works at all due to new infrastructure. You'll need to fix the dockerfile to be in line with the new infra.

visualfanatic:
@sr229 I tried to mess around with it, but it seems like issue in `scripts/	tasks.bash`. For some reason Docker is having issues with the scripts that are executed as part of `yarn build`, but it runs fine when executed on my macbook.

code-asher:
The `Dockerfile` should work:crossed_fingers:. I just tested it with:
```
docker build --network=host --build-arg vscodeVersion=1.36.1 .
docker run --rm --network=host -it $(docker images -q | head -n 1)
```
https://github.com/moby/moby/issues/19647 seems relevant.

LukasMasuch:
Thanks for the update. Unfortunately, the v2 version does not work with the nginx proxy configuration that I used for the previous version of code server. The web app is rendering fine, but the websocket cannot connect since it does not include the port. This is due to `authority` setting in the `head` of the index html not containing the port, see below:

```html
<meta id="vscode-workbench-web-configuration" data-settings='{"folderUri":{"$mid":1,"path":"/root","scheme":"code-server","authority":"localhost"},"remoteAuthority":"localhost","webviewEndpoint":"http://localhost:8443/webview/"}'>

<!-- Workaround to pass remote user data uri-->
<meta id="vscode-remote-user-data-uri" data-settings='{"$mid":1,"path":"/root/.config/code-server","scheme":"code-server","authority":"localhost"}'>
```
The `webviewEndpoint` is also set to the actual host/port that the code-server is started on and not the proxy address, but I don't know if this is causing any trouble as well. 

Our nginx configuration is a bit more complicated, but those are the proxy settings we currently set to proxy code-server:

```
proxy_set_header Host $host;
proxy_set_header X-Forwarded-Host $host;
proxy_set_header X-Forwarded-Server $host;
proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
proxy_http_version 1.1;
proxy_pass_request_headers on;

proxy_set_header Upgrade $http_upgrade;
proxy_set_header Connection $connection_upgrade;
proxy_store off;
proxy_redirect off;
```

sr229:
> Thanks for the update. Unfortunately, the v2 version does not work with the nginx proxy configuration that I used for the previous version of code server. The web app is rendering fine, but the websocket cannot connect since it does not include the port. This is due to `authority` setting in the `head` of the index html not containing the port, see below:
> 
> ```
> <meta id="vscode-workbench-web-configuration" data-settings='{"folderUri":{"$mid":1,"path":"/root","scheme":"code-server","authority":"localhost"},"remoteAuthority":"localhost","webviewEndpoint":"http://localhost:8443/webview/"}'>
> 
> 
> <meta id="vscode-remote-user-data-uri" data-settings='{"$mid":1,"path":"/root/.config/code-server","scheme":"code-server","authority":"localhost"}'>
> ```
> 
> The `webviewEndpoint` is also set to the actual host/port that the code-server is started on and not the proxy address, but I don't know if this is causing any trouble as well.
> 
> Our nginx configuration is a bit more complicated, but those are the proxy settings we currently set to proxy code-server:
> 
> ```
> proxy_set_header Host $host;
> proxy_set_header X-Forwarded-Host $host;
> proxy_set_header X-Forwarded-Server $host;
> proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
> proxy_http_version 1.1;
> proxy_pass_request_headers on;
> 
> proxy_set_header Upgrade $http_upgrade;
> proxy_set_header Connection $connection_upgrade;
> proxy_store off;
> proxy_redirect off;
> ```

You forgot to set the `--host` argument, hence its failing. Set the `--host` argument.

sr229:
We're getting some errors with some extensions. I'm listing a few:

 - `EditorConfig.EditorConfig` does not work, Stacktrace provided:
  ```
 Activating extension 'EditorConfig.EditorConfig' failed: Cannot find module 'lodash.get'.
```
- `icrawl.discord-vscode` does not work. Cannot find dist file.

**Comment by sr229** [comment]
Checking Remote Agent logs, we also get this:

```
[2019-08-08 21:49:10.557] [remoteagent] [info] Renamed to /home/chinodesuuu/.code-server/extensions/editorconfig.editorconfig-0.13.0
[2019-08-08 21:49:10.559] [remoteagent] [info] Installation completed. editorconfig.editorconfig
[2019-08-08 21:49:10.717] [remoteagent] [info] Extensions installed successfully: editorconfig.editorconfig
[2019-08-08 21:49:14.036] [remoteagent] [error] /src/build/code-server2d3637b69b09259ab9ecddfa05b73f2ac92c87be-vsc1.36.1-linux-x86_64-built/extensions/node_modules Cannot read file /src/build/code-server2d3637b69b09259ab9ecddfa05b73f2ac92c87be-vsc1.36.1-linux-x86_64-built/extensions/node_modules/package.json: undefined.
```

**Comment by sr229** [comment]
Testing on a actual remote VM: you can try a copy [here](https://www.katacoda.com/sr229/scenarios/code-server-playground)


Looks like WebSocket does fail to connect even if you set the host, it returns a `SSL_PROTOCOL_ERROR` in browser-side.

**Comment by sr229** [comment]
Confirmed b257c60 fixes @LukasMasuch's issue.

**Comment by code-asher** [comment]
I think those extension errors are due to the extensions being built incorrectly or incompletely in our marketplace since I get the same errors in code-server v1.

**Comment by sr229** [comment]
> I think those extension errors are due to the extensions being built incorrectly or incompletely in our marketplace since I get the same errors in code-server v1.

Explains the broken extensions, I'll try using the Microsoft marketplace to see if that's any better.

**Comment by leo60228** [comment]
The system clipboard doesn't work when using the Vim extension.

**Comment by sr229** [comment]
> The system clipboard doesn't work when using the Vim extension.

That isn't quite helpful, try digging some logs if possible.

**Comment by leo60228** [comment]
```
[Extension Host] TaskQueue: error: Error running task. Failed to handle key=p. not implemented.
log @ /codeserver/out/vs/workbench/workbench.web.api.js:18170
/codeserver/out/vs/workbench/workbench.web.api.js:127060 TaskQueue: Error running task. Failed to handle key=p. not implemented.
onDidNotificationChange @ /codeserver/out/vs/workbench/workbench.web.api.js:127060
/codeserver/out/vs/workbench/workbench.web.api.js:18170 [Extension Host] TaskQueue: error: Error running task. Failed to handle key=p. not implemented.
log @ /codeserver/out/vs/workbench/workbench.web.api.js:18170
```

**Comment by sr229** [comment]
Seems to be a limitation of VSCode Web itself and not us. Keep in mind VSCode Web is quite early stages (hence why VS Online is still in closed alpha).

**Comment by LukasMasuch** [comment]
I can also confirm that the preview 2 is working with my proxy settings (websocket is connecting). Thanks for the quick fix @code-asher. However, I discovered another issue that might be related: The webview iframe window that seems to be used within some parts, e.g. the extension description or the interactive window of the Python extension, is not loading because of an incorrect URL.  

image

image

This is most likely caused by the `webviewEndpoint` in the index html set incorrectly (`http://localhost/webview/` instead of `http://localhost:8091/vscode/webview/`) for my proxy setup. 

I have a found a few other extension-related issues, for those I will create separate issues.

**Comment by sr229** [comment]
> I can also confirm that the preview 2 is working with my proxy settings (websocket is connecting). Thanks for the quick fix @code-asher. However, I discovered another issue that might be related: The webview iframe window that seems to be used within some parts, e.g. the extension description or the interactive window of the Python extension, is not loading because of an incorrect URL.
> 
> image
> 
> image
> 
> This is most likely caused by the `webviewEndpoint` in the index html set incorrectly (`http://localhost/webview/` instead of `http://localhost:8091/vscode/webview/`) for my proxy setup.
> 
> I have a found a few other extension-related issues, for those I will create separate issues.

This is actually related since the last CLs should've fixed the webview problems.

**Comment by code-asher** [comment]
The webview URL is currently generated by using the host header plus the base path given to code-server when it starts, but we should probably patch VS Code to just use the current URL.

**Comment by shujaatak** [comment]
Does it fixes #233 or #791 ?

**Comment by sr229** [comment]
> Does it fixes #233 or #791 ?

Does not fix #233 and #791. You can't download directories or files in VSCode Web because its not part of the roadmap for it on VSO.

**Comment by sr229** [comment]
@code-asher Looks like there's a bug when performing a switch to folder ( Ctrl + K + O). It's not doing the query parameters properly (appends another `/` after `?folder=`, which causes errors, compared to previous behavior where it works properly).

**Comment by ViRb3** [comment]
Seems like Run Task doesn't do anything on preview 3

**Comment by ViRb3** [comment]
Pasting in terminal fails:
`Cannot read property 'triggerDataEvent' of undefined`

**Comment by sr229** [comment]
> Pasting in preview 3 terminal fails:
> `Cannot read property 'triggerDataEvent' of undefined`
> Should I be posting issues here or in the repo tracker?

Yes. You should.

I tested it with LKGR builds and I can't replicate this issue. Must be a build error in their end or there's something wrong with your system.

File: scripts/tasks.bash [hunk 10]
```diff
@@ -83,6 +83,17 @@ function build-code-server() {
 	cd "${buildPath}/out/vs/server" && yarn --production --ignore-scripts
 	rm "${buildPath}/out/vs/server/"{package.json,yarn.lock}
 
+	# onigasm 2.2.2 has a bug that makes it broken for PHP files so use 2.2.1.
+	# https://github.com/NeekSandhu/onigasm/issues/17
+	local onigasmPath="${buildPath}/node_modules/onigasm-umd"
+	rm -rf "${onigasmPath}"
+	git clone "https://github.com/alexandrudima/onigasm-umd" "${onigasmPath}"
+	cd "${onigasmPath}" && yarn && yarn add --dev onigasm@2.2.1 && yarn package
+	mkdir "${onigasmPath}-temp"
```
SuperSandro2000: You should add ``-p`` so that it does not fail when the directory already exists.

sabrehagen:
Could you push a Docker container build to DockerHub with the tag `2.preview.4-vsc1.37.0` please?

AjawadMahmoud:
I had difficulties building this image on Windows. Is this related by any chance to the "Linux builds only"?

I totally second @sabrehagen in asking you to provide the community with preview image via Docker Hub to allow more people to test and give feedback on your upcoming version.

code-asher:
I don't believe it's possible for this directory to exist because 
we remove and create the parent directory, copy the 
`package.json`, then run `yarn`. That means the only way for it to 
exist would be if there was a module called `onigasm-umd-temp` and 
the `package.json` listed it as a dependency. If that were the 
case, we wouldn't want to copy to or from it (since it would be a 
valid module) and failing on the `mkdir` would be preferable.

scripts/tasks.bash
code-asher: I don't believe it's possible for this directory to exist because
we remove and create the parent directory, copy the
`package.json`, then run `yarn`. That means the only way for it to
exist would be if there was a module called `onigasm-umd-temp` and
the `package.json` listed it as a dependency. If that were the
case, we wouldn't want to copy to or from it (since it would be a
valid module) and failing on the `mkdir` would be preferable.

SuperSandro2000: Anything in that list could fail and leave that directory behind.

SuperSandro2000:
> I had difficulties building this image on Windows. Is this related by any chance to the "Linux builds only"?

Yes. None of the build commands are ported to Windows and I don't think this has any priority as the Windows binaries are cross build.

> I totally second @sabrehagen in asking you to provide the community with preview image via Docker Hub to allow more people to test and give feedback on your upcoming version.

Just make sure to not tag it latest.

scripts/tasks.bash
code-asher: We remove the parent directory so once the `mkdir` line is reached that directory won't exist unless it was created by `yarn`.

https://github.com/cdr/code-server/blob/6737384d27f70f6dbc5cfcd1f87301ca6c27fa1f/scripts/tasks.bash#L69-L70

code-asher:
I pushed 2.preview.5-vsc1.37.0 to the Docker Hub. :tada:

sr229:
> I had difficulties building this image on Windows. Is this related by any chance to the "Linux builds only"?
> 
> I totally second @sabrehagen in asking you to provide the community with preview image via Docker Hub to allow more people to test and give feedback on your upcoming version.

If it says Linux only it means its "Linux-only", please keep that in mind, they don't have proper plans to port this until everything is working @AjawadMahmoud

sr229:
> I had difficulties building this image on Windows. Is this related by any chance to the "Linux builds only"?
> 
> I totally second @sabrehagen in asking you to provide the community with preview image via Docker Hub to allow more people to test and give feedback on your upcoming version.

If it says Linux only it means its "Linux-only", please keep that in mind, they don't have proper plans to port this until everything is working @AjawadMahmoud

scripts/tasks.bash
sr229: You can resolve this if this is a invalid suggestion @code-asher. I do agree this is unnecessary since we already define it at those parts.

sr229:
@code-asher We're getting CORS issues with `code-server` when using a marketplace API that isn't the default one.

Reproduction possible by doing this:

```sh
$ SERVICE_URL=https://marketplace.visualstudio.com/_apis/public/gallery ITEM_URL=https://az764295.vo.msecnd.net /opt/cdr/code-server --port=9000
```

code-asher:
@sr229 I set those variables and experimented with installing and 
uninstalling extensions and viewing their READMEs but it all seems 
to work as expected. Any idea what else I could do to replicate 
the issue?

sr229:
> @sr229 I set those variables and experimented with installing and uninstalling extensions and viewing their READMEs but it all seems to work as expected. Any idea what else I could do to replicate the issue?

I ran this on a Crostini environment with a `penguin.linux.test` domain. CrOS users will get failed XHR requests when the env var is set.

code-asher:
> Edit: looks like 400s are showing instead.

Looks like there is a difference in the IDs returned by our API vs Microsoft's. Ours returns something like `vscodevim.vim` while Microsoft's returns a UUID. So when an extension is installed with one API and then the API is swapped, it makes requests against the API using the ID obtained from the previous API which results in a 400 because the ID is invalid.

sr229:
> > Edit: looks like 400s are showing instead.
> 
> Looks like there is a difference in the IDs returned by our API vs Microsoft's. Ours returns something like `vscodevim.vim` while Microsoft's returns a UUID. So when an extension is installed with one API and then the API is swapped, it makes requests against the API using the ID obtained from the previous API which results in a 400 because the ID is invalid.

That's quite bad if that's the case. People who might actually use the Microsoft API might end up expecting parity if that's the case.

code-asher:
I'm not sure how we'd get the UUID aside from scraping Microsoft's 
marketplace which sounds like it could be a TOS violation.

It does seem likely that some people would switch to the Microsoft 
marketplace and then their existing extensions would never update 
unbeknownst to them. Maybe we should catch that "invalid id" error 
and warn the user that it looks like they have extensions from a 
different API and should re-install them if they want automatic 
updates.

Actually I'm not sure a simple re-install of the extensions would 
fix this. I should definitely double check that.

alippai:
Maybe @amandasilver or somebody from MS can push for creating an official extension API so there is no ToS violation. Did you asked them already or is this something off the table?

sr229:
> Maybe @AmandaSilver or somebody from MS can push for creating an official extension API so there is no ToS violation. Did you asked them already or is this something off the table?

Not gonna lie, I want Microsoft to consider VSCode-based deriveratives to be able to access the marketplace. This will be great ecosystem wise, and it'll prevent reinventing the wheel too much.

AjawadMahmoud:
@code-asher, you are the one! I gave the preview version an extensive test and I can very well say I've finally found the perfect IDE for the cloud. For the record. I ran it on a Python project with huge number of submodules. Had to install Python (v3.6), and pip and pylint separately (Da!). Installed Python extension and pylint was enabled automatically and had the full locale experience from the browser. Performance was perfect. 

Thanks for everyone working on this project.

sr229:
So new problem found while testing v2 again, looks like it doesn't replace the value of `?folder=` when switching to another folder, but rather appends to the `?folder=` that currently exists. I don't think that's intended.

code-asher:
@sr229 Curious, I thought I fixed that with 
https://github.com/cdr/code-server/pull/857/commits/6737384d27f70f6dbc5cfcd1f87301ca6c27fa1f. 
Would you check the commit you're running with --version?

sr229:
> @sr229 Curious, I thought I fixed that with [6737384](https://github.com/cdr/code-server/pull/857/commits/6737384d27f70f6dbc5cfcd1f87301ca6c27fa1f). Would you check the commit you're running with --version?

I'll try updating to see if that fixes it, might be me using outdated LKGR tree again

alippai:
Using setting `http.proxyAuthorization` results in `Refused to set unsafe header "Proxy-Authorization"` by workbench.web.api.js:1550

weikinhuang:
Hi, just tested this branch, with pusher/oauth2_proxy as my auth proxy, the websocket upgrade requests doesn't seem to be passing through the cookies, so in my case it never connects. Awesome work BTW

Request Headers:
```
GET wss://ide.test.example/?reconnectionToken=e5ac3033-a7d0-4bf5-9b5e-fa50ab3a9de9&reconnection=true&skipWebSocketFrames=false HTTP/1.1
Host: ide.test.example
Connection: Upgrade
Pragma: no-cache
Cache-Control: no-cache
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36
Upgrade: websocket
Origin: https://ide.test.example
Sec-WebSocket-Version: 13
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Sec-WebSocket-Key: Qbv10zJotk3uDIk53zBYqA==
Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits
```

sr229:
@code-asher Also might be new bug folder query param should redo into `&folder=/uwu/uwu` If there's already a first query param and code-server should be able to read wherever `folder` is located in the parameters.

sr229:
> Hi, just tested this branch, with pusher/oauth2_proxy as my auth proxy, the websocket upgrade requests doesn't seem to be passing through the cookies, so in my case it never connects. Awesome work BTW
> 
> Request Headers:
> 
> ```
> GET wss://ide.test.example/?reconnectionToken=e5ac3033-a7d0-4bf5-9b5e-fa50ab3a9de9&reconnection=true&skipWebSocketFrames=false HTTP/1.1
> Host: ide.test.example
> Connection: Upgrade
> Pragma: no-cache
> Cache-Control: no-cache
> User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36
> Upgrade: websocket
> Origin: https://ide.test.example
> Sec-WebSocket-Version: 13
> Accept-Encoding: gzip, deflate, br
> Accept-Language: en-US,en;q=0.9
> Sec-WebSocket-Key: Qbv10zJotk3uDIk53zBYqA==
> Sec-WebSocket-Extensions: permessage-deflate; client_max_window_bits
> ```

There will be work laid out to make auth from OAuth2 possible, it doesn't work yet, so please be patient. I suggest to not intercept calls from `/` since it will break websocket.

weikinhuang:
@sr229 I actually don't want to use any built in auth for code server, as I already have centralized auth on all my projects already with haproxy & oauth2_proxy acting as a reverse proxy. On the v1 branch this works perfectly fine, it is however broken on this branch.

sr229:
> @sr229 I actually don't want to use any built in auth for code server, as I already have centralized auth on all my projects already with haproxy & oauth2_proxy acting as a reverse proxy. On the v1 branch this works perfectly fine, it is however broken on this branch.

sr229:
> @sr229 I actually don't want to use any built in auth for code server, as I already have centralized auth on all my projects already with haproxy & oauth2_proxy acting as a reverse proxy. On the v1 branch this works perfectly fine, it is however broken on this branch.

Like I said, auth support is coming, you can plug your OAuth provider properly in the upcoming groundwork instead of intercepting it, which is 100% a bad idea anyways.

weikinhuang:
not true on the proxy being a bad idea, it's how i'm implementing zero trust in my case. as on top of that, I'm also using client cert auth. So the proxy is necessary regardless.

deansheather:
Opened a PR on sshcode cdr/sshcode#136 which allows for uploading a custom code-server binary. Seems to work out of the box with the most recent code-server preview.

sr229:
@code-asher not sure what happened but looks like CI is not outputting any `release/` directory. Is this intended?

```
0.01s$ ls $(pwd)
doc
Dockerfile
LICENSE
main.js
node_modules
package.json
README.md
scripts
src
tslint.json
typings
yarn.lock
dpl_0
```

code-asher:
Ah, I added a PACKAGE environment variable in an attempt to avoid 
the release to speed up the CI when we don't need it 
(unfortunately it seems to have a negligible effect).

sr229:
@code-asher not sure what happened but looks like applying the patches fails the script. 

https://travis-ci.com/sr229/cdr-web-lkgr/jobs/231172661#L550

code-asher:
@sr229 I updated VS Code again. I've been updating to master 
occasionally so I can try out the new changes faster but 
unfortunately there are often breaking changes in the patch. I 
hope it'll mellow out eventually.

sr229:
CC @deansheather looks like you guys need to rebase v2 after certifying it as GA. That conflicts are really concerning.

ddurham2:
@code-asher, how does this address #241?  I'm attempting to run the container in k8s under a path /code-server but cannot get it to work. Is the idea that I have to put an additional nginx reverse proxy after the LB just to make this happen? 

Thanks

code-asher:
@ddurham2 ahh yeah you still need some kind of reverse proxy. code-server needs to see / rather than /code-server.

#241 as I understand it was just about making code-server work behind a reverse proxy. This PR just added a `--base-path` flag that allowed code-server to properly make redirections.

sanjayme97:
> --base-path

how to give base path?

code-asher:
V3 no longer requires a base path so you won't need to pass it in.

If you're still using v2 and you're trying to access code-server at `<domain>/my/base/path` then you would do:

```
code-server --base-path /my/base/path
```

jrebey:
@code-asher I'm running 3.1.1 and while most of VSCode works, there are still things pointing to /static/* which are getting 404 since my code-server instance is being proxied via /vscode. Specifically, the Terminal will not load.

Here's one of the errors:
`DevTools failed to load SourceMap: Could not load content for https://mydomain/static/28e91ba70cd70fa9adf3f2e3e3b87631b5667ecf/lib/vscode/out/vs/workbench/services/extensions/worker/extensionHostWorker.nls.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE`

I'm running code-server behind an nginx proxy using the following:
```
proxy_pass http://code-server:8443/;
# Set WebSocket Proxy
proxy_set_header Upgrade $http_upgrade;
proxy_set_header Connection upgrade;
```
This could totally be something I did wrong... any ideas?

jrebey:
Looking at the vscode html returned... I see the following for all "/static/.*" links:

`href="./static/28e91ba70cd70fa9adf3f2e3e3b87631b5667ecf/src/browser/media/manifest.json"`

Shouldn't this be `href="static/..."`?

jrebey:
I was able to get this all running with the following changes to the nginx config... this is less than ideal though as it could interfere with locations in my / config...

```
location /static {
                include /config/nginx/oauth2-proxy.conf;
                proxy_pass http://code-server:8443/static/;
                proxy_set_header Host $host;
                proxy_set_header Accept-Encoding gzip;
                proxy_buffering off;
                proxy_http_version 1.1;
        }

        location /update {
                include /config/nginx/oauth2-proxy.conf;
                proxy_pass http://code-server:8443/update/;
                proxy_set_header Host $host;
                proxy_set_header Accept-Encoding gzip;
                proxy_buffering off;
                proxy_http_version 1.1;
        }
```

I'm sure I can simplify the config a bit with a location regex... but this was just for a test. I'm down to one remaining error... but that might be a vscode issue:

`DevTools failed to load SourceMap: Could not load content for https://mydomain/static/28e91ba70cd70fa9adf3f2e3e3b87631b5667ecf/lib/vscode/node_modules/vscode-textmate/release/main.js.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE`

code-asher:
`./static/` and `static/` are equivalent (both relative to the current URL). Are other static requests working fine? It's just the ones for `.map` files?

The `.map` files are only used for debugging code-server so if you don't plan on doing that there's no need to do anything about it.

If you do plan on doing that, it might have something to do with your oauth proxy? Browsers won't send cookies for `.map` requests unless the cookie's `SameSite` attribute is set to `lax` or `none`, so that could be the source of the issue.

You're right about the 404 error for the `vscode-textmate/release/main.js.map` file though. It seems to be a bug with `vscode-textmate`; the module doesn't contain that file when it's downloaded from npm so it looks like it might be incorrectly packaged.

jrebey:
It happens with any request under /static so the UI never loads since all the JavaScript and CSS are located there.

code-asher:
Bugger, I'm not sure what the issue could be in that case. What are you using to do the oauth? I'll try to run the same and see if I can replicate the issue.

jrebey:
I'm using https://github.com/oauth2-proxy/oauth2-proxy in docker.

I'm using the following containers:

linuxserver/code-server
linuxserver/letsencrypt
quay.io/oauth2-proxy/oauth2-proxy

FWIW, even when I completely disabled the oauth2-proxy on that handler (basically no auth_request directive) I'm still unable to get it working as a subfolder. The oauth stuff was only added once I had a working config by adding the /static and /update locations so I don't want those oauth lines to throw you off.

It didn't work with the following nginx config:

```
location /code-server {
  proxy_pass http://code-server:8443/;
  # Set WebSocket Proxy
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection upgrade;
}
```

When I switched to using a subdomain instead of subfolder, everything works as expected. So I'm not sure why code-server isn't trying to access https://host/code-server/static and instead tries to access http://host/static.

I have all 3 of these containers running within a separate network and the only host exposure is 80/443 in linuxserver/letsencrypt.

Let me know if a docker-compose.yml and some basic config files to replicate the issue in a github repo would help out and I can create that.

code-asher:
Ahh interesting, I'll try it out today with a subfolder and see what I can turn up.

code-asher:
Ohh wait is it the trailing slash? If you browse to: `https://host/code-server` vs `https://host/code-server/`?

With the former, a relative URL will result in `https://host/static` and with the latter a relative URL will result in `https://host/code-server/static`.

PR #885, opened 2019-08-05
fix: vscode.window.state in extHost always false

LeuisKen:
## Problem

I've an extension which rely on this `vscode.window.onDidChangeWindowState` extHost api, but since code-server did not implemented the `focus` and `blur` events in the custom `windowsService`, my extension doesn't work in the code-server. So I make this fix and it works.


LeuisKen:
Notice that your guys rewrite the code-server based on the vscode-remote version. 😂
I just test the code-server 2 preview.

deansheather:
@LeuisKen does this work in the v2 prerelease?

PR #894, opened 2019-08-10
Fix language

foxsir:


deansheather:
Hi, thanks for this PR! Unfortunately, we're already working on getting language support implemented in our v2 branch. The way we're doing v2 makes it much easier for us to implement language packs.

We don't plan on doing any releases between now and when we release v2 from my knowledge, so I'm closing this PR.

foxsir:
alright

Issue #108, opened 2019-03-07
Add environment variable support for existing runtime flags


Omeryl:
It was brought up in Discord by @PhilsLab that we should allow environment variables to be more easily configured in things like Docker containers. 

This is really simple, and already supported by the library used, we just need to figure out appropriate names. 

https://oclif.io/docs/flags is the documentation for the flags library, the key feature being `env`. 

We'd just need to add that to each of the flags in https://github.com/codercom/code-server/blob/master/packages/server/src/cli.ts#L20 like the following:
```ts
host: flags.string({ char: "h", default: "0.0.0.0", env: "HOST" }),
```

Thoughts are appreciated on this :).

kylecarbs:
LGTM. Feel free to throw up a PR for this, would be a great addition!

aslafy-z:
I'd love to see this land as I'm building a Helm Chart for `code-server`. It would consistently help setting settings and secrets.

Omeryl:
@aslafy-z I have a PR I'm going to submit tonight, I just had other stuff come up. Have to finish my workday first :).

amanvishnani:
Any update on this?

deansheather:
No updates, sorry. Feel free to open a PR!

zabawaba99:
As of https://github.com/cdr/code-server/commit/1697cc32a335b0510725621be6023cf2ec7cc497, the flags package is no longer used. That being said should still be relatively simple to allow people to use environment variables instead of options. I'll open a PR shortly to expose a few and we can have the discussion there as to what the others should be called.

nhooyr:
This should be filed against VS Code, we just use their parser.

PR #915, opened 2019-08-16
packages/server: Environment Variables

zabawaba99:
# Describe in detail the problem you had and how this PR fixes it
I need to be able to configure the server using environment variables because of how it's being deployed. The changes I made maintain the existing contracts (required variables and defaults) but allow users to set environment variables that are named just as the flags are.

# Is there an open issue you can link to?
#108

### DIFF
--- a/packages/server/src/cli.ts
+++ b/packages/server/src/cli.ts
@@ -20,27 +20,41 @@ const collect = <T>(value: T, previous: T[]): T[] => {
 	return previous.concat(value);
 };
 
+const split = (value: string | undefined): string[] => {
+	if (!value) {
+		return [];
+	}
+	return value.split(',');
+}
+
+const envVarBool = (value: string | undefined, defaultVal: boolean): boolean => {
+	if (value) {
+		return value.toLowerCase() === 'true'
+	}
+	return defaultVal;
+}
+
 commander.version(process.env.VERSION || "development")
 	.name("code-server")
 	.description("Run VS Code on a remote server.")
 	.option("--cert <value>")
 	.option("--cert-key <value>")
-	.option("-e, --extensions-dir <dir>", "Override the main default path for user extensions.")
-	.option("--extra-extensions-dir [dir]", "Path to an extra user extension directory (repeatable).", collect, [])
-	.option("--extra-builtin-extensions-dir [dir]", "Path to an extra built-in extension directory (repeatable).", collect, [])
-	.option("-d, --user-data-dir <dir>", "Specifies the directory that user data is kept in, useful when running as root.")
+	.option("-e, --extensions-dir <dir>", "Override the main default path for user extensions.", process.env.EXTENSIONS_DIR)
+	.option("--extra-extensions-dir [dir]", "Path to an extra user extension directory (repeatable).", collect, split(process.env.EXTRA_EXTENSIONS_DIR))
+	.option("--extra-builtin-extensions-dir [dir]", "Path to an extra built-in extension directory (repeatable).", collect, split(process.env.EXTRA_BULTIN_EXTENSIONS_DIR))
+	.option("-d, --user-data-dir <dir>", "Specifies the directory that user data is kept in, useful when running as root.", process.env.USER_DATA_DIR)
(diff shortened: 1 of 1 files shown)


zabawaba99:
I noticed the build is failing but don't really understand why. Any guidance on how to fix it?

packages/server/src/cli.ts
```diff
@@ -20,27 +20,41 @@ const collect = <T>(value: T, previous: T[]): T[] => {
...
-	.option("-H, --allow-http", "Allow http connections.", false)
-	.option("-P, --password <value>", "DEPRECATED: Use the PASSWORD environment variable instead. Specify a password for authentication.")
-	.option("--disable-telemetry", "Disables ALL telemetry.", false)
-	.option("--socket <value>", "Listen on a UNIX socket. Host and port will be ignored when set.")
-	.option("--trust-proxy", "Trust the X-Forwarded-For header, useful when using a reverse proxy.", false)
-	.option("--install-extension <value>", "Install an extension by its ID.")
-	.option("--bootstrap-fork <name>", "Used for development. Never set.")
+	.option("-N, --no-auth", "Start without requiring authentication.", envVarBool(process.env.NO_AUTH, false))
+	.option("-H, --allow-http", "Allow http connections.", envVarBool(process.env.ALLOW_HTTP, false))
+	.option("-P, --password <value>", "DEPRECATED: Use the PASSWORD environment variable instead. Specify a password for authentication.", process.env.PASSWORD)
```
deansheather: We want to avoid printing the password if it's supplied from the environment, so this line should be reverted.

packages/server/src/cli.ts
```diff
@@ -20,27 +20,41 @@ const collect = <T>(value: T, previous: T[]): T[] => {
...
-	.option("-d, --user-data-dir <dir>", "Specifies the directory that user data is kept in, useful when running as root.")
-	.option("--data-dir <value>", "DEPRECATED: Use '--user-data-dir' instead. Customize where user-data is stored.")
-	.option("-h, --host <value>", "Customize the hostname.", "0.0.0.0")
-	.option("-o, --open", "Open in the browser on startup.", false)
+	.option("-e, --extensions-dir <dir>", "Override the main default path for user extensions.", process.env.EXTENSIONS_DIR)
+	.option("--extra-extensions-dir [dir]", "Path to an extra user extension directory (repeatable).", collect, split(process.env.EXTRA_EXTENSIONS_DIR))
+	.option("--extra-builtin-extensions-dir [dir]", "Path to an extra built-in extension directory (repeatable).", collect, split(process.env.EXTRA_BULTIN_EXTENSIONS_DIR))
+	.option("-d, --user-data-dir <dir>", "Specifies the directory that user data is kept in, useful when running as root.", process.env.USER_DATA_DIR)
+	.option("--data-dir <value>", "DEPRECATED: Use '--user-data-dir' instead. Customize where user-data is stored.", process.env.DATA_DIR)
```
deansheather: Deprecated flags shouldn't have env variables. We don't want people to use them anymore.

packages/server/src/cli.ts
```diff
@@ -83,10 +97,6 @@ const bold = (text: string | number): string | number => {
 		readonly extraArgs?: string;
 	};
 
-	if (options.disableTelemetry) {
```
deansheather: If the user launches code-server with `--disable-telemetry`, the telemetry won't be disabled without this block. This should be reverted.

deansheather:
Regarding the broken build, it's not caused by any of your changes so it should be fine.

zabawaba99:
> This should be documented somewhere too, otherwise people won't know that they can use environment variables to configure code-server.

Where do you think this should go? I'm a bit unfamiliar with the project to not sure where the best place would be.

zabawaba99:
I'm fine either way. Does #857 address #108 as well?

sr229:
> I'm fine either way. Does #857 address #108 as well?

I don't think so. We use upstream VSCode compared to v1, but if you want you can rebase your PR to v2.

sr229:

Issue #914, opened 2019-08-16
Helm chart questions


paolomainardi:
Hi everyone and thanks for your hard work guys, the project is amazing.

As in the subject, i am developing an helm chart to contribute the project and i have some questions, let me start with the simplest one:

Do you see any contraindication to mount the persistence storage on `/home/coder` rather than the default path you suggest `/home/coder/project` ?

deansheather:
Thanks for working on a helm chart to contribute to the project! I don't have much experience with Kubernetes or Helm, but is there something wrong with the current [deployment file](https://github.com/cdr/code-server/blob/master/deployment/deployment.yaml) we have included in the repository?

paolomainardi:
Hi Dean! 

Nothing bad with that deployment file but as you can see everything is hardcoded, Ingress, persistence and security missing as well, all the aspect I want to cover with the helm chart, simpler to install and to adapt just by passing user’s custom vaiues.

deansheather:
In that case then, feel free to open a PR replacing the old chart. For the project storage, I'd recommend using `/home/coder/project` for consistency with the Docker image and possibly other deployment methods.

paolomainardi:
Ok thanks

sr229:
I think we @deansheather and @MichaelDesantis has worked already on this so I'll be closing this off.

InsOpDe:
What happened to the `chart`?

paolomainardi:
@InsOpDe it's my question too :)

MnrGreg:
Where does one find the helm chart now?

nhooyr:
No plans for a helm chart at the moment.

Is the docker container not enough?

paolomainardi:
Not enough of course, to deploy it on K8s cluster.
But the question is, why the one I contributed has been removed ?

PirateBread:
Hopefully not because extending this solution to be more useful steps on the toes of the enterprise paid version.

nhooyr:
> Hopefully not because extending this solution to be more useful steps on the toes of the enterprise paid version.

Not at all!

We only removed all these configs (k8s, helm, aws etc.) as no one was maintaining them in favour of a streamlined guide/docs. I think we could add a really simple k8s config but it's so easy to do when you know k8s, you just add a stateful set.

However, given the demand within the issue tracker for these configs, we were obviously wrong and I've created https://github.com/cdr/code-server/issues/2012 and to rectify this.

alexgorbatchev:
[...]

PR #917, opened 2019-08-18
#914 - New helm chart for v1

paolomainardi:
Refs to #914 

This PR adds an [helm chart](https://helm.sh/)  as another deployment option for Kubernetes.

I've made 1 opinionated decision for the persistence storage, by keeping persistent `/home/coder/project` and `/home/coder/.local/share/code-server` directory which basically contains all the vscode configurations, such as extensions, settings etc.

The helm chart includes an extensive README that should cover all the aspects to install and configure code-server.

The helm chart is now compatible with the latest stable version, but once the v2 will be out, the helm chart can be updated accordingly (it's just a matter of changing the port for now).

It still misses:

* [x] Define runtime variables from helm chart
* [x] Generate a k8s secret for PASSWORD
* [x] Make the args configurable


deployment/chart/README.md
```diff
@@ -0,0 +1,101 @@
+# code-server
```
deansheather: Could you please wrap this file at column 80? I've got another PR that changes the rest of the documentation to be wrapped at column 80.

deansheather: Don't worry about the table or codeblocks, just the paragraphs.

deployment/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+  name: {{ include "code-server.fullname" . }}
+  labels:
+    app.kubernetes.io/name: {{ include "code-server.name" . }}
+    helm.sh/chart: {{ include "code-server.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
```
sr229: Are you sure Recreate is the best strategy for this? I prefer a rolling update since it allows fallback when the container fails to deploy. Case and point, practice A/B

deployment/chart/values.yaml
```diff
@@ -0,0 +1,96 @@
...
+  ## code-server data Persistent Volume Storage Class
+  ## If defined, storageClassName: <storageClass>
+  ## If set to "-", storageClassName: "", which disables dynamic provisioning
+  ## If undefined (the default) or set to null, no storageClassName spec is
+  ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+  ##   GKE, AWS & OpenStack)
+  ##
+  # storageClass: "-"
+  accessMode: ReadWriteOnce
+  size: 8Gi
```
sr229: Setting 8Gi is a bit unsafe since we don't know what kind of persistence quota a dev/test K8s cluster has. Put it at least 2Gi or 1Gi (OpenShift quota)

deployment/chart/templates/pvc.yaml
```diff
@@ -0,0 +1,29 @@
...
+  accessModes:
+    - {{ .Values.persistence.accessMode | quote }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size | quote }}
+{{- if .Values.persistence.storageClass }}
+{{- if (eq "-" .Values.persistence.storageClass) }}
+  storageClassName: ""
+{{- else }}
+  storageClassName: "{{ .Values.persistence.storageClass }}"
```
sr229: have it OR to default for Minikube (I think Helm does it for you)

deployment/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,113 @@
...
+        command:
+          - sh
+          - -c
+          - |
+            chown -R {{ .Values.securityContext.runAsUser }}:{{ .Values.securityContext.fsGroup }} /home/coder
+        securityContext:
+          runAsUser: {{ .Values.volumePermissions.securityContext.runAsUser }}
+        volumeMounts:
+        - name: data
+          mountPath: /home/coder
```
sr229: mount to `/home/coder/project` not `/home/coder`. This will fail.

deployment/chart/templates/deployment.yaml
paolomainardi: Yes i agree with the theory, but when using a ReadWriteOnce volume more than one pod cannot mount the volume, that will cause the deployment to fail.
Another strategy would be to set `MaxSurge=0 and MaxUnavailable=1.` to the RollingUpdate strategy but it has the same effect of Recreate.

deployment/chart/values.yaml
paolomainardi: Ok

deployment/chart/templates/pvc.yaml
paolomainardi: If it's empty, it's the standard one.

deployment/chart/templates/deployment.yaml
paolomainardi: in this context is safe because we are just using a busybox image to fix the permissions on the entire disk

deployment/chart/README.md
paolomainardi: Done

deployment/chart/values.yaml
paolomainardi: Done

deployment/chart/templates/deployment.yaml
sr229: I don't think most dev/test K8s clusters has restrictions with PVCs, you might as well do ReadWriteMany if that's the case. If we find a lot of issues with this then we can revert back to RWO

paolomainardi: ReadWriteMany cannot be used for most of the cloud vendor disks, including AWS, GCP and Azure, unless you use a network file-system, this is the safest and most compatible choice.

paolomainardi: As you can see here: https://github.com/helm/charts/search?q=ReadWriteOnce&unscoped_q=ReadWriteOnce it's the common choice rather then https://github.com/helm/charts/search?q=ReadWriteMany&unscoped_q=ReadWriteMany which points directly to NFS-like charts.

deansheather:
Blocked on PGP signatures. I can merge this once you've signed all of your commits! :key:

paolomainardi:
@deansheather signed and squashed 👍

deansheather:
Thanks for contributing this!

paolomainardi:
Thanks for merging!

PR #927, opened 2019-08-22
Update README.md

mastappl:
D2C helps to deploy code-server to any cloud host on supported providers (DO, AWS, GCP, Vultr, UpCloud) or on own connected host with a singe-click. 
The deployment available with our free plan, so a user pays only for a host. It gives much more opportunity to deploy your solution to almost any host :) 
Check out the stack repo with a demo - https://github.com/d2cio/code-server-stack


mastappl:
> I don't feel like we should sponsor such when we already have a documentation outline for deploying code-server. We don't really want to compensate for a lot of use cases.

No problem. It just gives your users a wide choice of cloud providers (or a user's one) and it's totally free with one server of Code-server.

sr229:
Based on internal discussion we won't be including this PR in the master sadly. I hope you keep your integration updated however.

PR #931, opened 2019-08-25
Remove sensitive password logging

steven166:
- Passwords of failed login attempt are logged, these can still contains sensitive information. 
- Add external ip address to generated certificate, although it is self signed, the host verification will succeed when you trust the certificate. 
- Add basic auth in order to authenticate without session

No


sr229:
Blocked by GH-857. This will render this CL obsolete by 857. You can retarget against v2 instead once we have merged v2 to `master`.

packages/server/src/server.ts
```diff
@@ -94,7 +96,6 @@ export const createApp = async (options: CreateAppOptions): Promise<{
 						userAgent = userAgent.join(", ");
 					}
 					logger.info("Failed login attempt",
-						field("password", cookies.password),
```
nhooyr: I think it'd be ok to log the hash of the password versus the expected hash. See #952

packages/server/src/server.ts
```diff
@@ -137,9 +142,18 @@ export const createApp = async (options: CreateAppOptions): Promise<{
...
+				for(let ifName of Object.keys(networkInterfaces)){
+					let interfaces = networkInterfaces[ifName];
+					interfaces.filter(i => !i.internal && i.family === "IPv4").forEach(i => {
+						altNames.push(i.address);
+					});
+				}
 				const certs = await new Promise<pem.CertificateCreationResult>((res, rej): void => {
 					pem.createCertificate({
 						selfSigned: true,
+            altNames: altNames
```
nhooyr: I don't think this should be automatic. Since the certificate is self signed, your computer won't trust alt names anyway.

packages/server/src/server.ts
steven166: It is ok, but it wouldn't add any information other than "it doesn't match". Therefor it will only make the logs unreadable.

packages/server/src/server.ts
steven166: If you decide to trust the self-sigend certificate, the hostname/ip of the certificate still has to match in order to pass for a secure TLS connection.

nhooyr: Are you positive? We've been using this for a while now without any alt names and no issues have been reported.

steven166: I'm running this behind a proxy (on a different server) which reëncrypts the connection to code-server. The proxy trusts the self-signed certificate of code-server. But it is failing on the hostname verification, because the external ip of code-server is not matching the "localhost" which is in the certificate.

nhooyr: That's dangerous imo as you can get MITMed easily with a self signed certificate. It's best you generate the certificate yourself and add it as a trusted certificate to the proxy or use HTTP and move code-server onto the proxy server.

nhooyr: I guess my same reasoning applies in general to code-server generating self signed certificates. 🤔

sr229:

sr229:
Reopening to take over.

sr229:
I think this got deprecated by #835 so I'll just close this off.

Issue #565, opened 2019-04-22
Password in cookie stored in plain text!


StoneyEagle:
I don't think this needs any explaination.
![image](https://user-images.githubusercontent.com/45034970/56518770-95613980-6540-11e9-82b7-da41200d625e.png)

nol166:
related #515

Issue #515, opened 2019-04-17
Use bcrypt for password parameter


rcarmo:
It would be very helpful if the server could take `bcrypt` passwords, such as the output of `htpasswd -nB username` and use that username/cypher pair for authentication instead of having cleartext passwords all over the place... :)

lucacasonato:
I agree. At that point I think we should implement proper token based authentication. JWT and OAuth2 would be my preferred method. This would also enable different login providers later down the line.

boardfish:
+1 for OAuth2, especially with generic providers! I'll be using Keycloak if this becomes an option. i'll open a separate issue for that.

nhooyr:
The password is meant more of as a crutch, its best to put a reverse proxy in front for auth.

PR #952, opened 2019-09-05
Update password stored as a SHA256 hash in cookie

daTourist:
The authentication password is stored in _plain-text_ in a cookie named _password_. 

Some may not see this as a security issue for _automatically generated passwords_ (although bad practice) but it is definitively an issue when setting a _custom password_.

Therefore switching from a plain-text stored password to a SHA256 hash is a _quick first step_ in the right direction (aka. I agree it could be better, eg. salted hash).

I saw that there are pull requests providing other solutions (eg. #931 ) for authentication methods which were blocked by [GH-857](https://github.com/cdr/code-server/pull/857), but really think this pull request should be merged in for the next v1 release.

code-server-password-hash-poc

Issue #515 (Issue #565)

### DIFF
--- a/packages/app/browser/src/app.ts
+++ b/packages/app/browser/src/app.ts
@@ -2,6 +2,7 @@
 import { MDCTextField } from "@material/textfield";
 //@ts-ignore
 import { MDCCheckbox } from "@material/checkbox";
+import { createHash } from "crypto";
 import "material-components-web/dist/material-components-web.css";
 import "./app.scss";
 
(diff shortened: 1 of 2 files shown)


nhooyr:
This PR doesn't really add any security. Sure you can't tell what the original password is but you can still log into code-server with the hash which just acts as the password. Instead, we should hash the password server side and not store it in memory as plaintext and then hash the client side when comparing.

sr229:

PR #993, opened 2019-09-17
Updated Readme Fixed Online instructions

frank-dspeed:
Default port is 8443


code-asher:
Thanks for pointing this out! We changed the default port to 8080 in v2 but never got around to updating the Docker container. I modified our CI process to automatically deploy a v2 tag and updated the README to match.

PR #998, opened 2019-09-18
Move to GitHub Actions

sr229:

As discussed internally, we considered using GitHub Actions CI for automated releases. I implemented Binary releases for now.

**BREAKING CHANGE**: All Releases should now accompany a `CHANGELOG.md` WITHIN the repository.


sr229:
Blocking for now. Contacted GitHub Support since its a infrastructure issue within their CIs from my initial investigation.

nhooyr:
@sr229 what's the problem?

Also, I’d recommend circleci over github actions. github actions has no support for caching between builds which makes it very slow.

sr229:
> @sr229 what's the problem?
> 
> Also, I’d recommend circleci over github actions. github actions has no support for caching between builds which makes it very slow.

Seems the CI script fails when it shouldn't. Also regarding Docker caching - isn't that a paid plan for CircleCI?

nhooyr:
> Seems the CI script fails when it shouldn't. Also regarding Docker caching - isn't that a paid plan for CircleCI?

What does it fail with?

No, CircleCI is free for public projects.

sr229:
> > Seems the CI script fails when it shouldn't. Also regarding Docker caching - isn't that a paid plan for CircleCI?
> 
> What does it fail with?
> 
> No, CircleCI is free for public projects.

Encountering `invalid reference format` errors in the builds. Please refer on the GitHub CIs.

sr229:
CI is ready for deployment, we can replace Travis with Actions anytime now.

.github/workflows/build-commit.yml
```diff
@@ -0,0 +1,48 @@
...
+on: [push, pull_request]
+jobs:
+  build_amd64:
+    name: Build (AMD64 - CentOS 7 Target)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build code-server Binary
+        run: yarn && bash ./scripts/ci.bash
```
code-asher: We don't need to run yarn for any of this right? I think the scripts should execute properly without needing to run yarn since we download the cache which includes `node_modules`.

.github/workflows/build-commit.yml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build code-server Binary
+        run: yarn && bash ./scripts/ci.bash  
+        env:
+           VERSION: master
+           TARGET: 'linux'
+           VSCODE_VERSION: '1.38.1'
+           MINIFY: 'true'
```
code-asher: Could we skip minify/package for non-releases? Minifying can take quite some time and I'm not sure it's useful for a non-release.

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,75 @@
+name: Release
+on:
+  push:
+    tags:
+      - '*'
```
code-asher: Would this run on a tag on any branch? I'm thinking we'd want to only run for tags on the master branch.

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,75 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Build code-server Binary
+        run: yarn && bash ./scripts/ci.bash    
+        env:
+           MAJOR_VERSION: '2'
+           TARGET: 'linux'
+           VERSION: '${MAJOR_VERSION}.${GITHUB_SHA}'
```
code-asher: Should we use the tag for the version? I'm thinking the hash is going to be sort of an eyesore as far as versions go.

.github/workflows/build-release.yml
```diff
@@ -0,0 +1,75 @@
...
+           VSCODE_VERSION: '1.38.1'
+           MINIFY: 'true' 
+           PACKAGE: 'true'
+           
+      - uses: ncipollo/release-action@v1
```
code-asher: Would it be a good idea to add `draft: true` to these? I'd like to review the release before it actually gets published since a bunch of people will get emailed about it as soon as it gets made.

nhooyr:
Why is the build on pull_request and push?

.github/workflows/build-commit.yml
nhooyr: It's still a good idea to run `yarn` in case dependencies changed.

code-asher: We don't copy the `node_modules` into the Docker container so these would just go unused at least for the Linux builds. We could copy them in but the build script runs `yarn` anyway in the code-server directory so I don't think it's necessary.

And actually I misspoke a little, the cache only includes VS Code dependencies, not code-server. The caching situation could really use improvement...

code-asher: I think ideally we'd run yarn here with some sort of caching and copy it into the Docker container when building.

code-asher: Except we have to run yarn inside the Docker container to get the native modules built correctly.

nhooyr: I'd strongly recommend circleci over this GCP caching you guys are doing. It'll be much faster since the cache will be colocated with the circle ci infrastructure instead of fetching from gcp.

@sr229

code-asher: Yeah honestly that sounds like it'd be much nicer.

nhooyr:
Yea but why not just on push? Push applies to any commit pushed.

code-asher:
Ah I was under the impression that pull_request was the build merged with master but I could be wrong. I think for Travis at least that's what it is.

.github/workflows/build-commit.yml
nhooyr: Example for a Go library (but should apply similarly to code-server) at https://github.com/nhooyr/websocket/blob/27104bdeb98d981a335da9faeb0efd1b8d4f8b65/.circleci/config.yml

https://github.com/nhooyr/websocket/runs/215657031

nhooyr:
Yup, for travis it is but I don't think it's the case for GitHub.

https://help.github.com/en/articles/events-that-trigger-workflows#example-triggering-workflows-when-a-pull_request-is-assigned

> Triggered when a pull request is assigned, unassigned, labeled, unlabeled, opened, edited, closed, reopened, synchronize, ready_for_review, locked, unlocked or when a pull request review is requested or removed. See below to learn how this event works with forked repositories.

code-asher:
Ahhhh yup, certainly doesn't seem like it is.

code-asher:
I think the caching is going to be pretty critical especially since it's done manually right now so anything we can do to automate that would be clutch. I opened an issue to switch to CircleCI and set up caching here: https://github.com/cdr/code-server/issues/1004. CircleCI's caching mechanism looks pretty slick.

PR #1045, opened 2019-10-01
Create code-server.sls

korzq:
## Describe in detail the problem you had and how this PR fixes it
Their is no salt state for deployment using salt.

So I made one :P


korzq:
forgot to add some files, gonna make a new one. Apologies.

korzq:
I've also put the salt state in https://github.com/korzq/code-server-salt if it should be linked instead of being in the repo.

sr229:
Thanks for the conntribution, however, since we don't want to compensate for many use cases, we don't want to support such. I'll be closing this off for now. If its fine for you, just update your repo whenever we release something?

korzq:
Allright 👍

PR #1069, opened 2019-10-10
Add Quick-launch buttons

sr229:
Per-request by @kylecarbs - we're adding quick deploy buttons for GCP Cloud Run and Heroku. Azure will come later.


kondr1:
Can you try add button for now.sh https://zeit.co/docs/v2/more/deploy-button ?

PR #1074, opened 2019-10-14
ARM building support

### LINKED ISSUE #35: ARM binary release

sr229:
This allows building for ARM using Travis's new Multiarch support. This will supersede ALL PRs I had for ARM support.

Resolves #35 and requires cdr/nbin#25.


.travis.yml
```diff
@@ -15,14 +15,30 @@ before_install:
...
 jobs:
   include:
-    - name: "Linux build"
+    - name: "Linux build - x86"
       os: linux
+      arch: amd64
       dist: trusty
       env: TARGET="linux"
       if: tag IS blank
       script: scripts/ci.bash
```
code-asher: We can modify `ci.bash` to always build locally now right? We don't need the Docker images anymore.

.travis.yml
```diff
@@ -31,9 +47,17 @@ jobs:
...
+      os: linux
+      dist: trusty
+      arch: amd64
+      env: DOCKER_BUILD="true"
+      if: branch == master AND tag IS blank
+      script: docker build --build-arg githubToken="$GITHUB_TOKEN" --build-arg codeServerVersion="$VERSION" --build-arg vscodeVersion="$VSCODE_VERSION" -t codercom/code-server:"$TAG" -t codercom/code-server:v2 .
+    - name: "Docker build - ARM"
       os: linux
       dist: trusty
+      arch: arm64
```
code-asher: If we're building an arm64 version of the Dockerfile then we'll want to change the tag that automatically gets deployed as well otherwise it'll overwrite the amd64 version. Maybe set the `DOCKER_BUILD` environment variable to the arch or something like that.

.travis.yml
```diff
@@ -15,14 +15,30 @@ before_install:
 # Don't build on tags because we'll already have built the commit.
 jobs:
   include:
-    - name: "Linux build"
+    - name: "Linux build - x86"
```
code-asher: I'm thinking we should name these x86_64 to avoid potential confusion.

.travis.yml
code-asher: Oh unless we still need the Docker image for an older version of libc? And in that case would the image still work for arm64 since we build it in amd64?

code-asher: On codercom/nbin-centos I get:
```
ldd (GNU libc) 2.17
```

On ubuntu/trusty:
```
ldd (Ubuntu EGLIBC 2.19-0ubuntu6.15) 2.19
```

Although ubuntu/precise might work out:
```
ldd (Ubuntu EGLIBC 2.15-0ubuntu10.18) 2.15
```

sr229: To be honest we shouldn't need Docker anymore, we can just build via LXD and have the things emulated there if you wanna go local. There's really no point anymore to run Docker.

.travis.yml
sr229: we can append `-$(uname -m)` on the tag if needed.

.travis.yml
sr229: Will do 👍

.travis.yml
code-asher: I think that'd be dope.

code-asher: Why do we need LXD, by the way? Would it work to just build directly on the image Travis is providing us (maybe by targeting `precise` so we can drop to 2.15 to match VS Code)?

.travis.yml
```diff
@@ -8,21 +8,37 @@ before_install:
   - export MAJOR_VERSION="2"
   - export VSCODE_VERSION="1.39.2"
   - export VERSION="$MAJOR_VERSION.$TRAVIS_BUILD_NUMBER"
-  - export TAG="$VERSION-vsc$VSCODE_VERSION"
+  - export TAG="$VERSION-vsc$VSCODE_VERSION-$(uname -m)"
```
code-asher: I'm not sure we should append this here since that'd result in a separate tag for every version instead of having all the releases grouped under a single tag (e.g. https://github.com/cdr/code-server/releases/latest).

We could append it to the Docker tag only. There's also the v2 tag so we'd need to append to that as well. Should we leave `v2` pointing to amd64 and use `v2:arm64` for the arm64 version maybe? And then do the same for the version tag.

sr229: Not sure if I follow, so you're telling me that `$TAG` is used as well as a version string?

.travis.yml
sr229: if we still have Precise supported in Travis then I see no harm, we just need to get rid of the docker step.

.travis.yml
code-asher: Yup, `$TAG` is used to create the git tag and is where the releases will get attached, like this one here: https://github.com/cdr/code-server/releases/tag/2.1523-vsc1.38.1

With `uname -m` appended we'd end up with separate tags for every tar/zip instead of having them under one tag.

.travis.yml
code-asher: Ahhhh, but we'd still need something to build for Alpine. We could stick to Docker for that. The problem I was hoping to solve by getting rid of Docker is to more easily cache things in Travis (like the Yarn cache) since they're only inside the Docker container unless we mount or copy them out. Maybe I'm thinking about this all wrong?

code-asher: Resolving since we decided to stick with Docker.

SpinningK:
I want to turn my Android phone to a server and code on my ipad. I check this PR every day👻

mirsella:
me too @SpinningK, my rpi4. you can 1) activate notification when a new release is uploaded, and 2) you can also activate notification for all new message on this particular thread. hope it helps

sr229:
Might need to trial out ARM builds first in my LKGR repo first before merging this so if anyone's interested I'll be releasing builds there first.

Akarys42:
Hi, I hope this will be merged soon! @sr229 you said we can download build in your LKGR repo in the meantime, but I can't find it. Do you have any link please?

sr229:
> Hi, I hope this will be merged soon! @sr229 you said we can download build in your LKGR repo in the meantime, but I can't find it. Do you have any link please?

[Right here chief](https://github.com/sr229/cdr-web-lkgr). It'll take a while to run ARM builds though since we haven't built nbin to do package builds yet.

SpinningK:
Any progress?

deftdawg:
The Alpine Arm build failed because the environment is missing the ``libsecret-dev`` package...  I'm in the process if trying to manually build this now.

deftdawg:
On Android I can make it as far as ``yarn watch`` in the build instructions, but then gulp kills itself after exceeding the system file watch limits [ENOSPC](https://stackoverflow.com/questions/16748737/grunt-watch-error-waiting-fatal-error-watch-enospc).     Trying to ``yarn start`` it without having completed the compile also fails... 

Is there another way to get it to compile without watching all the files for changes (I assume that's what ``yarn watch`` is doing)?

Unfortunately, there's no easy way to override the ``fs.inotify.max_user_watches`` value on a non-rooted Android device.  

In any case, I'll try to build a Pi as well, if I get a chance. 😞

code-asher:
@deftdawg You might have better luck with the build instructions (https://github.com/cdr/code-server#build) instead of the development instructions. You won't need `yarn watch` if you're just building.

deftdawg:
@code-asher, I gave those instructions a try...  modifying package.json to point at @sr229 's nbin package, but unfortunately, it fails to find nbin and I'm too much of a node noob to understand why. 

These are my steps
```shell
git clone https://github.com/cdr/code-server.git
cd code-server
git checkout sr229/travis-arm
sed -ie 's=@coder/nbin": "^1.2.2=@coder/nbin": "github:sr229/nbin#sr229/travis-arm=' package.json
yarn build "1.39.2" "deftdawg-android-1.39.2"
```
And this is how yarn dies:
```
yarn run v1.16.0
$ yarn && yarn runner build 1.39.2 deftdawg-android-1.39.2
[1/4] Resolving packages...
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "linux" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning " > @coder/logger@1.1.8" has unmet peer dependency "@google-cloud/logging@^4.5.2".
warning "@coder/nbin > @coder/logger@1.1.11" has unmet peer dependency "@google-cloud/logging@^4.5.2".
[4/4] Building fresh packages...
success Saved lockfile.
$ cd ./scripts && node --max-old-space-size=32384 -r ts-node/register ./build.ts build 1.39.2 deftdawg-android-1.39.2
Error: Cannot find module '@coder/nbin'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Function.Module._load (internal/modules/cjs/loader.js:562:25)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/source/code-server/scripts/build.ts:1:1)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Module.m._compile (/source/code-server/node_modules/ts-node/src/index.ts:493:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/source/code-server/node_modules/ts-node/src/index.ts:496:12)
    at Module.load (internal/modules/cjs/loader.js:653:32)
```

Merith-TK:
@deftdawg interesting, it just states that it is missing a dependancy, try installing that dependancy and try again?

deftdawg:
> @deftdawg interesting, it just states that it is missing a dependancy, try installing that dependancy and try again?

@Merith-TK  yeah cool, like *how*?

I tried this:
```npm install "github:sr229/nbin#sr229/travis-arm" --loglevel verbose```
which installed after 64minutes on my phone, same error (both with and without -g global option).

I added:
```    "@google-cloud/logging": "^4.5.2",```
to packages.json's devDependencies and that made the warning about cloud logging go away, but still can't find nbin

I tried adding 
```    "@coder/nbin": "github:sr229/nbin#sr229/travis-arm",```
to both dependencies and devDependencies in scripts/packages.json, however that had no effect.

Feel like this should be a very easy thing to solve, but I don't really know _node_ very well and my google searches haven't turned up anything helpful...  just stuff like ``rm -rf node_modules`` which I have already tried.

**Comment by sr229** [comment]
There isn't much change in my ARM branch so don't point it there, I'll be working on a experimental distribution format for a bit for ARM since compiling it for the architecture requires a custom gcc toolchain.

Not a pleasant experience but I'll get around it soon.

**Comment by code-asher** [comment]
@deftdawg yeah I'd recommend just building from `master` and not modifying anything. You should be able to just do something similar to this:

```
git clone https://github.com/cdr/code-server.git
cd code-server
yarn build "1.39.2" "android"
node ./build/code-server*/out/vs/server/main.js
```

So basically you'll just run the compiled JavaScript files with node. You won't be able to package into a binary since nbin doesn't support that architecture yet.

Akarys42:
@code-asher I tried building from source like you said on my pi4, and I get a lot of errors because of missing gulp extensions. Is there a way to make sure sure I have all dependencies installed?

Akarys42:
Here is the build output I get while doing `sudo yarn build 1.39.2 arm` after adding a lot of missing packages 
CAC22BB7-3E9F-42D1-A5EA-641EB2EF693C

code-asher:
I think the build script swallows some error output. You could try running that command manually in the root VS Code directory and it might show you what's really going on: `yarn gulp compile-extensions-build --max-old-space-size=32384`.

code-asher:
You can also run `yarn` in the root VS Code directory when there are missing dependencies, although there should only be missing dependencies if the build is cancelled while it's installing because then on the next run it sees that `node_modules` exists and doesn't try to install again. Actually I'll fix that right now so it always runs `yarn`.

deftdawg:
@code-asher's fix commit is https://github.com/cdr/code-server/commit/015a99e87dab9c347d44f22eaaaf0a36a01b25e4... 

I'll cherry pick it to @sr229 's arm branch and have another go at attempting to build this with `nbin/master` this time.

code-asher:
@deftdawg this PR only changes the CI so there's no benefit in building against it locally. You could just build against master.

deftdawg:
@code-asher my bad, I'll have a go against master then.

Akarys42:
@code-asher I don’t really get more output with `yarn gulp compile-extensions-build --max-old-space-size=32384` (I also tried `yarn gulp compile-extensions-build`)
377B54FC-4C4A-499C-9377-644B187B9982

sr229:
@Akarys42 I was observing this behavior as well, along with a `Illegal Instruction` error. @coadler and I are gonna attempt it on a ARM64 VM and see if we can compile it.

Akarys42:
Nice, let me know if you can manage to build it!

>

deftdawg:
@Akarys42 I'm getting "SIGKILL" as well...  best I can figure it is the OS killing yarn because max-old-space-size is set to 32GB on my 4GB RPi4. 

image

I'm going to try with setting --max-old-space-size to 3096 (3GB) in package.json and see if that helps.

deftdawg:
Reducing max-old-space-size did not work.

I did however have success building it in a aarch64 alpine Docker container running under qemu-aarch64-static...  The x86-64 host machine had 64GB of RAM and I estimate the build required 10GB of memory to complete. 

The same machine only reported 2GB free with an arm Docker container, so will need to figure out how to get past that for an arm 32-bit build.

mirsella:
@deftawg make a 12GB swap can work ?

code-asher:
If I recall correctly I attempted a build on an ARM64 VM and it passed
once I updated Node to 10.15.1. Older versions of Node failed. The
master branch currently targets 12.14.0 but I haven't used that to build
on ARM64 yet so I can't confirm if it works.

deftdawg:
@code-asher 

Yarn complains about using node >8 or >11.   For my successful build, I used Alpine 3.9 inside a Docker container with qemu-aarch64-static registered to build Master on an x86-64.   Then I scp'ed the full code-server directory and started it on a Raspbery Pi 4 running aarch64.   

Newer versions of Alpine didn't work inside of Docker on x86-64, node will fail to compile while QEMU spits up (https://github.com/alpinelinux/docker-alpine/issues/48).

Screenshot of it running:

image

I'm not sure how we can get this built on Arm32v7 though, I think qemu-arm-static maybe has some 2GB memory limit or something on 32-bit...   I searched for Arm 32 machines with lots of ram and the only one I could find is an older QNAP NAS w/ a Coretex-A15 and DDR3 ram slots for up to 16gb.

sr229:
Closing since we found a better way now :3c

PR #1115, opened 2019-10-25
Add heartbeat file

### LINKED ISSUE #1050: [v2] discovering if user is actually using code-server (for idle check - can do in v1)
**Environment**:  A docker swarm running code-server development environments for multiple users simultaneously (teaching course).

**Previously**

code-server v1, will log to files in  ".cache/code-server/logs" added
a line every time the users browser contacts the running code-server.
This happens at least every 5 minutes (on the dot), while the user has
a open browser pointing to "code-server".

This let me implement a check looking at when these log files last
updated to determine when the user last had a browser open to
code-server.  This check could then shutdown the docker container
for users that have not used there code-server for some time.

That is we have implemented idle timeout for the docker instances.
This is vital for the system to keep things running smoothly for
the multi-user docker swarm.

_So far I have found no equivalent to implement idle timeout in
code-server v2._

**Suggestion.**

Provide a 'heartbeat' option/file, that is 'touched' (does not have to
write anything but it could, just have the file timestamp updated),
whenever "code-server" is contacted by the users browser.

This will make idle checking a lot easier too as it not need to search
for the latest log file, just the one heartbeat file.

However a log file file (as in v1) will be fine too.

Or prehaps you have some other suggestion.

PS:  Code-server v2 does generate logs whenever the user opens a file.  The log is for it attempting to find a 'git' repository for the file, and failing.
But it is not really a good idle/in-use indicator.

code-asher:
Fixes #1050.

@antofthy this will touch `~/.local/share/code-server/heartbeat` (or wherever the user data directory is located) once a minute as long as there is an active browser connection. No need to review the code (although I would welcome it if you feel like it) but I'm pinging you to verify this'll work out for your use case.

### DIFF
--- a/src/node/server.ts
+++ b/src/node/server.ts
@@ -469,6 +469,9 @@ export class MainServer extends Server {
 	private readonly proxyTimeout = 5000;
 
 	private settings: Settings = {};
+	private heartbeatTimer?: NodeJS.Timeout;
+	private heartbeatInterval = 60000;
+	private lastHeartbeat = 0;
 
 	public constructor(options: ServerOptions, args: ParsedArgs) {
 		super(options);
@@ -486,6 +489,7 @@ export class MainServer extends Server {
 	}
 
 	protected async handleWebSocket(socket: net.Socket, parsedUrl: url.UrlWithParsedQuery): Promise<void> {
+		this.heartbeat();
 		if (!parsedUrl.query.reconnectionToken) {
 			throw new Error("Reconnection token is missing from query parameters");
 		}
@@ -509,6 +513,7 @@ export class MainServer extends Server {
 		parsedUrl: url.UrlWithParsedQuery,
 		request: http.IncomingMessage,
 	): Promise<Response> {
+		this.heartbeat();
 		switch (base) {
 			case "/": return this.getRoot(request, parsedUrl);
 			case "/resource":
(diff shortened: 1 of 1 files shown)


antofthy:
That would be perfect.  Look forward to seeing this on the next code-server release.

antofthy:
Confirmed heartbeat working as expected in the last release  v2.1655-vsc1.39.2

shivanik6z:
currently, if I deploy code-server for the first time, till the time I don't open it heartbeat shows a very old time. Is it intended? shouldn't heartbeat time be the creation time?

antofthy:
Until it is updated heartbeat should either not exist or remain whatever it had been until then.

It is important that any monitoring also look at some other indicator that the user has only just started code-server, and perhaps has not opened there browser to connect to code-server.  In my own case we have another file that gets updated whenever the user starts up.

I also now have a log monitor that also updates that secondary file with date/time of last logged web activity for the users 'environment'.  This is different as code-server holds many of its web connections open.  As such code-server activity does NOT always mean new web-activity, that a proxy 'ingress' service can log, though code-server updated its heartbeat while such a connection is open.   Heartbeat is very important to monitor use of code-server!

NOTE Heartbeat is not strictly 'idle time', only a indication the user has code-server open on there browser.  Other software (Jupyter for example) does not report connections, only real idleness, which can cause users to become disconnected if they don't do anything substantial in there browser.

Issue #1117, opened 2019-10-27
Need new manual for nginx 


kondr1:
This doc request already was earlier but after update code-server i have `ERR_TOO_MANY_REDIRECTS` error again with my old configs. Whats change i must make to resolve this problem?

sr229:
Show us your nginx config so we know what's broken, thanks!

ecrode:
Had the same issue, here is a config that worked. After I figured out a config that worked, I fixed too many redirects by clearing site's cookies, right config was not enough.

server {
	listen 443 ssl;
	
	server_name ide.*;
	
	# ssl cert, in dockers linuxserver/letsencrypt format
	include /config/nginx/ssl.conf;

	client_max_body_size 0;

	location / {
		proxy_set_header   Host $host;
		proxy_redirect http://$host/ /;
		proxy_http_version 1.1;
		proxy_set_header Upgrade $http_upgrade;
		proxy_set_header Connection "upgrade";
		proxy_pass http://<ip>:8080;
	}
	
	location ~* \.(?:jpg|jpeg|gif|bmp|ico|png|swf)$ {
		proxy_set_header   Host $host;
		proxy_pass http://<ip>:8080;
		access_log off;
	}
}

sr229:
Requesting further information. Failure to reply means it was fixed in your end and I'll close this automatically after 3 days.

nhooyr:
We're removing the nginx manual.

PR #1156, opened 2019-11-07
Update server.ts

ecrode:
Change to avoid endless redirect whenever the password changes, while the old password is still set in a cookie.

Originally ran the container with a self generated password, which was accepted and recorded as a cookie. Updated the docker to use a custom password and a site started generating endless redirects between a home and a login directories, this has occurred due to an inability to clear the wrong password from memory.

Somewhat related, error #1117

### DIFF
--- a/src/node/server.ts
+++ b/src/node/server.ts
@@ -298,7 +298,10 @@ export abstract class Server {
 						return response;
 				}
 				if (!this.authenticate(request)) {
-					return { redirect: "/login" };
+					return { 
+			   			redirect: "/login",
+			   			headers: { "Set-Cookie": `password=` }
+			   		};
 				}
 				break;
 			case "/static":


code-asher:
I'm actually not super clear on how this fixes anything so if anyone has some insight I'd love to hear it. I thought maybe it had to do with a cookie being set to an invalid path or a dot-prefixed domain vs non-prefixed but I can't replicate the issue at all.

code-asher:
So I can trigger a loop if I redirect /login to /login/ since then the cookie will be set to /login instead of / which the root page can't read from. So clearing the cookie at / would solve the infinite loop although you still shouldn't be able to log in.

code-asher:
Eh nevermind ignore me that doesn't cause a loop either.

code-asher:
OK I think I figured it out we're all good. :smile:

Issue #1132, opened 2019-10-30
Modify default directory/workspace behavior


code-asher:
The current behavior uses this order:
1. Workspace query parameter.
2. Folder query parameter.
3. Last opened item (either workspace or folder).
4. Whatever was passed on the command-line (either workspace or folder).

Proposed behavior:
1. Workspace query parameter.
2. Folder query parameter.
3. Last opened item for that port (#1123)
4. Last opened item for any port.
5. Whatever was passed on the command-line.

The only remaining problem I can think of is that currently if you explicitly call code-server it'll favor the last directory instead of whatever you passed on the command line. It feels unexpected to me but we can't easily change that since if you run this in systemd for example you'd probably expect the opposite behavior when the service restarts (favor the last directory instead of what's passed to code-server).

pocka:
Using last opened workspace/directory, in other words, ignoring a path which a user specified explicitly is quite strange and frustrating at least to me. This prevents me to use code-server as an alternative for vscode because `code-server .` won't work.

Although we could open a specific directory by passing query parameters, I don't think it's a solution for this problem:

- We are still unable to specify an opening directory from a terminal.
- AFAIK not a few people using code-server in Chrome's app mode (in ChromeOS it's a "shortcut" but idk its correct name) which wouldn't allow us to edit the URL.

The port methods you proposed might work for the use-cases like #1123 , but not for the others.

The most simple way I think is to make persistent behavior an opt-in feature, like this:

```sh
# sorry for messy naming...
code-server ./dirA # opens ./dirA
code-server ./dirB # opens ./dirB
code-server ./dirC --prefer-last-opened # opens ./dirC
```

With this strategy, you can both 1) specify which directory to open and 2) use the last opened directory if it exists, though I'm not sure this could be implemented.

code-asher:
Yeah I was similarly frustrated when I was testing code-server locally but I wasn't sure it was a common thing because usually you would run code-server persistently on a remote machine, right? Is there a compelling use-case for spawning new instances of code-server on demand?

Soon we want to support opening things in existing code-server instances. That will use a flag so it'll be something like `code-server -r file/dir` and it'll use the running instance. So that might fulfill any need to run code-server on demand.

So I'm not sure yet if we should modify the default behavior since it means we'd have to make everyone aware they need to pass an extra flag (it's been my limited experience that users expect code-server to remember the last directory).

[...]

PR #1225, opened 2019-12-14
Updates manifest.json

TheHllm:
Improves usability by making use of fullscreen when installed as pwa.

### DIFF
--- a/src/media/manifest.json
+++ b/src/media/manifest.json
@@ -2,7 +2,7 @@
 	"name": "code-server",
 	"short_name": "code-server",
 	"start_url": ".",
-	"display": "standalone",
+	"display": "fullscreen",
 	"background-color": "#fff",
 	"description": "Run VS Code on a remote server.",
 	"icons": [{


sr229:
How does this changes anything? I think it's fine PWA-wise in `standalone mode`

TheHllm:
It makes it use "fullscreen" aka hides the url bar (on mobile browsers e.g. on a tablet) wich is so much better for use on the go.

nhooyr:
@code-asher standalone should work fine. This is a bug in the browsers, I'm going to revert.

code-asher:
Yeah it does work but fullscreen also hides the status bar which felt
like a more immersive editing experience to me. Is there an advantage to
having the status bar show persistently while using code-server?

nhooyr:
The original user mentioned the URL bar not being hidden.

The status bar should be shown imo as that's what normal apps do.

orditeck:
I don't know if my issue could be related to this but since `2.1698-vsc1.41.1`, I can't create multiple "shortcuts" from the same base URL in Chrome.

I was creating one shortcut per project, for example:
- https://code.mywebsite.com/?folder=/Sites/project1 = Shortcut opening in its standalone window
- https://code.mywebsite.com/?folder=/Sites/project2 = Shortcut opening in its standalone window
- https://code.mywebsite.com/?folder=/Sites/project3 = Shortcut opening in its standalone window

Now it always override the latest shortcut I made.

Before `2.1698`, `CrAppModeShortcutID` was always different = multiple shortcuts possible.
After `2.1698`, `CrAppModeShortcutID` is always the same = multiple shortcuts doesn't seem possible.

nhooyr:
I believe your problem is #1132

nhooyr:
Oh, nvm, that's interesting. I'm not sure what's going on. Can you file an issue?

PR #1275, opened 2020-01-13
Update the manifest start_url to point at index

AdaRoseCannon:
According to https://developer.mozilla.org/en-US/docs/Web/Manifest/start_url#Relative_URL

the start_url, if a relative URL like '.', should point to a path relative to the manifest.

After testing the install in Samsung Internet, this results in a broken link because it points to the folder containing the manifest:

`localhost:8080/static-336ee288886f5728e1191e9ada2aeba5eb0bbd76/out/vs/server/src/media/`

Which is the correct behaviour, but the expected behaviour is that it points to the root of the project.

An alternative change would be to: `../../../../../../`

After using `add to homescreen` in Samsung Internet on Android the window is pointing at the wrong path.

This fixes the issue by pointing to '/' instead of the manifest location '.'

image

### DIFF
--- a/src/media/manifest.json
+++ b/src/media/manifest.json
@@ -1,7 +1,7 @@
 {
 	"name": "code-server",
 	"short_name": "code-server",
-	"start_url": ".",
+	"start_url": "/",
 	"display": "standalone",
 	"background-color": "#fff",
 	"description": "Run VS Code on a remote server.",


code-asher:
Oh, nice catch. I assume `../../../../../../` would be better because
it's possible to host code-server at a path like `domain.com/vscode`?

AdaRoseCannon:
That would be the case, although setups like that i don't think are common since you lose the protections of running the 3rd party code on a subdomain.

In addition it would be very fragile and break if you were to change the level of nesting.

AdaRoseCannon:
An alternative would be to move the manifest.json to the root of the project and keep the '.' value for `start_url`.

Although I wouldn't know the best way to do that for this project.

code-asher:
It's true that scenario isn't common and I agree that subdomains are
superior but we have users with that setup and I don't want to break
anything for them. :)

Serving the manifest at the root seems like a solid idea. That's
actually how it used to be then I removed it to use the generic asset
URL without realizing this would break. This time we'll add a comment
saying not to remove it because of the start_url, haha.

AdaRoseCannon:
That sounds like a good idea, i'll close this R then and open an issue.

Issue #1303, opened 2020-01-28
Switch to proper semantic versioning


nhooyr:
We were discussing and the current version numbers are confusing as the build number doesn't communicate much.

cc @kylecarbs 

Major would be big changes, minor for new features and patch for bug fixes.

sr229:
I think we should also ditch the `-vsc-x.y.z` since it'll make it hard for automation software to cope up with latest releases. I think we can do that right now by editing the build script.

kylecarbs:
I agree the build number doesn't say much

PR #1310, opened 2020-01-28
scripts/build: ditch -vsc suffix

sr229:
This would fix some automation systems since keeping the vsc suffix with our new tag deployment system breaks some of the CIs most people use.

This should reproduce quite predictable builds like we did previously.

image

This compliments GH-1303.

### DIFF
--- a/scripts/build.ts
+++ b/scripts/build.ts
@@ -79,7 +79,7 @@ class Builder {
 
 		const vscodeSourcePath = path.join(this.outPath, "source", `vscode-${vscodeVersion}-source`);
 		const binariesPath = path.join(this.outPath, "binaries");
-		const binaryName = `code-server${codeServerVersion}-${target}-${arch}`;
+		const binaryName = `code-server-${codeServerVersion}-${target}-${arch}`;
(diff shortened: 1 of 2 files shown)


scripts/build.ts
```diff
@@ -231,7 +231,7 @@ class Builder {
 
 			const [productJson, packageJson] = await Promise.all([
 				merge("product", { commit, date }),
-				merge("package", { codeServerVersion: `${codeServerVersion}-vsc${vscodeVersion}` }),
+				merge("package", { codeServerVersion: codeServerVersion }),
```
code-asher: Not a big deal but this could be `{ codeServerVersion }`.

code-asher:
Also I think we should add a line to `--version` that shows the VS Code version.

scripts/build.ts
sr229: nice catch

sr229:
> Also I think we should add a line to `--version` that shows the VS Code version.

Point me to the right direction @code-asher, I don't think I know where.

sr229:
Just a little note I can't push to the repository at the moment, but I have the changes ready to merge.

nhooyr:
@code-asher says he already merged into #1338

PR #1318, opened 2020-01-29
Automatically push releases to GCS

deansheather:
The cron script doesn't work anymore with the new release architectures, it's a good time to switch to uploading automatically.

Don't know if there's an easy way to test it without creating a tag?

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -66,6 +66,16 @@ steps:
   when:
     event: tag
 
+- name: publish:gcs
+  image: plugins/gcs
+  settings:
+    source: gcs_bucket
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -66,6 +66,16 @@ steps:
   when:
     event: tag
 
+- name: publish:gcs
+  image: plugins/gcs
+  settings:
+    source: gcs_bucket
+    target: codesrv-ci.cdr.sh/
```
sr229: not sure if this is the correct bucket for this, I think this is the same we used for nbin AFAIK?

deansheather: This is the correct bucket

Issue #1340, opened 2020-02-05
Provide documentation for WSL


sr229:
People has been asking for Windows releases but since then we moved to Drone CI (#1258), we can no longer provide Windows-native builds. 

However, we can document deployment instructions for WSL 1 and WSL 2 just like we did over at Crostini, which I'll be doing shortly.

nhooyr:
We'll just provide windows releases as we remove nbin.

PR #1341, opened 2020-02-05
Add documentation for WSL/WSL2

sr229:
## Describe in detail the problem you had and how this PR fixes it

 I didn't go much into detail with WSL2 since it'll make the documentation really thorough, but I linked the respective MSDN documentation to guide the user what to do with their WSL2 environment

This should (hopefully) address concerns regarding people inquiring about Windows builds, which unofrtunately, we can't support at this moment.

This issue resolves GH-1340.


nhooyr:
Do not want to provide specific docs for WSL, this is out of scope.

PR #1346, opened 2020-02-06
Dockerfile: Combine two runs

SuperSandro2000:

The Dockerimages pulls to many layers and now it has one less.

no


nhooyr:
Thanks.

PR #1399, opened 2020-03-06
Optimize release Docker Image

SuperSandro2000:
This PR combines some RUNs to reduce the layer count and also really delte the apt cache in all layers.

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -16,8 +16,8 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/*
 
 # https://wiki.debian.org/Locale#Manually
-RUN sed -i "s/# en_US.UTF-8/en_US.UTF-8/" /etc/locale.gen
-RUN locale-gen
+RUN sed -i "s/# en_US.UTF-8/en_US.UTF-8/" /etc/locale.gen \
+  && locale-gen
 ENV LANG=en_US.UTF-8
 
 RUN chsh -s /bin/bash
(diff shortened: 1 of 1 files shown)


nhooyr:
This was intended for readability and debugging. I don’t think it’s a good idea to always decrease layers, they exist for a reason.

nhooyr:
There’s a bunch of old issues and PRs where this is discussed in detail. I do appreciate the pr @SuperSandro2000.

nhooyr:
I’ll add some comments to clarify this for anyone in the future.

SuperSandro2000:
Running ``RUN rm -rf /var/lib/apt/lists/*`` in its own layer rather then in the on ``apt-get update`` is run is useless and does not improve the overall download size in any way.

ci/release-image/Dockerfile
```diff
@@ -1,20 +1,28 @@
...
-RUN locale-gen
+RUN sed -i "s/# en_US.UTF-8/en_US.UTF-8/" /etc/locale.gen \
+  && locale-gen
 ENV LANG=en_US.UTF-8
 
 RUN chsh -s /bin/bash
 ENV SHELL=/bin/bash
 
-RUN apt-get install -y dumb-init sudo
-RUN apt-get install -y man procps vim nano htop ssh git
```
SuperSandro2000: For example having these two installs right after each other does not improve anything and just adds another layer which decreases performance.

nhooyr:
Yea in this specific case I was wrong, I'm sorry. The additional RUNs do not add to readability at all.

nhooyr:
Thanks!

Issue #1309, opened 2020-01-28
Allow forwarding ports via URL


nhooyr:
https://github.com/cdr/code-server/issues/512

So like `localhost:8080/port/8080` to access port 8080 view HTTP.

Rekrii:
Currently I do this via a reverse proxy (nginx or similar) on the host box to forward /8080 URL to :8080 the code-server instance, and then an nginx server in CS to host whatever the user (me) wants. Very neat and simple. Not sure how scaleable it is if you needed 10k+ ports. 

Comes down to if CS needs to be a swiss army knife of tools, or just a base IDE and you add stuff yourself (e.g. nginx as above).

sr229:
I think we'll have a lot of benefits having a port forwarding proxy since its been a long overdue request by the community for us. Though I'd prefer we do it as a VS Code extension since its more suitable to do it in  VS Code's own APIs AFAIK.

zpfei206:
@Rekrii Could you share the contents of the nginx.conf ? i am trounble on it now, Thans very much!

Rekrii:
@zpfei206 Sorry I was on holiday. See below:

I use something simple like this in an nginx instance running in code-server:
`server {
       listen 8080;
       server_name my.domain.com;
       root /config/workspace/MyProjectHTMLFiles/;
       index index.html;
}`

On the nginx reverse proxy i use: 
`location /8080/ {
      proxy_pass http://localhost:8080/;
      proxy_http_version 1.1;
      proxy_set_header Upgrade $http_upgrade;
      proxy_set_header Connection "upgrade";
}`

Also ensure all SSL config is secure; ciphers, etc are good (I use Qualsys SSL Labs test). 

As i mentioned above, this won't scale too well into a large number of ports. But I use this to test/host different prototype projects I'm working on - normally say 10 or so, so i can manually do this port forward setup for such a small number.

benz0li:
@Rekrii You might consider

```
location ~ /(\d+)/(.*) {
  proxy_pass http://localhost:$1/$2;
  proxy_http_version 1.1;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection "upgrade";
}
```

for general port forwarding.

kylecarbs:
Fixed once https://github.com/cdr/code-server/pull/1453 is merged.

benz0li:
Thanks for the great work on #1453.

[...]

PR #1453, opened 2020-03-24
HTTP proxy

code-asher:
This PR adds the ability to proxy HTTP ports through code-server.

To do this, you pass a domain on the command line. For example:

```
code-server --proxy-domain coder.com --proxy-domain test.coder.com
```

Domains can be in the format `coder.com` or `*.coder.com` and the flag can be repeated to add as many domains as you want.

When code-server encounters `[number].[domain]` in the host header it'll proxy to the port indicated by `[number]` if `[domain]` matches one of the provided proxy domains. For example, going to `8080.coder.com` in this example would proxy code-server to itself.

When authenticating the cookie will be set against the highest-level proxy. So in this example if you logged in at `8080.test.coder.com` the cookie would be set against `coder.com` since that's the highest level that matches in the provided list of domains. That way you only have to authenticate once.

**Edit**: `/proxy/[number]` will also work. No command-line flags are necessary to enable this.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -83,7 +83,7 @@ Now you can browse to `<port>.coder.com`. Note that this uses the host header so
 ensure your reverse proxy forwards that information if you are using one.
 
 ### Sub-paths
-Just browse to `/proxy/<port>`.
+Just browse to `/proxy/<port>/`.
 
 ## x86 releases?
 
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -65,6 +65,26 @@ only to HTTP requests.
 You can use [Let's Encrypt](https://letsencrypt.org/) to get an SSL certificate
 for free.
 
+## How do I access web services?
+code-server is capable of proxying to any port using either a subdomain or a
+subpath.
+
+### Sub-domains
+Set up a wildcard certificate for your domain and a wildcard DNS entry (or you
+can configure each subdomain individually for the ports you expect to use).
+
+Start code-server with the `--proxy-domain` flag set to your domain.
+
+```
+code-server --proxy-domain coder.com
+```
+
+Now you can browse to `<port>.coder.com`. Note that this uses the host header so
+ensure your reverse proxy forwards that information if you are using one.
+
+### Sub-paths
+Just browse to `/proxy/<port>`.
+
 ## x86 releases?
 
 node has dropped support for x86 and so we decided to as well. See
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -65,23 +65,29 @@ only to HTTP requests.
 You can use [Let's Encrypt](https://letsencrypt.org/) to get an SSL certificate
 for free.
 
-## How do I access web services?
+## How do I securely access web services?
 
 code-server is capable of proxying to any port using either a subdomain or a
-subpath.
+subpath which means you can securely access these services using code-server's
+built-in authentication.
 
 ### Sub-domains
 
-Set up a wildcard certificate for your domain and a wildcard DNS entry (or you
-can configure each subdomain individually for the ports you expect to use).
+You will need a DNS entry that points to your server for each port you want to
+access. You can either set up a wildcard DNS entry for `*.<domain>` if your domain
+name registrar supports it or you can create one for every port you want to
+access (`3000.<domain>`, `8080.<domain>`, etc).
+
+You should also set up TLS certificates for these subdomains, either using a
+wildcard certificate for `*.<domain>` or individual certificates for each port.
 
 Start code-server with the `--proxy-domain` flag set to your domain.
 
 ```
-code-server --proxy-domain coder.com
+code-server --proxy-domain <domain>
 ```
 
-Now you can browse to `<port>.coder.com`. Note that this uses the host header so
+Now you can browse to `<port>.<domain>`. Note that this uses the host header so
 ensure your reverse proxy forwards that information if you are using one.
 
 ### Sub-paths
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -66,10 +66,12 @@ You can use [Let's Encrypt](https://letsencrypt.org/) to get an SSL certificate
 for free.
 
 ## How do I access web services?
+
 code-server is capable of proxying to any port using either a subdomain or a
 subpath.
 
 ### Sub-domains
+
 Set up a wildcard certificate for your domain and a wildcard DNS entry (or you
 can configure each subdomain individually for the ports you expect to use).
 
@@ -83,6 +85,7 @@ Now you can browse to `<port>.coder.com`. Note that this uses the host header so
 ensure your reverse proxy forwards that information if you are using one.
 
 ### Sub-paths
+
 Just browse to `/proxy/<port>/`.
 
 ## x86 releases?
--- a/src/node/app/proxy.ts
+++ b/src/node/app/proxy.ts
@@ -49,14 +49,14 @@ export class ProxyHttpProvider extends HttpProvider implements HttpProxyProvider
     }
 
     // Ensure there is a trailing slash so relative paths work correctly.
-    const port = route.base.replace(/^\//, "")
-    const base = `${this.options.base}/${port}`
     if (this.isRoot(route) && !route.fullPath.endsWith("/")) {
       return {
-        redirect: `${base}/`,
+        redirect: `${route.fullPath}/`,
       }
     }
 
+    const port = route.base.replace(/^\//, "")
+    const base = `${this.options.base}/${port}`
     const payload = this.doProxy(route, request, response, port, base)
     if (payload) {
       return payload
--- a/src/node/app/proxy.ts
+++ b/src/node/app/proxy.ts
@@ -6,6 +6,10 @@ import * as querystring from "querystring"
 import { HttpCode, HttpError } from "../../common/http"
 import { HttpProvider, HttpProviderOptions, HttpProxyProvider, HttpResponse, Route } from "../http"
 
+interface Request extends http.IncomingMessage {
+  base?: string
+}
+
 /**
  * Proxy HTTP provider.
  */
@@ -24,6 +28,12 @@ export class ProxyHttpProvider extends HttpProvider implements HttpProxyProvider
     super(options)
     this.proxyDomains = proxyDomains.map((d) => d.replace(/^\*\./, "")).filter((d, i, arr) => arr.indexOf(d) === i)
     this.proxy.on("error", (error) => logger.warn(error.message))
+    // Intercept the response to rewrite absolute redirects against the base path.
+    this.proxy.on("proxyRes", (response, request: Request) => {
+      if (response.headers.location && response.headers.location.startsWith("/") && request.base) {
+        response.headers.location = request.base + response.headers.location
+      }
+    })
   }
 
   public async handleRequest(
@@ -41,14 +51,15 @@ export class ProxyHttpProvider extends HttpProvider implements HttpProxyProvider
     }
 
     // Ensure there is a trailing slash so relative paths work correctly.
-    const base = route.base.replace(/^\//, "")
-    if (isRoot && !route.originalPath.endsWith("/")) {
+    const port = route.base.replace(/^\//, "")
+    const base = `${this.options.base}/${port}`
+    if (isRoot && !route.fullPath.endsWith("/")) {
       return {
-        redirect: `/proxy/${base}/`,
+        redirect: `${base}/`,
       }
     }
 
-    const payload = this.doProxy(route.requestPath, route.query, request, response, base)
+    const payload = this.doProxy(route, request, response, port, base)
     if (payload) {
       return payload
     }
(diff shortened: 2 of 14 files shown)


kylecarbs:
We should add some docs on this (prob throw it in the README too). Super cool feature.

kylecarbs:
We should make sure that no caching on CDNs can take place: https://support.cloudflare.com/hc/en-us/articles/115003206852-Understanding-Origin-Cache-Control

kylecarbs:
If I try to connect to an invalid port, `code-server` crashes with:
```
Error: connect ECONNREFUSED 127.0.0.1:8080
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1129:14) {
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 8080
}
```

t-d-d:
Is it possible to use a (sub-)path instead or as well as the host name?

benz0li:
@t-d-d See https://github.com/cdr/code-server/issues/1309#issuecomment-605286385

> [...] I didn't mention it in the PR but /proxy/<number> will also work.

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+  public readonly proxyDomains: string[]
+  private readonly proxy = proxy.createProxyServer({})
+
+  /**
+   * Domains can be provided in the form `coder.com` or `*.coder.com`. Either
+   * way, `<number>.coder.com` will be proxied to `number`.
+   */
+  public constructor(options: HttpProviderOptions, proxyDomains: string[] = []) {
+    super(options)
+    this.proxyDomains = proxyDomains.map((d) => d.replace(/^\*\./, "")).filter((d, i, arr) => arr.indexOf(d) === i)
```
nhooyr, line 29: What is the filter and indexOf doing here?

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+    })
+  }
+
+  public async handleRequest(
+    route: Route,
+    request: http.IncomingMessage,
+    response: http.ServerResponse,
+  ): Promise<HttpResponse> {
+    const isRoot = !route.requestPath || route.requestPath === "/index.html"
```
nhooyr, line 44: Seen this duplicated across the codebase, a helper would be nice.

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+      throw new HttpError("Unauthorized", HttpCode.Unauthorized)
+    }
+
+    // Ensure there is a trailing slash so relative paths work correctly.
+    const port = route.base.replace(/^\//, "")
+    const base = `${this.options.base}/${port}`
+    if (isRoot && !route.fullPath.endsWith("/")) {
+      return {
+        redirect: `${base}/`,
```
nhooyr, line 58: Why not redirect to fullPath minus the trailing slash?

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+    }
+
+    const payload = this.doProxy(route, request, response, port, base)
```
nhooyr, line 62: Why does this take a response but also return a payload?

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+    }
+
+    // REVIEW: Absolute redirects need to be based on the subpath but I'm not
+    // sure how best to get this information to the `proxyRes` event handler.
+    // For now I'm sticking it on the request object which is passed through to
+    // the event.
+    ;(request as Request).base = base
+
+    const hxxp = response instanceof http.ServerResponse
```
nhooyr, line 171: `hxxp` is confusing, how about `isHTTP`?

src/node/entry.ts
```diff
@@ -118,6 +124,7 @@ const main = async (args: Args): Promise<void> => {
 
   if (typeof sshPort !== "undefined") {
```
nhooyr, line 125: Why do we have a SSH server?

src/node/app/proxy.ts
code-asher: That's filtering out items that were already seen in the array (making it unique).

src/node/app/proxy.ts
code-asher: Oh yeah very good point. Will do.

src/node/app/proxy.ts
code-asher: I might not be following what you mean by minus the trailing slash. We need the trailing slash to make relative paths on the page work (for example VS Code loads JS by doing `./path/to/file.js`. Or do you just mean to do this:
```
redirect: `${route.fullPath}/`,
``` 
If so that makes sense to me.

src/node/app/proxy.ts
```diff
@@ -0,0 +1,190 @@
...
+    if (!this.authenticated(request)) {
+      // Only redirect from the root. Other requests get an unauthorized error.
+      if (isRoot) {
+        return { redirect: "/login", query: { to: route.fullPath } }
+      }
+      throw new HttpError("Unauthorized", HttpCode.Unauthorized)
+    }
+
+    // Ensure there is a trailing slash so relative paths work correctly.
+    const port = route.base.replace(/^\//, "")
```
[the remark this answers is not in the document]
code-asher: Yup, `doProxy` takes a base path that it strips out just before sending the request through the proxy.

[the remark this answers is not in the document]
code-asher: Which is the `base` var just below this line.

src/node/app/proxy.ts
code-asher: Yeah I don't really like this either. `http-proxy` takes the response and deals with it by itself which is what the response is needed for. But the proxy might want to do a redirection or something in which case it needs to return a payload.

Oh maybe we can just return proxy information as part of the payload instead. Essentially moving `doProxy` out of `proxy.ts` and into `http.ts`.

src/node/entry.ts
code-asher: I believe this has something to do with the Chrome extension that was planned but I'm not 100% sure on the status of that. @kylecarbs @wbobeirne

wbobeirne: Yes, this was for the browser extension. That's pretty much on ice at this point, so I think it's probably best for safety / code cleanliness that we take it out since nothing is using it right now, and it's probably a liability to have such a powerful thing up by default. Or at the very least, disable by default.

Would be happy to make a PR to remove that if we're definitely not going to pursue the extension, but I'll defer to mr @kylecarbs on that.

src/node/app/proxy.ts
nhooyr: Let's use `[...new Set(array)]` instead.

code-asher: I like that; I might just keep it a set since I suppose there's no need for it to be an array.

doc/FAQ.md
```diff
@@ -65,6 +65,29 @@ only to HTTP requests.
...
 for free.
 
+## How do I access web services?
+
+code-server is capable of proxying to any port using either a subdomain or a
+subpath.
+
+### Sub-domains
+
+Set up a wildcard certificate for your domain and a wildcard DNS entry (or you
```
kylecarbs, line 75: We should also point out that you can manually set records for `3000.<domain>`, `8080.<domain>`, etc. Cloudflare doesn't support wildcard domains with TLS

kylecarbs:
@cmoog I'd love your thoughts on this. Especially in relation to the docs/setup.

doc/FAQ.md
code-asher: That's actually what I meant by the text in the parenthesis but I don't think it was very clear. What about something like:

> You will need a DNS entry that points to your server for each port you want to access. You can either set up a wildcard DNS entry for `*.<domain>` if your domain name registrar supports it or you can create one for every port you want to access (`3000.<domain>`, `8080.<domain>`, etc). 

> You should also set up TLS certificates for these subdomains, either using a wildcard certificate for `*.<domain>` or setting up individual certificates for each port.

doc/FAQ.md
```diff
@@ -65,6 +65,29 @@ only to HTTP requests.
 You can use [Let's Encrypt](https://letsencrypt.org/) to get an SSL certificate
 for free.
 
+## How do I access web services?
+
+code-server is capable of proxying to any port using either a subdomain or a
+subpath.
```
cmoog, line 71: Not sure that this sufficiently explains the motivations behind this, namely keeping other ports private/hidden. Might make it harder for someone to understand why this is useful given that most setups at the moment likely expose all ports. What about something like: 

```
## How do I keep other web services on my server password protected?
```

src/node/app/proxy.ts
nhooyr: Yea, so you don't have to calculate the full path yourself.

nhooyr: yea meant plus sry

src/node/app/proxy.ts
nhooyr: I like that.

doc/FAQ.md
code-asher: That makes sense. How do you feel about:

> How do I securely access web services?

And then below that:

> code-server is capable of proxying to any port using either a subdomain or a path which means you can access these services using code-server's authentication.

cmoog: Yeah that looks better to me 👍

Grimeton:
> This PR adds the ability to proxy HTTP ports through code-server.
> 
> To do this, you pass a domain on the command line. For example:
> 
> ```
> code-server --proxy-domain coder.com --proxy-domain test.coder.com
> ```
> 
> Domains can be in the format `coder.com` or `*.coder.com` and the flag can be repeated to add as many domains as you want.
> 
> When code-server encounters `[number].[domain]` in the host header it'll proxy to the port indicated by `[number]` if `[domain]` matches one of the provided proxy domains. For example, going to `8080.coder.com` in this example would proxy code-server to itself.
> 
> When authenticating the cookie will be set against the highest-level proxy. So in this example if you logged in at `8080.test.coder.com` the cookie would be set against `coder.com` since that's the highest level that matches in the provided list of domains. That way you only have to authenticate once.
> 
> **Edit**: `/proxy/[number]` will also work. No command-line flags are necessary to enable this.

Make sure to add some documentation on this and disable it by default. This is SUPER DANGEROUS. A lot of systems run on 127.0.0.1 without any security in place just relying on the loopback interface and the fact that net.ipv4.conf.all.route_localnet is set to 0 which disables routing of the 127/8 subnet, not even iptables will make anything happen here while this is 0.

When your proxy allows me to test all the ports on the local system just by iterating over /proxy/$number this will become really dangerous.

Yes, independent if the testing system has a cookie set or not.

I usually run code-server only behind nginx with the auth_request module against some http auth, e.g. Nextloud. (http://nginx.org/en/docs/http/ngx_http_auth_request_module.html)

This blocks everyone and everything from accessing anything code-server without auth.

nhooyr:
@Grimeton not sure I follow, you still have to be authorized, you can't just iterate over every port.

mklechan:
How do I disable this feature? Is that possible?

nhooyr:
> How do I disable this feature? Is that possible?

Why do you want to disable it?

sar:
It'd be great to add docs on nginx `rewrite, proxy_pass` rules for `/proxy/[number]` usage.

Currently this requires modifying app source as requests on `https://$host:$server_port/proxy/$1` will end up at base path `HTTP GET 404 /common.js` instead of `https://$host:$server_port/proxy/[number]/common.js`.

code-asher:
Ah yeah our proxy rewrites the base path out of the request.

I was thinking maybe we should just stop doing that since I think most
applications don't use relative paths.

nhooyr:
> I was thinking maybe we should just stop doing that since I think most
applications don't use relative paths.

Perhaps we do need to make this configurable. Just seems wrong to me that people would rather want to specify a base path in their app vs just doing this relatively like HTTP supports.

code-asher:
> Just seems wrong to me that people would rather want to specify a base path in their app vs just doing this relatively like HTTP supports.

Agreed! I thought the same thing when I initially wrote the proxy haha. But now that a few issues/comments have been made I'm starting to suspect that using relative paths is quite rare.

nhooyr:
> Agreed! I thought the same thing when we initially wrote the proxy haha. But now that a few issues/comments have been made I'm starting to suspect that using relative paths is quite rare.

If we made it configurable, then the path user applications expect is going to be coupled with the port they're listening on. and it'll be extremely weird to access on `localhost` lol.

I'm inclined to just write a FAQ entry on how why relative paths are what you should use instead.

nhooyr:
Opened #2485

erlloyd:
Just to further the conversation, some SPA frameworks (like react via create-react-app, for example) actually set up configuring the base path "in the app" (it's handled by react-scripts, in this case) by default. See my comments here on why this is problematic. I'd like to encourage you to still consider making this configurable at the very least: https://github.com/cdr/code-server/issues/2222#issuecomment-755001669
https://github.com/cdr/code-server/issues/2222#issuecomment-755008119

melissa-barca:
I am also interested in an option to disable this for reasons similar to those @Grimeton mentioned. The predictable path opens the door for CSRF attacks which occur when an authorized user is tricked into running code by an attacker: [https://owasp.org/www-community/attacks/csrf]( https://owasp.org/www-community/attacks/csrf ).

PR #1459, opened 2020-03-27
Add install helper scripts

### LINKED ISSUE #1396: Add npm package to make code-server super simple to install

cmoog:
Closes #1396 

I'm in agreement that a simple one-line install command would greatly improve the user experience of getting started. A transparent bash script seems better than depending on npm for the reasons stated in the issue.


cmoog:
Alternatively, I agree that `brew` would be far superior to this and to `npm`

nhooyr:
I would be in favour of a brew package but the reason for the `npm` package is so that anyone can use whatever node runtime they want as well and they can easily uninstall/reinstall it without depending on our scripts.

kylecarbs:
I'm in the middle on this. There seems to be a good amount of value in easily spinning up code-server without effort. If we did something like `curl https://code.cdr.dev | bash` for the complete install I could see code-server being spun up even for quick reasons.

cmoog:
Also-- it'd be nice to have it independent of node and npm, which is easy considering it's already packaged with its compatible node runtime. The ability to install coder on a bare ubuntu vm seems useful.

nhooyr:
I think most would just use `sshcode` for that though I can see the value in not having to install anything and just using curl.

nhooyr:
@cmoog we don't need a separate Mac/linux script, we can easily detect in a single script.

cmoog:
As kyle suggested, we'd probably want to add a proxy for `https://code.cdr.dev` to `https://raw.githubusercontent.com/cdr/code-server/master/install_helper.sh `, then update the README.

install_helper.sh
```diff
@@ -0,0 +1,104 @@
...
+  echo "-- Successfully installed code-server at $bin_path/code-server"
+
+}
+
+if [[ $OSTYPE == "linux-gnu" ]]; then
+  linux_install
+elif [[ $OSTYPE == "darwin"* ]]; then
+  mac_install
+else
+  echo "Unknown operating system. Not installing."
```
nhooyr, line 103: `exit 1`.

install_helper.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
+set -e
```
nhooyr, line 2: Why not `-euo pipefail`?

install_helper.sh
```diff
@@ -0,0 +1,104 @@
...
+
+get_releases() {
+  curl --silent "https://api.github.com/repos/cdr/code-server/releases/latest" |
+    grep '"browser_download_url":\|"tag_name":'
+}
+
+linux_install() {
+  bin_path=$HOME/bin
```
nhooyr, line 10: We shouldn't be installing it in ~/bin. It should be going into `/usr/local/bin`. Or just the current directory.

nhooyr:
Also, not sure if we should be using wget or curl.

install_helper.sh
cmoog: I was thinking it'd be nice to not require `sudo` perms... although requiring `sudo` seems better than installing everything into the current dir

nhooyr: We can also use /opt which does not require sudo iirc.

cmoog: Another option is to create a new hidden dir `~/.code-server` and put everything there. Then we could either tell the user to add it to their path or just do it ourselves?

nhooyr: Lets do whatever fzf does.

install_helper.sh
```diff
@@ -1,14 +1,14 @@
...
 get_releases() {
   curl --silent "https://api.github.com/repos/cdr/code-server/releases/latest" |
     grep '"browser_download_url":\|"tag_name":'
 }
 
 linux_install() {
-  bin_path=$HOME/bin
-  lib_path=$HOME/lib
+  bin_path=$HOME/.local/bin
+  lib_path=$HOME/.local/lib
```
nhooyr, line 11: Let's use ~/.local/share here instead as it's the XDG_DATA_HOME directory.

install_helper.sh
```diff
@@ -49,8 +49,11 @@ linux_install() {
 
   rm -rf -f $temp_path
 
+  if [ $bin_path != *"$PATH"* ]; then
+    echo "-- WARNING: ~/.local/bin is not in your \$PATH"
```
nhooyr, line 53: Let's substitute $bin_path instead of hard coding.

install_helper.sh
```diff
@@ -7,8 +7,8 @@ get_releases() {
 }
 
 linux_install() {
-  bin_path=$HOME/.local/bin
-  lib_path=$HOME/.local/lib
+  bin_path=$HOME/.local/share/code-server/bin
+  lib_path=$HOME/.local/share/code-server/lib
```
nhooyr, line 11: I don't think we need a subdirectory of lib here, let's just store the versions in `share/code-server/<version_number>`

cmoog: Then we can remove the "backup" step as well, correct?

cmoog: On second thought, let's keep it there in case someone reinstalls

nhooyr: Yea we can, we just keep all versions but replace the symlink to the binary in bin.

README.md
```diff
@@ -20,6 +20,12 @@ docker run -it -p 127.0.0.1:8080:8080 -v "$PWD:/home/coder/project" codercom/cod
 
 ## Getting Started
 
+### Installation
+
+```bash
+curl -s https://raw.githubusercontent.com/cdr/code-server/issue-1396/install_helper.sh | bash -s
```
nhooyr, line 26: Why the `-s` for `bash`?

install_helper.sh
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+RED='\033[0;31m'
```
nhooyr, line 4: I'd recommend using `tput setaf` instead of hard coding the escape codes.

install_helper.sh
```diff
@@ -0,0 +1,108 @@
...
+NC='\033[0m' # No Color
+
+get_releases() {
+  curl --silent "https://api.github.com/repos/cdr/code-server/releases/latest" |
```
nhooyr, line 8: Like the use of fully expanded options here vs shorthands.

install_helper.sh
```diff
@@ -0,0 +1,108 @@
...
+    grep '"browser_download_url":\|"tag_name":'
+}
+
+linux_install() {
```
nhooyr, line 12: Mac and linux install functions are pretty much duplicates. Let's have a single one that uses the values of bin_path and lib_path to install.

cmoog: The subtle differences make me think it'll be error-prone if we mix the logic. `.tgz` vs a `.zip`, the backup for mac, the release URL itself, etc.

README.md
cmoog: Specifies input should come from stdin-- it's possible it can be omitted

install_helper.sh
nhooyr: We can easily check for the extension and what command to use to extract.

I think we shouldn't worry about Mac. A home-brew formula is pretty simple and will be a much better option. Linux's 100 package systems is our only problem. Let's remove the Mac install.

README.md
nhooyr: Yea `-s` is only required if you want to pass in args.

install_helper.sh
cmoog: Agreed 👍

install_helper.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env bash
+# install_helper allows for easy installation of code-server on Linux systems
+# the latest official release is pulled from the Github API, where the proper asset
+# is then unpackaged and installed into ~/.local/share/code-server/<version>
+
+# The code-server binary is linked to ~/.local/share/code-server/bin/code-server
+
+set -euo pipefail
+
+RED=1
```
nhooyr, line 10: I would put this into a function instead of a variable just so you don't have to remember `tput setaf` everywhere else.

install_helper.sh
```diff
@@ -0,0 +1,164 @@
...
+    echo "$DETECTED_PROFILE"
+  fi
+}
+
+add_to_path() {
+  echo '-- Adding to $PATH...'
+  DETECTED_PROFILE="$(detect_profile)"
+  SOURCE_STR="\nexport PATH=\"\$HOME/.code-server/bin:\$PATH\"\n"
+
+  if [ -z "${DETECTED_PROFIL}" ]; then
```
ahhmarr, line 118: shouldn't it be `DETECTED_PROFILE`

cmoog: Yup! Nice catch.

install_helper.sh
```diff
@@ -5,24 +5,32 @@
...
+no_color() {
+  tput sgr 0
+}
 
 get_releases() {
   curl --silent "https://api.github.com/repos/cdr/code-server/releases/latest" |
     grep '"browser_download_url":\|"tag_name":'
 }
 
+bin_dir=$HOME/.code-server/bin
```
nhooyr, line 19: Why the switch back to `~/.code-server` over the XDG directories.

PR #1486, opened 2020-04-02
Back up old directory when updating

### LINKED ISSUE #1483: code-server deletes olds version when updating
Coder server just updated, and deleted my projects folder.

Is there anyway to retrieve it? I don't have a backup.
Infact, it delted all configs. It wiped the folder that coder was in an replaced it all.

code-asher:
Rename the old directory with a timestamp before moving the new directory into place.

Also has a minor fix for a notification saying the update failed when it didn't (`isLatest`
won't be true since that version is still technically the older version).

Fixes #1483 .

### DIFF
--- a/ci/vscode.patch
+++ b/ci/vscode.patch
@@ -512,7 +512,7 @@ index eab8591492..26668701f7 100644
  					options.logService.error(`${logPrefix} socketFactory.connect() failed. Error:`);
 diff --git a/src/vs/server/browser/client.ts b/src/vs/server/browser/client.ts
 new file mode 100644
-index 0000000000..a27a6f3a45
+index 0000000000..4f8543d975
 --- /dev/null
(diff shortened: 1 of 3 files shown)


ci/vscode.patch
```diff
@@ -696,10 +696,6 @@ index 0000000000..96fbd4b0bb
 +		}
 +
 +		const json = await response.json();
-+		if (!json.isLatest) {
```
nhooyr, line 699: Wouldn't this cause no errors to be reported if an update fails?

code-asher: The response will only be a 200 if the update succeeded so it should already have bailed out on the line above. The problem with this line is that `isLatest` was always false since the currently running code was still the old version.

Although, that means I should log the status text and not just "unexpected response". Will do that right now.

PR #1499, opened 2020-04-06
Adds dev container and docs

### LINKED ISSUE #1498: Dev container for code-server
Create a docker container for developing code-server in an isolated environment

cmoog:
Closes #1498

### DIFF
--- a/ci/dev-image/Dockerfile
+++ b/ci/dev-image/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:12
+
+RUN apt-get update && apt-get install -y \
+    curl \
+    iproute2 \
+    vim \
+    libgpgme-dev \
+    iptables \
+    net-tools \
+    libsecret-1-dev \
+    libx11-dev \
+    libxkbfile-dev
+
+CMD ["/bin/bash"]
--- a/ci/dev-image/Dockerfile
+++ b/ci/dev-image/Dockerfile
@@ -4,7 +4,6 @@ RUN apt-get update && apt-get install -y \
     curl \
     iproute2 \
     vim \
-    libgpgme-dev \
     iptables \
     net-tools \
     libsecret-1-dev \
--- a/ci/dev-image/exec.sh
+++ b/ci/dev-image/exec.sh
@@ -10,7 +10,7 @@ container_name=code-server-dev
 
 # Docker build is very verbose even when everything is cached.
 echo "--- Building $container_name"
-docker build -t $container_name -f ./ci/dev.Dockerfile . > /dev/null
+cd ../../ && docker build -t $container_name -f ./ci/dev-image/Dockerfile . > /dev/null
 
 set +e
 container_id=$(docker container inspect --format="{{.Id}}" $container_name 2>/dev/null)
@@ -42,7 +42,7 @@ CMD="$@"
 docker exec \
     $([[ -t 0 ]] && echo -it || true) \
     $container_id \
-    sh -c "$CMD"
+    bash -c "$CMD"
 
 docker exec \
     $([[ -t 0 ]] && echo -it || true) \
--- a/ci/dev-image/exec.sh
+++ b/ci/dev-image/exec.sh
@@ -8,21 +8,14 @@ git submodule update --init
 
 container_name=code-server-dev
 
-# Docker build is very verbose even when everything is cached.
-echo "--- Building $container_name"
-cd ../../ && docker build -t $container_name -f ./ci/dev-image/Dockerfile . > /dev/null
+enter() {
+  echo "--- Entering $container_name"
+  docker exec -it $container_id /bin/bash
+}
 
-set +e
-container_id=$(docker container inspect --format="{{.Id}}" $container_name 2>/dev/null)
-
-if [ $? -eq "0" ]; then
-    echo "--- Killing $container_name"
-    docker rm -f $container_name 2>/dev/null
-fi
-set -e
-
-echo "--- Spawning $container_name"
-container_id=$(docker run \
+run() {
+  echo "--- Spawning $container_name"
+  container_id=$(docker run \
     -it \
     --privileged \
     --name $container_name \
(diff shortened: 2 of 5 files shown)


nhooyr:
Nice, I like this.

ci/dev.Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM node:12
```
nhooyr, line 1: Let's put this in `ci/dev-image` along with the script to start it up and name that `start.sh` instead.

ci/dev.Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+RUN apt-get update && \
+    apt-get install -y curl iproute2 vim libgpgme-dev libdevmapper-dev iptables net-tools libsecret-1-dev libx11-dev libxkbfile-dev libxkbfile-dev
```
nhooyr, line 4: Put each of these on their own line.

nhooyr: Why do we need libdevmapper-dev?

nhooyr: Also libxkbfile-dev is duplicated

ci/dev.Dockerfile
nhooyr: Actually, how about `./exec.sh`?

cmoog: VSCode needs the `.Dockerfile` for language features.... it is not the same in other editors?

cmoog: VSCode needs the `.Dockerfile` for language features.... it is not the same in other editors?

ci/dev.Dockerfile
cmoog: Why not reduce the layering?

m
```diff
@@ -0,0 +1,50 @@
...
+
+# Docker build is very verbose even when everything is cached.
+echo "--- Building $container_name"
+docker build -t $container_name -f ./ci/dev.Dockerfile . > /dev/null
+
+set +e
+container_id=$(docker container inspect --format="{{.Id}}" $container_name 2>/dev/null)
+
+if [ $? -eq "0" ]; then
+    echo "--- Killing $container_name"
```
nhooyr, line 19: What's the purpose of this?

m
```diff
@@ -0,0 +1,50 @@
...
+    docker rm -f $container_name 2>/dev/null
+fi
+set -e
+
+echo "--- Spawning $container_name"
```
nhooyr, line 24: if the container is already running, we should just reuse it.

ci/dev.Dockerfile
nhooyr: I don't mean apt-get install for each, just the package names multiline. See the other docker files in the ci directory.

ci/dev.Dockerfile
nhooyr: ci/dev-image would be a directory with Dockerfile and exec.sh

cmoog: oh I see. sounds good 👍

m
nhooyr: Actually I guess that wouldn't be very useful since you can just use the VS Code terminal.

m
cmoog: To kill an existing instance of the dev container. much better than having to `exec bash` back into the container and kill all the watch processes

m
```diff
@@ -0,0 +1,50 @@
...
+echo "--- Executing: $@"
+
+CMD="$@"
+
+docker exec \
+    $([[ -t 0 ]] && echo -it || true) \
+    $container_id \
+    sh -c "$CMD"
+
+docker exec \
```
nhooyr, line 47: Why `sh -c` and bash?

m
nhooyr: Shouldn't the watch process be running in the foreground? Then u can just `ctrl+c`.

cmoog: Not if you get detached from the container, which happens to me all the time. When my ssh loses connection for example.

nhooyr: If you get detached, the container should exit. You'd want to pass in `-it` for that to work.

m
cmoog: Having to use the code-server terminal would not be optimal though considering it's constantly reloading, losing its history, etc.

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "--- Executing: $@"
+
+CMD="$@"
+
+docker exec \
+    $([[ -t 0 ]] && echo -it || true) \
+    $container_id \
+    bash -c "$CMD"
+
+docker exec \
```
nhooyr, line 47: Could you clarify why we have two docker execs here?

ci/dev-image/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:12
+
+RUN apt-get update && apt-get install -y \
+    curl \
+    iproute2 \
+    vim \
+    libgpgme-dev \
```
nhooyr, line 7: What's this for?

ci/dev-image/exec.sh
cmoog: First executes the command from the argument to `exec.sh`, then enters `bash`. Should probably combine them.

m
cmoog: Now thinking it's better leave this functionality out. It's now removed.

cmoog:
Anything preventing this from merging? @nhooyr

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,50 @@
...
+  echo "--- Spawning $container_name"
+  container_id=$(docker run \
+    -it \
+    --privileged \
+    --name $container_name \
+    "-v=$PWD:/code-server" \
+    "-w=/code-server" \
+    "-p=8080:8080" \
+    $([[ -t 0 ]] && echo -it || true) \
+    -d \
```
nhooyr, line 26: What is the `-d` for?

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,50 @@
...
+    $container_name)
+}
+
+build() {
+  echo "--- Building $container_name"
+  cd ../../ && docker build -t $container_name -f ./ci/dev-image/Dockerfile . > /dev/null
```
nhooyr, line 32: Separate lines.

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+set +e
```
nhooyr, line 35: You can use `|| true` next to the command that might fail instead of `set +e`.

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,50 @@
...
+enter() {
+  echo "--- Entering $container_name"
+  docker exec -it $container_id /bin/bash
+}
+
+run() {
+  echo "--- Spawning $container_name"
+  container_id=$(docker run \
+    -it \
+    --privileged \
```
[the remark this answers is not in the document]
cmoog: I think I was considering vms when I added that. But you're right– not needed for now.

ci/dev-image/exec.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
nhooyr, line 1: Let's add a comment explaining what the script does.

ci/dev-image/exec.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+# exec.sh opens an interactive bash session inside of a docker container 
+# for improved isolation during development
```
nhooyr, line 3: Could you elaborate that it reuses the container if it exists? And starts it if stopped.

PR #1561, opened 2020-04-27
Add basic rate limiting to login endpoint

### LINKED ISSUE #1320: Ratelimit code-server password auth
No rate limiting at the moment, which is embarrassing.

nhooyr:
Closes #1320

### DIFF
--- a/src/node/app/login.ts
+++ b/src/node/app/login.ts
@@ -1,4 +1,5 @@
 import * as http from "http"
+import * as limiter from "limiter"
 import * as querystring from "querystring"
 import { HttpCode, HttpError } from "../../common/http"
 import { AuthType, HttpProvider, HttpResponse, Route } from "../http"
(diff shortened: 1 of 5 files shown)


src/node/app/login.ts
```diff
@@ -108,3 +115,18 @@ export class LoginHttpProvider extends HttpProvider {
...
 }
+
+// RateLimiter wraps around the limiter library for logins.
+// It allows 2 logins every minute and 12 logins every hour.
+class RateLimiter {
+  private readonly minuteLimiter = new limiter.RateLimiter(2, "minute")
+  private readonly hourLimiter = new limiter.RateLimiter(12, "hour")
+
+  public try(): boolean {
+    console.info(this.minuteLimiter.getTokensRemaining(), this.hourLimiter.getTokensRemaining())
```
code-asher, line 126: Was this just for debugging? If not I think we might want to add some context. Although would it be a valid concern that it could be used to flood code-server's logs? Could put it behind a `debug` log instead I suppose.

nhooyr: Yea will remove sorry.

PR #1563, opened 2020-04-27
Remove SSH server

### LINKED ISSUE #1502: /ssh endpoint should be disabled by default
image

also, the changelogs on the release page do not state there is an ssh server, to begin with, and I think it is a decent security issue, because there may be people who don't use ssh for reasons or people who hold a STRICT ssh policy and this is just an ssh server without any user configurability by the looks of it, and it is enabled by default.

nhooyr:
Closes #1502


code-asher:
I think we can also remove all the `ssh2` dependencies from the `package.json`.

nhooyr:
You got it.

PR #1601, opened 2020-05-04
Add npm package and document/cleanup CI and build process

nhooyr:
Closes many issues that I'll go through later and link.


nhooyr:
We have debs/rpms now too.

Please review @code-asher 

I'll add alpine too.

nhooyr:
Actually lets not add alpine, just change docs to tell people to use npm package.

nhooyr:
@code-asher This moves the bundled node binary into `lib` and the code-server script into `bin`. That way you can easily add the bin directory to your $PATH and use code-server for the self contained releases.

ci/container/exec.sh
```diff
@@ -0,0 +1,25 @@
...
+    -e TRAVIS_TAG \
+    -e NPM_TOKEN \
+    -v "$(yarn cache dir):/usr/local/share/.cache/yarn/v6" \
+    -v "$PWD:/repo" \
+    -w /repo \
+    $(if [[ -t 0 ]]; then echo -it; fi) \
+    "$imageTag" \
+    "$*"
+
+  kill $pid
```
code-asher, line 22: I think this belongs in `static-release.sh`.

package.json
```diff
@@ -1,19 +1,29 @@
...
-    "runner": "cd ./ci && NODE_OPTIONS=--max_old_space_size=32384 ts-node ./build.ts",
-    "build": "yarn runner build",
-    "watch": "yarn runner watch"
+    "clean": "./ci/build/clean.sh",
+    "vscode": "./ci/dev/vscode.sh",
+    "vscode:patch": "cd ./lib/vscode && git apply ../../ci/dev/vscode.patch",
+    "vscode:diff": "cd ./lib/vscode && git diff HEAD > ../../ci/dev/vscode.patch",
+    "build": "./ci/build/build-code-server.sh && ./ci/build/build-vscode.sh",
+    "release": "./ci/build/build-release.sh",
+    "pkg": "./ci/build/build-static-pkgs.sh.sh",
```
code-asher, line 18: Little typo with `.sh.sh`.

ci/release-container/Dockerfile
```diff
@@ -26,18 +27,18 @@ ENV SHELL=/bin/bash
 RUN adduser --gecos '' --disabled-password coder && \
   echo "coder ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
 
-RUN curl -SsL https://github.com/boxboat/fixuid/releases/download/v0.4/fixuid-0.4-linux-amd64.tar.gz | tar -C /usr/local/bin -xzf - && \
+ARG ARCH=amd64
```
code-asher, line 30: This doesn't really matter but since I'm thinking about it what are your thoughts on this vs the `uname` + `sed` method of getting the arch used in the CI container?

ci/build/test-static-release.sh
```diff
@@ -1,21 +1,20 @@
...
-  cd "$(dirname "${0}")/.." || exit 1
+# Makes sure the release works.
+# This is to make sure we don't have Node version errors or any other
+# compilation-related errors.
+main() {
+  cd "$(dirname "${0}")/../.."
 
   local output
-  output=$(node ./build/out/node/entry.js --list-extensions 2>&1)
+  output=$(./release-static/code-server --list-extensions 2>&1)
```
code-asher, line 11: I think there needs to be a `bin` here.

nhooyr:
> This is looking super fly. rsync is a very nice touch.

<3

> I assume the commented-out tag parts of the Travis config are just for testing yeah?

Indeed.

ci/README.md
```diff
@@ -0,0 +1,80 @@
+# ci
```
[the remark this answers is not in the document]
nhooyr: ❤️

package.json
nhooyr: Nice catch!

ci/release-container/Dockerfile
nhooyr: uname + sed sounds better.

ci/build/test-static-release.sh
nhooyr: Good catch.

ci/release-container/Dockerfile
nhooyr: Though tbh it's nice right now to not clobber the dockerfile. Using uname + sed will make that line even harder to fall.

code-asher: Yeah I don't feel super strongly about it, it's mostly this odd feeling I get since technically you can build a broken Docker image if you provide the wrong arg (or no arg). 

But even in the theoretical world where the arg was wrong somehow that would mean the tag would also be wrong so it'd be broken in practice for users anyway. 

Which means the only place this could reasonably come into play is if someone was trying to build it on their own or something.

nhooyr: Fair, let's copy lib.sh in and source it and then just run the arch function.

code-asher: Dope, that sounds perfect.

ci/lib.sh
```diff
@@ -1,10 +1,44 @@
...
+      os="alpine"
+    fi
   fi
-  export VERSION=$code_server_version
+  echo "$os"
+}
+
+arch() {
+  case "$(uname -m)" in
+  aarch)
```
code-asher, line 33: I think we`ll want `aarch64` here.

PR #1602, opened 2020-05-04
Unset $PASSWORD after grabbing it

### LINKED ISSUE #1583: Remove $PASSWORD from ENV after reading it.
Hi,

patch the server so that $PASSWORD gets unset in the environment of the server after it has been read, because else it will get inherited to all the shells the server starts in a terminal...

KR,

G.

nhooyr:
Closes #1583

### DIFF
--- a/src/node/entry.ts
+++ b/src/node/entry.ts
@@ -90,6 +90,7 @@ const main = async (args: Args): Promise<void> => {
   } else {
     logger.info("  - No authentication")
   }
+  delete process.env.PASSWORD
 
   if (httpServer.protocol === "https") {
     logger.info(


src/node/entry.ts
```diff
@@ -90,6 +90,7 @@ const main = async (args: Args): Promise<void> => {
   } else {
     logger.info("  - No authentication")
   }
+  process.env.PASSWORD = undefined
```
code-asher, line 93: This doesn't really matter but since this will set it to the literal string "undefined" should we do `delete process.env.PASSWORD` instead?

nhooyr: Forgot, it's so dumb...

nhooyr:
In the future we ought to add a flag/config file option and remove docs on $PASSWORD to avoid issues like these.

Issue #1501, opened 2020-04-07
Can't find vscode version since V3


davezhouwa:
"V3 has some significant changes that will need to be accounted for in your
scripts.

We use semantic versioning now. The VS Code version will no longer be included
in the tag or the release file name. "

Some extensions are tie to particular vscode version,  so how can we find the version of vscode from the V3 releases?

code-asher:
The two ways I can think of are:
1. Use the "about" menu from VS Code.
2. Run code server with `code-server --version --json` which prints some extra version information in JSON format. It should look something like this (except the commit will be a hash):

```
  { codeServer: '3.0.2', commit: 'development', vscode: '1.43.2' }
 ```

Both of these require you to have already downloaded code-server though. If you're trying to figure out the version before downloading I think the only way is to click the tag from the release then go to `lib/vscode/package.json` and check the version listed there.

I think it might not be a bad idea for us to list the VS Code version in the release notes as well.

davezhouwa:
Please list the vscode version directly in the release notes

PR #1607, opened 2020-05-07
Switch to GH Actions

nhooyr:
Travis ARM is very unreliable :(


nhooyr:
I merged for testing automating publishing then docker image/npm. The GitHub Actions API requires the workflow file be in master for the API to work properly.

Please still review @code-asher

nhooyr:
Nvm I was wrong, got the API working not related to merge but still review @code-asher

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,89 @@
...
+        uses: actions/download-artifact@v2
+        with:
+          name: npm-package
+          path: ./release
+      - run: brew unlink node@12
+      - run: brew install node
+      - run: ./ci/steps/release-static.sh
+        env:
+          # Otherwise we get rate limited when fetching the ripgrep binary.
+          GITHUB_TOKEN: ${{ secrets.github_token }}
```
code-asher, line 84: Should we set this for the other jobs as well?

ci/README.md
```diff
@@ -6,6 +6,20 @@ Many of these scripts contain more detailed documentation and options in comment
...
 Any file and directory added into this tree should be documented here.
 
+## Publishing a release
+
+1. Change the version of code-server in `package.json` and push this commit.
+1. CI will run and generate an NPM package and release packages that you can download
+   as artifacts on Github Actions.
+1. Create a new draft release with the built release packages.
+1. Run some basic sanity tests on one of the released packages.
+1. Publish.
```
code-asher, line 16: I think we should mention to include the VS Code version in the release notes, see https://github.com/cdr/code-server/issues/1501#issuecomment-611031429. Including a changelog is of course obvious but maybe we should mention that as well (which at some point we should probably start generating automatically).

ci/README.md
```diff
@@ -6,6 +6,20 @@ Many of these scripts contain more detailed documentation and options in comment
...
+1. Download the built npm package and publish it.
+1. Place the debian releases into `./release-packages` and then push the docker
```
code-asher, line 18: I don't know if this is better or worse but my strat was to pull the versioned Docker image, tag it latest, then push it back up.

ci/README.md
```diff
@@ -68,15 +85,13 @@ This directory contains the container for CI.
 
 ## steps
 
-This directory contains a few scripts used in CI. Just helps avoid clobbering .travis.yml.
+This directory contains a few scripts used in CI.
+Just helps avoid clobbering .travis.yml.
```
code-asher, line 89: I suppose we don't have a `.travis.yml` anymore, maybe we can make this generic with `CI configuration file(s)` or something.

ci/README.md
```diff
@@ -68,15 +85,13 @@ This directory contains the container for CI.
...
 
 - [./steps/test.sh](./steps/test.sh)
   - Runs `yarn ci` after ensuring VS Code is patched
+- [./steps/release.sh](./steps/release.sh)
+  - Runs the full release process
+  - Generates the npm package at `./release`
 - [./steps/static-release.sh](./steps/static-release.sh)
```
code-asher, line 96: Looks like this is `release-static.sh` now.

ci/README.md
```diff
@@ -32,27 +46,30 @@ This directory contains scripts used for the development of code-server.
...
   - Useful to do a clean build.
 - [./build/code-server.sh](./build/code-server.sh)
   - Copied into static releases to run code-server with the bundled node binary.
-- [./build/archive-static-release.sh](./build/archive-static-release.sh)
-  - Archives `./release-static` into a tar/zip for CI with the proper directory name scheme
 - [./build/test-release.sh](./build/test-static-release.sh)
   - Ensures code-server in the `./release-static` directory runs
-- [./build/build-static-pkgs.sh](./build/build-static-pkgs.sh) (`yarn pkg`)
-  - Uses [nfpm](https://github.com/goreleaser/nfpm) to generate .deb and .rpm from a static release
+- [./build/build-packages.sh](./build/build-static-pkgs.sh) (`yarn package`)
```
code-asher, line 70: The names on the left don't match the right. Looks like on the first one the right is correct and on the second the left is correct.

ci/README.md
code-asher: Actually I see in your other PR the image is pushed after a release is created so I suppose we could also just push latest? Since at that point we'd have verified the release is good to go.

.github/workflows/ci.yaml
nhooyr: Only macOS for some reason needs it. The others never fail.

ci/README.md
nhooyr: You got it.

I'd be in favour of a non autogenerated changelog, commits can get verbose. I'd rather we summarize the big changes ourselves.

ci/README.md
nhooyr: Yea will add.

nhooyr:
> It occurred to me we'll need to update CONTRIBUTING.md to build a static release since you can't run the non-static release with node the way we describe since there are no VS Code node modules in there.

Fixing.

.github/workflows/ci.yaml
nhooyr: Documented this.

ci/README.md
code-asher: sgtm

PR #1623, opened 2020-05-11
Automate draft release

nhooyr:


ci/README.md
```diff
@@ -8,12 +8,16 @@ Any file and directory added into this tree should be documented here.
...
-1. CI will run and generate the `npm-package` and `release-packages` artifacts on the GH actions workflow
-1. Create a new draft release and attach all files in `release-packages`
-   1. Run some basic sanity tests on one of the released packages
-1. Summarize the major changes in the release notes and link to the relevant issues.
-   1. Make sure to mention the VS Code version in the release notes
+1. GitHub actions will generate the `npm-package` and `release-packages` artifacts
+1. Run `yarn release:github-draft` to create a GitHub draft release from the template with
+   the updated version.
+   1. Summarize the major changes in the release notes and link to the relevant issues.
+1. Wait for the artifacts in step 2 to build
```
code-asher, line 18: Should we number these incrementally instead of just with `1`? Normally I don't pay much attention but since we called out step 2 I was trying to count up in my head and got lost a couple times lol especially since Github is rendering a gap after 2 for some reason and I kept thinking it was a separate paragraph. I always read the readmes raw so maybe that's just a me problem.

ci/build/release-github-draft.sh
```diff
@@ -0,0 +1,21 @@
...
+main() {
+  cd "$(dirname "$0")/../.."
+  source ./ci/lib.sh
+
+  hub release create \
+    --file - \
+    --draft "${assets[@]}" "v$(pkg_json_version)" << EOF
+v$(pkg_json_version)
+
+VS Code v$(vscode_version)
```
code-asher, line 15: Very nice :+1:

ci/README.md
nhooyr: Usually in markdown you don't want to number them to avoid forgetting updates since when it renders it's immediately there but I agree it is nice when looking at the raw markdown.

nhooyr: Also I see no gap after 2, that's weird.

code-asher: It must have something to do with the specific screen size I have my browser at. My editor automatically updates the numbers so I was hoping there was something similar for yours.

nhooyr: Goland aint as cool as emacs :(

PR #1670, opened 2020-05-18
Fix paths from Windows client to non-Windows server

### LINKED ISSUE #1642: Accessing from windows is broken
Hi looks like latest build changes broke path separator.
1. Host machine windows
2. code-server machine linux in docker 
3. Relevant docker part
```
RUN go get github.com/derekparker/delve/cmd/dlv
RUN git clone https://github.com/cdr/code-server.git
WORKDIR code-server
RUN git submodule update --init
RUN yarn
RUN yarn vscode
RUN yarn build
RUN yarn build:vscode
RUN yarn release
WORKDIR release
RUN yarn --production
WORKDIR /
ENTRYPOINT node /code-server/release --disable-telemetry  --host 0.0.0.0 --port 3003 
```

code server is up and running in docker and accessible via 127.0.0.1:3003 on windows host .

Following request fails while ui is loading in browser
http://127.0.0.1:3003/vscode-remote-resource?path=%5Ccode-server%5Crelease%5Clib%5Cvscode%5Cextensions%5Ctheme-defaults%5Cthemes%5Cdark_plus.json&tkn=

Checked and confirmed that file is there.
changin request to http://127.0.0.1:3003/vscode-remote-resource?path=/code-server/release/lib/vscode/extensions/theme-defaults/themes/dark_plus.json&tkn=
returns json data.

### LINKED ISSUE #1659: Accessing from windows is broken
I am using static version of linux 64 bit build.

Program can start without any error. when accessing, code highlight is gone. File icon is gone. 

No error log, everything looks correct.

image

Look at the snap shot. The Java file is not highlighted.  File icon missing.

code-asher:
Fixes #1659
Fixes #1642

I tested:
- [x] Linux client to Linux host
- [x] Windows client to Windows host
- [x] Windows client to Linux host
- [x] Linux client to Windows host

The strat is to convert the path on the server end instead of on the browser end since the latter ends up using the browser's OS.

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -248,19 +248,17 @@ index 1e16cde724..0000000000
 -target "12.4.0"
 -runtime "node"
 diff --git a/src/vs/base/common/network.ts b/src/vs/base/common/network.ts
-index e4546b2cf6..9df12239fb 100644
+index e4546b2cf6..ad2c544e89 100644
 --- a/src/vs/base/common/network.ts
 +++ b/src/vs/base/common/network.ts
-@@ -94,16 +94,18 @@ class RemoteAuthoritiesImpl {
+@@ -94,16 +94,17 @@ class RemoteAuthoritiesImpl {
  		if (host && host.indexOf(':') !== -1) {
  			host = `[${host}]`;
  		}
 -		const port = this._ports[authority];
 +		// const port = this._ports[authority];
  		const connectionToken = this._connectionTokens[authority];
--		let query = `path=${encodeURIComponent(uri.path)}`;
-+		// NOTE@coder: Use fsPath for Windows support.
-+		let query = `path=${encodeURIComponent(uri.fsPath)}`;
+ 		let query = `path=${encodeURIComponent(uri.path)}`;
  		if (typeof connectionToken === 'string') {
  			query += `&tkn=${encodeURIComponent(connectionToken)}`;
  		}
@@ -1067,7 +1065,7 @@ index 0000000000..0d2e93edae
 +}
 diff --git a/src/vs/server/browser/worker.ts b/src/vs/server/browser/worker.ts
 new file mode 100644
-index 0000000000..8db1e17c38
+index 0000000000..a93381631a
 --- /dev/null
 +++ b/src/vs/server/browser/worker.ts
 @@ -0,0 +1,57 @@
@@ -1090,7 +1088,7 @@ index 0000000000..8db1e17c38
 +		scheme: self.location.protocol.replace(':', ''),
 +		authority: self.location.host,
 +		path: self.location.pathname.replace(/\/static\/([^\/]+)\/.*$/, '/static/$1\/'),
-+		query: `tar=${encodeURIComponent(module.extensionLocation.fsPath)}`,
++		query: `tar=${encodeURIComponent(module.extensionLocation.path)}`,
 +	});
 +	const response = await fetch(fetchUri.toString(true));
 +	if (response.status !== 200) {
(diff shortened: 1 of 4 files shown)


ci/dev/vscode.patch
```diff
@@ -248,19 +248,17 @@ index 1e16cde724..0000000000
...
+index e4546b2cf6..ad2c544e89 100644
 --- a/src/vs/base/common/network.ts
 +++ b/src/vs/base/common/network.ts
-@@ -94,16 +94,18 @@ class RemoteAuthoritiesImpl {
+@@ -94,16 +94,17 @@ class RemoteAuthoritiesImpl {
  		if (host && host.indexOf(':') !== -1) {
  			host = `[${host}]`;
  		}
 -		const port = this._ports[authority];
 +		// const port = this._ports[authority];
```
nhooyr, line 259: Should this be commented out?

ci/dev/vscode.patch
```diff
@@ -248,19 +248,17 @@ index 1e16cde724..0000000000
...
  		const connectionToken = this._connectionTokens[authority];
--		let query = `path=${encodeURIComponent(uri.path)}`;
-+		// NOTE@coder: Use fsPath for Windows support.
```
nhooyr, line 262: > We use

src/node/util.ts
```diff
@@ -224,3 +224,51 @@ export function extend(...args: any[]): any {
   }
   return c
 }
+
+/**
+ * Taken from vs/base/common/charCode.ts. Copied for now instead of importing so
+ * we don't have to set up a `vs` alias to be able to import with types (since
```
nhooyr, line 230: What's hard about setting up a vs alias to import? And why is it necessary? Why can't we just import the file?

I thought the reason for duplication was to avoid bringing in a ton of VS Code.

ci/dev/vscode.patch
code-asher: It's not used so it's commented out to prevent a build error about unused variables. Actually we could just remove the line instead of commenting it out.

As to why it's not used, VS Code normally sets the authority portion of the websocket to `host:port` but we change it to `window.location.host`. The problem is that they pull the host and port from the remote authority which is set based on the `host` header and that's not always accurate (reverse proxies, for example).

ci/dev/vscode.patch
code-asher: This is a removed line actually :smile: This is one of the reasons a fork might be better, reviewing changes to the patch is not fun.

ci/dev/vscode.patch
nhooyr: Why wasn't it causing a build error before?

ci/dev/vscode.patch
nhooyr: Oh yea lol.

src/node/util.ts
code-asher: Yeah that's the reasoning for `uri.ts` but `charCode.ts` doesn't import anything and has just the one enum so it wouldn't be too bad (although it's 425 lines).

To import directly we'd have to import from `out` so it works in the final build (`from "../../lib/vscode/out/vs/base/common/charCode`). But that causes type errors (`Could not find a declaration file for module`).

So we could do something like `from "vs/base/common/charCode"` but that means `vs` needs to point to `src` during development and `out` in the build. There are aliases for `tsc` but they don't affect the generated code (I think) so we'd need to hijack `require` or add a dependency to do it for us.

I feel like there might be a good alternative solution but I haven't come upon it yet.

nhooyr: > To import directly we'd have to import from out so it works in the final build

Why do we have to import from `out`? Why not `../../lib/vscode/src/vs/base/common/charCode`?

PR #1701, opened 2020-05-20
Add auto install script

nhooyr:


nhooyr:
```
~/src/cdr/code-server:auto-install$ ./install.sh -help
./install.sh [-d]

Autoinstalls code-server on any macOS or Linux system.

If ran on Ubuntu, Debian or Raspbian then the GitHub releases v3.3.1
deb package will be fetched and installed.

If ran on Fedora, CentOS, RHEL or openSUSE then the GitHub releases v3.3.1
rpm package will be fetched and installed.

If ran on macOS and Homebrew is installed then the Homebrew code-server
package will be installed. However, if Homebrew is not installed then
v3.3.1 of the macOS static archive on GitHub releases will be installed
into /usr/local/lib/code-server-3.3.1.

If ran on an unsupported architecture the npm package will be installed
with yarn or npm. Only amd64 and arm64 are currently supported.

If ran on any other Linux distro, v3.3.1 of the linux static archive on GitHub
releases will be installed into /usr/local/lib/code-server-3.3.1.

  -d Enables a dry run where where the steps that would have taken place
     are printed but do not actually execute.

  -s Forces the installation of a static release into /usr/local/lib/code-server-3.3.1
     Set the -p flag to change the installation prefix from /usr/local/lib

  -p Sets the installation prefix for a static release install.
```

nhooyr:
Need to add to README.md and guide.md

In the next release we'll add installation over SSH support and then deprecate sshcode.

nhooyr:
I think it'd be good to switch to long style flags and remove `getopt`.

install.sh
```diff
@@ -0,0 +1,345 @@
...
+  fi
+
+  sh_c="sh_c"
+  if [ ! -w "$STATIC_INSTALL_PREFIX" ]; then
+    sh_c="sudo_sh_c"
+  fi
+
+  "$sh_c" tar -C "$STATIC_INSTALL_PREFIX" -xzf "$CACHE_DIR/code-server-$VERSION-$OS-$ARCH.tar.gz"
+  # In case previously installed.
+  SKIP_LOG=1 "$sh_c" rm -Rf "$STATIC_INSTALL_PREFIX/code-server-$VERSION"
```
code-asher, line 208: Should we confirm this or do a `mv` instead? I feel like it's probably not going to bite anyone and this is probably how most or all other install scripts operate; I'm just thinking about the issue we had with that user storing their data in the code-server dir. That was probably a symptom of manually installing though and so might not be worth considering here.

install.sh
```diff
@@ -0,0 +1,345 @@
...
+    echo "+ Using cached $FILE from $URL"
+    return
+  fi
+
+  sh_c curl \
+    -#fL \
+    -Ro "$FILE.incomplete" \
+    -C - \
+    "$URL"
+  mv "$FILE.incomplete" "$FILE"
```
code-asher, line 151: I think this might need an `sh_c` to not fail a dry run.

nhooyr:
Also good idea to remove the hard coded version and always fetch latest in case people download and store the script.

install.sh
```diff
@@ -0,0 +1,345 @@
...
+
+install_arch() {
+  echo "Installing from the AUR."
+
+  fetch "https://aur.archlinux.org/cgit/aur.git/snapshot/code-server.tar.gz" "$CACHE_DIR/code-server-aur.tar.gz"
+
+  tmp_dir="$(mktemp -d)"
+  (
+    cd "$tmp_dir"
+    tar -xzf "$CACHE_DIR/code-server-aur.tar.gz" --strip-components 1
```
code-asher, line 180: Another sh_c needed for dry run I believe.

code-asher:
Agreed about the long option names, those are my preference in general. Fetching the latest version also seems like a good idea.

nhooyr:
And check for libc and libcxx #1706

install.sh
nhooyr: Yea will exit out just in case.

nhooyr:
@code-asher Please re-review the new changes.

Last thing I need to take care of is #1706

nhooyr:
Now that we know we have compat for glibc v2.17 I don't think it's worth the complexity in the install script to detect for anything below that since that's already 8 years old. And there is no other dependency anymore.

nhooyr:
tbh `--static` is kinda confusing.

How about `--method=archive` and then `--prefix` to adjust the installation prefix? @code-asher

nhooyr:
How about `--ignore-pkg-manager` and `--prefix`?

nhooyr:
I cracked it!

```
  --method [detect, archive]
      Choose the installation method. Defaults to detect.
      - detect detects the system package manager and tries to use it.
        Full reference on the process is further below.
      - archive installs a static release archive into ~/.local
        Add ~/.local/bin to your \$PATH to use it.
  --prefix=~/.local
      Sets the prefix used by static release archives. Defaults to ~/.local
      By default a release is unarchived into ~/.local/lib/code-server.X.X.X
      and the binary symlinked into ~/.local/bin/code-server.
      To install system wide pass ---prefix=/usr/local
```

code-asher:
`--method` and `--prefix` are great. Definitely more expressive than `--static`.

install.sh
```diff
@@ -0,0 +1,472 @@
...
+    echo "Unsupported package manager."
+    install_archive
+    ;;
+  esac
+}
+
+parse_arg() {
+  case "$1" in
+  *=*)
+    opt="${1#=*}"
```
code-asher, line 202: I think this should be `%` instead of `#`.

install.sh
```diff
@@ -0,0 +1,472 @@
...
+  FILE="$2"
+
+  echo "+ Downloading $URL"
+
+  if [ -e "$FILE" ]; then
+    echo "+ Using cached $FILE"
+    return
+  fi
+
+  SKIP_ECHO=1
```
code-asher, line 240: Should we leave out the `SKIP_ECHO` here and elsewhere? It's a little unfortunate that when I use `--dry-run` I don't see the actual commands being ran. 

Or we could update the description of `--dry-run` to say it lists steps and not necessarily commands. 

Or if `--dry-run` is passed then ignore the skip echo flag.

install.sh
```diff
@@ -0,0 +1,472 @@
...
+
+  fetch "https://github.com/cdr/code-server/releases/download/v$VERSION/code-server-$VERSION-$OS-$ARCH.tar.gz" \
+    "$CACHE_DIR/code-server-$VERSION-$OS-$ARCH.tar.gz"
+
+  sh_c="sh_c"
+  if [ ! -w "$ARCHIVE_INSTALL_PREFIX" ]; then
+    sh_c="sudo_sh_c"
+  fi
+
+  SKIP_ECHO=1 sh_c mkdir -p "$ARCHIVE_INSTALL_PREFIX/lib" "$ARCHIVE_INSTALL_PREFIX/bin"
```
code-asher, line 321: I think this should be `"$sh_c"`.

install.sh
```diff
@@ -0,0 +1,472 @@
...
+install_npm() {
+  if command_exists yarn; then
+    echo "Installing with yarn."
+    echo
+    sh_c yarn global add code-server --unsafe-perm
+    return
+  elif command_exists npm; then
+    echo "Installing with npm."
+    echo
+    sh_c npm install -g code-server --unsafe-perm
```
code-asher, line 346: Should we try detecting if we'll need to run this as root? We could check for write access to the directory echoed by `npm config get prefix`.

Yarn has a similar `yarn global bin` but idk if it's worth checking or if it always chooses a writable directory.

README.md
```diff
@@ -2,114 +2,45 @@
...
 
-You will need at least node v12 installed. See [#1633](https://github.com/cdr/code-server/issues/1633).
+First run to print out the install process:
 
 ```bash
-yarn global add code-server
-# Or: npm install -g code-server
-code-server
-# Now visit http://127.0.0.1:8080. Your password is in ~/.config/code-server/config.yaml
+curl -fsSL https://code-server.dev/install.sh | sh  -s -- --dry-run
```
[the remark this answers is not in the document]
nhooyr: Mhmm great catch, always point out stuff like this for sure, if user's notice it's embarrassing.

doc/FAQ.md
```diff
@@ -136,24 +143,19 @@ code-server --proxy-domain <domain>
...
+user a virtual machine instead of just a container.
 
-## Docker in code-server docker container?
+## Docker in code-server container?
 
-If you'd like to access docker inside of code-server, we'd recommend running a docker:dind container
-and mounting in a directory to share between dind and the code-server container at /var/run. After, install
-the docker CLI in the code-server container and you should be able to access the daemon as the socket
-will be shared at /var/run/docker.sock.
+If you'd like to access docker inside of code-server, we'd recommend running a `docker:dind` container
```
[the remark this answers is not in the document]
nhooyr: Ah very true, I'll change to just recommend mounting in `docker.sock` and leave it at that.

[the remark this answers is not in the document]
nhooyr: The advantage to `dind` is volume mounts will easily work cause you can share nearly any directory at the exact same path with the daemon.

doc/guide.md
```diff
@@ -60,79 +61,86 @@ Once you've signed up and created a GCP project, create a new Compute Engine VM
...
 
 **Never**, **ever** expose `code-server` directly to the internet without some form of authentication
 and encryption as someone can completely takeover your machine with the terminal.
 
-By default, code-server will enable password authentication which will
-require you to copy the password from the code-server config file to login. Since it
-cannot use TLS by default, it will listen on `localhost` to avoid exposing itself
-to the world. This is fine for testing but will not work if you want to access `code-server`
+By default, `code-server` will enable password authentication which will require you to copy the
+password from the`code-server`config file to login. It will listen on`localhost` to avoid exposing
```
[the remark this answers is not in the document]
nhooyr: Yea that was weird fixed.

doc/guide.md
```diff
@@ -220,32 +227,32 @@ sed -i.bak 's/bind-addr: 127.0.0.1:8080/bind-addr: 0.0.0.0:443/' ~/.config/code-
...
 systemctl --user restart code-server
 ```
 
 ### How do I securely access development web services?
 
-If you're working on a web service and want to access it locally, code-server can proxy it for you.
+If you're working on a web service and want to access it locally, `code-server` can proxy it for you.
 
-See [FAQ.md](https://github.com/cdr/code-server/blob/master/doc/FAQ.md#how-do-i-securely-access-web-services).
+See the [FAQ](https://github.com/cdr/code-server/blob/master/doc/FAQ.md#how-do-i-securely-access-web-services).
```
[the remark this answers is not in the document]
nhooyr: Fixed.

doc/install.md
```diff
@@ -0,0 +1,181 @@
...
+
+  - Add `~/.local/bin` to your `$PATH` to run code-server.
+
+- For macOS it will install the Homebrew package.
+
+  - If Homebrew is not installed it will install the latest static release into `~/.local`.
+  - Add `~/.local/bin` to your `$PATH` to run code-server.
+
+- If ran on an architecture with no binary releases, it will install the npm package with `yarn` or `npm`.
+  - We only have binary releases for amd64 and arm64 presently.
```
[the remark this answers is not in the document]
nhooyr: Maybe we should stick with `binary` tbh. I think users might understand it better than `static`.

install.sh
```diff
@@ -0,0 +1,472 @@
...
+
+- For macOS it will install the Homebrew package.
+  - If Homebrew is not installed it will install the latest static release
+    into ~/.local
+
+- If ran on an architecture with no binary releases, it will install the
+  npm package with yarn or npm.
+  - We only have binary releases for amd64 and arm64 presently.
+
+It will cache all downloaded assets into ~/.cache/code-server
```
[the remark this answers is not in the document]
nhooyr: Yea I'm gonna fix that actually. The AUR package is nothing more than a script to install code-server but we can cache the actual result of the script.

install.sh
nhooyr: Ohh yea, nice catch. It's funny it works because the only time `opt` is printed is when there is no argument provided so it reads normally.

code-asher: lol yup I noticed the same thing, the only way to trigger is `--arg=` then it says `--arg= requires an argument`.

install.sh
nhooyr: So this was intentional as the commands can be quite verbose. That's why I added some logs to explain what's being done. E.g. the `Downloading` log.

install.sh
nhooyr: Mhmm that's true. I forgot npm by default tries to use `/usr/local`. 🤔

install.sh
code-asher: That's fair, I mostly just feel like the current description of `--dry-run` is a bit misleading.

fwiw I thought being able to see the actual commands with `--dry-run` was pretty dope since you can fully audit what's going to happen. Skipping some commands somewhat lessons the usefulness of the flag imo (at least, in the way I was expecting to use it).

nhooyr: `--dry-run` prints out the process, not the commands. I don't think printing out the exact commands is useful when they're really verbose, like the curl's take up multiple lines.

If you really want the commands for tracing, you can use `TRACE=1`.

code-asher: That's true but I find `TRACE=1` really difficult to follow.

My impression was that `--dry-run` was a way for you to audit the script before running it but you can't fully audit the script if some commands are obfuscated, right?

I think part of the reason I have that impression is that the description says `Echo the commands` so if we change that to something like `Echo the steps` that might do it.

nhooyr: My bad, I thought it already said echo the process, will fix and make the logs a little more clear.

nhooyr: Ended up removing `SKIP_ECHO` and printing all install commands and piping all output through a sed that replaces `$HOME` with `~`. Removes the verbosity enough.

install.sh
nhooyr: Fixed.

nhooyr:
Only thing left is to rename static -> binary.

install.sh
```diff
@@ -290,18 +285,20 @@ install_aur() {
...
   tmp_dir="$(mktemp -d)"
-  cd "$tmp_dir"
 
-  echo "+ Downloading PKGBUILD from https://aur.archlinux.org/cgit/aur.git/snapshot/code-server.tar.gz"
-  SKIP_ECHO=1 sh_c 'curl -fsSL https://aur.archlinux.org/cgit/aur.git/snapshot/code-server.tar.gz | tar -xz --strip-components 1'
-  unset SKIP_ECHO
-  echo "+ Downloaded into $tmp_dir"
-  sh_c makepkg -si
+  echo "+ Downloading PKGBUILD into $tmp_dir from https://aur.archlinux.org/cgit/aur.git/snapshot/code-server.tar.gz"
+  curl -fsSL https://aur.archlinux.org/cgit/aur.git/snapshot/code-server.tar.gz | tar -xzC "$tmp_dir"
```
code-asher, line 291: Should this still use `sh_c` so it doesn't download during a dry run?

nhooyr: Has to to get the $VERSION on the AUR.

code-asher: Ah yeah that makes sense. We could set a dummy version but tbh this seems fine to me.

code-asher: Better to have the real version even in a dry run for sure.

nhooyr: Yea, otherwise you don't know what's being installed.

PR #1705, opened 2020-05-20
Update CONTRIBUTING.md

oonqt:
It is not mentioned that the "release" script requires "jq" to be installed on the system to build code-server releases


nhooyr:
@oonqt Thank you so much! <3

PR #1718, opened 2020-05-24
Update README.md

KolyaKorruptis:
Clarify right from the start that code-server is serving the OSS version of vscode, to prevent misunderstandings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # code-server
 
-Run [VS Code](https://github.com/Microsoft/vscode) on any machine anywhere and access it in the browser.
+Run [VS Code OSS](https://github.com/Microsoft/vscode) on any machine anywhere and access it in the browser.
 
 - **Code everywhere:** Code on your Chromebook, tablet, and laptop with a
   consistent dev environment. Develop on a Linux machine and pick up from any


nhooyr:
I think it might be more confusing than helpful. I think an FAQ entry might be a better option here.

KolyaKorruptis:
I agree that a FAQ section would be useful to explain the details. 

But I don't see how correctly stating the basis and goal of this project would be more confusing than the current wording. You can call it "Community Builds of Visual Studio Code" (like headmelted does) or "VSCode OSS" (like MS does) or differentiate it from the commercial MS product in any other way. Just make it clear what you are offering.

nhooyr:
> But I don't see how correctly stating the basis and goal of this project would be more confusing than the current wording. You can call it "Community Builds of Visual Studio Code" (like headmelted does) or "VSCode OSS" (like MS does) or differentiate it from the commercial MS product in any other way. Just make it clear what you are offering.

Only power users are aware of the difference. As far as most people are concerned, there's only VS Code.

KolyaKorruptis:
That is a why you should make people aware of the differences upfront. Otherwise they run into confusion and problems lateron, when their favourite extensions don't seem to exist, or aren't up to date. Even power users will want to know what they get right away.

nhooyr:
Discussed with @code-asher and we think it'd be best to avoid user confusion by making it so apparent. Very few users are aware and it'd be more confusing than helpful.

It's best that the first FAQ question be `What are the differences from normal VS Code?`.

nhooyr:
I hope https://github.com/cdr/code-server/pull/1736 addresses your concerns well enough.

Issue #1738, opened 2020-05-28
issues including libstdc++ (or how to get eaten by a grue)


geiseri:
I just installed the Debian package of code server on Ubuntu Eoan and any C++ application that tries to run as a child of code server is now broken with the following error:

```
cmake: /usr/lib/code-server/bin/../lib/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by cmake)
cmake: /usr/lib/code-server/bin/../lib/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by cmake)
cmake: /usr/lib/code-server/bin/../lib/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by cmake)
cmake: /usr/lib/code-server/bin/../lib/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by cmake)
cmake: /usr/lib/code-server/bin/../lib/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by /lib/x86_64-linux-gnu/libjsoncpp.so.1)
```
This is because the libstdc++ included is only at ABI level `CXXABI_1.3.8` and in my case, `libjsoncpp` was linked against a symbol first only found in  `CXXABI_1.3.9`.  If I remove that libstdc++ from code it all works again.

Traditionally statically linking (or just flat out including) libstdc++ is a disaster. 

\<rant> 
Why it is 2020 and this has been a problem for since the first time we had dynamic linking can only be attributed to the C++ standards committee hating people who ship software and the GCC authors being sadists who enjoy torturing kitten when they are not torturing developers.  While #1706 was a noble dream, it will break so much more than you will fix, your hampster will leave you, your house will burn down, you will die alone, and more than likely be writing a bug report like this in 5 years for someone else's project.  
 \</rant>

[...]

PR #1740, opened 2020-05-28
Trim LD_LIBRARY_PATH on startup

nhooyr:
Not sure if this will work or we have to somehow set these variables
every time the bundled node binary is ran and then trim them on startup.

Updates #1738

### DIFF
--- a/src/node/entry.ts
+++ b/src/node/entry.ts
@@ -122,25 +122,7 @@ const main = async (args: Args, cliArgs: Args, configArgs: Args): Promise<void> 
   }
 }
 
-function trimLDLibraryPath(): void {
-  let ldVar: string
-  if (process.platform === "linux") {
-    ldVar = "LD_LIBRARY_PATH"
-  } else if (process.platform === "darwin") {
-    ldVar = "DYLD_LIBRARY_PATH"
-  } else {
-    return
-  }
-
-  // Removes the leading path added by ./ci/build/code-server.sh to use our bundled
-  // dynamic libraries. See ci/build/build-standalone-release.sh
-  // This is required to avoid child processes using our bundled libraries.
-  process.env[ldVar] = process.env[ldVar]?.replace(path.dirname(process.execPath) + ":", "")
-}
-
 async function entry(): Promise<void> {
-  trimLDLibraryPath()
-
   const tryParse = async (): Promise<[Args, Args, Args]> => {
     try {
       const cliArgs = parse(process.argv.slice(2))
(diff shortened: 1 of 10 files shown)


src/node/entry.ts
```diff
@@ -126,19 +122,40 @@ const main = async (cliArgs: Args): Promise<void> => {
...
+  } else if (process.platform === "darwin") {
+    ldVar = "DYLD_LIBRARY_PATH"
+  } else {
+    return
+  }
+
+  // Removes the leading path added by ./ci/build/code-server.sh to use our bundled
+  // dynamic libraries. See ci/build/build-standalone-release.sh
+  // This is required to avoid child processes using our bundled libraries.
+  process.env[ldVar] = process.env[ldVar]?.replace(path.dirname(process.execPath) + ":", "")
```
geiseri, line 138: I am not sure that `LD_LIBRARY_PATH` is populated at this point so it might be empty.  You can confirm correct operation if you run the following from a child process: 

```
ldd  $(which cmake) | grep code-server
```

if you see any output it is still pointing at code server then it might not have taken.  I think you might need to do something like AppImage where they rewrite the linker path for the binary to their library paths, that way you will not pollute downstream.  I am not sure if this https://medium.com/@nehckl0/creating-relocatable-linux-executables-by-setting-rpath-with-origin-45de573a2e98 is feasible.   That might be the best hope since you can put that in the installer when the location is selected.

code-asher: I built off this PR and can confirm there's no output for that command whereas there was before.

src/node/entry.ts
```diff
@@ -126,19 +122,40 @@ const main = async (cliArgs: Args): Promise<void> => {
   }
 }
 
+function trimLDLibraryPath(): void {
+  let ldVar: string
+  if (process.platform === "linux") {
```
code-asher, line 127: I think we might want to check if the variable exists or something rather than use the platform because `process.platform` returns non-`linux` values even if `uname` normally returns `Linux`. https://nodejs.org/api/process.html#process_process_platform

code-asher: Actually that might only be the case for Android. :thinking: The rest seem to match up in terms of `uname` and `process.platform`.

nhooyr:
Closing as this isn't going to work and be a good solution.

There's no good way to ensure every time the node binary is invoked, we have our `LD_LIBRARY_PATH` set but when any plugin for example invokes an external process, it doesn't have our variables set.

The plan is to use a static binary on macOS and switch to building the static release on CentOS which should be enough to point everyone else to just use the npm package if they're on something older.

PR #1761, opened 2020-06-03
Stop bundling libraries in release

nhooyr:
- Instead we now use CentOS 7 for the static build to guarantee
that we only depend on libc v2.17

- For macOS we now pull in a static node binary and bundle that instead.

Updates #1738


ci/images/centos7/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+RUN yum groupinstall -y 'Development Tools'
+RUN yum install -y python2 libsecret-devel libX11-devel libxkbfile-devel
+
+RUN npm config set python python2
+
+RUN yum install -y epel-release && yum install -y jq
+RUN yum install -y rsync
+
+# Copied from ../Dockerfile
```
code-asher, line 15: I think this may mean to point to `../debian8/Dockerfile`.

code-asher:
You probably already saw the build error, but I'll make note there's a missing `/` just in case.

code-asher:
Er, I mean `\`.

nhooyr:
Mhm yea, pushing rn.

meain:
Are you planning on making a release with this fix on top of 3.4.0 soon?

nhooyr:
Yes 3.4.1 should be out today.

nhooyr:
New release is out!

https://github.com/cdr/code-server/releases/tag/v3.4.1

Issue #1758, opened 2020-06-02
Release nomenclature inconsistency


aptalca:
There is a general inconsistency between versioning nomenclature. Some releases are in the format of `X.X.X` and some are `vX.X.X`. In the past, stable releases used one format and pre-releases used another. But now there are two stable releases with the same version number, one with and one without the `v`.

Preferably code-server would pick one nomenclature and run with that for everything (including the npm packages). Constant changes to version and filename formats of bins/packages without any notice make it very difficult for downstream automation.

nhooyr:
I agree there have been mistakes here and I'm sorry!

That second release only exists to make autoupdates work from `3.2.0` work. 

Background at https://github.com/cdr/code-server/releases/tag/v3.4.0

- We made some changes to make autoupdates from `v3.2.0` work again.
   - To recap we removed autoupdating in `v3.3.0` and restructured
      the releases a bit which caused autoupdates on `v3.2.0` to break. See  #1664
   - To fix this:
      1. We added a symlink from `code-server` to `bin/code-server` to the root of the release.
      2. We now upload duplicate release assets for the release naming scheme used by `v3.2.0`.
          - i.e we have `macos-amd64` and `darwin-x86_64` as well as `linux-amd64` and `linux-x86_64`.
  - In the next few weeks the symlinks and the duplicate release assets will be removed so
     please update your scripts to use the new release scheme.

All main releases tags will indeed be prefixed with `v`, we're just trying to update users on the old release structure.

nhooyr:
Updated the release description to make this more clear.

Issue #1750, opened 2020-05-31
user-data-dir, extensions-dir  in config file still doesn't work


silenceleaf:
user-data-dir, extensions-dir 

in config file still doesn't work

please test it.

Version: 3.4.0

nhooyr:
For posterity, this was fixed by https://github.com/cdr/code-server/commit/2c2a6498afe6588e1c2221d2fc2d748a5589c919

nhooyr:
New release is out!

Issue #1746, opened 2020-05-29
404 when code-server is downloading a newer release


keiththomps:
When prompted to install a new version through the UI I'm receiving a 404 error downloading the newest release. It appears to be caused by the versions 3.30 and later being tagged with a `v3.3.0` but the associated release file name doesn't include the `v`.

Here's the URL being used to download the newest release vs the newest release's actual URL:

Attempted (404 error): https://github.com/cdr/code-server/releases/download/v3.4.0/code-server-v3.4.0-linux-x86_64.tar.gz
Actual: https://github.com/cdr/code-server/releases/download/v3.4.0/code-server-3.4.0-linux-x86_64.tar.gz

This could be "fixed" in the code [here](https://github.com/cdr/code-server/blob/master/src/node/app/update.ts#L332), but it might just be a consistency issue that can be addressed moving forward by adjusting the tag names and files to match.

- Web Browser: Chrome
- Local OS: macOS 10.15.4
- Remote OS: CentOS 7.8.2003
- Remote Architecture: x86_64
- `code-server --version`: 3.2.0

hasip-timurtas:
The same for me

liaocyintl:
The same for me from +1

nhooyr:
Should just work now!

nhooyr:
Published https://github.com/cdr/code-server/releases/tag/3.4.0

Sorry about this!

KolyaKorruptis:
I think this is a misunderstanding. The URL requested by code-server has a "v" before the version number, inside the file name. And that's what we need to update. A filename with "v3..4.0" in it.  
Removing the "v" from the tag won't help unfortunately.

When this is fixed, these URLs should work: 
https://github.com/cdr/code-server/releases/download/v3.4.0/code-server-v3.4.0-linux-x86_64.tar.gz
https://github.com/cdr/code-server/releases/download/v3.4.0/code-server-v3.4.0-linux-arm64.tar.gz

nhooyr:
It has a v because it’s filling the latest release tag into the template. With the v removed it should just work.

nhooyr:
Argh this is going to break the install script though. 🤦🏻‍♂️

nhooyr:
Fixing.

KolyaKorruptis:
In any case, it didn't find an arm64 release.
![20200602_232548](https://user-images.githubusercontent.com/4076947/83571732-95e04800-a528-11ea-8ef0-910c8d6b53f2.jpg)
[...]

Issue #1693, opened 2020-05-20
Code server missing globalStorage folder


nathanrpage97:
code server is missing a globalStorage folder on configuration:

For linux default location:

    ~/.local/share/code-server/User

Missing this issue causes extensions that rely on this folder to error (e.g., todo tree).

Edit: [Reference to vs code api](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalStoragePath)

nathanrpage97:
Actually this is something that todo tree should do based off of API.

nathanrpage97:
I missed the final part

    However, the parent directory is guaranteed to be existent

penuts7644:
As a quick fix you could just make a `globalStorage` directory under `~/.local/share/code-server/User/` and then reinstall Todo Tree.

nathanrpage97:
Yea thats what I did as a temp fix, but this was just for reporting a spec oversight.

nhooyr:
cc @code-asher

nathanrpage97:
Actually this is fixed in v3.3.1.

nhooyr:
Awesome!

dvyas1:
> Actually this is fixed in v3.3.1.

I am on v3.4 and this folder is still missing.

PR #1766, opened 2020-06-04
v3.4.1

nhooyr:
VS Code v1.45.1

# Bug Fixes

- Hopefully this one fixes autoupdates from `3.2.0` once and for all.
   - We're now publishing duplicate releases without the `v` prefix so that auto updates from `3.2.0` work. See #1746 and #1758 for background.
   - Note there is no auto updating in any of the recent releases including this one.

- The config file's `extension-dir` and `data-dir` parameters are now properly obeyed #1750
   - This was supposed to be fixed in the last release but because of a subtle bug it turned out it was printing the correct directories but not actually using them.
   - It should now use `extension-dir` and `data-dir` as set in the config file

- We do not bundle any dynamic libraries in releases anymore #1738
   - It was causing bizarre issues for users and we couldn't make it work perfectly.
   - On Linux we now compile on CentOS 7 so that the minimum glibc version is v2.17 and glibcxx is v3.4.18
   - On macOS we now pull in a fully static release of node.

- Using the deprecated root `code-server` symlink now warns on usage that it will be removed in the next few weeks.
   - The entry script now also handles recursive symlinks correctly.

- `globalStorageHome` is now created on startup #1693 so that extensions that rely on it being there should work now

### DIFF
--- a/ci/README.md
+++ b/ci/README.md
@@ -16,7 +16,7 @@ Make sure you have `$GITHUB_TOKEN` set and [hub](https://github.com/github/hub) 
 
 1. Update the version of code-server and make a PR.
    1. Update in `package.json`
-   2. Update in [install.sh](../install.sh)
+   2. Update in [./doc/install.md](../doc/install.md)
 2. GitHub actions will generate the `npm-package`, `release-packages` and `release-images` artifacts.
 3. Run `yarn release:github-draft` to create a GitHub draft release from the template with
    the updated version.
--- a/doc/install.md
+++ b/doc/install.md
@@ -77,8 +77,8 @@ commands presented in the rest of this document.
 ## Debian, Ubuntu
 
 ```bash
-curl -fOL https://github.com/cdr/code-server/releases/download/v3.4.0/code-server_3.4.0_amd64.deb
-sudo dpkg -i code-server_3.4.0_amd64.deb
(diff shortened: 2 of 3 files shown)


GeorgeDavis-TM:
Still seeing issues with v3.4.1.

`ubuntu@ip-172-31-70-10:~$ code-server -v`
`info  Using config file ~/.config/code-server/config.yaml` 
`3.4.1 48f7c2724827e526eeaa6c2c151c520f48a61259`

config.yaml looks like this

`bind-addr: 0.0.0.0:8080`
`auth: password`
`password: "346y3vh2"`
`cert: false`
`extension-dir: ~/vscode/ext`
`data-dir: ~/vscode/user-data`

Error message reads
`error reading /home/ubuntu/.config/code-server/config.yaml: Unknown option --extension-dir=~/vscode/ext`
`error reading /home/ubuntu/.config/code-server/config.yaml: Unknown option --data-dir=~/vscode/user-data`

@nhooyr  Any comments? Thank you.

code-asher:
@GeorgeDavis-TM Those options are a little bit off. Here's the relevant output from `--help`:
```
      --user-data-dir       Path to the user data directory.
      --extensions-dir      Path to the extensions directory.
```

PR #1767, opened 2020-06-04
FAQ Updates

nhooyr:
- Explain differences compared to Theia
- Explain $SERVICE_URL and $ITEM_URL in more detail


nhooyr:
For anyone in the future, this was merged into master, just through the CLI.

PR #1802, opened 2020-06-12
Add FreeBSD support for an npm based install

zathras777:
With this small change code-server can be installed via npm on a FreeBSD amd64 install.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -360,6 +360,9 @@ os() {
   Darwin)
     echo macos
     ;;
+  FreeBSD)
+    echo freebsd
+    ;;
   esac
 }
 


zathras777:
It may well be the standalone release that's installed, but with this
change the install script can at least be used to perform an install.
Without this patch the script simply does nothing :-(

$ curl -fsSL https://code-server.dev/install.sh | sh -s -- --dry-run
Unsupported OS FreeBSD.

nhooyr:
> It may well be the standalone release that's installed, but with this
change the install script can at least be used to perform an install.
Without this patch the script simply does nothing :-(

As FreeBSD does not use glibc, the standalone release won't work.

I'll modify the install script to instead install via npm when an unrecognized unix system is detected instead of erring out.

nhooyr:
See https://github.com/cdr/code-server/issues/1803

nhooyr:
Thanks again for finding this @zathras777!

zathras777:
Whoops, sent a bit too soon :-) Following the patch,

$ sh ./install.sh
FreeBSD 12.1-RELEASE-p3
No precompiled releases for amd64.
Installing with npm.

nhooyr:
Hmm. That's confusing, we do have precompiled releases for amd64? What does `uname -m` give you?

zathras777:
FreeBSD

$ uname -m
amd64

Linux,

$ uname -m
x86_64

Think the confusion comes from arch() being used for the logic
decision and then plain old "uname -m" in the echo statement.

ARCH="$(arch)"
  if [ ! "$ARCH" ]; then
    if [ "$METHOD" = standalone ]; then
      echoerr "No releases available for the architecture $(uname -m)."
      echoerr 'Please rerun without the "--method standalone" flag to
install from npm.'
      exit 1
    fi
    echoh "No precompiled releases for $(uname -m)."
    install_npm
    return
  fi

That probably explains why with the simple patch the script installs
an npm build :-)

nhooyr:
Yea that makes sense. `uname -m` on linux always returns x86_64. I'll fix in #1803

GogoFC:
Couldn't install on FreeBSD.

With npm version 8 installed, gave error:
```sh
Please install npm or yarn to install code-server!
You will need at least node v12 and a few C dependencies.
See the docs https://coder.com/docs/code-server/latest/install#yarn-npm
```
After installing yarn, error:
```sh
[1/4] Resolving packages...
[2/4] Fetching packages...
error code-server@3.12.0: The engine "node" is incompatible with this module. Expected version "= 14". Got "16.10.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/global for documentation about this command.
```

PR #1858, opened 2020-07-02
Update install.md

clandestinz:
docker run point to different repo. updated so that coresponds to the link provided below


nhooyr:
Hi, thanks for contributing. I believe the docs are correct as is.

The below image is linked only for arm32.

> For `arm32` support there is a popular community maintained alternative:

Issue #770, opened 2019-06-13
Reverse Proxy Issue


AmansRevenger:
- `code-server` version: v1.1156-vsc1.33.1
- OS Version: Ubuntu 18.04 

## Description

Reverse Proxy doesnt work at all.

It either results in failed SSL Handshake or

this:

![image](https://user-images.githubusercontent.com/18306225/59458684-41c1ec80-8e1b-11e9-9c9e-a4fcdc235e66.png)

![image](https://user-images.githubusercontent.com/18306225/59458746-66b65f80-8e1b-11e9-9a2d-3447f7a2cbe0.png)

Reverse Proxy Settings with working grafana above:

![image](https://user-images.githubusercontent.com/18306225/59458842-a4b38380-8e1b-11e9-8893-4735ae4befc5.png)

What am I missing?

RyoshiKayo:
Your proxy pass needs to be `http://localhost:8443` not `http://giovanni:8443` I believe. I'm not sure where you got `giovanni` from.

AmansRevenger:
giovanni is the hostname of the server in the network where the docker containers and services run, nginx itself is a docker container, so localhost wont work...

fbartels:
To me it seems the issue is that you are trying to proxy Code-Server as a subfolder. AFAIK it has to run in root of the domain.

AmansRevenger:
So mywebsite.com/vscode would never work?

fbartels:
No, at least not how it is currently designed. You would need to setup vscode.mywebsite.com.

A lot of programs are designed this way today. With sni and free certs from let's encrypt this is also not much work. Only potential roadblock would be DNS providers not allowing additional DNS entries.

AmansRevenger:
I already use letsEncrypt, but I dont think I can setup a whole subdomain...

deansheather:
At the moment (until #722 is merged), code-server doesn't work when it's not mounted at root (`/`) in your reverse proxy.

You should be able to easily issue a new certificate for a subdomain on your domain through letsencrypt very easily, and then create a new virtual host in nginx.

abubakar-bilal:
@deansheather I just created a subdomain and added configuration for proxy. But it gives me following error

Nginx conf

```
server {
  listen 443 ssl;
  server_name ide.mysite.com;

  location / {
    proxy_pass http://localhost:8443;
    proxy_set_header Host            $host:$server_port;
    proxy_set_header X-Forwarded-For $remote_addr;
[...]

PR #1867, opened 2020-07-06
Add Nginx instructions to guide

Niek:
Added Nginx instructions for people who prefer to use this instead of Caddy

See for example issue #770

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -244,6 +244,25 @@ To avoid the warnings, you can use [mkcert](https://mkcert.dev) to create a self
 trusted by your OS and then pass it into `code-server` via the `cert` and `cert-key` config
 fields.
 
+### Nginx reverse proxy
+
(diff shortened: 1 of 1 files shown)


nhooyr:
My reasoning to leave this out was that any user that can't setup Nginx without an example should be using Caddy instead. It's better on every front.

nhooyr:
cc @code-asher

nhooyr:
We've had so many issues filed from people just not understanding how to configure nginx.

code-asher:
Yeah maybe it's worth having it just on account of all the issues. If so, we should add a trailing slash to the `proxy_pass` line since that makes sub-paths work and I've had to explain that at least a few times in various issues I think.

Niek:
IMHO it never hurts to have *too* much documentation. Even if Caddy is good for 95% of the users, I think those 5% with existing nginx installs would appreciate it.

I've added the trailing / now.

code-asher:
Thanks for the contribution!

PR #1874, opened 2020-07-09
Update VS Code to 1.47.0

code-asher:
Pretty straightforward update.

- Refactor `readFileStream` in file system provider to use VS Code's instead of copying the code.
- Update fix that allows any extension to be installed since that area was refactored in 1.47.0.
- Update location of Firefox scm viewlet fix (it was moved to another file).
- Remove csp fix since it's included in 1.47.0.

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -12,12 +12,12 @@ index e73dd4d9e8..e3192b3a0d 100644
  coverage/
 diff --git a/.yarnrc b/.yarnrc
 deleted file mode 100644
-index 1406d749d7..0000000000
+index 135e10442a..0000000000
 --- a/.yarnrc
 +++ /dev/null
 @@ -1,3 +0,0 @@
 -disturl "https://atom.io/download/electron"
--target "7.3.1"
+-target "7.3.2"
 -runtime "electron"
 diff --git a/build/gulpfile.reh.js b/build/gulpfile.reh.js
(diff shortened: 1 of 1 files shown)


code-asher:
Also added a trailing slash to the webview endpoint. It seems this passed to extensions as the source for CSP tags now, and without the trailing slash CSP refuses to load the resources for some reason. This was causing the pdf extension to break, for example.

ci/dev/vscode.patch
```diff
@@ -3248,57 +3233,38 @@ index cfac383e8a..c535d38296 100644
...
-+			// NOTE@coder: Allow extensions of any kind.
- 			return this.extensionManagementServerService.remoteExtensionManagementServer.extensionManagementService.installFromGallery(gallery);
+@@ -237,6 +237,11 @@ export class ExtensionManagementService extends Disposable implements IExtension
+ 			return this.extensionManagementServerService.webExtensionManagementServer.extensionManagementService.installFromGallery(gallery);
  		}
- 		return Promise.reject('No Servers to Install');
+ 
++		// NOTE@coder: Fall back to installing on the remote server.
++		if (this.extensionManagementServerService.remoteExtensionManagementServer) {
++			return this.extensionManagementServerService.remoteExtensionManagementServer.extensionManagementService.installFromGallery(gallery);
```
nhooyr, line 3245: good god 🤮

Issue #1898, opened 2020-07-21
Update code-server packages


sreya92:
The following packages need to be updated:

- `http-proxy` >= 1.18.1
-  Pin `node` to 12.18.1

nhooyr:
[v3.5.0 with VS Code v1.48.2 has been released, fixing this issue!](https://github.com/cdr/code-server/releases/tag/v3.5.0)

PR #1905, opened 2020-07-22
Update dependencies in package.json

nhooyr:
Updates #1898


nhooyr:
Yea not sure what happened there, will retry updating.

nhooyr:
So yea we do actually.

PR #1920, opened 2020-07-28
feat: persist route query to local

fxxjdedd:
Provide a way for the shell script running in the docker container to get the url query.

### DIFF
--- a/src/node/app/vscode.ts
+++ b/src/node/app/vscode.ts
@@ -183,14 +183,9 @@ export class VscodeHttpProvider extends HttpProvider {
       }),
     ])
 
-    let promise = Promise.resolve()
-    if (startPath) {
-      promise = settings.write({ lastVisited: startPath })
-    }
-    // `settings.write` depends on `settings.read` internally. To avoid race conditions, a promise is added here to synchronize.
-    promise.then(() => {
-      // the query should not be extends, but should be replaced directly.
-      settings.write({ query: route.query }, true)
+    settings.write({
+      lastVisited: startPath || lastVisited, // If startpath is undefined, then fallback to lastVisited
+      query: route.query,
     })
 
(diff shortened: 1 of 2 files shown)


code-asher:
Thanks for the contribution! Could you expound on the use case?

fxxjdedd:
Thank you for your reply.
The use case is: the URL of our web ide will carry some dynamic parameters, such as user, project, cluster, etc. These parameters need to be obtained in the debugger and terminal.

code-asher:
Interesting, thanks for the additional details. We already have a file that contains the last visited path in `~/.local/share/code-server/coder.json` (just a bit further down in `getRoot`). Would it work for your use case to add the query variables to the `lastVisited` object in that file instead?

fxxjdedd:
Of course, it will help a lot！


------------------ Original ------------------

code-asher:
Awesome, let's go that route then.

fxxjdedd:
@code-asher 

I made some changes and it works fine locally.

fxxjdedd:
emmm, I have some different opinions:
Although await is clearer, but it&nbsp;will blocks all the code below, but promise.then will not.&nbsp;
On the other hand, if they are written together, the query will be updated using extends, so that when the query becomes empty, the data in the settings will not be updated correctly


------------------ Original ------------------

code-asher:
That's fair, let's leave out the `await`.

We can pass `true` when updating both to use the shallow update. That's actually preferable anyway because the `lastVisited` object should be replaced entirely.

code-asher:
In fact I'm thinking about removing `extends` entirely because I'm not sure we even need a recursive merge.

fxxjdedd:
cool, let‘s have a try!

------------------ Original ------------------

PR #1926, opened 2020-07-30
Cannot find module 'rimraf'

suisrc:
fix ci for package error: Cannot find module 'rimraf'


suisrc:
docker image 🌰 for vsvode 1.47.3 in debian 10: suisrc/vscode:1.47.3

code-asher:
Thank you for the PR! It turns out the packaging step is still failing for a different reason so I patched that and decided to patch rimraf there instead of adding it as a package to the root in https://github.com/cdr/code-server/pull/1931.

PR #1927, opened 2020-07-30
Dead code

code-asher:
Removes all the code not currently in use.


src/node/app/update.ts
```diff
@@ -27,7 +18,7 @@ export interface LatestResponse {
 }
 
 /**
- * Update HTTP provider.
+ * HTTP provider for checking updates (does not download/install them).
```
nhooyr, line 21: Let's clarify this exists due to history. I think it'd make more sense to just check for updates client side.

src/node/http.ts
```diff
@@ -502,9 +508,15 @@ export class HttpServer {
...
+  /**
+   * Stop and dispose everything. Return an array of disposal errors.
+   */
+  public async dispose(): Promise<Error[]> {
     this.socketProvider.stop()
-    this.providers.forEach((p) => p.dispose())
+    const providers = Array.from(this.providers.values())
+    // Catch so all the errors can be seen rather than just the first one.
+    const responses = await Promise.all<Error | undefined>(providers.map((p) => p.dispose().catch((e) => e)))
+    return responses.filter<Error>((r): r is Error => typeof r !== "undefined")
```
nhooyr, line 519: Why do we need to use `typeof` here versus `!== undefined`?

src/node/http.ts
```diff
@@ -518,41 +530,51 @@ export class HttpServer {
...
         auth: this.options.auth || AuthType.None,
-        base: `/${endpoint}`,
         commit: this.options.commit,
         password: this.options.password,
       },
       ...args,
     )
-    this.providers.set(`/${endpoint}`, p)
-    return p
+    const endpoints = (typeof endpoint === "string" ? [endpoint] : endpoint).map((e) => e.replace(/^\/+|\/+$/g, ""))
```
nhooyr, line 562: What's this doing exactly? Seems much more involved than before. Would be good to have a quick comment here.

src/node/app/update.ts
code-asher: When we have a dashboard I was thinking it would be helpful to show whether your version of code-server was out of date. Currently all the HTML is generated server-side so if we keep that paradigm we'll need this to be server-side as well. 

The other thing is that having this as an endpoint lets the checking code be shared between vscode and code-server without having to copy it or bundle it into both code-server and vscode, or something like that, if that makes sense.

nhooyr: Makes sense.

src/node/http.ts
code-asher: Practically no difference, just a stylistic choice. It might be because I'm thinking of it in terms of the type rather than the value. It parallels what I'd do if I wanted to filter strings instead, and I like the consistency. It also works for undeclared variables, further enforcing the consistency.

It might also be partly force of habit from back when you could overwrite undefined on the global object. :stuck_out_tongue:

But even though the global undefined is read-only now it still feels weird to me that I can do `const undefined = "something"` and now my checks have a different meaning.

src/node/http.ts
code-asher: Agreed, it's a bit of a mess. It's the same code (trimming slashes on the end) but with the addition of wrapping the endpoint in an array if it's not already an array. I put it on all one line for some reason. :facepalm:

PR #1934, opened 2020-07-31
Plugins

code-asher:
Plugins provide an external way to add endpoints to code-server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ release-gcp/
 release-images/
 node_modules
 node-*
+/plugins
--- a/ci/build/build-code-server.sh
+++ b/ci/build/build-code-server.sh
@@ -18,7 +18,7 @@ main() {
   fi
 
   parcel build \
-    --public-url "/static/$(git rev-parse HEAD)/dist" \
+    --public-url "." \
     --out-dir dist \
     $([[ $MINIFY ]] || echo --no-minify) \
     src/browser/register.ts \
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -21,6 +21,10 @@ main() {
   rsync README.md "$RELEASE_PATH"
   rsync LICENSE.txt "$RELEASE_PATH"
   rsync ./lib/vscode/ThirdPartyNotices.txt "$RELEASE_PATH"
+
+  # Keep these types since code-server's exported types use them.
+  mkdir -p "$RELEASE_PATH/lib/vscode/src/vs/server"
+  rsync ./lib/vscode/src/vs/server/ipc.d.ts "$RELEASE_PATH/lib/vscode/src/vs/server"
 }
 
 bundle_code_server() {
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -22,7 +22,9 @@ main() {
   rsync LICENSE.txt "$RELEASE_PATH"
   rsync ./lib/vscode/ThirdPartyNotices.txt "$RELEASE_PATH"
 
-  # Keep these types since code-server's exported types use them.
+  # code-server exports types which can be imported and used by plugins. Those
+  # types import ipc.d.ts but it isn't included in the final vscode build so
+  # we'll copy it ourselves here.
   mkdir -p "$RELEASE_PATH/lib/vscode/src/vs/server"
   rsync ./lib/vscode/src/vs/server/ipc.d.ts "$RELEASE_PATH/lib/vscode/src/vs/server"
 }
--- a/src/browser/pages/vscode.html
+++ b/src/browser/pages/vscode.html
@@ -24,21 +24,17 @@
     <meta id="vscode-remote-nls-configuration" data-settings="{{NLS_CONFIGURATION}}" />
 
     <!-- Workbench Icon/Manifest/CSS -->
-    <link rel="icon" href="{{BASE}}/static/{{COMMIT}}/src/browser/media/favicon.ico" type="image/x-icon" />
-    <link
-      rel="manifest"
-      href="{{BASE}}/static/{{COMMIT}}/src/browser/media/manifest.json"
-      crossorigin="use-credentials"
-    />
+    <link rel="icon" href="{{CS_STATIC_BASE}}/src/browser/media/favicon.ico" type="image/x-icon" />
+    <link rel="manifest" href="{{CS_STATIC_BASE}}/src/browser/media/manifest.json" crossorigin="use-credentials" />
     <!-- PROD_ONLY
-    <link data-name="vs/workbench/workbench.web.api" rel="stylesheet" href="{{BASE}}/static/{{COMMIT}}/lib/vscode/out/vs/workbench/workbench.web.api.css">
+    <link data-name="vs/workbench/workbench.web.api" rel="stylesheet" href="{{CS_STATIC_BASE}}/lib/vscode/out/vs/workbench/workbench.web.api.css">
     END_PROD_ONLY -->
-    <link rel="apple-touch-icon" href="{{BASE}}/static/{{COMMIT}}/src/browser/media/pwa-icon-384.png" />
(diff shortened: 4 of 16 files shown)


src/browser/pages/vscode.html
```diff
@@ -48,10 +44,6 @@
 
   <!-- Startup (do not modify order of script tags!) -->
   <script>
-    const parts = window.location.pathname.replace(/^\//g, "").split("/")
```
nhooyr, line 51: Ah I see how it made lots of things simpler by making the path directly correspond to the file.

src/node/app/static.ts
```diff
@@ -32,17 +31,25 @@ export class StaticHttpProvider extends HttpProvider {
   /**
    * Return a resource with variables replaced where necessary.
    */
-  protected async getReplacedResource(route: Route): Promise<HttpResponse> {
+  protected async getReplacedResource(request: http.IncomingMessage, route: Route): Promise<HttpResponse> {
     // The first part is always the commit (for caching purposes).
```
nhooyr, line 35: Shouldn't this be removed.

src/node/app/static.ts
```diff
@@ -32,17 +31,25 @@ export class StaticHttpProvider extends HttpProvider {
...
     const split = route.requestPath.split("/").slice(1)
 
+    const resourcePath = path.resolve("/", ...split)
+
+    // Make sure it's in code-server or a plugin.
+    const validPaths = [this.rootPath, process.env.PLUGIN_DIR]
+    if (!validPaths.find((p) => typeof p !== "undefined" && p.length > 0 && resourcePath.startsWith(p))) {
```
nhooyr, line 42: I think we can just use `p &&` instead of the explicit undefined and length check.

src/node/http.ts
```diff
@@ -240,16 +240,17 @@ export abstract class HttpProvider {
     response: HttpStringFileResponse,
     extraOptions?: Omit<T, "base" | "csStaticBase" | "logLevel">,
   ): HttpStringFileResponse {
+    const base = this.base(route)
     const options: Options = {
-      base: this.base(route),
-      commit: this.options.commit,
+      base,
+      csStaticBase: base + "/static/" + this.options.commit + this.rootPath,
```
nhooyr, line 246: Oh nvm, ah I see, the commit is still here but we store the base instead of just the commit.

src/node/plugin.ts
```diff
@@ -49,4 +49,8 @@ export const loadPlugins = async (httpServer: HttpServer, args: Args): Promise<v
       logger.warn(error.message)
     }
   }
+
+  if (process.env.PLUGIN_DIR) {
+    await loadPlugin(process.env.PLUGIN_DIR, httpServer, args)
```
nhooyr, line 54: Perhaps it'd be better to call `_loadPlugins` with this directory instead to allow multiple external plugins in an arbitrary directory?

src/common/util.ts
```diff
@@ -40,21 +39,28 @@ export const trimSlashes = (url: string): string => {
...
 }
 
+/**
+ * Resolve a relative base against the window location. This is used for
+ * anything that doesn't work with a relative path.
+ */
+export const resolveBase = (base?: string): string => {
+  // After resolving the base will either start with / or be an empty string.
+  if (!base || base.startsWith("/")) {
+    return base || ""
```
vapurrmaid, line 49: Could use `??` here since we're defaulting to `""`

src/common/util.ts
```diff
@@ -40,21 +39,28 @@ export const trimSlashes = (url: string): string => {
...
+  }
+  const parts = location.pathname.replace(/^\//g, "").split("/")
```
vapurrmaid, line 51: Looks to me like we don't need `.replace(/^\//g, "")` here because we return if `base.startsWith("/")`

.gitignore
```diff
@@ -10,3 +10,4 @@ release-gcp/
 release-images/
 node_modules
 node-*
+/plugins
```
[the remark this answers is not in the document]
code-asher: Fair, is there any downside to being explicit though?

ci/build/build-code-server.sh
```diff
@@ -18,7 +18,7 @@ main() {
   fi
 
   parcel build \
-    --public-url "/static/$(git rev-parse HEAD)/dist" \
+    --public-url "." \
```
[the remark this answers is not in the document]
code-asher: I think this only affects the source map URL (at least for our setup; I couldn't find any other place where it was used or output in the compiled files) so now the source maps will be loaded relative to their source rather than from an absolute path. My thinking was this was a bit more flexible since if the static path changes we don't have to change it here as well. 

I think the absolute path would also cause an issue if you were serving code-server from a sub-path like `domain.tld/code-server/`.

ci/build/build-release.sh
```diff
@@ -21,6 +21,10 @@ main() {
   rsync README.md "$RELEASE_PATH"
   rsync LICENSE.txt "$RELEASE_PATH"
   rsync ./lib/vscode/ThirdPartyNotices.txt "$RELEASE_PATH"
+
+  # Keep these types since code-server's exported types use them.
```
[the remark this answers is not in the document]
code-asher: Will rework this comment :+1:

src/node/plugin.ts
```diff
@@ -0,0 +1,52 @@
...
+const loadPlugin = async (pluginPath: string, httpServer: HttpServer, args: Args): Promise<void> => {
+  try {
+    const plugin: Plugin = require(pluginPath)
+    plugin.activate(httpServer, args)
+    logger.debug("Loaded plugin", field("name", path.basename(pluginPath)))
+  } catch (error) {
+    if (error.code !== "MODULE_NOT_FOUND") {
+      logger.warn(error.message)
+    } else {
+      logger.debug(error.message)
```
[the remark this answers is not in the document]
code-asher: Oh yeah that's true. Will change.

src/node/plugin.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+
+export type Activate = (httpServer: HttpServer, args: Args) => void
+
+export interface Plugin {
+  activate: Activate
+}
+
+const originalLoad = require("module")._load
```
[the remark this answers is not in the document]
code-asher: Will do.

src/node/app/static.ts
code-asher: The comment? I think it's still accurate, the first part is sliced off because it's the commit.

src/node/app/static.ts
code-asher: Oh yeah duh. Good point.

src/node/app/static.ts
code-asher: Oh just saw your comment below.

src/node/plugin.ts
code-asher: I had a very similar thought but I wasn't sure how to make that work with the current x11wasm structure. Currently it's just a `code-server` directory in the `x11wasm` repo so it would try to load every directory in there. Maybe we could have a `code-server-plugins` directory inside `x11wasm`?

code-asher: Another thought is to require symlinking into `plugins` in the `code-server` directory instead of an environment variable. Or maybe make `PLUGIN_DIR` able to contain multiple paths.

src/common/util.ts
code-asher: Oh snap good catch.

.gitignore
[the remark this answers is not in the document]
nhooyr: No, in fact I guess we should be doing this for all ignores that aren't global. Feel free to leave this as is.

code-asher:
I'm thinking about merging this in soon and if there are any tweaks we want to make to how plugins are loaded we can do that as we experiment with it. Lemme know if there are any objections!

vapurrmaid:
All good on my end!

PR #1954, opened 2020-08-09
Simplify HttpError

vapurrmaid:
This change is cosmetic (issue not linked due to being extremely minor). As part of reading the codebase for the first time, I noticed `details` is not used anywhere.

* Remove public keyword - public by default
* Remove details - this was unused throughout the entire codebase

### DIFF
--- a/src/common/http.ts
+++ b/src/common/http.ts
@@ -9,7 +9,7 @@ export enum HttpCode {
 }
 
 export class HttpError extends Error {
-  public constructor(message: string, public readonly code: number, public readonly details?: object) {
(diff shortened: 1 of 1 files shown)


code-asher:
Ah, this actually exists so plugins can add details if they throw errors. It's true none of the code in code-server itself uses it (although maybe it should if there is a spot where some extra details could be useful).

vapurrmaid:
Wasn't aware, am now - TY @code-asher

code-asher:
We should probably create an explicit file with the types we want to export rather than just exporting everything so we know what's safe to change.

Issue #1587, opened 2020-05-01
Switch to git subtree for lib/vscode instead of a submodule


nhooyr:
Our patch is massive, a subtree would be a better option at this point.

Discussed and approved by @code-asher

nhooyr:
Also see discussion at https://github.com/cdr/code-server/pull/1982

SPGoding:
I just experimented a bit and it seems pretty simple to migrate to `git subtree` :eyes:

1. Remove the Submodule: delete `.gitmodules` and `lib/vscode` and commit.
2. Add Subtree: `git subtree add -P lib/vscode https://github.com/microsoft/vscode.git release/1.51 --squash`
	- This added all the content of Vanilla VS Code's `release/1.51` branch to `code-server/lib/vscode` in a single commit. 
3. Bye Bye Patch:
	1. `git apply --directory=lib/vscode ci/dev/vscode.patch`
		- All the meaty changes are now applied to the local `lib/vscode` directory.
		- Note: `yarn vscode:patch` won't work here unless you add `--directory=lib/vscode` to the script, which isn't worth it as we will delete the script really soon.
	2. Commit something like `Apply patch`.
	3. Delete `ci/dev/vscode.patch`.
	4. Delete `ci/dev/patch-vscode.sh` and npm `vscode:patch` script.
	5. Delete `ci/dev/diff-vscode.sh` and npm `vscode:diff` script.
		- It's actually still possible to diff between the local VS Code and the Vanilla VS Code.
		- First set up remotes: `git remote add vscode https://github.com/microsoft/vscode.git && git fetch vscode`
		- Then `git diff vscode/release/1.51 HEAD:lib/vscode`
	6. Commit. Perhaps I missed a lot of scripts and workflows that should be changed though.

Now every changes made to `lib/vscode` can be reviewed like how it should be. It just works™.

If you want to update the local `lib/vscode` with Vanilla VS Code, run `git subtree pull -P lib/vscode <VS Code remote name> <VS Code remote branch name> --squash`.

nhooyr:
Thank you for writing down the process @SPGoding! Going to make closing this issue much easier.

jsjoeio:
Closed via #2476

PR #1982, opened 2020-08-18
VS Code 1.48.0

code-asher:


nhooyr:
Thoughts on merging into a `dev` branch and then merging `dev` into `master` when we release?

Would avoid confusion re the version in package.json and documentation updates for things that haven't made it into master.

nhooyr:
Lack of a git subtree is also contributing to massive commits to the patch.

code-asher:
Smaller patchfiles do seem like would help reviewability, and maybe even with conflict resolution. I think many of the core issues remain though so I'd be happier without any patches. :smile:

nhooyr:
Lets hash it out in https://github.com/cdr/code-server/pull/1982

nhooyr:
Whoops, meant to link #1587

PR #1984, opened 2020-08-19
Added /healthz JSON response for heartbeat data. #1940

### LINKED ISSUE #1940: healthcheck path
It looks like any http request will cause the heartbeat file to be updated.
Is it possible to add a health checking path that just returns an "OK" response without updating the heartbeat?
That way a reverse proxy can check to see if the server is up without looking like user activity.

JacobGo:
I added a JSON object with heartbeat information by passing the server's Heart to the new HealthHttpProvider. I wasn't quite sure [what status code to use](https://stackoverflow.com/questions/25389261/which-http-status-code-should-i-use-for-a-health-check-failure). I was considering `503` for expired health checks but settled on `200` with state in the response content. 
I haven't written TypeScript before, perhaps a type for the `result` object would be better?
I made sure not to heartbeat for onRequest with the `/healthz` route.
Closes #1940

Side note: when can we expect a new release? 3.4.1 is two months old and **still** has a broken heart 🙃  (zombie connections fixed by ebef18d6262b310953293860ab2a55f3bee3d112)


code-asher:
As for the new release, we should have one out this week.

code-asher:
One interesting note is that if you use a browser then it requests the service worker which triggers a heart beat. But since this is going to be used by things that aren't browsers I think it should be fine.

nhooyr:
I'm so sorry about this! Deleted the dev branch was gonna switch this back to base on master but GitHub doesn't let you do that....

🤦 

Please open a new PR with these changes! Once again, I'm so sorry!

nhooyr:
Nvm, got it!

src/node/app/health.ts
```diff
@@ -0,0 +1,43 @@
...
+export class HealthHttpProvider extends HttpProvider {
+
+    public constructor(
+        options: HttpProviderOptions,
+        private readonly heart: Heart
+      ) {
+        super(options)
+    }
+
+    private alive(): Boolean {
```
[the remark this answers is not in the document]
JacobGo: Sure, I went ahead and put `alive()` in Heart.

JacobGo:
Figured out how to run the linter 😅
Hopefully this will make it into 3.5.1! Love the project, I might try to tackle some more milestones, time permitting.

code-asher:
Everything looks great. Thank you for the contribution!

src/node/app/health.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Check the heartbeat.
+ */
+export class HealthHttpProvider extends HttpProvider {
+  public constructor(options: HttpProviderOptions, private readonly heart: Heart) {
+    super(options)
+  }
+
+  public async handleRequest(route: Route, request: http.IncomingMessage): Promise<HttpResponse> {
+    if (!this.authenticated(request)) {
```
nhooyr, line 14: Shouldn't this endpoint work regardless of whether or not we're authenticated? @code-asher

code-asher: Ah crap you're right.

PR #1991, opened 2020-08-21
install.sh: Add remote install support

nhooyr:
Closes cdr/sshcode#185 and pretty much archives that
repository.

PR #1996, opened 2020-08-25
install.sh: Add remote install support

nhooyr:
Closes cdr/sshcode#185, #1996, and pretty much archives that repository.

See also #1991

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,26 +17,44 @@ usage() {
 Installs code-server for Linux, macOS and FreeBSD.
 It tries to use the system package manager if possible.
 After successful installation it explains how to start using code-server.
+
+Pass --start to startup code-server immediately, print the URL it can be
+accessed at and the initial password. Then the script will tail code-server's logs.
+
+Pass in user@host to install code-server on user@host over ssh.
+Pass --start to forward the code-server port and start it so that
+you can immediately access it.
+
+If you rerun the script, code-server will be updated only if necessary.
 ${not_curl_usage-}
 Usage:
 
-  $arg0 [--dry-run] [--version X.X.X] [--method detect] [--prefix ~/.local]
+  $arg0 [--dry-run] [--version X.X.X] [--method detect] [--prefix ~/.local] [--start] [user@host]
 
   --dry-run
       Echo the commands for the install process without running them.
+
   --version X.X.X
       Install a specific version instead of the latest.
+
   --method [detect | standalone]
       Choose the installation method. Defaults to detect.
       - detect detects the system package manager and tries to use it.
         Full reference on the process is further below.
       - standalone installs a standalone release archive into ~/.local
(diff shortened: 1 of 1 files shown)


nhooyr:
Yea I think I'ma update all the docs/messages when installing remotely and modify fetch to fetch locally and then copy over ssh so that it works in closed off environments.

install.sh
```diff
@@ -17,27 +17,44 @@ usage() {
...
+
+Pass in user@host to install code-server on user@host over ssh.
+Pass --start to forward the code-server port so that you can immediately access it.
+
+If you rerun the script, code-server will be updated only as required.
 ${not_curl_usage-}
 Usage:
 
-  $arg0 [--dry-run] [--version X.X.X] [--method detect] [--prefix ~/.local]
+  $arg0 [--dry-run] [--version X.X.X] [--method detect] [--prefix ~/.local] [--start] [user@host]
```
kylecarbs, line 31: In `--method detect` the `detect` keyword is variable.

In `--start` we have `user@host` separated. Is this intentional?

nhooyr: Yes, `--start` can be used without `user@host`.

kylecarbs: Ah I understand.

install.sh
```diff
@@ -446,18 +470,36 @@ arch() {
...
+  fi
   if [ ! "${DRY_RUN-}" ]; then
+    sh_f "$@"
+  fi
+}
+
+# Always runs.
+sh_f() {
+  if [ "$SSH_ARGS" ]; then
+    mkdir -p ~/.ssh/sockets
```
SuperSandro2000, line 490: The ``~/.ssh`` directory requires to have 700 for openssh to accept it. It should probably be chmoded to that.

install.sh
```diff
@@ -494,10 +513,10 @@ echoerr() {
   echoh "$@" >&2
 }
 
-# humanpath replaces all occurances of " $HOME" with " ~"
-# and all occurances of '"$HOME' with the literal '"$HOME'.
+# humanpath replaces all occurances of " $RHOME" with " ~"
+# and all occurances of '"$RHOME' with the literal '"$RHOME'.
 humanpath() {
-  sed "s# $HOME# ~#g; s#\"$HOME#\"\$HOME#g"
+  sed "s# $RHOME# ~#g; s#\"$RHOME#\"\$RHOME#g"
```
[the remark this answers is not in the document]
SuperSandro2000: ``~`` is not supported in early init scripts before the shell started. ``$HOME`` is preferred.

install.sh
nhooyr: Good catch!

nhooyr:
This got messy quickly, opening a follow up PR without `--start` and no support for hosts without internet which I think is fine for now.

Issue #1528, opened 2020-04-14
Provide VS Code flags in VSCODE_OPTIONS


kylecarbs:
`VSCODE_OPTIONS` can be an environment variable providing flags to code-server much like `NODE_OPTIONS`.

nhooyr:
When do you see this being useful over flags?

nhooyr:
Oh nvm I misunderstood the issue.

softov:
I'm trying to implement/use AHP within vscode. 

When doing $ code-server --agent-host-bridge-port 12345 --agent-host-bridge-host 10.10.10.10.
I get a "Unknown option".

More like missing options/flags. I have made some changes to enable it..

```
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -56,6 +56,14 @@
   "session-socket"?: string
   "cookie-suffix"?: string
   "link-protection-trusted-domains"?: string[]
+  // Agent host bridge. These belong to the VS Code server, which already
+  // understands them; code-server only has to accept them so `toCodeArgs`
+  // can pass them through. Without the declaration the parser rejects them
+  // with "Unknown option" even though the bundled server supports them.
+  "agent-host-bridge-port"?: string
+  "agent-host-bridge-path"?: string
+  "agent-host-bridge-host"?: string
+  "agent-host-bridge-connection-token"?: string
   // locale is used by both VS Code and code-server.
   locale?: string
 }
@@ -322,6 +330,25 @@
       "Override the reconnection grace time in seconds. Clients who disconnect for longer than this duration will need to \n" +
       "reload the window. Defaults to 10800 (3 hours).",
   },
+  "agent-host-bridge-port": {
+    type: "string",
+    description:
+      "Bridge agent-host traffic to an already-running agent host listening on this port. Does not spawn one; "  +
+      "start one with `code agent host --port <port>`.",
+  },
+  "agent-host-bridge-path": {
+    type: "string",
+    description:
+      "Bridge agent-host traffic to an already-running agent host listening on this socket path. Does not spawn one.",
+  },
+  "agent-host-bridge-host": {
+    type: "string",
+    description: "Host the externally-running agent host is reachable at. Defaults to localhost.",
+  },
+  "agent-host-bridge-connection-token": {
+    type: "string",
+    description: "Connection token required by the externally-running agent host.",
+  },
 }
 
 export const optionDescriptions = (opts: Partial<Options<Required<UserProvidedArgs>>> = options): string[] => {
```

[...]

PR #2013, opened 2020-08-27
Remove open-in flag

code-asher:
- Always launch into the existing instance if there is a socket
  - VS Code sets this variable when the socket is created so if this is set it should mean the socket exists
- Unset for `yarn watch` so you can still develop code-server inside code-server
  - Otherwise it would just send to the socket instead of launching
- Avoid the config info line when sending to an existing instance

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "lint": "./ci/dev/lint.sh",
     "test": "./ci/dev/test.sh",
     "ci": "./ci/dev/ci.sh",
-    "watch": "NODE_OPTIONS=--max_old_space_size=32384 ts-node ./ci/dev/watch.ts"
+    "watch": "VSCODE_IPC_HOOK_CLI= NODE_OPTIONS=--max_old_space_size=32384 ts-node ./ci/dev/watch.ts"
   },
   "main": "out/node/entry.js",
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,7 +26,7 @@
     "lint": "./ci/dev/lint.sh",
     "test": "./ci/dev/test.sh",
     "ci": "./ci/dev/ci.sh",
-    "watch": "NODE_OPTIONS=--max_old_space_size=32384 ts-node ./ci/dev/watch.ts"
+    "watch": "VSCODE_IPC_HOOK_CLI= NODE_OPTIONS=--max_old_space_size=32384 ts-node ./ci/dev/watch.ts"
```
nhooyr, line 29: ez

code-asher: pz

shayne:
What do you think about making another `bin` in `package.json` called `code` that hooks into `entry.ts`? We can work on expanding the functionality to support all the possible flags that the standard `code` command provides. This `code` could get its own usage/help as well.

code-asher:
I think that may be a good idea to provide an extra layer of compatibility. We want to add a `VSCODE_OPTIONS` environment variable that will accepting anything and pass it as-is down to VS Code's CLI parser so we don't have to end up adding all the flags manually to code-server (https://github.com/cdr/code-server/issues/1528). 

So with that in mind we could directly import VS Code's CLI parser when invoking `code` then run our `entry.ts` after setting `VSCODE_OPTIONS` (or we could send it via IPC). That way the flags and help would always be up to date without any extra work on our part.

nhooyr:
Lets discuss in this a new issue. Opened https://github.com/cdr/code-server/issues/2023

PR #2019, opened 2020-08-28
Link to code-server job listing

nhooyr:


README.md
```diff
@@ -60,6 +60,9 @@ you're in North America or Europe.
 
 Please get in [touch](mailto:jobs@coder.com) with your resume/github if interested.
 
+We're also hiring someone specifically to help maintain code-server.
```
kylecarbs, line 63: No need to put `also` here IMO.

nhooyr: Doesn't flow otherwise. Above we're talking generally but here we're mentioning that we specifically want someone for code-server as well.

PR #2033, opened 2020-08-31
v3.5.1

nhooyr:
- Fixes path of systemd system service in nfpm


nhooyr:
Gonna just update the binaries in place while we can.

PR #2041, opened 2020-09-02
Fix npm post install to fetch extension dependencies

### LINKED ISSUE #1961: npm module doesn't install extension dependencies
Did a fresh install of the docker image and keep running into the following error:
`Running the contributed command: 'emmet.expandAbbreviation' failed.`

My only setting is:
`{
    "emmet.triggerExpansionOnTab": true
}`

No other extensions installed/enabled. Straight from installation to opening a new html file and trying to expand 
`!` 
or 
`ul>li*5`
results in that error. 

Current version details: 

- code-server: v3.4.0
- VS Code: v1.45.1
- Commit: 69ad52907e8ea109345831d29da5425cb2a55047
- Date: 2020-05-27T23:38:22Z (2 mos ago)

nhooyr:
Closes #1961

### DIFF
--- a/ci/build/npm-postinstall.sh
+++ b/ci/build/npm-postinstall.sh
@@ -36,6 +36,13 @@ vscode_yarn() {
   yarn --production --frozen-lockfile
   cd extensions
   yarn --production --frozen-lockfile
+  for ext_dir in */; do
+    cd "$ext_dir"
+    if [ -f package.json ]; then
+      yarn --production --frozen-lockfile
+    fi
+    cd -
+  done
 }
 
 main "$@"


code-asher:
It seems that the final `package.json` that VS Code ships for each builtin extension has the `dependencies` portion removed so `yarn` ends up not installing anything.

We could prevent that deletion: https://github.com/microsoft/vscode/blob/1.48.2/build/lib/extensions.ts#L73

Or we could only add back the dependencies specifically marked as `external` in the extension's webpack configuration since the rest will have been bundled.

nhooyr:
> We could prevent that deletion: https://github.com/microsoft/vscode/blob/1.48.2/build/lib/extensions.ts#L73

That's what I'll do!

Issue #2021, opened 2020-08-28
globalState is not persisted across extension updates


Veetaha:
Our extension fails to work on code-server (https://github.com/rust-analyzer/rust-analyzer/issues/5892).
The problem is that we use [`globalState`](https://code.visualstudio.com/api/extension-capabilities/common-capabilities#data-storage). We expect the contents of `globalState` to persist across the extension version updates. Even though this is not mentioned in `vscode` API docs, but original `vscode` does persist it, but `code-server` doesn't.

For example we check for new nightly extension release and record it in `globalState` [like this](https://github.com/rust-analyzer/rust-analyzer/blob/8146700f82b75cd3cda5191108b4277a9908c744/editors/code/src/persistent_state.ts#L10-L18).
But when we [programmatically install the extension from VSIX](https://github.com/rust-analyzer/rust-analyzer/blob/8146700f82b75cd3cda5191108b4277a9908c744/editors/code/src/main.ts#L201) and reload the window, code-server clears the persistent storage.

# Meta
- Web Browser: `Google Chrome Version 84.0.4147.125 (Official Build) (64-bit)`
- Local OS: `Ubuntu 20.04.1 LTS`
- Remote OS: (idk, just running `code-server` on my local PC, so probably same as Local OS)
- Remote Architecture: `x86_64` my PC
- `code-server --version`: `3.4.1` 48f7c2724827e526eeaa6c2c151c520f48a61259

nhooyr:
cc @code-asher

code-asher:
I can replicate with:

```
await context.globalState.update("key", "value")
vscode.commands.executeCommand("workbench.action.reloadWindow");
```

I think that (in the web version of VS Code at least) the storage file isn't written to immediately but rather after a delay and reloading for some reason interrupts that delay causing it to never write (further supported by the fact that adding a 1s timeout around the reload makes it work).

Hopefully will be able to track down why the write gets canceled and get it fixed today.

Veetaha:
Should we close this?

code-asher:
Recently we've switched to closing the issues once a release with the fix goes out so let's leave it open for now!

code-asher:
I think we're now closing the issues so we can tell how far the milestone has progressed so I'll go ahead and close this out.

PR #2049, opened 2020-09-02
Patch VS Code to wait for storage write

code-asher:
Confirmed that it fixes https://github.com/rust-analyzer/rust-analyzer/issues/5892 (reported in https://github.com/cdr/code-server/issues/2021).

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -696,6 +696,49 @@ index 2185bb5228c..35463ca6520 100644
  			(err: any, socket: ISocket | undefined) => {
  				if (err || !socket) {
  					options.logService.error(`${logPrefix} socketFactory.connect() failed. Error:`);
+diff --git a/src/vs/platform/storage/browser/storageService.ts b/src/vs/platform/storage/browser/storageService.ts
+index 59b1baf912..cf9805554b 100644
+--- a/src/vs/platform/storage/browser/storageService.ts
++++ b/src/vs/platform/storage/browser/storageService.ts
+@@ -116,8 +116,8 @@ export class BrowserStorageService extends Disposable implements IStorageService
+ 		return this.getStorage(scope).getNumber(key, fallbackValue);
+ 	}
+ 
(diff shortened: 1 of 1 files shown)


ci/dev/vscode.patch
```diff
@@ -696,9 +696,52 @@ index 2185bb5228c..35463ca6520 100644
...
+diff --git a/src/vs/platform/storage/common/storage.ts b/src/vs/platform/storage/common/storage.ts
+index 1623957cb1..d366438d54 100644
+--- a/src/vs/platform/storage/common/storage.ts
++++ b/src/vs/platform/storage/common/storage.ts
+@@ -83,7 +83,7 @@ export interface IStorageService {
+ 	 * The scope argument allows to define the scope of the storage
+ 	 * operation to either the current workspace only or all workspaces.
+ 	 */
+-	store(key: string, value: string | boolean | number | undefined | null, scope: StorageScope): void;
++	store(key: string, value: string | boolean | number | undefined | null, scope: StorageScope): Promise<void> | void;
```
nhooyr, line 723: Why `| void`?

nhooyr (changes requested):
What's causing so many lines in the patch to change?

ci/dev/vscode.patch
code-asher: There are a bunch of tests that implement this interface. We could modify them all to return promises instead of void but it would increase the patch size a decent amount.

code-asher:
It looks like my version of git outputs index hashes that are 10 characters in length while the last patch generator's version used 11 characters. As far as I can tell this isn't configurable yet (it will be in the next version).

But I'm thinking we should modify the patch script to use `--full-index`, then no matter what version of git you use the hashes will be the same length.

code-asher:
I removed the extraneous hash updates. :smile:

nhooyr:
> But I'm thinking we should modify the patch script to use --full-index, then no matter what version of git you use the hashes will be the same length.

Sounds good, go ahead and do that!

nhooyr:
In a separate commit of course.

nhooyr (approved):

PR #2055, opened 2020-09-03
Don't require auth for healthz

code-asher:


nhooyr (approved):

nhooyr:
@code-asher Can you add some docs for this as well?

code-asher:
Will do.

nhooyr (approved):

PR #2057, opened 2020-09-03
docker: Fix $DOCKER_USER

### LINKED ISSUE #2056: Fix $DOCKER_USER to use cp and not mv and only convert once
See https://github.com/cdr/code-server/discussions/2051

nhooyr:
We do not try renaming $HOME anymore as there is no good way
to do it.

We also only try to convert if the user hasn't been changed.

Finally I added usage to the docker docs in install.md

Closes #2056


code-asher (commented):
Ran it and it's working for me. The only potential issue is that changes I make to the mounted directories don't show in the Docker container (and vice-versa). Would something like a symlink work, maybe?

code-asher (approved):
Works like a charm!

PR #2074, opened 2020-09-09
Minor grammar fixes

dharkness:
This PR contains two minor grammar fixes to the README.

Not in this PR is that one of the sentences at the start uses the Oxford comma while three or four after it do not. I recommend being consistent, but I understand this is more of a personal preference.


code-asher (approved):
Thanks for the corrections! Good point about consistency. I'm an Oxford comma fan but I think this is might be the only use in the repo so maybe we opt to go without it.

PR #2086, opened 2020-09-11
Integrate Coder Cloud Agent

nhooyr:
Will allow users the following flow:

1. Install code-server.
2. Run code-server.
3. Run `code-server --expose my-name`.
4. Authenticate with GitHub and then access their code-server instance over HTTPS at `https://<github-user>-<instance-name>.cdr.dev`.

Still a WIP and things above may change. We also need to open source the Go agent.

Sort of an alternative to Microsoft's Self Hosted Azure Codespaces. See https://github.com/cdr/code-server/discussions/2077


geiseri:
It would be interesting to see this on gitlab too.  Might be a better value, than competing against github.

sr229:
This might be prone to name collisions as there might be cases you'll want to run it on a seperate system should they choose not to use a container - we should add at least a randomized identifier which is how Codespaces has done it.

nhooyr:
> This might be prone to name collisions as there might be cases you'll want to run it on a seperate system should they choose not to use a container - we should add at least a randomized identifier which is how Codespaces has done it.

You get to pick the name every time. We'd rather have a short URL. If there is a conflict, we can append some random text to the name ourselves @sreya

sreya:
As @nhooyr said you get to pick the name everytime you register a new server, the names won't conflict across different users because it is a unique key composed of the server name and your github username. For now as a user, you cannot register two servers with the same name.

nhooyr:
> It would be interesting to see this on gitlab too. Might be a better value, than competing against github.

We only use GitHub for auth, we can add multiple options down the line.

src/node/coder-cloud.ts
```diff
@@ -0,0 +1,72 @@
...
+  })
+}
+
+export function coderCloudProxy(addr: string) {
+  // addr needs to be in host:port format.
+  // So we trim the protocol.
+  addr = addr.replace(/^https?:\/\//, "")
+
+  const _proxy = async () => {
+    const agent = spawn(coderCloudAgent, ["proxy", "--code-server-addr", addr], {
```
sreya, line 39: i think this is `bind` now right?

src/node/coder-cloud.ts
```diff
@@ -0,0 +1,72 @@
...
+        rej({
+          message: `coder cloud agent exited with ${code}`,
+        })
+        return
+      }
+      res()
+    })
+  })
+}
+
```
sreya, line 32: this can be removed

src/node/coder-cloud.ts
sreya, line 39: based on our convo this would become `"bind", "--code-server-addr", addr, [name]`. `name` can be empty now, we'll generate one based on the hostname

code-asher (approved):

nhooyr:
Made it clear `--coder-bind` is in beta and it's not shown by default in help output unless `CS_BETA` is non empty in the environment.

PR #2099, opened 2020-09-15
Make opening in an existing instance work outside code-server

### LINKED ISSUE #2036: Allow opening files in existing code-server instance from a terminal outside code-server
running code-server <file> attempts to start another instance of code-server instead of opening the file

Web Browser: Chrome
Local OS: Arch Linux
Remote OS: OSX
Remote Architecture: x86_64
code-server --version: 3.5.0

code-asher:
Closes #2036 

Refactored the wrapper a little bit as well so we don't have to do all the socket checking stuff in both the parent and the child.

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -529,9 +529,10 @@ export const shouldOpenInExistingInstance = async (args: Args): Promise<string |
   }
 
   // It's possible the user is trying to spawn another instance of code-server.
-  // Check if any unrelated flags are set (add one for `_` which always exists),
-  // that a file or directory was passed, and that the socket is active.
-  if (Object.keys(args).length === openInFlagCount + 1 && args._.length > 0) {
+  // Check if any unrelated flags are set (check against one because `_` always
+  // exists), that a file or directory was passed, and that the socket is
+  // active.
+  if (Object.keys(args).length === 1 && args._.length > 0) {
     const socketPath = await readSocketPath()
     if (socketPath && (await canConnect(socketPath))) {
       return socketPath
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -327,6 +327,21 @@ export const parse = (
 
   logger.debug("parsed command line", field("args", args))
 
+  return args
+}
+
+export async function setDefaults(args: Args): Promise<Args> {
+  args = { ...args }
+
+  if (!args["user-data-dir"]) {
+    await copyOldMacOSDataDir()
+    args["user-data-dir"] = paths.data
+  }
+
+  if (!args["extensions-dir"]) {
+    args["extensions-dir"] = path.join(args["user-data-dir"], "extensions")
+  }
+
   // --verbose takes priority over --log and --log takes priority over the
   // environment variable.
   if (args.verbose) {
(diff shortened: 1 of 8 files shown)


arturoiam (approved):

src/node/cli.ts
```diff
@@ -453,3 +449,51 @@ async function copyOldMacOSDataDir(): Promise<void> {
...
+    return args[cur as keyof Args] ? prev + 1 : prev
+  }, 0)
+  if (openInFlagCount > 0) {
+    return readSocketPath()
+  }
+
+  // It's possible the user is trying to spawn another instance of code-server.
+  // Check if any unrelated flags are set (add one for `_` which always exists),
+  // that a file or directory was passed, and that the socket is active.
+  if (Object.keys(args).length === openInFlagCount + 1 && args._.length > 0) {
```
nhooyr, line 491: Wouldn't `openInFlagCount` always be zero here?

src/node/entry.ts
```diff
@@ -165,83 +259,23 @@ async function entry(): Promise<void> {
...
+  }
+
+  const wrapper = new WrapperProcess(require("../../package.json").version)
+  return wrapper.start()
 }
 
-entry()
+entry().catch((error) => {
+  logger.error(error.message)
+  ipcMain.exit(error)
```
nhooyr, line 280: Why do we override `process.exit` and use this instead?

src/node/wrapper.ts
```diff
@@ -71,6 +65,19 @@ export class IpcMain {
     }
   }
 
+  /**
+   * Ensure we control when the process exits.
+   */
+  public preventExit(): void {
```
nhooyr, line 71: Ahh this makes way more sense now, we're only doing it in the child right?

nhooyr (commented):
refactor makes the code a lot clearer!

src/node/entry.ts
code-asher: The main reason is so 3rd party code can't call `process.exit` and kill code-server (plugins for example).

code-asher: Probably not a huge concern though, I think if anything happened it would be more likely to be buggy code than a malicious attempt to kill code-server.

code-asher: Also we might want to think about running 3rd party stuff in a separate process anyway.

src/node/wrapper.ts
code-asher: Yup! We don't run any 3rd party code in the parent so we don't need it there.

src/node/cli.ts
code-asher: Oh snap that's right. Nice catch!

src/node/wrapper.ts
nhooyr: Yea we were overriding it before which was causing the `process.exit`'s caused by the cloud agent exiting to not `process.exit` the main process.

code-asher: Ahhh that makes sense.

code-asher:
Rebased, also added two commits (remove `tryParse` and update the standalone test).

nhooyr (approved):

PR #2125, opened 2020-09-22
Update guide.md for iPad simple setup

AckerApple:
I found it so simple and beautiful to use a different method to get VS Code on my iPad running

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -7,6 +7,7 @@
   - [Google Cloud](#google-cloud)
 - [2. Install code-server](#2-install-code-server)
 - [3. Expose code-server](#3-expose-code-server)
+  - [iPad Simple Setup](#ipad-simple-setup)
   - [SSH forwarding](#ssh-forwarding)
   - [Let's Encrypt](#lets-encrypt)
     - [NGINX](#nginx)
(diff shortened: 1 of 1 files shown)


Mel-5569 (approved):

doc/guide.md
```diff
@@ -115,6 +116,12 @@ There are several approaches to securely operating and exposing `code-server`.
 
 **tip**: You can list the full set of `code-server` options with `code-server --help`
 
+### iPad Simple Setup
+
+Now that you have an exposed code-server instance with an IP and port number, you are one simple step away.
```
nhooyr, line 121: What's the point of this app vs just visiting code-server with safari?

AckerApple: So you know how the documentation says iPads don’t work easy and you need a cloud or self signed ssl injection for iPads to support copy/paste and other commands....

Ya, you don’t need any of that with Serveditor.

With this app you don’t have to cloud setup or config a secure certificate setup for a full featured Visual Studio.

This has everything to do with iOS and its https requirements of a true ssl

nhooyr: Wouldn't it just be over `http` then at that point? Or does it access code-server over `ssh`?

I've got self signed certificates working on iPad now! See #2255

Will be in next release.

AckerApple: Whatever protocol it does use, it just works including copy paste and additional features that code server does NOT support unless on https

doc/guide.md
```diff
@@ -115,6 +116,12 @@ There are several approaches to securely operating and exposing `code-server`.
 
 **tip**: You can list the full set of `code-server` options with `code-server --help`
 
+### iPad Simple Setup
```
nhooyr, line 119: Let's add a note re this app in `doc/ipad.md` and then I think we can merge!

AckerApple: Ok, I see your callout to edit this file:

https://github.com/cdr/code-server/blob/v3.7.1/doc/ipad.md

I will do just that

nhooyr:
Thank you @AckerApple!

AckerApple:
Thank you so very much. I’ve been so meaning to get back to this. I program every day on code server on my iPad. I can help anyone with more questions

PR #2147, opened 2020-09-30
Add variables to better customize plugin directories

### LINKED ISSUE #2122: Allow multiple plugins to be specified
`PLUGIN_DIR` only accepts a single directory, so if we were to have multiple plugins, we could only choose to use 1 at a time. We could either just split on comma, have the variable accept a JSON array, or change it to a flag that can be specified multiple times (e.g. `--plugin /path/to/plugin1 --plugin /path/to/plugin2`)

code-asher:
- Still loads all plugins from the `plugins` directory as before
  - But now `PLUGIN_DIR` will set that directory instead of being used to specify an individual plugin
  - This is the directory I envision being used for distributing code-server with plugins or where plugins would be placed for running instances of code-server in production by users
- `PLUGIN_DIRS` can be used to set multiple individual plugins if you don't want to move them into a single directory
  - Really only intended for development

Lemme know if we wanna change this up, this is just based off the VS Code plugin structure (except they use command line flags and not environment variables).

Closes #2122.

### DIFF
--- a/src/node/plugin.ts
+++ b/src/node/plugin.ts
@@ -66,25 +66,23 @@ const _loadPlugins = async (pluginDir: string, httpServer: HttpServer, args: Arg
 }
 
 /**
- * Load all plugins from the `plugins` directory and the directory specified by
- * `PLUGIN_DIR`.
-
- * Also load any individual plugins found in `PLUGIN_DIRS` (colon-separated).
- * This allows you to test and develop plugins without having to move or symlink
- * them into one directory.
+ * Load all plugins from the `plugins` directory, directories specified by
+ * `CS_PLUGIN_PATH` (colon-separated), and individual plugins specified by
+ * `CS_PLUGIN` (also colon-separated).
  */
 export const loadPlugins = async (httpServer: HttpServer, args: Args): Promise<void> => {
+  const pluginPath = process.env.CS_PLUGIN_PATH || `${path.join(paths.data, "plugins")}:/etc/code-server/plugins`
+  const plugin = process.env.CS_PLUGIN || ""
   await Promise.all([
     // Built-in plugins.
     _loadPlugins(path.resolve(__dirname, "../../plugins"), httpServer, args),
     // User-added plugins.
-    _loadPlugins(
-      path.resolve(process.env.PLUGIN_DIR || path.join(paths.data, "code-server-extensions")),
-      httpServer,
-      args,
-    ),
-    // For development so you don't have to use symlinks.
-    process.env.PLUGIN_DIRS &&
-      (await Promise.all(process.env.PLUGIN_DIRS.split(":").map((dir) => loadPlugin(dir, httpServer, args)))),
+    ...pluginPath.split(":").map((dir) => _loadPlugins(path.resolve(dir), httpServer, args)),
+    // Individual plugins so you don't have to symlink or move them into a
+    // directory specifically for plugins. This lets you load plugins that are
+    // on the same level as other directories that are not plugins (if you tried
+    // to use CS_PLUGIN_PATH code-server would try to load those other
+    // directories as plugins). Intended for development.
+    ...plugin.split(":").map((dir) => loadPlugin(path.resolve(dir), httpServer, args)),
   ])
 }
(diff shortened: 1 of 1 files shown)


code-asher:
Default plugin dir should actually use `XDG` I think. Thoughts? Having it inside code-server is weird, except maybe for built-in plugins. So maybe we should load both.

code-asher:
I went ahead and did that.

code-asher:
Should note this isn't backwards compatible with the previous method so maybe it would be better keep maintain the old `PLUGIN_DIR` behavior and use different variables for the new stuff (or use command line flags instead).

nhooyr (commented):
Can we use `CS_PLUGIN_PATH` instead?

Where plugin path consists of colon separated directories in which plugins are loaded from.

And defaults to: `$HOME/.config/code-server/plugins:/etc/code-server/plugins`.

code-asher:
`CS_PLUGIN_PATH` is a much better name.

src/node/plugin.ts
```diff
@@ -23,38 +27,58 @@ require("module")._load = function (request: string, parent: object, isMain: boo
...
+ */
 const loadPlugin = async (pluginPath: string, httpServer: HttpServer, args: Args): Promise<void> => {
   try {
     const plugin: Plugin = require(pluginPath)
     plugin.activate(httpServer, args)
-    logger.debug("Loaded plugin", field("name", path.basename(pluginPath)))
+
+    logger.debug(
+      "Loaded plugin",
+      field("name", path.basename(pluginPath)),
```
nhooyr, line 40: Maybe this should be from `package.json` as well?

nhooyr: And then a separate field for the path.

src/node/plugin.ts
```diff
@@ -23,38 +27,58 @@ require("module")._load = function (request: string, parent: object, isMain: boo
...
+ */
+const _loadPlugins = async (pluginDir: string, httpServer: HttpServer, args: Args): Promise<void> => {
   try {
-    await _loadPlugins(httpServer, args)
+    const files = await util.promisify(fs.readdir)(pluginDir, {
+      withFileTypes: true,
+    })
+    await Promise.all(files.map((file) => loadPlugin(path.join(pluginDir, file.name), httpServer, args)))
   } catch (error) {
     if (error.code !== "ENOENT") {
```
nhooyr, line 58: Wouldn't loadPlugin never throw?

src/node/plugin.ts
```diff
@@ -23,38 +27,58 @@ require("module")._load = function (request: string, parent: object, isMain: boo
...
-  if (process.env.PLUGIN_DIR) {
-    await loadPlugin(process.env.PLUGIN_DIR, httpServer, args)
-  }
+/**
+ * Load all plugins from the `plugins` directory, directories specified by
+ * `CS_PLUGIN_PATH` (colon-separated), and individual plugins specified by
+ * `CS_PLUGIN` (also colon-separated).
+ */
+export const loadPlugins = async (httpServer: HttpServer, args: Args): Promise<void> => {
+  const pluginPath = process.env.CS_PLUGIN_PATH || `${path.join(paths.data, "plugins")}:/etc/code-server/plugins`
```
nhooyr, line 70: Since we're using the data dir, we probably want to use `/usr/share/code-server/plugins` instead of `/etc/code-server/plugins`.

See https://wiki.archlinux.org/index.php/XDG_Base_Directory

src/node/plugin.ts
code-asher: Ooo yeah much better.

src/node/plugin.ts
code-asher: Yup, it's the `readdir` part that might throw in this case.

src/node/plugin.ts
code-asher: Ahhh good point. I see now your original comment used the config dir, do you think that would be more appropriate than the data dir? My instinct was to use data but I'm not sure if that's how most editors do it.

nhooyr: data-dir makes most sense as plugins won't be user editable.

src/node/plugin.ts
nhooyr: Gotchu!

nhooyr (approved):

PR #2167, opened 2020-10-07
Allow building release without deleting existing node_modules

nhooyr:
```
KEEP_MODULES=1 ./ci/steps/release.sh && node ./release --auth=none
```

Can be ran to build & test a release quickly.


code-asher (approved):
:fire:

Issue #1729, opened 2020-05-27
Install script should be able to install remotely


nhooyr:
Waiting on #1701 to be merged, then fix any release blocking bugs and release the next release and then I'll do this to deprecate sshcode.

Also see https://github.com/cdr/sshcode/issues/185

nhooyr:
- [ ] `--start` functionality which starts code-server and forwards port if necessary
- [ ] `--coder-bind` functionality
- [ ] support for hosts without internet
- [ ] support for non ssh shells
- [ ] support for rerunning install script and only installing if update required

nhooyr:
See #2183 and https://github.com/cdr/code-server/issues/2219

PR #2183, opened 2020-10-09
install.sh: Allow installing directly onto a remote host

nhooyr:
Updates #1729

To fully close that issue see the various TODOs.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,21 +17,28 @@ usage() {
 Installs code-server for Linux, macOS and FreeBSD.
 It tries to use the system package manager if possible.
 After successful installation it explains how to start using code-server.
+
+Pass in user@host to install code-server on user@host over ssh.
+The remote host must have internet access.
 ${not_curl_usage-}
 Usage:
 
-  $arg0 [--dry-run] [--version X.X.X] [--method detect] [--prefix ~/.local]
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -100,9 +107,17 @@ main() {
     METHOD \
     STANDALONE_INSTALL_PREFIX \
     VERSION \
-    OPTIONAL
+    OPTIONAL \
+    ALL_FLAGS
 
+  ALL_FLAGS=""
```
code-asher, line 113: We'll need to set `SSH_ARGS` as well I think to avoid unbound errors.

install.sh
```diff
@@ -132,16 +147,33 @@ main() {
...
+      break
+      ;;
     esac
 
     shift
   done
 
+  if [ "$SSH_ARGS" ]; then
+    echoh "Installing remotely with ssh $SSH_ARGS"
+    curl -fsSL https://code-server.dev/install.sh | prefix "$SSH_ARGS" ssh "$SSH_ARGS" sh -s -- "$ALL_FLAGS"
```
code-asher, line 173: Ahh no need to mess with `sh_c` and all that with this method. Nice!!

code-asher (changes requested):
I gave it a spin and it works like a dream. :fire:

code-asher (approved):

PR #2185, opened 2020-10-09
Extension divergence help information in the UI

### LINKED ISSUE #2132: Add help information to the extension search view
Show a yellow or red call to action in the extension search view informing the user of all alternative extension installation methods.

cmoog:
Adds the following UI to the extension search view:

Screen Shot 2020-10-13 at 12 06 47 PM

Closes #2132

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -3165,6 +3165,51 @@ index 18ea0bfedb4492327429a38237b05915b29f6dd0..d59a17c17f4fffa23d786ce36b4ff624
  				this._filenameKey.set(value ? basename(value) : null);
  				this._langIdKey.set(value ? this._modeService.getModeIdByFilepathOrFirstLine(value) : null);
  				this._extensionKey.set(value ? extname(value) : null);
+diff --git a/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts b/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
+index 33ab9f5a25e4124706ef38d6bf39f77e0ccafb32..8e479f579ee70ba5781b68f6b460ce223ec32e07 100644
+--- a/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
++++ b/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
+@@ -408,6 +408,40 @@ export class ExtensionsViewPaneContainer extends ViewPaneContainer implements IE
+ 		addClass(parent, 'extensions-viewlet');
+ 		this.root = parent;
+ 
++		// NOTE@coder this UI element helps users understand the extension marketplace divergence
++		const extensionHelperLocalStorageKey = 'coder.extension-help-message';
++		if (localStorage.getItem(extensionHelperLocalStorageKey) === null) {
++			const helperHeader = append(this.root, $('.header'));
++			helperHeader.id = 'codeServerMarketplaceHelper';
++			helperHeader.style.height = 'auto';
++			helperHeader.style.border = '3px solid red';
++			helperHeader.style.borderRadius = '10px';
++			helperHeader.style.fontWeight = '600';
++			helperHeader.style.padding = 'padding: 5px 16px';
++			helperHeader.style.position = 'relative';
++			helperHeader.innerHTML = `
++			<div>
++			<p>
++			NOTE: code-server has its own marketplace of open-source extensions due to
++			<a href="https://github.com/cdr/code-server/blob/master/doc/FAQ.md#differences-compared-to-vs-code" target="_blank">legal distribution restrictions</a>.
++			VSIX assets from the <a href="https://marketplace.visualstudio.com/vscode" target="_blank">official marketplace</a> can be
++			downloaded and installed by uploading through the dropdown above.
++			</p>
++			</div>
++					`;
++			const dismiss = append(helperHeader, $('span'));
++			dismiss.innerHTML = 'Dismiss';
++			dismiss.style.position = 'absolute';
++			dismiss.style.bottom = '5px';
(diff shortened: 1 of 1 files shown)


ci/dev/vscode.patch
```diff
@@ -31,32 +31,6 @@ index f2ea1bd37010b1eb8a43ce9beaae4a88810f6e2d..3f660f9981921ec465d2b8809a1a5ea5
  	const yarnrc = fs.readFileSync(path.join(REPO_ROOT, 'remote', '.yarnrc'), 'utf8');
  	const target = /^target "(.*)"$/m.exec(yarnrc)[1];
  	return target;
-diff --git a/build/lib/extensions.js b/build/lib/extensions.js
```
nhooyr, line 34: Something went wrong with the diff here. I think you need to rebase your branch and clean.

nhooyr (changes requested):
lgtm otherwise

kylecarbs (changes requested):
```
NOTE: code-server has it's own marketplace of open-source extensions due to 
<a href="https://github.com/cdr/code-server/blob/master/doc/FAQ.md#differences-compared-to-vs-code" target="_blank">legal distribution restrictions</a>.
VSIX assets from the official marketplace can be 
<a onClick="MAKE THIS EXECUTE THE UPLOAD VSIX COMMAND">uploaded<a/>.
```
Thoughts on something like this?

cmoog:
@kylecarbs I do think a link to the official marketplace is worth _a lot_.

kylecarbs:
Oh yeah I'm down to include that.

Just tryna reduce the size of the copy as much as possible.

code-asher:
Possibly dumb question, is downloading extensions to use in
non-Microsoft products a TOS violation? That's the sense I had but I'm
no lawyer.

ci/dev/vscode.patch
```diff
@@ -3140,6 +3114,50 @@ index 18ea0bfedb4492327429a38237b05915b29f6dd0..d59a17c17f4fffa23d786ce36b4ff624
...
++		helperHeader.style.height = 'auto';
++		helperHeader.style.border = '3px solid red';
++		helperHeader.style.borderRadius = '10px';
++		helperHeader.style.fontWeight = '600';
++		helperHeader.style.padding = 'padding: 5px 16px';
++		helperHeader.style.position = 'relative';
++		helperHeader.innerHTML = `
++<div>
++<p>
++NOTE: the code-server extension marketplace is different from the Microsoft VSCode Extension Marketplace. code-server offers it's own
```
code-asher, line 3136:
```suggestion
+NOTE: the code-server extension marketplace is different from the Microsoft VSCode Extension Marketplace. code-server offers its own
```

code-asher (commented):
Looks good, +1 from me on reducing the copy size as much as possible. Should we store the dismissal state for subsequent page loads or would it be more beneficial have it show more often?

sr229:
My nitpick here is use the same caution dialogue indicator used in Git when you're exceeding the 50-character limit so the user will see it immediately, it will add a sense of visual priority of what to read first.

code-asher:
Hmm, that's an interesting idea, it looks like it pops up below the textbox as you type. I lean toward keeping the current method because that popup floats and would probably cover an extension or two in the search results which could be frustrating. We could only show it if there weren't any results, but I think having results but not the right ones might be a common case (just a guess though) and those cases would never get the help text.

ci/dev/vscode.patch
```diff
@@ -3140,6 +3140,51 @@ index 18ea0bfedb4492327429a38237b05915b29f6dd0..d59a17c17f4fffa23d786ce36b4ff624
...
  				this._extensionKey.set(value ? extname(value) : null);
+diff --git a/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts b/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
+index 33ab9f5a25e4124706ef38d6bf39f77e0ccafb32..8e479f579ee70ba5781b68f6b460ce223ec32e07 100644
+--- a/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
++++ b/src/vs/workbench/contrib/extensions/browser/extensionsViewlet.ts
+@@ -408,6 +408,40 @@ export class ExtensionsViewPaneContainer extends ViewPaneContainer implements IE
+ 		addClass(parent, 'extensions-viewlet');
+ 		this.root = parent;
+ 
++		// NODE@coder this UI element helps users understand the extension marketplace divergence
```
code-asher, line 3151:
```suggestion
+		// NOTE@coder this UI element helps users understand the extension marketplace divergence
```

ci/dev/vscode.patch
```diff
@@ -3140,6 +3140,51 @@ index 18ea0bfedb4492327429a38237b05915b29f6dd0..d59a17c17f4fffa23d786ce36b4ff624
...
++		const extensinHelperLocalStorageKey = 'coder.extension-help-message';
```
code-asher, line 3152:
```suggestion
+		const extensionHelperLocalStorageKey = 'coder.extension-help-message';
```

code-asher (approved):
Looks good to me. BTW if we still want to add a link for jumping directly to the upload we can do it with something like:
```
uploadButton.addEventListener("click", () => {
  this.instantiationService.createInstance(InstallVSIXAction, InstallVSIXAction.ID, InstallVSIXAction.LABEL).run()
});
```

cmoog:
Closing this PR as it's become clear that linking to or referencing the existence of the official marketplace is a no-go. Suggesting users to upload via the dropdown has the same problem. Keeping the helper text without including instructions for resolution will cause more confusion than it's worth. Eventually, if we ever move to open-source our extension marketplace, an overlay of this nature will be appropriate. 

😢 I'll open a new PR with a revised issue template for extension requests.

PR #2186, opened 2020-10-09
Allow login while proxing a sub-domain

### LINKED ISSUE #2128: Catch unauthenticated POST requests to the proxy and return an error instead of redirecting to vs code
I'm trying to proxy POST calls for a graphql api running on my dev instance, but when I do POST requests, I always recieve:
`{"error":"Unsupported method POST"}`
Visiting pages / GET requests work fine.

I assume this to be a restriction in the proxy, but didn't take a deeper look yet.

code-asher:
Fixes #2128.

### DIFF
--- a/src/node/http.ts
+++ b/src/node/http.ts
@@ -927,7 +927,29 @@ export class HttpServer {
     }
 
     // Must be authenticated to use the proxy.
-    route.provider.ensureAuthenticated(request)
+    if (!route.provider.authenticated(request)) {
+      // Attempt to determine if it's the user browsing the root and if so fall
+      // through to allow the login flow.
+      if (request.headers["content-type"] !== "application/json") {
+        switch (route.providerBase) {
+          case "/":
+          case "/static":
+            if (request.method === "GET") {
+              return undefined
+            }
+            break
+          case "/login":
+            if (request.method === "GET" || request.method === "POST") {
+              return undefined
+            }
+            break
+        }
+      }
+
+      // Assume anything else is some kind of request from the proxied
+      // application and return an unauthorized message.
+      throw new HttpError("Unauthorized", HttpCode.Unauthorized)
+    }
 
     return {
       proxy: {


code-asher:
tbh this feels a little jank to me but I'm not sure how else this could be implemented.

src/node/http.ts
```diff
@@ -927,7 +927,29 @@ export class HttpServer {
     }
 
     // Must be authenticated to use the proxy.
-    route.provider.ensureAuthenticated(request)
+    if (!route.provider.authenticated(request)) {
+      // Attempt to determine if it's the user browsing the root and if so fall
+      // through to allow the login flow.
+      if (request.headers["content-type"] !== "application/json") {
```
nhooyr, line 933: Let's check explicitly for `text/html` here.

src/node/http.ts
```diff
@@ -927,7 +927,29 @@ export class HttpServer {
     }
 
     // Must be authenticated to use the proxy.
-    route.provider.ensureAuthenticated(request)
+    if (!route.provider.authenticated(request)) {
+      // Attempt to determine if it's the user browsing the root and if so fall
```
nhooyr, line 931:
```suggestion
      // Attempt to determine if it's the user browsing the route and if so fall
```

src/node/http.ts
```diff
@@ -927,7 +927,29 @@ export class HttpServer {
...
+      // through to allow the login flow.
+      if (request.headers["content-type"] !== "application/json") {
+        switch (route.providerBase) {
```
nhooyr, line 934: I'm confused by this, don't we pretty always want to fall through to the login flow if `text/html` is asked for but the user isn't authenticated?

nhooyr (changes requested):

src/node/http.ts
code-asher: I wanted to do that but Firefox and Chromium don't seem to set this at all. When I logged it the value was `undefined`.

src/node/http.ts
code-asher: I mean root, for example `8080.mydomain.com` and not `8080.mydomain.com/some/page`. Maybe we should always redirect though even if it's not the root.

src/node/http.ts
code-asher: Yup, if there was a way to reliably detect the request was asking for HTML (and ideally also that it wasn't an xhr request) that would 100% be the way to do it. All this jank is to work around not being able to do that (at least as far as I know, I hope I'm wrong though).

src/node/http.ts
code-asher: Well I guess we can't do that safely, it would end up causing the problem in #2128 again on account of the whole not being able to detect HTML nonsense.

code-asher:
Pretty much all of this conflicts with the Express rewrite so I'm going to bundle it into that PR.

PR #2204, opened 2020-10-14
Update VS Code to 1.50.0

code-asher:
Planning on making a 3.6.1 release with this.

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -929,11 +929,13 @@ index 0000000000000000000000000000000000000000..3c0703b7174ad792a4b42841e96ee937
 +};
 diff --git a/src/vs/server/browser/extHostNodeProxy.ts b/src/vs/server/browser/extHostNodeProxy.ts
 new file mode 100644
-index 0000000000000000000000000000000000000000..ed7c078077b0c375758529959b280e091436113a
+index 0000000000000000000000000000000000000000..6c6b87a05610417d73635c5a151845000f216d28
 --- /dev/null
 +++ b/src/vs/server/browser/extHostNodeProxy.ts
-@@ -0,0 +1,46 @@
+@@ -0,0 +1,52 @@
++import { VSBuffer } from 'vs/base/common/buffer';
 +import { Emitter } from 'vs/base/common/event';
++import { UriComponents } from 'vs/base/common/uri';
 +import { createDecorator } from 'vs/platform/instantiation/common/instantiation';
 +import { ExtHostNodeProxyShape, MainContext, MainThreadNodeProxyShape } from 'vs/workbench/api/common/extHost.protocol';
 +import { IExtHostRpcService } from 'vs/workbench/api/common/extHostRpcService';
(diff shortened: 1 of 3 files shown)


ci/dev/vscode.patch
```diff
@@ -57,23 +44,6 @@ index 7e529f17cb84d28d84de4ff64fa9fb8fc48135a9..462d699dc485369c74a4d9fdfefa48ba
...
-@@ -5,11 +5,8 @@
-  *--------------------------------------------------------------------------------------------*/
- Object.defineProperty(exports, "__esModule", { value: true });
- const path = require("path");
--const fs = require("fs");
- const root = path.dirname(path.dirname(__dirname));
--const yarnrcPath = path.join(root, 'remote', '.yarnrc');
--const yarnrc = fs.readFileSync(yarnrcPath, 'utf8');
--const version = /^target\s+"([^"]+)"$/m.exec(yarnrc)[1];
-+const version = process.versions.node;
```
nhooyr, line 73: Was this removal intended?

ci/dev/vscode.patch
```diff
@@ -659,10 +626,10 @@ index 575b2aafc3802cd6f5f943930e30de9f2c2690de..873181f967856759e3dc001e5bbe06e2
 +	}
  }
 diff --git a/src/vs/platform/product/common/product.ts b/src/vs/platform/product/common/product.ts
-index bb33203d1727b1c076efac9113afc3b2580cdbd9..c53cea338cdaa0f0ac15542c129e1572b3f13b80 100644
+index 798faa74ae825de7449b74609ed649912531ec0f..487abd285a42407f669ce5f5396423644f822a83 100644
 --- a/src/vs/platform/product/common/product.ts
 +++ b/src/vs/platform/product/common/product.ts
-@@ -30,6 +30,12 @@ if (isWeb) {
+@@ -37,6 +37,12 @@ if (isWeb || typeof require === 'undefined' || typeof require.__$__nodeRequire !
```
nhooyr, line 632: Why was this change required?

ci/dev/vscode.patch
```diff
@@ -1049,6 +1024,21 @@ index 0000000000000000000000000000000000000000..0d2e93edae2baf34d27b7b52be0bf496
 +		}
 +	}
 +
++	async $fetchExtension(extensionUri: UriComponents): Promise<VSBuffer> {
```
nhooyr, line 1027: Why do we have to implement this?

ci/dev/vscode.patch
```diff
@@ -20,32 +20,19 @@ index 3c6eccfb102f2084d16395d70d65f05a91b6d47b..00000000000000000000000000000000
...
-diff --git a/build/lib/extensions.js b/build/lib/extensions.js
-index 9cc40c4e1befd38886dc5880581d6f462a38dd3a..34e1fc89a8ac1c273a5cb41f19a088a8ec759d24 100644
---- a/build/lib/extensions.js
-+++ b/build/lib/extensions.js
-@@ -66,7 +66,7 @@ function fromLocal(extensionPath, forWeb) {
-     if (isWebPacked) {
-         input = updateExtensionPackageJSON(input, (data) => {
-             delete data.scripts;
--            delete data.dependencies;
-+            // https://github.com/cdr/code-server/pull/2041#issuecomment-685910322
```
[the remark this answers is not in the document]
code-asher: Yup, for all of these .js build files, they're no longer committed.

https://github.com/cdr/code-server/pull/2204/commits/e3699cf258d3c25785239aacd8c51432b608e97f

ci/dev/vscode.patch
code-asher: This is just the context of the patch, so VS Code made this change. This patch is seriously too hard to diff.

ci/dev/vscode.patch
code-asher: I made a comment on https://github.com/cdr/code-server/pull/2204/commits/f20f7ac166b40b7420a6114968a3506362280b54 but the gist is that the worker is now in a separate origin so it can't fetch the extension (cross origin domain issues) and even if you resolve that by setting the appropriate headers it still doesn't work because it doesn't send the cookie. So instead I have it asking the main thread to fetch the extension. Since this mechanism already exists for the extension to communicate with the main thread (for example to make file read calls) I figured it was a decent solution.

code-asher: I should comment this in the code and not just in the commit.

ci/dev/vscode.patch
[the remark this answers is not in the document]
nhooyr: We have to move the patches into the `.ts` files then right?

nhooyr (changes requested):
Looks good to me now other than the patch removals of the build files. We need to apply the same patch to the `.ts` files now.

ci/dev/vscode.patch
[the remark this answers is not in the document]
code-asher: Yup, they're already there. They don't show up in the diff though since we already had 'em.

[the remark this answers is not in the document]
nhooyr: Ohhh, nice!

nhooyr (approved):

PR #2220, opened 2020-10-21
install.sh: Allow customizing remote shell for remote installation

### LINKED ISSUE #1729: Install script should be able to install remotely

nhooyr:
Closes #1729

Last TODO left on that issue. I removed making remote installation on a non internet connected instance work as I don't think that's a very common use case and if it is, manual installation isn't hard anyway. It'll require lots of churn and make the install script quite abstract and hard to understand.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -24,7 +24,7 @@ ${not_curl_usage-}
 Usage:
 
   $arg0 [--dry-run] [--version X.X.X] [--method detect] \
-        [--prefix ~/.local] [user@host]
+        [--prefix ~/.local] [--rsh ssh] [user@host]
 
   --dry-run
       Echo the commands for the install process without running them.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -109,7 +112,8 @@ main() {
     VERSION \
     OPTIONAL \
     ALL_FLAGS \
-    SSH_ARGS
+    RSH_ARGS \
+    RSH
```
code-asher, line 116: Looks like we need to set the default for `RSH` to `ssh`.

code-asher (changes requested):

install.sh
nhooyr: Embarrassing. Fixed!

code-asher (approved):

Issue #2213, opened 2020-10-19
whoami errors out when UID doesn't map to a user


nhooyr:
See https://github.com/cdr/code-server/discussions/2199#discussioncomment-99174

PR #2223, opened 2020-10-22
release-image: Remap UID within the image before handling $DOCKER_USER

nhooyr:
If do not update the UID within the passwd database to match whatever
uid the container is being ran as, then sudo will not work when renaming
the user to match $DOCKER_USER as it will complain about the current
user being non-existent.

Updates #2213


code-asher (approved):
Awesome, I ran with a custom UID and username with no problems.

PR #2238, opened 2020-10-26
Convert code-server http code to use Express

### LINKED ISSUE #1899: Express-ify
Replace custom HTTP handlers with Express.

### LINKED ISSUE #2128: Catch unauthenticated POST requests to the proxy and return an error instead of redirecting to vs code

code-asher:
Story details: https://app.clubhouse.io/coder/story/1385

Couple items still remaining but existing code is ready to review:
- [x] I keep getting log directories created in the code-server directory over and over with new timestamps.
- [x] Need a way to register disposables for when the application is killed.

Closes #2128
Closes #1899


src/node/routes/vscode.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  const [content, options] = await Promise.all([
+    await fs.readFile(path.join(rootPath, "src/browser/pages/vscode.html"), "utf8"),
+    vscode
+      .initialize(
+        {
+          args: req.args,
+          remoteAuthority: req.headers.host || "",
+        },
+        req.query,
```
wbobeirne, line 29: Getting this error here, any idea why that might be?
```
[tsc] src/node/routes/vscode.ts:29:9 - error TS2345: Argument of type 'ParsedQs' is not assignable to parameter of type 'Query'.        │
[tsc]   Index signatures are incompatible.                                                                                              │
[tsc]     Type 'string | string[] | ParsedQs | ParsedQs[] | undefined' is not assignable to type 'string | string[] | undefined'.       │
[tsc]       Type 'ParsedQs' is not assignable to type 'string | string[] | undefined'.                                                  │
[tsc]         Type 'ParsedQs' is not assignable to type 'string'.                                                                       │
[tsc] 29         req.query,                                                                                                             │
```

src/node/routes/vscode.ts
```diff
@@ -0,0 +1,99 @@
...
+    .digest("base64")
+  socket.write(
+    [
+      "HTTP/1.1 101 Switching Protocols",
+      "Upgrade: websocket",
+      "Connection: Upgrade",
+      `Sec-WebSocket-Accept: ${reply}`,
+    ].join("\r\n") + "\r\n\r\n",
+  )
+  await vscode.sendWebsocket(socket, req.query)
```
wbobeirne, line 71: Similar to above
```
[tsc] src/node/routes/vscode.ts:71:38 - error TS2345: Argument of type 'ParsedQs' is not assignable to parameter of type 'Query'.       │
[tsc] 71   await vscode.sendWebsocket(socket, req.query)
```

src/node/routes/index.ts
```diff
@@ -0,0 +1,123 @@
...
+      res.status(err.status || 500).send(
+        replaceTemplates(req, content)
+          .replace(/{{ERROR_TITLE}}/g, err.status || "Error")
+          .replace(/{{ERROR_HEADER}}/g, err.status || "Error")
+          .replace(/{{ERROR_BODY}}/g, err.message),
+      )
+    } catch (error) {
+      next(error)
+    }
+  })
```
wbobeirne, line 122: Looks like app.use is for some reason ambiguous in this case, one way to fix is explicitly define the type of the handler
```suggestion
  // Handle errors.
  const errorHandler: ErrorRequestHandler = async (err, req, res, next) => {
    const resourcePath = path.resolve(rootPath, "src/browser/pages/error.html")
    res.set("Content-Type", getMediaMime(resourcePath))
    try {
      const content = await fs.readFile(resourcePath, "utf8")
      if (err.code === "ENOENT" || err.code === "EISDIR") {
        err.status = HttpCode.NotFound
      }
      res.status(err.status || 500).send(
        replaceTemplates(req, content)
          .replace(/{{ERROR_TITLE}}/g, err.status || "Error")
          .replace(/{{ERROR_HEADER}}/g, err.status || "Error")
          .replace(/{{ERROR_BODY}}/g, err.message),
      )
    } catch (error) {
      next(error)
    }
  }
  app.use(errorHandler)
```
requires you to also import `ErrorRequestHandler` from `express`.

src/node/cli.ts
```diff
@@ -466,7 +543,7 @@ function bindAddrFromArgs(addr: Addr, args: Args): Addr {
   return addr
 }
 
-export function bindAddrFromAllSources(cliArgs: Args, configArgs: Args): [string, number] {
+function bindAddrFromAllSources(cliArgs: Args, configArgs: Args): [string, number] {
   let addr: Addr = {
```
TeffenEllis, line 547: I might've misunderstood this function but there might be some clean up here
```ts
function bindAddrFromAllSources(...argsConfig: Args[]): Addr {
  let addr: Addr = {
    host: "localhost",
    port: 8080,
  }

  for (const args of argsConfig) {
    addr = bindAddrFromArgs(addr, args)
  }
  
  return addr
}
```

src/node/entry.ts
```diff
@@ -215,44 +144,43 @@ const main = async (args: Args, configArgs: Args): Promise<void> => {
...
-      await coderCloudBind(serverAddress!, args.link.value)
+      await coderCloudBind(serverAddress, args.link.value)
+      logger.info("  - Connected to cloud agent")
     } catch (err) {
       logger.error(err.message)
       ipcMain.exit(1)
     }
   }
+
+  if (serverAddress && !args.socket && args.open) {
```
TeffenEllis, line 162: Is the check for `serverAddress` redundant if `ensureAddress` throws?

src/node/entry.ts
```diff
@@ -215,44 +144,43 @@ const main = async (args: Args, configArgs: Args): Promise<void> => {
...
+    // The web socket doesn't seem to work if browsing with 0.0.0.0.
+    const openAddress = serverAddress.replace(/:\/\/0.0.0.0/, "://localhost")
```
TeffenEllis, line 164: Nit: regex is ambiguous with unescaped `.` 
```suggestion
    const openAddress = serverAddress.replace("://0.0.0.0", "://localhost")
```

src/node/entry.ts
```diff
@@ -215,44 +144,43 @@ const main = async (args: Args, configArgs: Args): Promise<void> => {
...
+    await open(openAddress).catch((error: Error) => {
+      logger.error("Failed to open", field("address", openAddress), field("error", error))
+    })
+    logger.info(`Opened ${openAddress}`)
```
TeffenEllis, line 168: The `catch` method still allows the next logger line to print

```suggestion
    try {
      await open(openAddress)
    } catch (error) {
      logger.error("Failed to open", field("address", openAddress), field("error", error))
      ipcMain.exit(1)
    }

    logger.info(`Opened ${openAddress}`)
```

src/node/routes/proxy.ts
```diff
@@ -0,0 +1,43 @@
...
+  // Absolute redirects need to be based on the subpath when rewriting.
+  ;(req as any).base = `${req.baseUrl}/${req.params.port}`
+
+  proxy.web(req, res, {
+    ignorePath: true,
+    target: getProxyTarget(req, true),
+  })
+})
+
+router.ws("/(:port)(/*)?", (socket, head, req) => {
```
wbobeirne, line 38: I could be totally wrong here, but I thought that Express ran handlers in order. So shouldn't `all` be hitting first here? Or does `ws` somehow pick up things that `all` is missing?

src/node/routes/update.ts
```diff
@@ -0,0 +1,34 @@
...
+// This route will force a check.
+router.get("/check", async (_, res) => {
+  const update = await provider.getUpdate(true)
+  res.json({
+    checked: update.checked,
+    latest: update.version,
+    current: version,
+    isLatest: provider.isLatestVersion(update),
+  })
+})
```
wbobeirne, line 34: Great example of a file where the routes are super easy to reason about, and all the business logic is contained elsewhere. Definitely makes me feel really confident about this approach!

src/node/heart.ts
```diff
@@ -0,0 +1,46 @@
...
+      this.lastHeartbeat = Date.now()
+      if (typeof this.heartbeatTimer !== "undefined") {
+        clearTimeout(this.heartbeatTimer)
+      }
+      this.heartbeatTimer = setTimeout(() => {
+        this.isActive()
+          .then((active) => {
+            if (active) {
+              this.beat()
+            }
```
TeffenEllis, line 38: Nit: Use short-circuit:

```ts
if (this.alive()) return
```

src/node/http.ts
```diff
@@ -1,938 +1,253 @@
...
-      },
-    }
+/**
+ * Patch Express routers to handle web sockets and async routes (since we have
+ * to patch `get` anyway).
+ *
+ * Not using express-ws since the ws-wrapped sockets don't work with the proxy
+ * and wildcards don't work correctly.
+ */
+function patchRouter(): void {
```
wbobeirne, line 190: I am sufficiently spooked by this whole thing, but I'm not sure I have the full context here. This all _seems_ like stuff that should be handleable through middlewares, and not actually mutating the underlying `Router` implementation. If this approach actually is necessary, is there any way for us to do this in a non-mutative way to `Router`? Can we extend the class / singleton and provide a new one to our express app? I feel confident there's got to be a better way here.

wbobeirne (changes requested):
There's a lot in this PR that I don't have the context to parse, but I caught a couple of items. VSCode seems to fire up just fine for me though, no bugs that I found with some cursory usage.

The `patchRouter` is the only big unknown to me. I haven't worked with modifying the underlying implementation of Express to know enough about where to do that, but I'm pretty sure this ain't it.

src/node/proxy.ts
```diff
@@ -0,0 +1,97 @@
...
+  // The ideal would be to have a reliable way to detect if this is a request
+  // for (or originating from) our root or login HTML. But requests for HTML
+  // don't seem to set any content type.
+  return (
+    req.headers["content-type"] !== "application/json" &&
+    ((req.originalUrl.startsWith("/") && req.method === "GET") ||
+      (req.originalUrl.startsWith("/static") && req.method === "GET") ||
+      (req.originalUrl.startsWith("/login") && (req.method === "GET" || req.method === "POST")))
+  )
+}
```
TeffenEllis, line 62:
```suggestion
const shouldFallThrough = (req: Request): boolean => {
const shouldFallThrough = ({ originalUrl, method, headers }: Request): boolean => {
  // The ideal would be to have a reliable way to detect if this is a request
  // for (or originating from) our root or login HTML. But requests for HTML
  // don't seem to set any content type.

  if (headers["content-type"] === "application/json") return false

  return (
    (originalUrl.startsWith("/") && method === "GET") ||
    (originalUrl.startsWith("/static") && method === "GET") ||
    (originalUrl.startsWith("/login") && (method === "GET" || method === "POST"))
  )
}
```

src/node/routes/index.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (args.cert && !(req.connection as tls.TLSSocket).encrypted) {
+      return res.redirect(`https://${req.headers.host}${req.originalUrl}`)
+    }
+
+    // Return robots.txt.
+    if (req.originalUrl === "/robots.txt") {
+      const resourcePath = path.resolve(rootPath, "src/browser/robots.txt")
+      res.set("Content-Type", getMediaMime(resourcePath))
+      return res.send(await fs.readFile(resourcePath))
+    }
```
TeffenEllis, line 77: Nit: this can likely be handled via [express static](https://expressjs.com/en/starter/static-files.html).

src/node/routes/index.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  await loadPlugins(app, args)
+
+  app.use(() => {
+    throw new HttpError("Not Found", HttpCode.NotFound)
+  })
+
+  // Handle errors.
+  // TODO: The types are broken; says they're all implicitly `any`.
+  app.use(async (err: any, req: any, res: any, next: any) => {
```
TeffenEllis, line 105: Express types are especially broad since `use` can encounter many situations:

```ts
import express, { RequestHandler, ErrorRequestHandler } from 'express'

const errorHandler: ErrorRequestHandler = (err, req, res, next) => {
  logger.error(err.toString())
  const status: number = err.status ?? err.statusCode ?? 500
  
  res.status(status).json(err)
}
```

src/node/routes/static.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as zlib from "zlib"
+import { HttpCode, HttpError } from "../../common/http"
+import { rootPath } from "../constants"
+import { authenticated, replaceTemplates } from "../http"
+import { getMediaMime, pathToFsPath } from "../util"
+
+export const router = Router()
+
+// The commit is for caching.
+router.get("/(:commit)(/*)?", async (req, res) => {
```
TeffenEllis, line 16: possibly replaceable with express-static

src/node/proxy.ts
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Determine if the user is browsing /, /login, or static assets and if so fall
+ * through to allow the redirect and login flow.
+ */
+const shouldFallThrough = (req: Request): boolean => {
+  // The ideal would be to have a reliable way to detect if this is a request
+  // for (or originating from) our root or login HTML. But requests for HTML
+  // don't seem to set any content type.
+  return (
+    req.headers["content-type"] !== "application/json" &&
```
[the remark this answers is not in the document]
code-asher: Ahhhh thank you!!

src/node/routes/vscode.ts
code-asher: You might need to re-apply the VS Code patch. It accepts the query and I had to modify it to match Express.

src/node/routes/index.ts
code-asher: Oh dope thank you for this, I forgot to go back and look into it. <3

src/node/routes/proxy.ts
code-asher: With my patched router if the request is for a websocket then **only** the `ws` methods will be called. If it's not a websocket request then the `ws` methods will **never** be called.

src/node/routes/update.ts
code-asher: Agreed!

src/node/http.ts
code-asher: I know we discussed on Slack but just so it's here as well this is the method that `express-ws` uses in case that affects our decision.

I'm not certain middleware can have the same effect because this patch will send websocket requests to `ws` handlers and everything else to the other handlers as per usual. With middleware we can modify the request but we can't change where it goes, I think? We also can't add a new `ws` method that way.

Some alternatives that came to mind as I was weaving this black magic:

  - Instead of a separate `ws` function every handler checks for `ws` on the `req` object and if it exists handle it appropriately (or error it if there's no websocket functionality at that route). My main objection with this is that the really clean registration interface Express provides gets a little muddled but it's not that bad.

  - Provide a wrapped router that everyone uses. `import { Router } from "express"` would be forbidden. This was actually the route I tried first but I aborted because I couldn't figure out a way to enforce it with code (I tried to override express.Router but it failed saying something about not having a getter) and it felt unfortunate that I (or a plugin) couldn't just import and use Express like usual. But I'm not totally opposed, ultimately it's not really that much more effort, just change the `from` on the import.

  - Have something completely separate. Like `import { wsRouter} from "./ws-router"` which has an Express-like syntax but maybe a bit simpler. I kind of like this one.

I'm happy to go with whatever we're most comfortable with!

src/node/routes/index.ts
code-asher: Ah yes! Agreed. I skipped for now just since we already had the code but will definitely look into this.

src/node/routes/index.ts
code-asher: tyty

src/node/entry.ts
code-asher: Oo nice catch.

src/node/update.ts
```diff
@@ -0,0 +1,133 @@
...
+        logger.debug("Making request", field("uri", uri))
+        const httpx = uri.startsWith("https") ? https : http
+        const client = httpx.get(uri, { headers: { "User-Agent": "code-server" } }, (response) => {
+          if (
+            response.statusCode &&
+            response.statusCode >= 300 &&
+            response.statusCode < 400 &&
+            response.headers.location
+          ) {
+            ++redirects
```
nhooyr: Nuts to me that we have to handle this.

src/node/update.ts
```diff
@@ -0,0 +1,133 @@
...
+            if (redirects > maxRedirects) {
+              return reject(new Error("reached max redirects"))
+            }
+            response.destroy()
```
nhooyr: Destroy should the above `reject`.

src/node/update.ts
```diff
@@ -0,0 +1,133 @@
...
+            return request(url.resolve(uri, response.headers.location))
+          }
+
+          if (!response.statusCode || response.statusCode < 200 || response.statusCode >= 400) {
```
nhooyr: If we move this above the previous status code check, then we only need a single `!response.statusCode`.

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+  public constructor() {
+    this.vsRootPath = path.resolve(rootPath, "lib/vscode")
+    this.serverRootPath = path.join(this.vsRootPath, "out/vs/server")
+  }
+
+  public async dispose(): Promise<void> {
+    if (this._vscode) {
+      const vscode = await this._vscode
+      vscode.removeAllListeners()
+      this._vscode = undefined
```
nhooyr, line 25: Should this be before `remoteAllListeners` in case that throws?

nhooyr: Or after `vscode.kill` I mean.

nhooyr: i.e if either throw, do we want to try again on a subsequent dispose, or do we want to just ignore?

src/node/proxy.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (!authenticated(req)) {
+    if (shouldFallThrough(req)) {
+      return next()
+    }
+    throw new HttpError("Unauthorized", HttpCode.Unauthorized)
+  }
+
+  proxy.web(req, res, {
+    ignorePath: true,
+    target: `http://127.0.0.1:${port}${req.originalUrl}`,
```
TeffenEllis, line 83: Binding all interfaces with `0.0.0.0` is sometimes necessary when accessing code-server over a local network, or when used with Dev URLs in Coder Enterprise. Will this cause any issues?

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+      vscode.kill()
+    }
+  }
+
+  public async initialize(
```
nhooyr, line 30: Still need to review this file and the final commit.

nhooyr (changes requested):
I know it's a lot but I tried to really be thorough as this is a big change and I wanted to make sure I fully understood the details as well.

TeffenEllis:
@code-asher,

One bit I noticed was a requirement to verify authentication for multiple router handlers. 

Screen Shot 2020-10-28 at 11 55 11

IMO, this can be moved into a single wrapper:

```ts
import * as express from "express"
import * as expressCore from "express-serve-static-core"
import qs from "qs"
//...

/**
 * Throw an error if not authorized.
 */
export const ensureAuthenticated: expressCore.RequestHandler<any, any, any, any> = (req): void => {
  if (!authenticated(req)) {
    throw new HttpError("Unauthorized", HttpCode.Unauthorized)
  }
}

export const authenticatedHandler = <
  P = expressCore.ParamsDictionary,
  ResBody = any,
  ReqBody = any,
  ReqQuery = qs.ParsedQs
>(
  cb: expressCore.RequestHandler<P, ResBody, ReqBody, ReqQuery>,
): expressCore.RequestHandler<P, ResBody, ReqBody, ReqQuery> => {
  const handler: expressCore.RequestHandler<P, ResBody, ReqBody, ReqQuery> = (...args) => {
    ensureAuthenticated(...args)
    cb(...args)
  }

  return handler
}
```

```ts
router.get(
  "/resource(/*)?",
  authenticatedHandler(async (req, res) => {
    if (typeof req.query.path === "string") {
      res.set("Content-Type", getMediaMime(req.query.path))
      res.send(await fs.readFile(pathToFsPath(req.query.path)))
    }
  }),
)
```

src/node/proxy.ts
nhooyr: @TeffenEllis this code is used to proxy to a localhost port. If you bind to `0.0.0.0` then it should be available on localhost as well!

@code-asher Perhaps it is a good idea to instead redirect to `0.0.0.0` so that any port on any interface is accessible.

src/node/http.ts
```diff
@@ -1,938 +1,253 @@
...
-    return {
-      proxy: {
-        port,
-      },
-    }
+/**
+ * Patch Express routers to handle web sockets and async routes (since we have
+ * to patch `get` anyway).
+ *
+ * Not using express-ws since the ws-wrapped sockets don't work with the proxy
```
nhooyr, line 187: Let's link to the upstream issue for this.

src/node/proxy.ts
code-asher: Sounds good to me!

src/node/cli.ts
```diff
@@ -325,13 +325,37 @@ export const parse = (
...
+  }
+
   logger.debug("parsed command line", field("args", args))
 
   return args
 }
 
-export async function setDefaults(args: Args): Promise<Args> {
-  args = { ...args }
+export interface DefaultedArgs extends ConfigArgs {
```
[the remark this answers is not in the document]
code-asher: Everything in `Args` is optional so this new interface reflects that some of these properties will definitely exist after setting the defaults. 

The practical benefit is that you can directly use things like `someFunc(args.host)` instead of having to do something like `args && someFunc(args.host) || throw new Error("host wasn't set")` or `someFunc(args.host!)`.

src/node/cli.ts
```diff
@@ -381,7 +405,52 @@ export async function setDefaults(args: Args): Promise<Args> {
...
+
+  // If we're being exposed to the cloud, we listen on a random address and
+  // disable auth.
+  if (args.link) {
+    args.host = "localhost"
+    args.port = 0
+    args.socket = undefined
+    args.cert = undefined
+
+    if (args.auth !== AuthType.None) {
```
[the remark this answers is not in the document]
code-asher: Fair!

src/node/cli.ts
```diff
@@ -381,7 +405,52 @@ export async function setDefaults(args: Args): Promise<Args> {
...
+
+  if (args.cert && !args.cert.value) {
+    const { cert, certKey } = await generateCertificate()
+    args.cert = {
+      value: cert,
+    }
+    args["cert-key"] = certKey
+  }
+
+  const usingEnvPassword = !!process.env.PASSWORD
```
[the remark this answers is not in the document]
code-asher: This is used by the login page to tell if the password was set from the environment since we `delete process.env.PASSWORD`.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+ * Get the relative path that will get us to the root of the page. For each
+ * slash we need to go up a directory. For example:
+ * / => .
+ * /foo => .
+ * /foo/ => ./..
+ * /foo/bar => ./..
+ * /foo/bar/ => ./../..
+ */
+export const relativeRoot = (req: express.Request): string => {
+  const depth = (req.originalUrl.split("?", 1)[0].match(/\//g) || []).length
```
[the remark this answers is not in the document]
code-asher: Splits off the query portion so we have just the path. I would have used `req.path` but they strip off the trailing slash for some reason and that information is required to get the proper relative URL. I'm not super familiar with Express though so maybe there's a better way to do this.

test/cli.test.ts
```diff
@@ -232,6 +238,71 @@ describe("parser", () => {
...
+  })
+
+  it("should use env var password", async () => {
+    process.env.PASSWORD = "test"
+    const args = parse([])
+    assert.deepEqual(args, {
+      _: [],
+    })
+
+    assert.deepEqual(await setDefaults(args), {
```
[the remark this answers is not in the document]
code-asher: Ahh yeah good point. Actually a few things in this function feel kind of weird to me because they aren't really defaults per-se. What do you think about renaming it entirely?

I'm having trouble coming up with a good name though. `prepareArgs`?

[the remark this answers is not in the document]
code-asher: I'm good with two functions as well though.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+  override: expressCore.Query = {},
+): void => {
+  const query = Object.assign({}, req.query, override)
+  Object.keys(override).forEach((key) => {
+    if (typeof override[key] === "undefined") {
+      delete query[key]
+    }
+  })
+
+  const relativePath = normalize(`${relativeRoot(req)}/${to}`, true)
```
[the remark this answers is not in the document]
code-asher: It doesn't seem to have anything in the way of helper methods like this. But I might have just missed it! Although it's possible `res.redirect` will normalize for you (I haven't tested).

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
-    }
-    this.proxy.on("error", (error, _request, response) => {
-      response.writeHead(HttpCode.ServerError)
-      response.end(error.message)
-    })
-    // Intercept the response to rewrite absolute redirects against the base path.
-    this.proxy.on("proxyRes", (response, request: ProxyRequest) => {
-      if (response.headers.location && response.headers.location.startsWith("/") && request.base) {
-        response.headers.location = request.base + response.headers.location
+export const getCookieDomain = (host: string, proxyDomains: string[]): string | undefined => {
```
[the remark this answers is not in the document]
code-asher: Yup, I believe you can send multiple cookies with different domains in them in the same request.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+    // Might be blank/missing, so there's nothing more to do.
+    !host ||
+    // IP addresses can't have subdomains so there's no value in setting the
+    // domain for them. Assume anything with a : is ipv6 (valid domain name
+    // characters are alphanumeric or dashes).
+    host.includes(":") ||
+    // Assume anything entirely numbers and dots is ipv4 (currently tlds
+    // cannot be entirely numbers).
+    !/[^0-9.]/.test(host) ||
+    // localhost subdomains don't seem to work at all (browser bug?).
```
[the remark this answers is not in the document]
code-asher: Oof yeah this comment did not make the problem clear. With localhost domains the cookies don't seem to filter down. So if my domain is `dev.localhost` and I go to `8080.dev.localhost` this code would try to set the cookie at `dev.localhost` but it can't be read at `8080.dev.localhost` which means you can never log in there.

We could instead set a cookie at `8080.dev.localhost` and `dev.localhost`, maybe that's what you were thinking with the multiple cookie question? That seems decent to me although I suppose if we ever implement a logout it could be a bit tricky getting all the cookies since we won't know which sub-domains they've been set for. (I think?)

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+    host.endsWith(".localhost") ||
+    // It might be localhost (or an IP, see above) if it's a proxy and it
```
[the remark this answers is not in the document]
code-asher: Hopefully my above comment explains this better. Just to note, this isn't preventing setting a cookie, just that it's going to set the cookie without a `domain` attribute which will set it at the current domain.

The problem I had was that if I set a cookie with `domain=localhost` it just wouldn't get set at all.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
-      response.end(error.message)
-    })
-    // Intercept the response to rewrite absolute redirects against the base path.
-    this.proxy.on("proxyRes", (response, request: ProxyRequest) => {
-      if (response.headers.location && response.headers.location.startsWith("/") && request.base) {
-        response.headers.location = request.base + response.headers.location
+export const getCookieDomain = (host: string, proxyDomains: string[]): string | undefined => {
+  const idx = host.lastIndexOf(":")
+  host = idx !== -1 ? host.substring(0, idx) : host
+  if (
```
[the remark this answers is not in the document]
code-asher: Hopefully my above comment explains a bit better, the idea is that if you are visiting `8080.codeserver.com` then with this code the cookie will be set at `codeserver.com` instead of `8080.codeserver.com` so that you have a single cookie and only have to log in once.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+    socket: net.Socket,
+    head: Buffer,
+    req: express.Request,
+    next: express.NextFunction,
+  ) => void | Promise<void>
+
+  type WebsocketMethod<T> = (route: expressCore.PathParams, ...handlers: WebsocketRequestHandler[]) => T
+
+  interface WithWebsocketMethod {
+    ws: WebsocketMethod<this>
```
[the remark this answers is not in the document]
code-asher: To be honest it's just lifted from `@types/express-ws`. But I think it's required to make the return type `this`.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+export const handleUpgrade = (app: express.Express, server: http.Server): void => {
+  server.on("upgrade", (req, socket, head) => {
+    socket.on("error", () => socket.destroy())
+
+    req.ws = socket
+    req.head = head
+    req._ws_handled = false
+
+    const res = new http.ServerResponse(req)
+    res.writeHead = function writeHead(statusCode: number) {
```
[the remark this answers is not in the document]
code-asher: It's called every time. It's a bit weird because we're sending the request through Express which doesn't have any concept of web sockets, then at the end of the chain it calls this. We then just do nothing unless there was an error in which case we destroy the socket.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+    route: expressCore.PathParams,
+    ...handlers: express.WebsocketRequestHandler[]
+  ) {
+    originalGet.apply(this, [
+      route,
+      ...handlers.map((handler) => {
+        const wrapped: express.Handler = (req, _, next) => {
+          if ((req as any).ws) {
+            ;(req as any)._ws_handled = true
+            Promise.resolve(handler((req as any).ws, (req as any).head, req, next)).catch(next)
```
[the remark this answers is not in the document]
code-asher: The handler might not return a promise so `.catch` wouldn't work. We could rewrite to make this handler async and use await + try/catch though.

src/node/http.ts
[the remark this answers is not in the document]
wbobeirne: Worth noting this code is unsafe, if `req.originalUrl` is missing a `?` this will crash with `[0].match`

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
-      const existingProvider = this.providers.get(`/${endpoint}`)
-      this.providers.set(`/${endpoint}`, p)
-      if (existingProvider) {
-        logger.debug(`Overridding existing /${endpoint} provider`)
-        // If the existing provider isn't registered elsewhere we can dispose.
-        if (!Array.from(this.providers.values()).find((p) => p === existingProvider)) {
-          logger.debug(`Disposing existing /${endpoint} provider`)
-          existingProvider.dispose()
-        }
+    // Send the request off to be handled by Express.
```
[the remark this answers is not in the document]
code-asher: I'm not 100% sure but I think `http.Server.on("upgrade"` just always gives us the upgraded socket. I'm not sure there's a way to check the route first. Lemme know if I've misunderstood.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
+          } else {
+            next()
+          }
+        }
+        return wrapped
+      }),
+    ])
+    return this
+  }
+  // Handle async responses for `post` as well since we're in here anyway.
```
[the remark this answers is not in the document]
code-asher: Right?! Seems like it would be very easy for them to support. But I tested and async errors are lost to the void.

If we don't call next then if it hasn't written anything the connection will just hang. For example if there's an `fs.readFile` that fails we won't ever get to the final error page. Maybe there's a way to detect if the connection has been written to?

src/node/proxy.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Return the port if the request should be proxied. Anything that ends in a
+ * proxy domain and has a *single* subdomain should be proxied. Anything else
+ * should return `undefined` and will be handled as normal.
+ *
+ * For example if `coder.com` is specified `8080.coder.com` will be proxied
+ * but `8080.test.coder.com` and `test.8080.coder.com` will not.
+ *
+ * Throw an error if proxying but the user isn't authenticated.
```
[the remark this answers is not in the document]
code-asher: Ah crap thank you I moved it out. A comment is a lie waiting to happen.

src/node/routes/index.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+  app.use("/proxy", proxy.router)
+  app.use("/static", _static.router)
+  app.use("/update", update.router)
+  app.use("/vscode", vscode.router)
+
+  await loadPlugins(app, args)
+
+  app.use(() => {
+    throw new HttpError("Not Found", HttpCode.NotFound)
```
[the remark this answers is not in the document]
code-asher: Yup, latest on npm is 4.17.1 though. We could use the alpha though? `5.0.0-alpha.6` looks like the latest.

src/node/entry.ts
```diff
@@ -111,15 +111,14 @@ const main = async (args: DefaultedArgs): Promise<void> => {
...
 
   const [app, server] = await createApp(args)
   const serverAddress = ensureAddress(server)
-
-  // TODO: register routes
-  await loadPlugins(app, args)
+  await register(app, server, args)
 
   logger.info(`Using config file ${humanPath(args.config)}`)
   logger.info(`HTTP server listening on ${serverAddress} ${args.link ? "(randomized by --link)" : ""}`)
```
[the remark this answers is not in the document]
code-asher: Aw crap sorry about that, I should have realized.

src/node/routes/static.ts
```diff
@@ -1,73 +1,61 @@
...
-      return this.getTarredResource(request, pathToFsPath(route.query.tar))
-    }
-
-    const response = await this.getReplacedResource(request, route)
-    if (!this.isDev) {
-      response.cache = true
-    }
-    return response
+// The commit is for caching.
+router.get("/(:commit)(/*)?", async (req, res) => {
```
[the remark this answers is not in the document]
code-asher: Same as before, for example `/static/-/path/to/file`. We could have a separate route if we wanted to avoid the extra `-`.

src/node/routes/static.ts
```diff
@@ -1,73 +1,61 @@
...
-  /**
-   * Tar up and stream a directory.
-   */
-  private async getTarredResource(request: http.IncomingMessage, ...parts: string[]): Promise<HttpResponse> {
-    const filePath = path.join(...parts)
-    let stream: Readable = tarFs.pack(filePath)
-    const headers: http.OutgoingHttpHeaders = {}
-    if (request.headers["accept-encoding"] && request.headers["accept-encoding"].includes("gzip")) {
-      logger.debug("gzipping tar", field("filePath", filePath))
+  const tar = Array.isArray(req.query.tar) ? req.query.tar[0] : req.query.tar
```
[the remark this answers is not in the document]
code-asher: Loading extensions in the browser! Like Vim.

src/node/routes/vscode.ts
```diff
@@ -1,237 +1,99 @@
...
-          ? resolve(message.options)
-          : reject(new Error("Unexpected response during initialization"))
-      })
-      vscode.once("error", reject)
-      vscode.once("exit", (code) => reject(new Error(`VS Code exited unexpectedly with code ${code}`)))
-      this.send({ type: "init", id, options }, vscode)
+router.get("/", async (req, res) => {
+  if (!authenticated(req)) {
+    return redirect(req, res, "login", {
+      to: req.baseUrl || "/",
```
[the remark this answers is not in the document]
code-asher: Nice catch! Was left over from before when the left side could be undefined.

src/node/routes/vscode.ts
```diff
@@ -1,237 +1,99 @@
...
-
-    response.content = response.content
+  ])
+
+  options.productConfiguration.codeServerVersion = version
+
+  res.send(
+    replaceTemplates(
+      req,
+      // Uncomment prod blocks if not in development. TODO: Would this be
```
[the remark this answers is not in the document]
code-asher: Agreed! I wanted to switch to a templating library at some point. I'm good with the JS strings, I agree it's an improvement over the current situation.

src/node/routes/vscode.ts
```diff
@@ -1,237 +1,99 @@
...
+
+router.get("/resource(/*)?", async (req, res) => {
+  ensureAuthenticated(req)
+  if (typeof req.query.path === "string") {
+    res.set("Content-Type", getMediaMime(req.query.path))
+    res.send(await fs.readFile(pathToFsPath(req.query.path)))
+  }
+})
+
+router.get("/vscode-remote-resource(/*)?", async (req, res) => {
```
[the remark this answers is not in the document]
code-asher: I was recalling that there's this weird case somewhere that does `/vscode-remote-resource/something/something?path=/path`. I haven't double-checked it so I just added the wildcard for now.

src/node/update.ts
```diff
@@ -0,0 +1,133 @@
...
+  ) {}
+
+  /**
+   * Query for and return the latest update.
+   */
+  public async getUpdate(force?: boolean): Promise<Update> {
+    // Don't run multiple requests at a time.
+    if (!this.update) {
+      this.update = this._getUpdate(force)
+      this.update.then(() => (this.update = undefined))
```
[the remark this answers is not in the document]
code-asher: Fair! `this._getUpdate` has a try/catch and so should never fail but should we catch anyway just in case?

src/node/update.ts
```diff
@@ -0,0 +1,133 @@
...
+    const now = Date.now()
+    try {
+      let { update } = !force ? await this.settings.read() : { update: undefined }
+      if (!update || update.checked + this.updateInterval < now) {
+        const buffer = await this.request(this.latestUrl)
+        const data = JSON.parse(buffer.toString()) as LatestResponse
+        update = { checked: now, version: data.name.replace(/^v/, "") }
+        await this.settings.write({ update })
+      }
+      logger.debug("got latest version", field("latest", update.version))
```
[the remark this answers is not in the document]
code-asher: I'm down for that.

src/node/vscode.ts
code-asher: Good point, I'll move it after.

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+      })
+
+      this._vscode = new Promise((resolve, reject) => {
+        vscode.once("message", (message: ipc.VscodeMessage) => {
+          logger.debug("got message from vs code", field("message", message))
+          return message.type === "ready"
+            ? resolve(vscode)
+            : reject(new Error("Unexpected response waiting for ready response"))
+        })
+        vscode.once("error", reject)
```
nhooyr, line 96: Is it a good idea/easy to unbind such event handlers? As after a resolve, if vscode errors or exits, this event handler doesn't have to be called. I'm not sure.

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    return this._vscode
+  }
+
+  /**
+   * VS Code expects a raw socket. It will handle all the web socket frames.
+   */
+  public async sendWebsocket(socket: net.Socket, query: ipc.Query): Promise<void> {
+    // TODO: TLS socket proxy.
```
nhooyr, line 108: Could you expand on this?

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+    const vscode = await this._vscode
+    this.send({ type: "socket", query }, vscode, socket)
+  }
+
+  private send(message: ipc.CodeServerMessage, vscode?: cp.ChildProcess, socket?: net.Socket): void {
```
nhooyr, line 113: `vscode` is always present here.

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+      throw new Error("vscode is not running")
+    }
+    vscode.send(message, socket)
+  }
+
+  /**
+   * Choose the first non-empty path.
+   */
+  private async getFirstPath(
+    startPaths: Array<{ url?: string | string[] | ipc.Query | ipc.Query[]; workspace?: boolean } | undefined>,
```
nhooyr, line 124: Some documentation on these different parameters would be really helpful.

src/node/vscode.ts
```diff
@@ -0,0 +1,150 @@
...
+      { url: query.folder, workspace: false },
+      options.args._ && options.args._.length > 0
+        ? { url: path.resolve(options.args._[options.args._.length - 1]) }
+        : undefined,
+      lastVisited,
+    ])
+
+    settings.write({
+      lastVisited: startPath,
+      query,
```
nhooyr, line 46: Some docs on this file and what we persist would be good too, but that's probably for a separate PR. Opened #2249

nhooyr (changes requested):

src/node/update.ts
code-asher: Very true. There's probably a library we could import.

src/node/vscode.ts
code-asher: Seems like a good idea to me. Originally I thought `error` only fired if the process fails to start up at all but I just checked and that's not the case. I'll unbind it when the promise resolves/rejects.

src/node/vscode.ts
code-asher: It might not be, `sendWebsocket` for example can call `this.send` with `undefined` (if `this._vscode` is undefined). I think I'll move that check to `sendWebsocket` though because it's only used in one other place and it definitely does exist there.

code-asher: Ah well maybe not since then I have two places where it throws `vscode is not running` which feels a little unfortunate.

src/node/http.ts
code-asher: I suspect they don't have any plans on letting raw sockets pass through since the library seems focused on specifically wrapping the sockets with `ws`. Still, maybe they could make it optional. The library was last updated in 2018 so idk how active they still are.

I removed the bit about the wildcard since I think that can actually be worked around although I think the method used here is still superior.

src/node/http.ts
code-asher: Well this sucks, this method doesn't work on Express 5.

src/node/http.ts
```diff
@@ -1,885 +1,253 @@
...
-    }
+        return wrapped
+      }),
+    ])
+    return this
+  }
+  // Overwrite `get` so we can distinguish between websocket and non-websocket
+  // routes. While we're at it handle async responses.
+  ;(express.Router as any).get = function get(route: expressCore.PathParams, ...handlers: express.Handler[]) {
+    originalGet.apply(this, [
```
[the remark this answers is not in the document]
code-asher: Unfortunately `this` will be undefined, I think because `originalGet` is from the `Router()` prototype and not the actual router.
```
    var route = this.route(path)
                     ^
TypeError: Cannot read property 'route' of undefined
```

src/node/http.ts
code-asher: I got it working on Express 5 but I opted to use a separate app to handle web socket routes because there's middleware on the main app that doesn't make any sense to have on websocket routes, for example the cookie or body parser, the error html returned if routes throw an error, and so on. There's no jank or patching in this new version.

src/node/routes/proxy.ts
code-asher: Was wrong about this, would need to patch `all` as well to fix this.

src/node/vscode.ts
```diff
@@ -51,14 +51,20 @@ export class VscodeProvider {
 
     logger.debug("setting up vs code...")
     return new Promise<ipc.WorkbenchOptions>((resolve, reject) => {
+      const onExit = (code: number | null) => reject(new Error(`VS Code exited unexpectedly with code ${code}`))
+
       vscode.once("message", (message: ipc.VscodeMessage) => {
         logger.debug("got message from vs code", field("message", message))
+        vscode.off("error", reject)
```
nhooyr, line 58: Why do we want to unbind if an error occurs with the vscode process? Wouldn't we still want to be notified?

nhooyr (changes requested):
commit separation is next level wow

src/node/vscode.ts
code-asher: Here all we do is reject if an error occurs, so once the promise has fulfilled keeping this handler will call `reject` but have no effect so we might as well unbind it.

But we do bind a different handler to `error` during spawn and log the error there as well as unset the process so we can spawn another one.

src/node/proxy.ts
```diff
@@ -50,15 +50,18 @@ const maybeProxy = (req: Request): string | undefined => {
...
-  // The ideal would be to have a reliable way to detect if this is a request
-  // for (or originating from) our root or login HTML. But requests for HTML
-  // don't seem to set any content type.
-  return (
-    req.headers["content-type"] !== "application/json" &&
-    ((req.originalUrl.startsWith("/") && req.method === "GET") ||
-      (req.originalUrl.startsWith("/static") && req.method === "GET") ||
-      (req.originalUrl.startsWith("/login") && (req.method === "GET" || req.method === "POST")))
-  )
+  // See if it looks like a request for the root or login HTML.
```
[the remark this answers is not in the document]
code-asher: It was in response to this issue: https://github.com/cdr/code-server/issues/2128

The gist is that if we return a 40x (if we let all requests fall through) or a 30x (if we redirect all of them) then it masks the real problem (that the request isn't authenticated). 

So in the above issue the application makes xhr POST requests against the root but by default it doesn't include cookies. So at the time it fell through and got a 400 for those requests which made it look like code-server wasn't proxying correctly.

If we instead redirected all the requests to the login I imagine it'd try to parse it as JSON and would error with something like "unexpected token <".

I wasn't confident there wouldn't be more cases like this so I let through the absolute minimum required to log in.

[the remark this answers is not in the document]
nhooyr: > If we instead redirected all the requests to the login I imagine it'd try to parse it as JSON and would error with something like "unexpected token <".

If the requester accepts html, we should redirect to login every time. If they do not, then we should return a 401 with a error message that there is no code-server cookie.

src/node/http.ts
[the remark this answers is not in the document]
code-asher: No longer called in Express 5 so I've removed it.

src/node/proxy.ts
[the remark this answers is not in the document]
code-asher: Fair enough, I didn't do this initially because I was concerned some framework somewhere uses it in xhr requests or some noob wouldn't set the right header and blame code-server. Seems like a safe enough bet though, and it means you don't have to be at the root of your app to get redirected which is definitely nice.

src/node/vscode.ts
nhooyr: I guess I'm confused why we're binding twice? What's wrong with the earlier bind? I understand rejecting the promise but why duplicate `onExit`?

code-asher: By duplicate `exit` binds do you mean the one here in `initialize` and the bind in `fork`?

nhooyr: Yes!

nhooyr: Ah I see, you wait for the `"ready"` message and then `"options"` message. I think we can definitely structure this code to reduce the duplication. Feel free to defer though for a later PR.

code-asher: Mmm yeah `fork` happens once when spawning VS Code and `initialize` happens every time we load the page to send and receive initialization options. Def agree that we could create some kind of generalized wrapper around a process that handles all the binds for you whenever you want to spawn or communicate with it since there's quite a bit of overlap. For now I'll defer to later :smile:

code-asher: Ahh yeah I re-read your comment and realized it's not just "quite a bit of overlap" it's literally the same code lol just a different message.

code-asher: Gonna just rewrite it now actually.

src/node/wsRouter.ts
```diff
@@ -0,0 +1,57 @@
+import * as express from "express"
+import * as expressCore from "express-serve-static-core"
+import * as http from "http"
+import * as net from "net"
+
+export const handleUpgrade = (app: express.Express, server: http.Server): void => {
```
nhooyr, line 6: yea this is a lot nicer! big fan of this commit!

src/node/wsRouter.ts
```diff
@@ -0,0 +1,57 @@
...
+  server.on("upgrade", (req, socket, head) => {
+    socket.on("error", () => socket.destroy())
```
nhooyr, line 8: Perhaps we destroy the socket with the error passed in? I'm surprised node doesn't do this automatically wtf.

nhooyr: Actually, it does! Any error will cause the socket to become closed.

https://nodejs.org/api/net.html#net_event_error_1

> Emitted when an error occurs. **The 'close' event will be called directly following this event.**

src/node/routes/index.ts
```diff
@@ -117,4 +142,11 @@ export const register = async (app: Express, server: http.Server, args: Defaulte
   }
 
   app.use(errorHandler)
+
+  const wsErrorHandler: express.ErrorRequestHandler = async (err, req) => {
+    logger.error(`${err.message} ${err.stack}`)
+    ;(req as WebsocketRequest).ws.destroy(err)
```
nhooyr, line 158: I believe we want to use `end` here to ensure at least everything written to the connection so far is flushed correctly.

src/node/wsRouter.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    req.ws = socket
+    req.head = head
+    req._ws_handled = false
+
+    // Send the request off to be handled by Express.
+    ;(app as any).handle(req, new http.ServerResponse(req), () => {
+      if (!req._ws_handled) {
+        socket.destroy(new Error("Not found"))
```
nhooyr, line 17: `end`.

nhooyr: Shouldn't we write a 404 here actually?

src/node/update.ts
nhooyr: Yea if we can I'd much rather we use a library here. It's really error prone to write this all ourself. I'm sure we're probably missing a few HTTP gotchas too.

src/node/routes/index.ts
```diff
@@ -121,23 +121,29 @@ export const register = async (
...
-  const errorHandler: express.ErrorRequestHandler = async (err, req, res, next) => {
-    const resourcePath = path.resolve(rootPath, "src/browser/pages/error.html")
-    res.set("Content-Type", getMediaMime(resourcePath))
-    try {
+  const errorHandler: express.ErrorRequestHandler = async (err, req, res) => {
+    if (err.code === "ENOENT" || err.code === "EISDIR") {
+      err.status = HttpCode.NotFound
+    }
+
+    const status = err.status ?? err.statusCode ?? 500
```
nhooyr, line 133: Any reason we don't just use `||`? Is it best practice now to use `??` to avoid `0` being treated as false?

src/node/proxy.ts
```diff
@@ -71,9 +52,27 @@ router.all("*", (req, res, next) => {
...
-    if (shouldFallThrough(req)) {
+    // Let the assets through since they're used on the login page.
+    if (req.path.startsWith("/static/") && req.method === "GET") {
       return next()
     }
+
+    // Assume anything that explicitly accepts text/html is a user browsing a
+    // page (as opposed to an xhr request). Don't use `req.accepts()` since
+    // *every* request that I've seen (in Firefox and Chromium at least)
+    // includes `*/*` making it always truthy.
```
nhooyr: appreciate the comment here.

nhooyr (changes requested):

src/common/emitter.ts
```diff
@@ -32,8 +38,21 @@ export class Emitter<T> {
   /**
    * Emit an event with a value.
    */
-  public emit(value: T): void {
-    this.listeners.forEach((cb) => cb(value))
+  public async emit(value: T): Promise<void> {
+    let resolve: () => void
+    const promise = new Promise<void>((r) => (resolve = r))
```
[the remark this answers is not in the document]
code-asher: I think we have to separate it because if we call `cb` in the promise callback then `promise` will still be undefined as it hasn't been set yet (since these callbacks are ran synchronously).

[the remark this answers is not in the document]
code-asher: Like if we did something like:
```
const promise = Promise.all(listeners.map(async (cb) => {
  await cb(value, promise)
})
```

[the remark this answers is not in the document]
nhooyr: Ohhh, I gotchu.!

src/node/routes/index.ts
code-asher: I'm not sure if it's best practice, but the `0` is why I used it. Although that's probably not a valid HTTP code anyway?

src/node/wsRouter.ts
code-asher: I'm a bit nervous with `end` because that's what I used for the VS Code socket and it resulted in the connections lingering causing the heartbeat file to be inaccurate: https://github.com/cdr/code-server/commit/9ff0e455c366449b2ba02674d878303fb967ef36

Maybe this was a symptom of some other problem though? No idea. I think maybe it's possible to keep the socket open even after `end` in some cases but I don't know exactly what.

Can we write a 404 to a socket? Node gives them to us already upgraded. My knowledge around how exactly this all works is pretty fuzzy.

src/node/wsRouter.ts
code-asher: Oh dope! Will remove.

src/node/routes/index.ts
code-asher: That makes sense, I'll change this and the other one to `end`.

src/node/update.ts
code-asher: Fair.

src/node/routes/index.ts
nhooyr: Yea it isnt.

nhooyr: nbd, just asking questions.

src/node/wsRouter.ts
nhooyr: Yea so we're always writing a 101 above right. We shouldn't be doing that. Only when we find a handler should the 101 be written, otherwise it should be a normal 404.

If that's not clear, see https://tools.ietf.org/html/rfc6455#section-1.2 for details on how the handshake process works

code-asher: I think I just don't know what I'm talking about lol. We handle the upgrade, not Node. I think I have my terminology mixed up. So we can just `socket.write` the 404 if I understand correctly.

code-asher: I think this is sorted now. We only 101 on the VS Code endpoint, otherwise if it's not handled it's a 404.

code-asher:
I think all comments have been addressed except the following which I think we could tackle a bit later:

- Store last update internally instead of to disk
- Use a library to fetch the latest version from GitHub
- Split or rename `setDefaults`

Lemme know if I missed something.

One problem though is that the error handler doesn't seem to work. It never gets called so we just get the default Express error page. I'm not sure if it has something to do with the Express 5 update (I didn't see anything relevant in the migration guide: https://expressjs.com/en/guide/migrating-5.html) or what.

nhooyr:
> One problem though is that the error handler doesn't seem to work. It never gets called so we just get the default Express error page. I'm not sure if it has something to do with the Express 5 update (I didn't see anything relevant in the migration guide: https://expressjs.com/en/guide/migrating-5.html) or what.

Let's make that a release blocking issue.

src/node/routes/domainProxy.ts
```diff
@@ -55,7 +55,7 @@ router.all("*", (req, res, next) => {
       }
       // Redirect all other pages to the login.
       return redirect(req, res, "login", {
-        to: req.path,
+        to: req.path !== "/" ? req.path : undefined,
```
nhooyr, line 58: Do we not need to worry about `baseUrl` here?

src/node/routes/vscode.ts
```diff
@@ -22,18 +22,14 @@ router.get("/", async (req, res) => {
...
-    vscode
-      .initialize(
-        {
-          args: req.args,
-          remoteAuthority: req.headers.host || "",
-        },
-        req.query,
-      )
-      .catch((error) => {
+    (async () => {
```
nhooyr, line 25: 👍🏼 much easier to read.

nhooyr (changes requested):
Approval after these last few concerns.

src/node/vscode.ts
```diff
@@ -88,32 +81,71 @@ export class VscodeProvider {
...
+  /**
+   * Listen to a single message from a process. Reject if the process errors,
+   * exits, or times out.
+   *
+   * `fn` is a function that determines whether the message is the one we're
+   * waiting for.
+   */
+  private onMessage<T extends ipc.VscodeMessage>(
+    proc: cp.ChildProcess,
+    fn: (message: ipc.VscodeMessage) => message is T,
```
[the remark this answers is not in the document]
code-asher: I only wish it was a liiiittle bit smarter and could infer the callback return type instead of making me use `message is ipc.ReadyMessage`.

src/node/vscode.ts
```diff
@@ -88,32 +81,71 @@ export class VscodeProvider {
...
+      const reject = (error: Error) => {
+        clearTimeout(timeout)
+        _reject(error)
+      }
+
+      const onExit = (code: number | null) => {
+        reject(new Error(`VS Code exited unexpectedly with code ${code}`))
+      }
+
+      const timeout = setTimeout(() => {
```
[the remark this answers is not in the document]
code-asher: Fair fair, just so I don't look totally incompetent this works as well. :smile: Just realized though that we need to unbind `error` and `exit` on timeout though.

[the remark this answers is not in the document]
code-asher: Oof and I unbound too early in the `message` handler.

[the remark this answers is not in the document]
code-asher: I retract my statement about not being incompetent. :stuck_out_tongue:

src/node/vscode.ts
```diff
@@ -52,33 +53,25 @@ export class VscodeProvider {
...
+          startPath,
         },
-        vscode,
-      )
+      },
+      vscode,
+    )
+
+    const message = await this.onMessage<ipc.OptionsMessage>(vscode, (message): message is ipc.OptionsMessage => {
+      // There can be parallel initializations so wait for the right ID.
```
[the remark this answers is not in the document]
code-asher: Loading the page in two separate tabs is the main way I think. If we had some sort of collaborative mode it would probably be a bigger problem then.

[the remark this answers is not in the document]
code-asher: I guess in theory it could also happen if you refresh a page before it finishes loading to send off two almost-parallel requests although I don't think I've hit it that way before.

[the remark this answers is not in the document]
nhooyr: > Loading the page in two separate tabs is the main way I think. If we had some sort of collaborative mode it would probably be a bigger problem then.

Shouldn't we prevent sending parallel `init`'s at the same time?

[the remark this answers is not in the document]
nhooyr: Or actually, doesn't each websocket get it's own vscode process? Or is that only its own extension host?

src/node/routes/domainProxy.ts
code-asher: I didn't add it since `baseUrl` is empty when the route is on the root but I think it's a good idea to add it since the route should work no matter where it's mounted.

src/node/vscode.ts
```diff
@@ -104,7 +104,7 @@ export class VscodeProvider {
       dispose()
     })
 
-    this._vscode = this.onMessage<ipc.ReadyMessage>(vscode, (message): message is ipc.ReadyMessage => {
```
nhooyr: 👌🏼

nhooyr (approved):
🚀

src/node/vscode.ts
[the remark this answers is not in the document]
code-asher: There's a single VS Code process for all websockets and an extension host for each websocket. Each websocket also gets its own management connection where we initialize all the services VS Code uses (like the file service, environment service, language initialization, and so on).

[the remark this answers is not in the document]
code-asher: The services are shared too though, come to think. Language stuff isn't though, and there are some other differences in the initialization like which folder to open up (depending on the query parameter for example).

Issue #2247, opened 2020-10-29
Submit button styling issue


jw-12138:
- Web Browser: Safari
- Local OS: iOS / iPadOS
- Remote OS: Linux / Ubuntu 20.04 LTS
- Remote Architecture: x86-64
- `code-server --version`: 3.6.1

---

Can say it's a bug, but here we are...🌝

![image](https://user-images.githubusercontent.com/29943110/97517376-0acc8680-19d0-11eb-8486-cf3f09f86377.png)

The submit button is not well styled.

just that.

PR #2248, opened 2020-10-29
Fix a styling issue on iOS/iPadOs

jw-12138:
Related issue: [https://github.com/cdr/code-server/issues/2247](https://github.com/cdr/code-server/issues/2247)

So, instead of using button elements( including `input[type="submit"]`)  directly on page, I added a `label` element which preserved the style while remaining the same functionality.

### Before
image

### After 
image

Over all, not a big problem but still, an improvement.

### DIFF
--- a/src/browser/pages/login.css
+++ b/src/browser/pages/login.css
@@ -26,7 +26,7 @@ body {
   border: 1px solid #ddd;
   box-sizing: border-box;
   color: black;
-  flex: 1;
+  width: 100%;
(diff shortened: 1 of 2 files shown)


nhooyr:
Hi @jw-12138, thanks for reporting this bug and contributing. ❤️ 

I've opened https://github.com/cdr/code-server/pull/2339 with your commit included for credits but a simpler fix applied where we just disable webkit's native input styles.

PR #2252, opened 2020-10-30
Plugin API to add more applications to code-server

nhooyr:
This PR introduces the commits that implement the plugin system we've discussed.

It is described in detail in [./typings/plugin.d.ts](./typings/plugin.d.ts).

The overlay is not yet implemented but I'm working on it.


src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
+      const p = this._loadPlugin(dir, packageJSON)
+      this.plugins.push(p)
+    } catch (err) {
+      if (err.code !== "ENOENT") {
+        this.logger.warn(`failed to load plugin: ${err.message}`)
+      }
+    }
+  }
+
+  private _loadPlugin(dir: string, packageJSON: PackageJSON): Plugin {
```
wbobeirne, line 138: This nomenclature is pretty confusing, it's unclear from their names and without reading the code fully as to why I would call `loadPlugin` vs `_loadPlugin`, especially since they're both private.

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
+  name: string
+  version: string
+  description: string
+  engines: {
+    "code-server": string
+  }
+}
+
+function q(s: string): string {
+  if (s === undefined) {
```
wbobeirne, line 185: This typing doesn't line up.

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
+    }
+    if (!packageJSON.name) {
+      throw new Error("plugin missing name")
+    }
+    if (!packageJSON.version) {
+      throw new Error("plugin missing version")
+    }
+    if (!packageJSON.description) {
+      throw new Error("plugin missing description")
+    }
```
wbobeirne, line 155: This doesn't feel like it's worth `throw`ing over.

typings/plugin.d.ts
```diff
@@ -0,0 +1,110 @@
...
+export interface Plugin {
+  /**
+   * init is called so that the plugin may initialize itself with the config.
+   */
+  init(config: PluginConfig): void
+
+  /**
+   * Returns the plugin's router.
+   */
+  router(): express.Router
```
wbobeirne, line 67: ? This wasn't in the proposed spec. While I see what you're going for here, and preventing router collisions like this makes sense, this will be pretty disruptive to the current routes and bares further discussion.

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
-export const loadPlugins = async (app: Express, args: Args): Promise<void> => {
-  const pluginPath = process.env.CS_PLUGIN_PATH || `${path.join(paths.data, "plugins")}:/usr/share/code-server/plugins`
-  const plugin = process.env.CS_PLUGIN || ""
-  await Promise.all([
+  /**
+   * mount mounts all plugin routers onto r.
+   */
+  public mount(r: express.Router): void {
+    for (const p of this.plugins) {
+      r.use(`/${p.name}`, p.router())
```
wbobeirne, line 72: This is more of a minor concern, but the routes are going to be really ugly if these plugins come from npm. They'll all be something like `/code-server-plugin-jupyter/ide` or something to that effect.

wbobeirne (changes requested):
Code overall looks good, but deviates from the agreed spec. We should probably catch up about the changes.

src/node/plugin.ts
nhooyr: This is a pretty standard idiom. `_` is just the counterpart to the main exposed function and is less "coupled". i.e here it returns the `Plugin` so that the other function can add it into the list of registered plugins.

You wouldn't call either, it's an implementation detail.

src/node/plugin.ts
nhooyr: Fair enough!

typings/plugin.d.ts
nhooyr: I'm confused. This has no semantic difference from what was in the spec. Returning a router is less code than duplicating the express API in our interface.

src/node/plugin.ts
nhooyr: That's fair. Perhaps I should move name back into the Plugin interface then.

typings/plugin.d.ts
nhooyr: So talked to asher and only vs code has routes at the top level for backwards compatibility and we'll add redirects to code-server directly for those.

For everything else, under a subpath is the way to go imo.

src/node/plugin.ts
nhooyr: Actually moving this into the plugin interface. I think only the version should be taken from `package.json`.

src/node/plugin.ts
nhooyr: I what would you prefer instead?

code-asher: I've also seen the pattern `loadPlugin` and `doLoadPlugin` although idk if I like that better personally. One idea is to move the `_` methods internal to the others as `const` arrow functions (potentially even dropping the `_`) which might help make more clear what they're for.

test/test-plugin/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "private": true,
+  "name": "test-plugin",
+  "version": "1.0.0",
+  "description": "Fake plugin for testing code-server's plugin API",
+  "engines": {
+    "code-server": "^3.6.0"
+  },
+  "main": "out/index.js",
```
code-asher, line 9: Any thoughts on how we should build this before running `yarn test`? What comes to mind for me is:

- Require you `cd` in and build it first (maybe with an entry in `CONTRIBUTING.md`)
- Do it automatically as part of the `yarn test` script
- Remove `main` and move index.ts to the root of the plugin so it works with both Typescript directly or with Javascript

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
-    logger.debug(
-      "Loaded plugin",
-      field("name", packageJson.name || path.basename(pluginPath)),
-      field("path", pluginPath),
-      field("version", packageJson.version || "n/a"),
-    )
-  } catch (error) {
-    logger.error(error.message)
+export class PluginAPI {
+  private readonly plugins = new Array<Plugin>()
```
code-asher, line 34: Why `new Array` instead of `[]`?

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
+  private readonly logger: Logger
+
+  public constructor(
+    logger: Logger,
```
code-asher, line 38: Is there an advantage to passing this in over `import`ing it?

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
+   */
+  public async applications(): Promise<Application[]> {
+    const apps = new Array<Application>()
+    for (const p of this.plugins) {
+      const pluginApps = await p.applications()
+
+      // Add plugin key to each app.
+      apps.push(
+        ...pluginApps.map((app) => {
+          return { ...app, plugin: p }
```
code-asher, line 60: I'm not sure this really matters but it feels a bit strange to me that the whole plugin is here (even though it's not on the types). My thinking is that it might cause some unexpected behavior (for example like how you had to `stringify` in the tests).

src/node/plugin.ts
```diff
@@ -1,92 +1,189 @@
...
-    // directory specifically for plugins. This lets you load plugins that are
-    // on the same level as other directories that are not plugins (if you tried
-    // to use CS_PLUGIN_PATH code-server would try to load those other
-    // directories as plugins). Intended for development.
-    ...plugin
-      .split(":")
-      .filter((p) => !!p)
-      .map((dir) => loadPlugin(path.resolve(dir), app, args)),
-  ])
+    await this._loadPlugins(path.join(__dirname, "../../plugins"))
```
code-asher, line 82: Awwww I liked this code. :cry: Is this to load plugins in a deterministic order?

typings/plugin.d.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * code-server uses this to collect the list of applications that
+   * the plugin can currently provide.
+   * It is called when /api/applications is hit or the overlay needs to
+   * refresh the list of applications
+   *
+   * Ensure this is as fast as possible.
+   */
+  applications(): Application[] | Promise<Application[]>
```
kylecarbs, line 77: Why is this sometimes a promise?

src/node/plugin.ts
nhooyr: The type is in the `new Array<Plugin>` here. Can switch if you prefer the other style. I'm not sure which to prefer.

src/node/plugin.ts
nhooyr: Yes, it  can be named for tests for example. i.e test specific logging at some point.

Can revert if you'd prefer importing.

src/node/plugin.ts
nhooyr: Fair enough! I'll be explicit here.

src/node/plugin.ts
nhooyr: Yes!

typings/plugin.d.ts
nhooyr: See above discussion with Will. Not all plugins need an async applications function.

nhooyr: e.g. vscode

test/test-plugin/package.json
nhooyr: My bad, forgot to build this before `yarn test`.

The reason it has its own build process is that the top level `tsconfig.json` has its top level dir as `src`. But I'll try add it there instead if possible.

src/node/plugin.ts
nhooyr: Switching to a map anyway.

src/node/plugin.ts
nhooyr: Added some comments here anyway.

nhooyr:
So addressed every comment + added `/api/applications` endpoint + plugins/apps now have a homepage field.

Should be good to merge now!

nhooyr:
Example endpoint output (also in docs):

```json
[
  {
    "name": "Test App",
    "version": "4.0.0",
    "iconPath": "/test-plugin/test-app/icon.svg",
    "path": "/test-plugin/test-app",
    "description": "This app does XYZ.",
    "homepageURL": "https://example.com",
    "plugin": {
      "name": "test-plugin",
      "version": "1.0.0",
      "modulePath": "/Users/nhooyr/src/cdr/code-server/test/test-plugin",
      "displayName": "Test Plugin",
      "description": "Plugin used in code-server tests.",
      "routerPath": "/test-plugin",
      "homepageURL": "https://example.com"
    }
  }
]
```

src/node/plugin.ts
code-asher: Sounds good!

I've seen some weird behavior with object wrappers in comparison with literals and I know Crockford advises against them (and recommended they be deprecated at one point) so I've always avoided them but idk if it really matters in the end.

src/node/plugin.ts
code-asher: Nah, that's fair.

src/node/plugin.ts
```diff
@@ -1,92 +1,228 @@
...
+            `ignoring duplicate plugin ${q(p.name)} at ${q(dir)}, using previously loaded ${q(p.modulePath)}`,
+          )
+          return
+        }
+      }
+      const p = this._loadPlugin(dir, packageJSON)
+      this.plugins.set(p.name, p)
+    } catch (err) {
+      if (err.code !== "ENOENT") {
+        this.logger.warn(`failed to load plugin: ${err.message}`)
```
wbobeirne, line 157: Should include the full stack trace, running into this and it's quite difficult to debug without it

src/node/plugin.ts
```diff
@@ -1,92 +1,228 @@
...
+   * loads the plugin now that we know there is no duplicate
+   * and that the package.json has been read.
+   */
+  private _loadPlugin(dir: string, packageJSON: PackageJSON): Plugin {
+    dir = path.resolve(dir)
+
+    const logger = this.logger.named(packageJSON.name)
+    logger.debug("loading plugin", field("plugin_dir", dir), field("package_json", packageJSON))
+
+    if (!semver.satisfies(version, packageJSON.engines["code-server"])) {
```
wbobeirne, line 173: This errors out in a hard to understand way if you don't specify an `engines` in your package.json:
```
[2020-11-04T22:14:07.857Z] warn  pluginapi failed to load plugin: TypeError: Cannot read property 'code-server' of undefined
```

test/test-plugin/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "private": true,
+  "name": "test-plugin",
+  "version": "1.0.0",
+  "engines": {
+    "code-server": "^3.6.0"
```
wbobeirne, line 6: This annoyingly warns on every npm command, perhaps we should use a custom key?
```
warning code-server-example-plugin@0.0.1: The engine "code-server" appears to be invalid.
```

wbobeirne (changes requested):
Started creating a test plugin for this and ran into some quirks.

I also think this may be going too hard on the plugins === applications angle, but we can loosen that up later if something like an rsync plugin or other "headless" plugin became useful to consider.

typings/pluginapi.d.ts
```diff
@@ -0,0 +1,183 @@
...
+ *
+ * Plugins are required as soon as they are found and then initialized.
+ * See the Plugin interface for details.
+ *
+ * If two plugins are found with the exact same name, then code-server will
+ * use the first one and emit a warning.
+ *
+ */
+
+/* Programmability
```
code-asher, line 48: To match the other blocks:
```suggestion
/**
 * Programmability
```

code-asher (approved):
:+1:

test/test-plugin/src/index.ts
```diff
@@ -0,0 +1,34 @@
+import * as express from "express"
+import * as fspath from "path"
+import * as pluginapi from "../../../typings/pluginapi"
+
+export const displayName = "Test Plugin"
+export const routerPath = "/test-plugin"
```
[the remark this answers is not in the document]
nhooyr: I like the `/` as when you glance over the code you see it as a path. I'll modify plugin loading to require at least two characters and a prefix of `/` for the router path.

test/test-plugin/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+export const homepageURL = "https://example.com"
+export const description = "Plugin used in code-server tests."
```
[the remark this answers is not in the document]
nhooyr: Yea I agree it's weird. Is there any way to type a bunch of your top level exports?

If not, I'll just make the top level export a single `plugin` and then that'll be easy.

test/test-plugin/package.json
nhooyr: So this is how vscode does things as well. I'm leaning in favour of keeping it the same to keep with the idiom but also ok changing to our own key.

cc @code-asher

typings/pluginapi.d.ts
nhooyr: Embarrasing for me tbh

nhooyr:
> I also think this may be going too hard on the plugins === applications angle, but we can loosen that up later if something like an rsync plugin or other "headless" plugin became useful to consider.

Agreed things can definitely easily change later. I'm not sure what would be useful to consider or how it'd work so I can't design for them. I do think that they should be able to easily work with just the `init` function, report zero applications and register zero routes. Perhaps we'd need another field to indicate they provide no applications/are headless so we can indicate that in the overlay.

test/test-plugin/package.json
wbobeirne: Cool, no worries if it's already convention

test/test-plugin/src/index.ts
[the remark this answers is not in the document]
wbobeirne: There's not, to my knowledge.

src/node/plugin.ts
```diff
@@ -1,92 +1,243 @@
...
-export const loadPlugins = async (app: Express, args: Args): Promise<void> => {
-  const pluginPath = process.env.CS_PLUGIN_PATH || `${path.join(paths.data, "plugins")}:/usr/share/code-server/plugins`
-  const plugin = process.env.CS_PLUGIN || ""
-  await Promise.all([
+  /**
+   * mount mounts all plugin routers onto r.
+   */
+  public mount(r: express.Router): void {
+    for (const [, p] of this.plugins) {
+      r.use(`/${p.routerPath}`, p.router())
```
wbobeirne, line 89: This makes it accessible on e.g. `//test-plugin`, not `/test-plugin`.
```suggestion
      r.use(`${p.routerPath}`, p.router())
```

wbobeirne (changes requested):

test/test-plugin/package.json
code-asher: It does seem to be a convention in use, as mentioned VS Code does it this way and I could have sworn I've seen it elsewhere although the details escape me. `--ignore-engines` can ignore it; I wonder if that's a valid setting for a `.yarnrc` or something.

code-asher:
We could have just `init` that returns all the info and optionally a router and an applications func (or pass `init` something that lets you do something like `api.registerRoute()`) then if it doesn't have a router skip the `use` and if it doesn't have `applications` skip it when getting all the apps.

src/node/plugin.ts
nhooyr: Ah interesting, I figured express would normalize it.

nhooyr:
How about we just make the applications and router callback optional? I prefer that to putting things into `init` as it's not really something that will change per init, it's something the plugin either implements or doesn't.

PR #2255, opened 2020-10-30
Fix self signed certificate for iPad

### LINKED ISSUE #1778: Document that generated self signed certificates go into the data directory
1. ~Users can specify paths of the cert and its key in `config.yaml` if HTTPS is enabled, instead of using `--cert` and `--cert-key`.~
2. ~In the `doc` directory, I can't find a documentation explaining entries in `config.yaml`. Luckily, now the `config.yaml` has a very small number of entries. However, the lack of documentation is always confusing, especially after the file becomes complicated in the future.~ Document that generated self signed certificates go into the data directory

BTW, I have another question: When HTTPS is enabled but the `--cert` argument is not used, which certificate will be used?

Thanks.

nhooyr:
See composing commits.

### DIFF
--- a/doc/ipad.md
+++ b/doc/ipad.md
@@ -3,13 +3,21 @@
 # iPad
 
 - [iPad](#ipad)
-  - [How to access code-server with a self signed certificate on iPad](#how-to-access-code-server-with-a-self-signed-certificate-on-ipad)
+  - [Known Issues](#known-issues)
+  - [How to access code-server with a self signed certificate on iPad?](#how-to-access-code-server-with-a-self-signed-certificate-on-ipad)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 # iPad
 
-## How to access code-server with a self signed certificate on iPad
+## Known Issues
+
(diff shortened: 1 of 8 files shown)


nhooyr:
Second commit here is very juicy!

doc/ipad.md
```diff
@@ -0,0 +1,48 @@
...
+
+## How to access code-server with a self signed certificate on iPad
+
+Accessing a self signed certificate on iPad isn't as easy as accepting through all
+the security warnings. Safari will prevent WebSocket connections unless the certificate
+is installed as a profile on the device.
+
+The below assumes you are using the self signed certificate that code-server
+generates for you. If not, that's fine but you'll have to make sure your certificate
+abides by the following guidelines from apple: https://support.apple.com/en-us/HT210176
```
code-asher, line 28:
```suggestion
abides by the following guidelines from Apple: https://support.apple.com/en-us/HT210176
```

code-asher (approved):
This is :fire:, wonderful commit separation too.

PR #2258, opened 2020-10-31
Fix path

mberbero:
Fixing rsync path


code-asher:
I don't believe this is the correct path, see: https://github.com/cdr/code-server/tree/v3.6.1/src/browser/media

mberbero:
> I don't believe this is the correct path, see: https://github.com/cdr/code-server/tree/v3.6.1/src/browser/media

Fix for this cdr:handlebars-templates branch.
https://github.com/cdr/code-server/tree/handlebars-templates

code-asher:
Ohhh I didn't check the target branch name! Cool, I'll come back to this
after I've reviewed the handlebars PR.

code-asher:
Thanks again for this! We ended up changing the directory altogether during code review and I think the rsync path in that branch is correct now.

Issue #1791, opened 2020-06-09
Silently attempt automatic reconnect before displaying popup for reload.


GrahamDumpleton:
Right now if the web socket connection to the backend is lost, a popup will be displayed looking like:

![image](https://user-images.githubusercontent.com/507637/84211117-e8f46500-aafd-11ea-820a-eb176fde7fc8.png)

If you don't click on any buttons it will disappear after about five seconds and reconnect anyway.

This behaviour of displaying the popup immediately when the connection is lost is very frustrating with certain deployments.

In particular when code-server is deployed to Kubernetes, where Kubernetes is using nginx as the ingress controller, this can result in the popup displayed in the worst case about every 4 to 5 minutes.

The reason for this is that every time a new ingress is added to the Kubernetes cluster, nginx has to reload the configuration, and nginx in doing a configuration reload will shutdown any active connections after a grace period. That default grace period is 240 seconds.

This means in a very active Kubernetes cluster, such as one hosting a training portal or workspace environment where the code-server editor is used and people come and go very frequently, all code-server sessions on the cluster get a stream of these popups occurring, which can get frustrating quite quickly when people are active in the editor a lot of the time.

Suggestion as a new feature therefore is to change the behaviour such that if the connection is lost, instead of immediately displaying the popup, attempt to reconnect with a delay of 100-250ms, with retries up to a period of 1 second. Only after multiple failures to reconnect when the 1 second expires, then display the popup.

This means reconnection can be handled transparently without the user needing to know, in cases where the loss of the connection is transient due to a reload in any routing layer in front of the backend. Users will therefore not be interrupted by the popup all the time.

If concerned about there being no indication at all that a reload was performed, then add a little icon in the bottom status bar which indicates connection state. When the connection is lost first change the status icon to indicate loss of connection. If the automatic retries within the 1 second period work, then set the status icon back to things being okay. If the 1 second expires then display the popup.

Since can't seem to find any settings in code-server which otherwise control the behaviour around reconnects, the only other option to reduce the annoyance of the popups that know of is to suggest to people to override the Kubernetes ingress controller configuration globally for the cluster to increase the grace periods from 240s to a longer duration.

* https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#worker-shutdown-timeout

[...]

PR #2284, opened 2020-11-09
Implement background reconnection

mgmachado:
Issue this PR solves: #1791

On a connection loss, the client attempts to reconnect up to twice in the background before displaying the reconnection pop-up.

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -689,7 +689,7 @@ index 3715cbb8e6ee41c3d9b5090918d243b723ae2d00..c65de8ad37e727d66da97a8f8b170cbc
 -
 -
 diff --git a/src/vs/platform/remote/common/remoteAgentConnection.ts b/src/vs/platform/remote/common/remoteAgentConnection.ts
-index 18d3d04fd20335975293e37b3b641120dd92da20..4e49f9d63623da6c84624144765f76ec127ea526 100644
+index 18d3d04fd20335975293e37b3b641120dd92da20..e072fc38ccd950462f42fdf112c8e7e673f351f5 100644
 --- a/src/vs/platform/remote/common/remoteAgentConnection.ts
 +++ b/src/vs/platform/remote/common/remoteAgentConnection.ts
 @@ -92,7 +92,7 @@ async function connectToRemoteExtensionHostAgent(options: ISimpleConnectionOptio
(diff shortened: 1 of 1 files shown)


ci/dev/vscode.patch
```diff
@@ -689,6 +689,53 @@ index 18d3d04fd20335975293e37b3b641120dd92da20..4e49f9d63623da6c84624144765f76ec
  			(err: any, socket: ISocket | undefined) => {
  				if (err || !socket) {
  					options.logService.error(`${logPrefix} socketFactory.connect() failed. Error:`);
+@@ -411,16 +411,23 @@ abstract class PersistentConnection extends Disposable {
+ 		}
+ 		const logPrefix = commonLogPrefix(this._connectionType, this.reconnectionToken, true);
+ 		this._options.logService.info(`${logPrefix} starting reconnecting loop. You can get more information with the trace log level.`);
+-		this._onDidStateChange.fire(new ConnectionLostEvent());
```
nhooyr, line 696: Only concern I have is whether it's ok to not fire this here. Maybe it prevents VS Code from trying to do things on the connection?

But I'm not sure. Just something to note. cc @code-asher

nhooyr (changes requested):
Thank you for contributing @mgmachado, this looks really awesome!

🔥 

Going to need approval from @code-asher

code-asher (changes requested):
Thank you for the PR!

To echo @nhooyr's comment, there can be some unintended consequences if we modify the underlying events.

For example, we rely on these events for our Node proxy service and this could break the service because we need to re-establish the proxy when the connection goes down and comes back up but if that happens in less than three attempts we won't get the necessary events.

Instead we should modify the popup itself, I think. It looks like it's in /lib/vscode/src/vs/workbench/contrib/remote/browser/remote.ts.

Potential implementation: keep a counter that increments each time `ReconnectionWait` happens and resets to zero every time `ConnectionGain` happens then use that to determine whether to show the dialogs.

Another potential solution is to change `ProgressLocation.Dialog` to `ProgressLocation.Notification` so it appears in the bottom right corner instead of in the middle where it hijacks everything. The advantage here is that it lets the user be more aware of when the connection goes down while still being able to use the editor normally but there's already a persistent connection status in the bottom left so it's probably not a big deal. I'm good either way.

ci/dev/vscode.patch
choyrim: @nhooyr @mgmachado @code-asher 
a search in the vscode repo indicates 3 subscribers to `PersistentConnectionEventType`:
- `RemoteAgentConnectionStatusListener` - [github link](https://github.com/microsoft/vscode/blob/f3a23697a90f49276cdb4f5531e931f9d6134086/src/vs/workbench/contrib/remote/browser/remote.ts#L770-L821)
- `RemoteExtensionHostEnvironmentUpdater` - [github link](https://github.com/microsoft/vscode/blob/8fcf778ba251131780f2a110d311161f39b97b03/src/vs/workbench/contrib/remote/electron-sandbox/remote.contribution.ts#L82-L91)
- `ExtensionService` - [github link](https://github.com/microsoft/vscode/blob/522391c3b5a04e2b31ca2a03d31658b79c4b8ba3/src/vs/workbench/services/extensions/electron-browser/extensionService.ts#L317-L324)

The `ExtensionService` does act on the `ConnectionLost` event - so that it can `_clearResolvedAuthority`. The `RemoteExtensionHostEnvironmentUpdater` is only interested in the `ConnectionGain` event - also related to the remote authority.

An alternative approach could be to modify `RemoteAgentConnectionStatusListener` to handle the events differently. In the same file, it seems like changing [VisibleProgress](https://github.com/microsoft/vscode/blob/f3a23697a90f49276cdb4f5531e931f9d6134086/src/vs/workbench/contrib/remote/browser/remote.ts#L609-L675) with some sort of delayed activation might be cleaner (after a name change perhaps).

mgmachado:
@choyrim @nhooyr @code-asher thank you for the feedback. I agree that modifying the behavior within the popup is a better approach than suppressing the events.

Should I convert this PR to draft while I work on the suggestions posted here?

nhooyr:
> Should I convert this PR to draft while I work on the suggestions posted here?

tbh, I don't really have a preference. I believe the conversion to draft is a somewhat new feature. I suppose it makes sense in this case to indicate this is a WIP.

mgmachado:
Hey code server devs, I implemented a better solution, but it caused a merge check to fail. I'm working on resolving that - it's the one that tests that the release standalone image can install python and jupyter extensions correctly.

nhooyr:
> Hey code server devs, I implemented a better solution, but it caused a merge check to fail. I'm working on resolving that - it's the one that tests that the release standalone image can install python and jupyter extensions correctly.

I believe that failure is unrelated. Feel free to push your changes up and just make sure you're rebased on `master`.

mgmachado:
The suggestion to convert `Dialog` to `Notification` in the ConnectionGain event handler in remote.ts was easy to implement. However, this caused the notifications to keep popping up in the bottom left. This might be ok if the socket drop happens every now and then, but it's distracting when socket drops happen more frequently (in a busy K8s cluster or when accessed  through a flaky VPN).

The behavior I've implemented now is: distraction free reconnects the first 2 times, and a Dialog popup at the third attempt to bring the re-connection failure to the user's attention, without suppressing the firing of the connection events.

choyrim:
@nhooyr @code-asher do you still have concerns about this PR after the refactoring?

or do you need us to resolve the latest conflicts before you can review?

Let us know.

nhooyr:
No this looks great! I'll test it today and if all works well, it'll be in the next release!

ci/dev/vscode.patch
```diff
@@ -3199,6 +3275,84 @@ index 94e7e7a4bac154c45078a1b5034e50634a7a43af..8164200dcef1efbc65b50eef9c270af3
...
+ 					disposableListener = null;
+ 				}
++				let suppressPopup = (typeof e.connectionAttempt == 'number' && e.connectionAttempt < SHOW_POPUP_ON_ATTEMPT)
++				let forceDialog = (typeof e.connectionAttempt == 'number' && e.connectionAttempt == SHOW_POPUP_ON_ATTEMPT)
+ 				switch (e.type) {
+ 					case PersistentConnectionEventType.ConnectionLost:
+-						if (!visibleProgress) {
+-							visibleProgress = showProgress(ProgressLocation.Dialog, [reconnectButton, reloadButton]);
++						if (suppressPopup) {
++							hideProgress()
```
nhooyr, line 3301: Why are these calls necessary? Wouldn't we not have made progress visible if the connection attempt was lower than `SHOW_POPUP_ON_ATTEMPT`?

nhooyr:
Other than that last comment, this looks great!

Thanks for contributing @mgmachado @choyrim

Greatly appreciated 🚀

ci/dev/vscode.patch
mgmachado: @nhooyr Did you mean "Wouldn't we not have made progress visible if the connection attempt was **greater** than `SHOW_POPUP_ON_ATTEMPT`?" Then yes, these calls are redundant. The code could be simplified to 
```typescript
		switch (e.type) {
			case PersistentConnectionEventType.ConnectionLost:
				hideProgress();
				break;
```
Or did you mean something else?

nhooyr: Yes sorry, I didn't mean to add the extra `not` in there.

Yup in that case this looks good to me, will merge. I'll do a follow up commit later to simplify what I can.

Thanks again!

mgmachado: Thanks a lot for the feedback and accepting the PR!

PR #2300, opened 2020-11-13
node/routes: Fix error handling

### LINKED ISSUE #2297: Express does not use error handler
Since updating to Express 5 our error pages are no longer used.

nhooyr:
We should always send HTML if the user agent expects it.

If they do not, they should clearly indicate such via the Accept header.

Closes #2297


code-asher (approved):

nhooyr:
Removed the fixup in master!

PR #2048, opened 2020-09-02
Update Kubernetes Helm Chart

Matthew-Beckett:
Fixes #2012, #914

PR #2303, opened 2020-11-13
Move helm from root and fix stuff in README

nhooyr:
See @sreya's review in #2048 

Only thing I did not add is templating the mount directory but we can wait for someone else to contribute that at some point.

Also, I fixed the examples for helm v3 which does not have `--name`.

cc @alexgorbatchev @Matthew-Beckett


ci/helm-chart/README.md
```diff
@@ -10,8 +10,8 @@ This chart is community maintained by [@Matthew-Beckett](https://github.com/Matt
 ## TL;DR;
 
 ```console
-$ git clone https://github.com/cdr/code-server.git
-$ helm install code-server/charts/code-server
+$ git clone https://github.com/cdr/code-server
+$ helm install code-server code-server/ci/helm-chart
```
sreya, line 14: what about just `/ci/helm` the `chart` suffix seems implicit

ci/helm-chart/README.md
```diff
@@ -97,9 +98,9 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 install`. For example,
 
 ```console
-$ helm install --name my-release \
-  --set persistence.enabled=false \
-    deployment/chart
+$ helm install code-server \
```
sreya, line 101: we could potentially do `helm upgrade --install` to both install and upgrades

sreya (approved):

nhooyr:
Moving this to v3.6.4 now, I don't think it's wise to merge as is. I'm not sure if this chart is for helm v2 or v3 now given the use of `--name` in the examples.

sreya:
@nhooyr imo we should really only promote use of v3

nhooyr:
Actually moved the release to tomorrow instead.

ci/helm-chart/README.md
nhooyr: For me at least it helps to be clear that the directory itself is the chart. It's not like a bunch of scripts to generate the chart or anything.

Matthew-Beckett:
@nhooyr this is a Helm v3 compatible chart.

nhooyr:
> @nhooyr imo we should really only promote use of v3.

Absolutely!

But since the examples use `--name`, I wonder if @Matthew-Beckett used v2 to create it. Want to give him a chance to clarify and perhaps we can use `v3` to generate the base which has improvements.

ci/helm-chart/README.md
Matthew-Beckett: I agree, in most other repositories that maintain charts they are usual in a charts directory at the root, it's unusual for it to be called CI, it might be confusing since it's not build scripts or CI related wizardry.

Matthew-Beckett:
> > @nhooyr imo we should really only promote use of v3.
> 
> Absolutely!
> 
> But since the examples use `--name`, I wonder if @Matthew-Beckett used v2 to create it. Want to give him a chance to clarify and perhaps we can use `v3` to generate the base which may have improvements.

I used v3 to generate the template and cherry-picked the code-server related elements from the Helm Chart that was deprecated.

nhooyr:
Ok awesome! Then yea this should be good to go!

ci/helm-chart/README.md
nhooyr: > I agree, in most other repositories that maintain charts they are usual in a charts directory at the root, it's unusual for it to be called CI, it might be confusing since it's not build scripts or CI related wizardry.
 
At some point we will want to automate publishing the helm chart so it'll at least be somewhat used in CI.

We tend to keep all our release related stuff in the `ci` directory to prevent root pollution. In terms of discovery, I think the link I added in `install.md` should be enough.

nhooyr:
Awesome, thanks again @alexgorbatchev @Matthew-Beckett @sreya <3

PR #2331, opened 2020-11-18
Fix log message

piousdeer:


code-asher (approved):
Nice catch! Thank you for the fix!

PR #2334, opened 2020-11-18
Separate process wrappers and pass arguments

### LINKED ISSUE #2316: Env var PASSWORD is not honored in v3.7.1
- `uname -a`: `Linux arch2011 5.9.6-arch1-1 #1 SMP PREEMPT Thu, 05 Nov 2020 21:00:46 +0000 x86_64 GNU/Linux`
- `code-server --version`: `3.7.1 2bf91ff6a6d1f29a8a93f97684177ccf5388cf34`

Environment variable `PASSWORD` is not honored any more in code-server v3.7.1:

```
$ PASSWORD=abcdefg code-server
[2020-11-17T00:18:41.502Z] info  code-server 3.7.1 2bf91ff6a6d1f29a8a93f97684177ccf5388cf34
[2020-11-17T00:18:41.506Z] info  Using user-data-dir ~/.local/share/code-server
[2020-11-17T00:18:41.515Z] info  Using config file ~/.config/code-server/config.yaml
[2020-11-17T00:18:41.516Z] info  HTTP server listening on http://127.0.0.1:8080
[2020-11-17T00:18:41.516Z] info    - Authentication is enabled
[2020-11-17T00:18:41.516Z] info      - Using password from ~/.config/code-server/config.yaml
[2020-11-17T00:18:41.516Z] info    - Not serving HTTPS
```

Compared to v3.6.2:

```
$ PASSWORD=abcdefg code-server
[2020-11-17T00:20:34.807Z] info  Using user-data-dir ~/.local/share/code-server
[2020-11-17T00:20:34.812Z] info  code-server 3.6.2 9bde62fbd611a7a91c5f327fa43e0d06f1379169
[2020-11-17T00:20:34.813Z] info  Using config file ~/.config/code-server/config.yaml
[2020-11-17T00:20:34.817Z] info  HTTP server listening on http://127.0.0.1:8080
[2020-11-17T00:20:34.818Z] info      - Using password from $PASSWORD
[2020-11-17T00:20:34.818Z] info      - To disable use `--auth none`
[2020-11-17T00:20:34.818Z] info    - Not serving HTTPS
```

Content of `config.yaml` is auto generated:

```yml
$ cat ~/.config/code-server/config.yaml
bind-addr: 127.0.0.1:8080
auth: password
password: 28cc3d30b23383e05b070ea0
cert: false
```

Even if I delete `config.yaml` manually and then start code-server, it won't work as code-server will auto generate the config again.

```
[2020-11-17T00:17:13.039Z] info  Wrote default config file to ~/.config/code-server/config.yaml
```

code-asher:
This is to fix #2316.

Previously we only deleted `PASSWORD` in the child process but now it's deleted in the parent (which I think is good since that means it'll be hidden from the VS Code CLI spawn as well and anything else we might spawn from the parent in the future). 

But as a consequence the child no longer has access to `PASSWORD` and can't use it when parsing the arguments. This PR fixes that by passing the arguments from the parent to the child.

As a bonus this means we only have to do the parsing once, including if it gets restarted via a SIGUSR1.

Since it was getting a little convoluted with all the if statements I split out the child and parent process wrappers which I think makes it all a little more clear. Also was able to re-use `onMessage` from the VS Code wrapper.

### DIFF
--- a/src/node/wrapper.ts
+++ b/src/node/wrapper.ts
@@ -1,8 +1,9 @@
-import { Logger, field, logger } from "@coder/logger"
+import { field, Logger, logger } from "@coder/logger"
 import * as cp from "child_process"
 import * as path from "path"
 import * as rfs from "rotating-file-stream"
 import { Emitter } from "../common/emitter"
+import { DefaultedArgs } from "./cli"
 import { paths } from "./util"
 
 const timeoutInterval = 10000 // 10s, matches VS Code's timeouts.
@@ -58,7 +59,12 @@ export function onMessage<M, T extends M>(
   })
 }
 
-interface HandshakeMessage {
+interface ParentHandshakeMessage {
+  type: "handshake"
+  args: DefaultedArgs
+}
+
+interface ChildHandshakeMessage {
   type: "handshake"
 }
 
@@ -67,7 +73,8 @@ interface RelaunchMessage {
   version: string
 }
 
-type Message = RelaunchMessage | HandshakeMessage
+type ChildMessage = RelaunchMessage | ChildHandshakeMessage
+type ParentMessage = ParentHandshakeMessage
 
 class ProcessError extends Error {
   public constructor(message: string, public readonly code: number | undefined) {
@@ -164,15 +171,16 @@ class ChildProcess extends Process {
   /**
    * Initiate the handshake and wait for a response from the parent.
    */
-  public async handshake(): Promise<void> {
+  public async handshake(): Promise<DefaultedArgs> {
     this.send({ type: "handshake" })
-    await onMessage<Message, HandshakeMessage>(
+    const message = await onMessage<ParentMessage, ParentHandshakeMessage>(
       process,
-      (message): message is HandshakeMessage => {
+      (message): message is ParentHandshakeMessage => {
         return message.type === "handshake"
       },
       this.logger,
     )
+    return message.args
   }
 
   /**
@@ -185,7 +193,7 @@ class ChildProcess extends Process {
   /**
    * Send a message to the parent.
    */
-  private send(message: Message): void {
+  private send(message: ChildMessage): void {
     if (!process.send) {
       throw new Error("not spawned with IPC")
     }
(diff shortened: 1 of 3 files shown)


src/node/wrapper.ts
```diff
@@ -256,18 +266,52 @@ export class WrapperProcess {
...
+/**
+ * Process wrapper.
+ */
+export const wrapper =
+  typeof process.env.CODE_SERVER_PARENT_PID !== "undefined"
+    ? new ChildProcess(parseInt(process.env.CODE_SERVER_PARENT_PID))
+    : new ParentProcess(require("../../package.json").version)
+
+export function isChild(proc: ChildProcess | ParentProcess): proc is ChildProcess {
+  return proc instanceof ChildProcess
```
nhooyr, line 353: Would it be better to just directly do `proc instanceof ChildProcess`?

src/node/wrapper.ts
```diff
@@ -111,19 +164,15 @@ class ChildProcess extends Process {
...
-          process.removeListener("message", onMessage)
-          resolve()
-        }
-      }
-      // Initiate the handshake and wait for the reply.
-      process.on("message", onMessage)
-      this.send({ type: "handshake" })
-    })
+  public async handshake(): Promise<void> {
+    this.send({ type: "handshake" })
```
nhooyr, line 175: Shouldn't we register the handler before sending? To avoid the message being lost.

src/node/wrapper.ts
```diff
@@ -253,21 +268,17 @@ export class ParentProcess extends Process {
...
-    // If we have a process then we've already bound this.
-    if (!this.child) {
-      process.on("SIGUSR1", async () => {
-        this.logger.info("Received SIGUSR1; hotswapping")
-        this.relaunch()
-      })
-    }
+  public start(args: DefaultedArgs): Promise<void> {
+    // Store for relaunches.
+    this.args = args
```
nhooyr, line 276: Why not put this in the constructor?

nhooyr (changes requested):

src/node/wrapper.ts
code-asher: I'm down for that, the main reason I didn't was so I didn't have to export `ChildProcess` since it won't be used anywhere else.

src/node/wrapper.ts
code-asher: I don't think it's possible since it's single-threaded so nothing will be able to run in between the send and when we attach the message listener. I'm down to change it just in case though!

src/node/wrapper.ts
code-asher: The only reason is because it's constructed in this file where we don't have args yet.

code-asher: Could definitely move the `new` parts into `entry.ts` and the `process.stdout.isTTY` and `uncaughtException` bits as well.

src/node/wrapper.ts
nhooyr: Fair! let's leave as is then.

src/node/wrapper.ts
nhooyr: All good.

src/node/wrapper.ts
nhooyr: Ah I gotchu, fine to leave as is then.

nhooyr (approved):

Issue #2329, opened 2020-11-18
Markdown previews not working


ctlajoie:
When I try to open a preview on a markdown file I get `command 'markdown.showPreviewToSide' not found`

I am not sure if it is relevant to this issue but I am also experiencing #1593, which was an issue I reported a while back. It had been fixed (known to be working in 3.4.1) but appears to have returned in recent builds and that issue was reopened.

- Web Browser: Chrome
- Local OS: Linux
- Remote OS: Linux
- Remote Architecture: x86_64
- `code-server --version`: 3.7.1 2bf91ff6a6d1f29a8a93f97684177ccf5388cf34

code-asher:
The markdown extension linked in that issue is still broken I believe, I
think it isn't being built correctly in our marketplace so we'll need to
take a look and see why.

The built-in markdown preview should work though. I just tried with
3.7.1 and Chrome (via right-click > open preview on the markdown file
tab) and it worked for me.

Are there any other errors in the browser console?

In my case even though the markdown preview is showing I'm still getting
a bunch of errors on the webview endpoint; that might be related.

ctlajoie:
I was able to get preview to load up after I updated chrome, which was several months out of date... my bad. I did run into those 404s though. Looks like you already have a PR for that.

PR #2336, opened 2020-11-18
Fix 404 webviews and tar endpoint

code-asher:
Related: #2329

### DIFF
--- a/src/node/routes/static.ts
+++ b/src/node/routes/static.ts
@@ -7,13 +7,33 @@ import * as tarFs from "tar-fs"
 import * as zlib from "zlib"
 import { HttpCode, HttpError } from "../../common/http"
 import { rootPath } from "../constants"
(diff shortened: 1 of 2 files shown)


src/node/routes/static.ts
```diff
@@ -7,13 +7,33 @@ import * as tarFs from "tar-fs"
...
 import { getMediaMime, pathToFsPath } from "../util"
 
 export const router = Router()
 
 // The commit is for caching.
 router.get("/(:commit)(/*)?", async (req, res) => {
+  // Used by VS Code to load extensions into the web worker.
+  const tar = Array.isArray(req.query.tar) ? req.query.tar[0] : req.query.tar
+  if (typeof tar === "string") {
+    ensureAuthenticated(req)
```
nhooyr, line 20: Def need to mention this bug in release notes.

nhooyr (approved):

PR #2348, opened 2020-11-20
Use file system for user data

code-asher:


ci/dev/vscode.patch
```diff
@@ -3797,18 +3797,25 @@ index 85d83f37da179a1e39266cf72a02e971f590308e..0659738b36df1747c9afcabf8d9abf26
...
+@@ -119,8 +119,25 @@ export class BrowserWorkbenchEnvironmentService implements IWorkbenchEnvironment
  	@memoize
  	get logFile(): URI { return joinPath(this.options.logsPath, 'window.log'); }
  
-+	// NOTE@coder: Use the regular path for extensions that write directly to disk
-+	// instead of using the VS Code API.
++	// NOTE@coder: Use the same path in // ../../../../platform/environment/node/environmentService.ts
++	// and don't use the user data scheme. This solves two problems:
++	//  1. Extensions running in the browser (like Vim) might use these paths
++	//     directly instead of using the file service and most likely can't write
```
nhooyr, line 3810: What do you mean by

> most likely can't write to `/User` on disk.

They can't at all right since they're running in the browser.

nhooyr (approved):
Tested and things work as expected now!

ci/dev/vscode.patch
code-asher: They can since client-side extensions run in our shimmed Node environment which calls out to the server for things like `fs.write` etc.

nhooyr: Oh gotchu.

PR #2358, opened 2020-11-24
vscode: Show notification when upgrade is available

nhooyr:

### DIFF
--- a/ci/dev/vscode.patch
+++ b/ci/dev/vscode.patch
@@ -769,10 +769,10 @@ index 096b9e23493539c9937940a56e555d95bbae38d9..ef37e614004f550f7b64eacd362f6894
  	remove(key: string, scope: StorageScope): void {
 diff --git a/src/vs/server/browser/client.ts b/src/vs/server/browser/client.ts
 new file mode 100644
-index 0000000000000000000000000000000000000000..c6eef331346ebc244a26e8b1e5919d192225b971
+index 0000000000000000000000000000000000000000..385b9da491d38a9f5d10fab6e4666c84a892f49d
 --- /dev/null
 +++ b/src/vs/server/browser/client.ts
-@@ -0,0 +1,237 @@
+@@ -0,0 +1,240 @@
 +import { Emitter } from 'vs/base/common/event';
 +import { URI } from 'vs/base/common/uri';
 +import { localize } from 'vs/nls';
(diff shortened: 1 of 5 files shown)


nhooyr:
Will add a ss in the AM.

nhooyr:
Screen Shot 2020-11-24 at 1 06 49 PM

frenck:
How can I disable this update check?

ci/dev/vscode.patch
```diff
@@ -922,8 +924,57 @@ index 0000000000000000000000000000000000000000..3c0703b7174ad792a4b42841e96ee937
 +		});
 +	}
 +
++	const logService = (services.get(ILogService) as ILogService);
++	const storageService = (services.get(IStorageService) as IStorageService);
++	// We set this here first in case the path changes.
++	const updateCheckEndpoint = path.join(window.location.pathname, "/update/check")
```
code-asher, line 930: I think we'll want to use `options.base` at some point to make this independent of the vscode endpoint (for example if you access it at `/vscode` instead of `/`).

nhooyr:
> How can I disable this update check?

Is there a reason you need to? code-server will never actually upgrade itself. It's only a notification. Perhaps we should make it a weekly/monthly notification instead?

ci/dev/vscode.patch
nhooyr: There is options.base anymore. It was removed.

frenck:
Basically 2 reasons:

- It is repacked in another Docker container as a Home Assistant add-on. All those releases are tested before releasing it to our userbase, your update cycle doesn't match ours. Neither does the versioning. This will be confusing to them.

- As for my personal system; I'm not interested in having it phone home every couple of hours for an update check.

nhooyr:
> It is repacked in another Docker container as a Home Assistant add-on. All those releases are tested before releasing it to our userbase, your update cycle doesn't match ours. Neither does the versioning. This will be confusing to them.
> I simply don't want a notification. I rather control the things I'm notified of. It is distracting while working actually.

Fair enough, I've opened https://github.com/cdr/code-server/issues/2361 for the next release.

> As for my personal system; I'm not interested in having it phone home every couple of hours for an update check.

It doesn't hit any server hosted by us. It just checks against the latest release on GitHub.

frenck:
Thank you!

ci/dev/vscode.patch
nhooyr: I'll add it back though.

futurist:
How to disable the update notification? I see the PR don't have any options/command-args to disable that!

PR #2381, opened 2020-11-30
vscode: Reconnect in the background up to 5 seconds

### LINKED ISSUE #1791: Silently attempt automatic reconnect before displaying popup for reload.

nhooyr:
Based on the previous commits by @mgmachado but simplified.

I also changed the threshold to error after a single attempt as the
connection has likely been borked and the user should be in the know if
they couldn't reconnect after 5 seconds.

Closes #1791


nhooyr:
Fixed!

mgmachado:
LGTM. Much simpler!

code-asher (approved):

PR #2389, opened 2020-12-02
Add coder-log path to FAQ's 'How do I debug...' section

csaska:
Previously, the _How do I debub issues with code-server?_ section of FAQ was referencing `~/.local/share/code-server/logs`. For the more recent verions of code-server, it appears this log directory was renamed to `~/.local/share/code-server/coder-logs`

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -232,7 +232,7 @@ code-server --log debug
 Once this is done, replicate the issue you're having then collect logging
 information from the following places:
 
-1. stdout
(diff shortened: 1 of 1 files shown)


code-asher:
Thank you for the PR!

We should actually probably reference both. `logs` contains the VS Code
logs while `coder-logs` contains a copy of code-server's `stdout` and
`stderr`.

Let's change `1.` to say something like:

```
The most recent files from `~/.local/share/code-server/coder-logs`.
```

code-asher (approved):
Thanks!!

PR #2398, opened 2020-12-04
Remove the "install Xcode" bit

ed789d0:
I have not tested this myself, but I am pretty sure you don't need the full Xcode installed to build native node modules, just the command line tools. https://github.com/nodejs/node-gyp/blob/master/macOS_Catalina.md#solutions


nhooyr (approved):
Agreed, great catch @ed789d0!

PR #2400, opened 2020-12-04
Use proxy-agent to support $HTTP_PROXY

### LINKED ISSUE #124: HTTP proxy environment variables support
vscode use http_proxy and https_proxy environment variables if the proxy setting not set. It seems not to work when using code-server.

nhooyr:
Closes #124

This works by monkey patching the http and https modules's default agent
at runtime to the one given by the proxy-agent package.


src/node/update.ts
```diff
@@ -106,12 +106,13 @@ export class UpdateProvider {
...
         const client = httpx.get(uri, { headers: { "User-Agent": "code-server" } }, (response) => {
           if (!response.statusCode || response.statusCode < 200 || response.statusCode >= 400) {
+            response.destroy()
             return reject(new Error(`${uri}: ${response.statusCode || "500"}`))
           }
 
           if (response.statusCode >= 300) {
-            ++redirects
             response.destroy()
+            ++redirects
```
kylecarbs, line 115: What's with these being moved? Feels strangely out of scope

kylecarbs (changes requested):

src/node/update.ts
nhooyr: @code-asher and I now separate things cleanly into separate commits so it's best to review our PRs in commit order. The reasoning for this is in commit #2 `update.ts: Fix response memory leak` and is for consistency with the `response.destroy` I added above.

ik it's random and can go into another PR but just reducing churn.

nhooyr:
Confirmed working, this should be good now!

src/node/proxy_agent.ts
```diff
@@ -0,0 +1,53 @@
...
+ * - lib/vscode/src/vs/base/node/proxy_agent.ts
+ * The second is a symlink to the first.
+ */
+
+/**
+ * monkeyPatch patches the node HTTP/HTTPS library to route all requests through our
+ * custom agent from the proxyAgent package.
+ */
+export function monkeyPatch(vscode: boolean): void {
+  if (!process.env.HTTP_PROXY) {
```
code-asher, line 22: Should we also take `HTTPS_PROXY`? VS Code also seems to use the lowercase versions, not sure if we should as well. https://github.com/microsoft/vscode/blob/1.51.1/src/vs/platform/request/node/proxy.ts#L11

code-asher (changes requested):
Tested here as well and everything looks swell. I'm ecstatic to have this implemented at long last!! Per-commit review is :fire: as usual. :smile:

ci/dev/vscode.patch
```diff
@@ -2644,7 +2644,7 @@ index 0000000000000000000000000000000000000000..fb7999ee343ddfd5f225a5ef727c9e55
 +		this.logger.trace('Spawning extension host...');
 +		const proc = cp.fork(
 +			FileAccess.asFileUri('bootstrap-fork', require).fsPath,
-+			[ '--type=extensionHost' ],
```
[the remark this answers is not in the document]
nhooyr: Will revert and add a comment!

src/node/proxy_agent.ts
nhooyr: `process.env` is case insensitive! I think uppercase is canonical for environment variables so I think we should prefer that.

Yea I suppose we can just copy what VS Code's doing here.

nhooyr: I might be confused actually, not sure if it's case sensitive or not.

nhooyr: Ah yes, only on windows. I guess I'll just copy the VS Code function for this.

nhooyr: Nvm, it's confusing. There's no difference between `HTTP_PROXY` and `HTTPS_PROXY` for us as `proxy-agent` just takes a single URL and automatically uses the right protocol for whatever is required. VS Code lets you separate the two protocols out but there's no real value in that as far as I can tell. So going to leave it as is with a comment and add the lowercase version.

code-asher (approved):

PR #2409, opened 2020-12-06
Add hashedPassword config

### LINKED ISSUE #2225: Password hash
Hi,

As far as I understand, the current password setup stores the password as plaintext in the user config file. It would probably be safer to have it stored as a hash and do a hash comparison upon login.

Is this feature already available? Else, would you accept a PR adding this behavior?

Thanks,
Best,

SPGoding:
Resolve #2225.

### DIFF
--- a/doc/guide.md
+++ b/doc/guide.md
@@ -297,6 +297,9 @@ and then restart `code-server` with:
 sudo systemctl restart code-server@$USER
 ```
 
+Alternatively, you can specify the SHA-256 of your password at the `hashedPassword` field in the config file.
+The `hashedPassword` field takes precedence over `password`.
+
 ### How do I securely access development web services?
 
 If you're working on a web service and want to access it locally, `code-server` can proxy it for you.
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -29,6 +29,7 @@ export interface Args extends VsArgs {
   config?: string
   auth?: AuthType
   password?: string
+  hashedPassword?: string
   cert?: OptionalString
   "cert-host"?: string
   "cert-key"?: string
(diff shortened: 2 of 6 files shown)


code-asher (approved):
Thank you for the PR! Works like a dream. :cloud:

nhooyr:
This broke CI on master unfortunately.

@jsjoeio @code-asher @oxy Github actions do not run on PRs from forks. Moving forward make sure `yarn ci` passes before merging!

nhooyr:
Never mind, I was wrong. It passed on CI, I'm not sure why CI is failing for me locally with:

```
$ ./ci/dev/lint.sh
src/node/routes/login.ts:33:23 - error TS2339: Property 'usingEnvHashedPassword' does not exist on type 'DefaultedArgs'.

33   } else if (req.args.usingEnvHashedPassword) {
                         ~~~~~~~~~~~~~~~~~~~~~~

src/node/routes/login.ts:71:16 - error TS2339: Property 'hashedPassword' does not exist on type 'DefaultedArgs'.

71       req.args.hashedPassword
                  ~~~~~~~~~~~~~~

src/node/routes/login.ts:72:57 - error TS2339: Property 'hashedPassword' does not exist on type 'DefaultedArgs'.

72         ? safeCompare(hash(req.body.password), req.args.hashedPassword)
                                                           ~~~~~~~~~~~~~~


Found 3 errors.
```

Maybe the typescript cache?

nhooyr:
Yup, things work perfectly after `yarn clean`. sorry for the noise.

ZacJW:
Why is this using SHA256 for password hashing. Dustin Moris Gorski has a [short write up](https://dusted.codes/sha-256-is-not-a-secure-password-hashing-algorithm) explaining why not to use SHA256 for password hashing. TLDR is SHA is designed to be fast and modern hardware can crunch them very quickly, which is the exact opposite of what you want for password hashing. A password hash should be as slow as is bearable to make brute forcing it as difficult as possible.

Maybe use bcrypt with a user configurable work factor as it fixes this problem. There is however also the issue that this password hash is stored in a file with default read access for every user on the system. Makes this choice of hash all the more problematic.

code-asher:
Makes a lot of sense. I'm down to use bcrypt.

PR #2422, opened 2020-12-08
doc: Switch screenshot to use Chrome

nhooyr:
Safari runs code-server horribly. We shouldn't even pretend that it sort
of works.

See the issues tagged browser-safari!


nhooyr:
You can see the difference live here: https://github.com/cdr/code-server/tree/ss-20c2

.github/CODEOWNERS
```diff
@@ -1,3 +1 @@
-* @code-asher @nhooyr
```
oxy, line 1: Not sure if removing CODEOWNERS here was intentional!

oxy (commented):
Screenshot looks good, not sure what happened with `.github/CODEOWNERS`

.github/CODEOWNERS
nhooyr: It was! We strive to break down each PR into individual commits. Makes it much easier to review. See the justification in the second commit.

Creating a new PR for every little change tends to get unwieldy.

nhooyr: But yea know what actually, I'll try to do that from now on. It's def confusing to mesh things together in unrelated PRs.

PR #2423, opened 2020-12-08
install.sh: Use $ID_LIKE to detect distro

### LINKED ISSUE #2343: Install script uses standalone for Amazon Linux instead of rpm
The install script picks the standalone version instead of the rpm based version when installing on an Amazon Linux 2 ec2 instance. I manually had to use rpm to install, which is included in Amazon Linux 2. The script should also use rpm so that the systemctl service is installed as well

Running: ` curl -fsSL https://code-server.dev/install.sh | sh -s -- --dry-run` gives the following output
```
Amazon Linux 2
Unsupported package manager.
Installing standalone release archive v3.7.2 from GitHub releases.

+ Reusing ~/.cache/code-server/code-server-3.7.2-linux-amd64.tar.gz
+ mkdir -p ~/.local/lib ~/.local/bin
+ tar -C ~/.local/lib -xzf ~/.cache/code-server/code-server-3.7.2-linux-amd64.tar.gz
+ mv -f ~/.local/lib/code-server-3.7.2-linux-amd64 ~/.local/lib/code-server-3.7.2
+ ln -fs ~/.local/lib/code-server-3.7.2/bin/code-server ~/.local/bin/code-server

Standalone release has been installed into ~/.local/lib/code-server-3.7.2
Please extend your path to use code-server:
  PATH="$HOME/.local/bin:$PATH"
Then you can run:
  code-server
```

oxy:
Update the install script to use `$ID_LIKE` where possible to detect similar distros.
Script now detects Amazon Linux as a .rpm-based distro; also de-duplicated checks for other distros.

Tested to work on:
- Alpine Linux, `ash`
- Amazon Linux, `bash`
- Arch Linux, `bash`
- Ubuntu, `dash`

Fixes #2343.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -238,10 +238,10 @@ main() {
   macos)
     install_macos
     ;;
-  ubuntu | debian | raspbian)
+  debian)
     install_deb
     ;;
-  centos | fedora | rhel | opensuse)
+  fedora | opensuse)
     install_rpm
     ;;
   arch)
@@ -425,14 +425,16 @@ os() {
 }
 
 # distro prints the detected operating system including linux distros.
+# Also parses ID_LIKE for common distro bases.
 #
 # Example outputs:
-# - macos
-# - debian, ubuntu, raspbian
-# - centos, fedora, rhel, opensuse
-# - alpine
-# - arch
-# - freebsd
+# - macos -> macos
+# - freebsd -> freebsd
+# - ubuntu, raspbian, debian ... -> debian
+# - amzn, centos, rhel, fedora, ... -> fedora
+# - opensuse-{leap,tumbleweed} -> opensuse
+# - alpine -> alpine
+# - arch -> arch
 #
 # Inspired by https://github.com/docker/docker-install/blob/26ff363bcf3b3f5a00498ac43694bf1c7d9ce16c/install.sh#L111-L120.
 distro() {
@@ -444,12 +446,15 @@ distro() {
   if [ -f /etc/os-release ]; then
     (
       . /etc/os-release
-      case "$ID" in opensuse-*)
-        # opensuse's ID's look like opensuse-leap and opensuse-tumbleweed.
-        echo "opensuse"
-        return
-        ;;
-      esac
+      if [ "${ID_LIKE-}" ]; then
+        for id_like in $ID_LIKE; do
+          case "$id_like" in debian | fedora | opensuse)
+            echo "$id_like"
+            return
+            ;;
+          esac
+        done
+      fi
 
       echo "$ID"
     )


nhooyr:
This looks awesome!

Though, your git email doesn't match your GitHub email. And also make sure to setup commit signing. https://docs.github.com/en/free-pro-team@latest/github/authenticating-to-github/signing-commits

install.sh
```diff
@@ -444,12 +444,14 @@ distro() {
   if [ -f /etc/os-release ]; then
     (
       . /etc/os-release
-      case "$ID" in opensuse-*)
-        # opensuse's ID's look like opensuse-leap and opensuse-tumbleweed.
-        echo "opensuse"
-        return
-        ;;
-      esac
+      if [ -n "${ID_LIKE-}" ]; then
```
nhooyr, line 447: No need for `-n` here.

install.sh
```diff
@@ -444,12 +444,14 @@ distro() {
...
+        for distro_like in $ID_LIKE; do
```
nhooyr, line 448: Let's use lowercase `id_like` here.

nhooyr (changes requested):

install.sh
oxy: `$ID_LIKE` is case-senstiive as it comes from `/etc/os-release` - perhaps we should uppercase `distro_like` instead?

nhooyr: it's idiomatic shell to use lowercase variables for function local variables. I should have clarified, `distro_like` should be renamed to `id_like`.

oxy: Ah, thanks for the clarification! And done! :+1:

install.sh
```diff
@@ -425,16 +425,16 @@ os() {
...
 #
 # Example outputs:
-# - macos
-# - debian, ubuntu, raspbian
-# - centos, fedora, rhel, opensuse
-# - alpine
-# - arch
-# - freebsd
-#
-# Inspired by https://github.com/docker/docker-install/blob/26ff363bcf3b3f5a00498ac43694bf1c7d9ce16c/install.sh#L111-L120.
```
nhooyr, line 437: Let's keep the inspired by line.

install.sh
```diff
@@ -425,16 +425,16 @@ os() {
 }
 
 # distro prints the detected operating system including linux distros.
+# also parses ID_LIKE for common distro bases
```
nhooyr, line 428:
```suggestion
# Also parses ID_LIKE for common distro bases.
```

nhooyr (changes requested):

nhooyr (approved):
🚀

oxy:
Just ran `yarn ci` - it reformatted the script, so I force-pushed it with changes - now it all passes, and we should be good!

nhooyr (approved):

Issue #1669, opened 2020-05-18
Allow to enter password in URL instead of form


momiji:
Before anything : **Fantastic work, impressive...**

I have a "small" improvement request...

We're several people using code-server on our servers, and most of the time we're using it under the same unix account. It is not problem as the port number is different for each user, as well as the generated random password.

We have a shell script that search for a free port, generates a random port, prints the URL to connect and the password, and finally starts code-server. The user just have to click on the link, and enter the password.

The idea of this improvement would be that it is no more necessary to enter the password and have it directly in the URL. This is not a security concerns in my use case, because the browser uses an SSH tunnel directly to the server.

For example, if I connect to my server `dev01.localhost` and I run my script, it gives me the url with a random port like `12345`: `http://dev01.localhost:12345` and the generated password `S3cr3t`.
I would like to be able to have only the url, like `http://dev01.localhost:12345/login?password=S3cr3t`

By the way, browsing to different SSH tunnels is very easy once you install a dynamic proxy extension. Personaly, I'm using SwitchyOmega for this in Chrome and Firefox. With the above example, I would add a rule like dev01.localhost to go through a SOCKS proxy on the port configured in the SSH connection. This is a top!

dmratcliffe:
Isn't there a way to disable the password all together?

Also, I feel a "one time login" would be a better implementation of this. I don't thinks its the best to drop a password in logs, even if its a one time use.

nhooyr:
@momiji If you're using an SSH tunnel you can disable password authentication with `--auth=none` or `auth: none` in the config file.

Issue #1285, opened 2020-01-20
How to put the password in the request url?


ladyrick:
For example, can I use http://127.0.0.1:8080/?password=balabala to login instead of input the password manually?

Jupyter notebook does it well: http://127.0.0.1:8080/?token=balabala

sr229:
If I remember, we use `Authorization`.

sr229:
We set the authentication via `Set-Cookie`. there's a value in there called `key=`. That's how we set your password in there. Look in your network requests under `/login`.

ladyrick:
But set-cookie is not so convenient for user to open it directly.

ladyrick:
> We set the authentication via `Set-Cookie`. there's a value in there called `key=`. That's how we set your password in there. Look in your network requests under `/login`.

I type the vscode-server url in address bar and open it.
After input the password, I can get access to it.

Then every time I open the url, password is not asked.

But, when I put the vscode-server url in a button
`<a href="http://10.20.30.40:9900" target="_blank">code-server</a>`
and click it to open vscode-server, the password is always required.

Why is that?

ladyrick:
> > We set the authentication via `Set-Cookie`. there's a value in there called `key=`. That's how we set your password in there. Look in your network requests under `/login`.
> 
> I type the vscode-server url in address bar and open it.
> After input the password, I can get access to it.
> 
> Then every time I open the url, password is not asked.
> 
> But, when I put the vscode-server url in a button
> `<a href="http://10.20.30.40:9900" target="_blank">code-server</a>`
> and click it to open vscode-server, the password is always required.
> 
> Why is that?

Solved.
It is because `SameSite=strict`

Issue #1236, opened 2019-12-23
Passing Authorization Credentials Through URL 


Ceknaz:
## Description
Is there any possible way to add the key into the URL when `--auth password` is activated and redirecting to main page if the key is correct, instead of inserting the password on `/login` page manually?

sr229:
It's not possible, if you want such functionality, I suggest looking at the community's input with their implementations in #792. We won't support this upstream for now.

sr229:
It's not possible, if you want such functionality, I suggest looking at the community's input with their implementations in #792. We won't support this upstream for now.

PR #2428, opened 2020-12-09
Add GET authentication

JammSpread:
Adds config option `enable-get-requests` to enable authentication in the URL.
(i.e. logging in on someone going to a link like, **192.168.1.49:8080/login?pass=SamplePassword**)
This pr adds both the password and pass queries with the password query taking higher precedence in the URL.

This may have functionality when sharing a workspace with someone trusted.

Issues that had mentioned this ability:
- #1236
- #1285
- #1669

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -150,7 +150,7 @@ const options: Options<Required<Args>> = {
   socket: { type: "string", path: true, description: "Path to a socket (bind-addr will be ignored)." },
   "enable-get-requests": {
     type: "boolean",
-    description: `Enable authentication via the url with queries. (Usage: ?pass=[password] after the rest of the url)`
+    description: `Enable authentication via the url with a query parameter. (Usage: ?pass=[password] after the rest of the url.)`
   },
   version: { type: "boolean", short: "v", description: "Display version information." },
   _: { type: "string[]" },
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -562,6 +562,31 @@ export async function readConfigFile(configPath?: string): Promise<ConfigArgs> {
   }
 }
 
+export async function writeConfigFile(configPath?: string, data? : object) {
+  if (!configPath) {
+    configPath = process.env.CODE_SERVER_CONFIG
+    if (!configPath) {
+      configPath = path.join(paths.config, "config.yaml")
+    }
+  }
+
+  if (!(await fs.pathExists(configPath))) {
+    await fs.outputFile(configPath, await defaultConfigFile())
+    logger.info(`Wrote default config file to ${humanPath(configPath)}`)
+  }
+
+  const configFile = await fs.readFile(configPath)
+  const config = yaml.safeLoad(configFile.toString(), {
+    filename: configPath,
+  })
+  if (!config || typeof config === "string") {
+    throw new Error(`invalid config: ${config}`)
+  }
+
+  const dumpedData = yaml.safeDump({...config, ...data});
+  await fs.outputFile(configPath, dumpedData)
+}
+
 function parseBindAddr(bindAddr: string): Addr {
   const u = new URL(`http://${bindAddr}`)
   return {
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -47,6 +47,7 @@ export interface Args extends VsArgs {
   tokens?: string[]
   "generate-token"?: boolean
   "list-tokens"?: boolean
+  "revoke-token"?: string
   version?: boolean
   force?: boolean
   "list-extensions"?: boolean
@@ -167,6 +168,10 @@ const options: Options<Required<Args>> = {
   "generate-token": {
     type: "boolean",
     description: "Generate a new token for quick access."
+  },
+  "revoke-token": {
+    type: "string",
+    description: "Remove and disable a specific token from use."
   },
   version: { type: "boolean", short: "v", description: "Display version information." },
   _: { type: "string[]" },
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -43,6 +43,7 @@ export interface Args extends VsArgs {
   port?: number
   "bind-addr"?: string
   socket?: string
+  "enable-get-requests"?: boolean
   version?: boolean
   force?: boolean
   "list-extensions"?: boolean
@@ -147,6 +148,10 @@ const options: Options<Required<Args>> = {
   port: { type: "number", description: "" },
 
   socket: { type: "string", path: true, description: "Path to a socket (bind-addr will be ignored)." },
+  "enable-get-requests": {
+    type: "boolean",
+    description: `Enable authentication via the url with queries. (Usage: ?pass=[password] after the rest of the url)`
+  },
   version: { type: "boolean", short: "v", description: "Display version information." },
   _: { type: "string[]" },
 
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -44,6 +44,9 @@ export interface Args extends VsArgs {
   "bind-addr"?: string
   socket?: string
   "enable-get-requests"?: boolean
+  tokens?: string[]
+  "generate-token"?: boolean
+  "list-tokens"?: boolean
   version?: boolean
   force?: boolean
   "list-extensions"?: boolean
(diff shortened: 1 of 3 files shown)


src/node/routes/login.ts
```diff
@@ -98,4 +88,35 @@ router.post("/", async (req, res) => {
   } catch (error) {
     res.send(await getRoot(req, error))
   }
+}
+
+export const router = Router()
+
+router.use((req : any, res : any, next : any) => {
```
code-asher, line 95:
```suggestion
router.use((req: Request, res: Response, next: NextFunction) => {
```

For this to work you'll need to import `NextFunction` from `express` at the top of the file.

src/node/routes/login.ts
```diff
@@ -98,4 +88,35 @@ router.post("/", async (req, res) => {
...
+    return redirect(req, res, to, {
+      to: undefined, 
+      password: undefined,
+      pass: undefined,
+    })
+  }
+  next()
 })
+
+router.get("/", async (req : any, res : any) => {
```
code-asher, line 107:
```suggestion
router.get("/", async (req: Request, res: Response) => {
```

src/node/routes/login.ts
```diff
@@ -98,4 +88,35 @@ router.post("/", async (req, res) => {
...
+      return await login(req, res, req.query.password as string)
+    }
+    else if (req.query.pass) {
+      return await login(req, res, req.query.pass as string)
+    }
+  }
+  res.send(await getRoot(req))
+})
+
+router.post("/", async (req : any, res : any) => {
```
code-asher, line 120:
```suggestion
router.post("/", async (req: Request, res: Response) => {
```

src/node/routes/login.ts
```diff
@@ -98,4 +88,35 @@ router.post("/", async (req, res) => {
...
+    })
+  }
+  next()
 })
+
+router.get("/", async (req : any, res : any) => {
+  if (req.args["enable-get-requests"]) {
+    // `?password` overrides `?pass`
+    if (req.query.password) {
+      return await login(req, res, req.query.password as string)
```
code-asher, line 111: The query variables here can be objects or arrays so it might be safer to check if it's the right type instead of typecasting.

```suggestion
    if (req.query.password && typeof req.query.password === "string") {
      return await login(req, res, req.query.password)
```

src/node/routes/login.ts
```diff
@@ -98,4 +88,35 @@ router.post("/", async (req, res) => {
...
+    }
+    else if (req.query.pass) {
+      return await login(req, res, req.query.pass as string)
```
code-asher, line 114: Minor linter error here due to the newline. Same comment here about typecasting, too.

```suggestion
} else if (req.query.pass && typeof req.query.pass === "string") {
      return await login(req, res, req.query.pass)
```

code-asher (changes requested):
This looks great!

Maybe we should add the ability to generate keys/tokens? Then you could give out tokens to people which would let you revoke them without having to change your password.

For example we could have something like:
```
$ code-server --generate-token
<token here>
```

And it would insert into a `tokens` array in the `config.yaml`.

Maybe out of scope for this PR though.

src/node/cli.ts
```diff
@@ -147,6 +148,10 @@ const options: Options<Required<Args>> = {
   port: { type: "number", description: "" },
 
   socket: { type: "string", path: true, description: "Path to a socket (bind-addr will be ignored)." },
+  "enable-get-requests": {
```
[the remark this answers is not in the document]
code-asher: If we went the token route we could call it `enable-token-auth` or something.

src/node/entry.ts
```diff
@@ -125,6 +125,10 @@ const main = async (args: DefaultedArgs): Promise<void> => {
     logger.info(`  - Authentication is disabled ${args.link ? "(disabled by --link)" : ""}`)
   }
 
+  if (args["enable-get-requests"]) {
+    logger.info(`  - Login via GET is enabled ${args.auth === AuthType.None ? "(however auth is disabled)" : ""}`)
```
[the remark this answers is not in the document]
JammSpread: I thought that it might seem intrusive to people who are just using normal POST passwords.

src/node/routes/login.ts
```diff
@@ -43,45 +43,35 @@ const getRoot = async (req: Request, error?: Error): Promise<string> => {
...
-  }
-  next()
-})
-
-router.get("/", async (req, res) => {
-  res.send(await getRoot(req))
-})
-
-router.post("/", async (req, res) => {
+const login = async (req : Request, res : any, password : string) => {
```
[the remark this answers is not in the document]
JammSpread: Oh, that makes sense!

src/node/cli.ts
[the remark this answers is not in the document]
JammSpread: That seems quick and concise! I might try that. :) Yeah, I agree that tokens seem to be safer.

src/node/routes/login.ts
JammSpread: Didn't think of that thanks!

JammSpread:
Should GET passwords be completely gone if tokens are added?

JammSpread:
Should tokens persist sessions if not revoked?

JammSpread:
I've made and tested the `generate-tokens` method. However I noticed that whenever it appends a token to the token array after two tokens are already in the config file it writes to the config like this:
```yaml
tokens:
  - '73713b95623914257d6799e3,39cb10ecd3e758848417a376'
  - c2dfe443affe447df13ecd43
```
I am currently wondering why this would occur.

code-asher:
@JammSpread I feel absolutely terrible for doing this to you after you've written all of this (really great!) code. :sob: 

We had an internal discussion about whether we want to support additional authentication methods (such as GET tokens or GET passwords) and while we think it's a cool idea ultimately we came to the conclusion that these are a little out of code-server's scope and it would be more secure to rely on external authentication mechanisms.

For example SSH or something like https://github.com/oauth2-proxy/oauth2-proxy.

I'm sorry I didn't check whether we wanted to integrate this feature before reviewing! Please let me know if you have any questions/concerns/counterpoints.

JammSpread:
Oh that's fine. If that's what might be best for `code-server` then that's great! I agree things might get bloated (and can vastly complicate things) if we focus too much on some features and keeping things light is what the experience of VSC is known for. Simplicity is sometimes far more appealing! 😄

jhgoebbert:
This patch would make it possible to call `code-server` from the launchpad of a JupyterLab (using the extension jupyter-server-proxy).
You can find the `code-server`-proxy [here](https://pypi.org/project/jupyter-codeserver-proxy) . At the moment we need to disable any authentication [[code-details]](https://github.com/FZJ-JSC/jupyter-codeserver-proxy/blob/main/jupyter_codeserver_proxy/__init__.py#L93) as we cannot forward the password of `code-server` using url parameters.

Perhaps you might want to rethink this PR once more :thinking:

jhgoebbert:
Cool. That would be fantastic!
  
It would be great, if `code-server` could be started with something like `--generate-token` (or in perhaps even with `--token <hash>`) which could be used then in the URL like `https://myserver.org/code-server?token=<hash>` to bypass the login-screen.

In the case of an integration into JupyterLab's launchpad the request for a token would then be added [HERE](https://github.com/FZJ-JSC/jupyter-codeserver-proxy/blob/main/jupyter_codeserver_proxy/__init__.py#L92). And this token would be passed to the code-server's launchpad-button-link [HERE](https://github.com/FZJ-JSC/jupyter-codeserver-proxy/blob/main/jupyter_codeserver_proxy/__init__.py#L115).

code-asher:
The way upstream does it is that you can pass `--connection-token <token>` or `--connection-token-file <file>` so that sounds like it should meet the requirements.

I just realized there would be overhead though because code-server has its own routes so we would need to update our authentication to support the token.  Unfortunate but I think since upstream supports token-based authentication that is a strong reason for us to support it as well.

jhgoebbert:
Do you have any idea what priority this token-based authentication has for `code-server`?  
Code-Server currently is the only solution which supports to run behind a configurable-http-proxy and can therefore handle URLs like:
`https://jupyter-jsc-staging.fz-juelich.de/user/j.goebbert.fz-juelich.de/de47c0193b364c36b9da34e681f14911/code-server`

code-asher:
Priority for implementation?  I will add it to the backlog but I 
have no
idea if/when we will get to it.

PR #2434, opened 2020-12-09
Add --ignore-last-opened option

### LINKED ISSUE #1951: Start code-server with no project directory
I have a single instance code server, but I like to have multiple projects open.  It used to be that if you didn't supply a root directory it would not have any folder open by default.  Is there a way to get this back, or make this a flag at startup?

JammSpread:
Added --no-project or -empty option to code-server

Fixes #1951

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -52,7 +52,7 @@ export interface Args extends VsArgs {
   "proxy-domain"?: string[]
   locale?: string
   _: string[]
-  "no-project"?: boolean
+  "ignore-last-opened"?: boolean
   "reuse-window"?: boolean
   "new-window"?: boolean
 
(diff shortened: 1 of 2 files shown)


JammSpread:
Should I add or modify -empty to be -e for brevity?

code-asher:
Thank you for implementing this!

Since this only affects the "last opened" behavior I think we might want to name the flag something to match. Something like `--no-remember-last-opened` although it's a bit long. The description could be something like "Do not remember the last opened directory or workspace."

JammSpread:
We could follow how VSC did it and use something like a `restore` option that defaults to true. Since it still remembers the last folder. We could also do something like `no-restoration` though that seems somewhat vague.

JammSpread:
Do you think we could use something like `ignore-last-opened`?
I've been thinking this since we are still remembering the `last-opened` and we want to integrate it into the name.

code-asher:
That's true, we do still store the last opened. `ignore` is definitely
better than `no-remember`.

I really like `restore` too but I think `restore` could be in reference
to the directory, or the file tabs, or the terminals, etc, so overall
`ignore-last-opened` seems like it describes the behavior more
accurately to me.

code-asher (approved):
Thank you again! <3

PR #2453, opened 2020-12-14
FAQ.md: Add mention of hashedPassword

### LINKED ISSUE #2451: Mention hashedPassword in FAQ
See https://github.com/cdr/code-server/issues/2225#issuecomment-742082627

Just would be nice to demonstrate how you can use  `sha256sum` to generate the hash.

nhooyr:
Closes #2451

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -11,6 +11,7 @@
 - [Where are extensions stored?](#where-are-extensions-stored)
 - [How is this different from VS Code Codespaces?](#how-is-this-different-from-vs-code-codespaces)
 - [How should I expose code-server to the internet?](#how-should-i-expose-code-server-to-the-internet)
+- [Can I store my password hashed?](#can-i-store-my-password-hashed)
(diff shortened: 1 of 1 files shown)


oxy (approved):
LGTM! Not sure if we want to put a warning that the password once hashed can't be recovered - should be obvious but may not be.

nhooyr:
> Not sure if we want to put a warning that the password once hashed can't be recovered - should be obvious but may not be.

I think we're good. That's the point of hashing after all.

immanuelfodor:
Will this work with Docker environment variable like `PASSWORD` but as `HASHEDPASSWORD`?

immanuelfodor:
Nevermind, I've found it in the binary options: 

--password             The password for password authentication (can only be passed in via $PASSWORD or the config file).                                     

--hashed-password      The password hashed with SHA-256 for password authentication (can only be passed in via $HASHED_PASSWORD or the config file).                                 Takes precedence over 'password'.

PR #2454, opened 2020-12-14
cli: hashedPassword -> hashed-password 

nhooyr:
Capital letters in the CLI are evil.

cc @code-asher @JammSpread

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -163,7 +163,7 @@ Again, please follow [./guide.md](./guide.md) for our recommendations on setting
 
 ## Can I store my password hashed?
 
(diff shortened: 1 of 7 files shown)


code-asher (approved):

JammSpread (approved):
This all seems good! Just a question, although wouldn't it also make sense to rename the occurrences of `usingEnvPassword` and `usingEnvHashedPassword`?

nhooyr:
> wouldn't it also make sense to rename the occurrences of usingEnvPassword and usingEnvHashedPassword?

We only use hyphens on the CLI flags, on all other variables we've been using camel case.

PR #2455, opened 2020-12-14
release-image: Change default coder shell to /bin/bash

nhooyr:
Closes #2410

Issue #2410, opened 2020-12-06
Keyboard shortcuts not working in terminal


Mearman:
For some reason the remote terminal isn't correctly receiving non-character keyboard inputs. E.g. `up`, `down`, `esc`, `Ctrl+c`, etc.

![image](https://user-images.githubusercontent.com/1331872/101294728-5c5b0300-3811-11eb-8b2d-9c44fef0dd34.png)

- Web Browser: Chrome
- Local OS: Windows 10
- Remote OS: Docker/unRAID
- `code-server --version`: 3.7.4

code-asher:
What do you get for `printenv SHELL`? You might be running a shell that
doesn't support the arrow keys. Assuming that's the problem, does
installing and running `bash` make them work?

Mearman:
I get nothing from `printenv SHELL` I'm just running the docker unRAID template.
I was just using the default `sh` shell.
`sh`, `bash`, `rbash`, and `dash` appear to be installed by default. 
`bash` or `rbash` resolve the issue, but it persists in `sh` and `dash`

nhooyr:
Those keys you see are the correct codes for the arrow keys. This is expected behaviour in `sh`! Try locally or over SSH, it should behave the same.

Mearman:
Ahh that's annoying. Is it possible the default shell to be initially set to bash in the then to be more user-friendly?

nhooyr:
@Mearman The default shell isn't controlled by `code-server`. It's whatever you've got it set for your user.

nhooyr:
Oh nvm I see the problem! The dockerfile doesn't set the `coder` user's default shell to `bash`.

https://github.com/cdr/code-server/blob/40b345d297c33d71534d3725e1d32f834d2e3893/ci/release-image/Dockerfile#L24-L25

Mearman:
Sorry I should I have specified it was in Docker.

nhooyr:
@Mearman What docker image are you using?

I was wrong regarding our default image. It does in fact use bash by default.

Mearman:
Ahh it's LinuxServer's didn't realise they were rolling their own Dockerfile.
https://forums.unraid.net/topic/81306-support-linuxserverio-code-server/
https://github.com/linuxserver/docker-code-server/blob/master/Dockerfile

Would putting in a PR with the addition of `ENV SHELL=/bin/bash` be sufficient?

nhooyr:
You'd need to call `chsh` as well!

[...]

PR #2457, opened 2020-12-14
release-image: Remove unnecessary call to chsh

nhooyr:
Confused me in #2410, see #2455

debian:10 defaults to bash as $SHELL


jsjoeio (approved):
Assuming we can ignore the arbitrary GitHub Action failure on one of the builds, then LGTM! 🚢

image


Thanks for linking to the discussion - that was helpful context.

PR #2458, opened 2020-12-14
release-image: Ignore $DOCKER_USER on restart

### LINKED ISSUE #2267: Docker: Restarting container
- Web Browser:
- Local OS: MacOS
- Remote OS:
- Remote Architecture:
- `code-server --version`: Docker (latest)

I setup Docker as described here: https://github.com/cdr/code-server/blob/v3.6.2/doc/install.md#docker

```
mkdir -p ~/.config
docker run -it --name code-server -p 127.0.0.1:8080:8080 \
  -v "$HOME/.config:/home/coder/.config" \
  -v "$PWD:/home/coder/project" \
  -u "$(id -u):$(id -g)" \
  -e "DOCKER_USER=$USER" \
  codercom/code-server:latest
```

The initial run works perfectly. I then stop the container using `ctrl+c`. I try to start the container again using: `docker start code-server` and nothing happens.

When I run `docker logs code-server` I get the following: `usermod: user 'coder' does not exist`

nhooyr:
Closes #2267

### DIFF
--- a/ci/release-image/entrypoint.sh
+++ b/ci/release-image/entrypoint.sh
@@ -5,7 +5,7 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ]; then
+if [ "${DOCKER_USER-}" ] && [ "$DOCKER_USER" != "$USER" ]; then
   echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd > /dev/null
   # Unfortunately we cannot change $HOME as we cannot move any bind mounts
   # nor can we bind mount $HOME into a new home as that requires a privileged container.


ci/release-image/entrypoint.sh
```diff
@@ -5,7 +5,7 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ]; then
+if [ "${DOCKER_USER-}" != "${USER-}" ]; then
```
code-asher, line 8: I think we'll need to check if `DOCKER_USER` otherwise this branch will run if it's empty.

Actually...we used to have it this way, right? Yeah: https://github.com/cdr/code-server/commit/79443c14ff41f38d3f05e0a53e4aa592e3c0e418

Any idea why we changed it? I feel like there was a reason but I might need to test it to remember exactly why.

nhooyr: omfg haha yess!! Yea so that commit linked looks like I just made a dumb mistake. I removed that branch thinking it was simpler but i was wrong.

code-asher (approved):

Issue #1826, opened 2020-06-24
FreeBSD installation not working error "Cannot reconnect. Please reload the window."


siuleinad01:
I have been trying to get this to work on FreeBSD 12.1, code-servers installs with some warnings but when you load the screen, authenticate and code actually opens after a few seconds I get the above error, and in the terminal I get the following"

Error: Could not locate the bindings file. Tried:
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/build/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/build/Debug/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/build/Release/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/out/Debug/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/Debug/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/out/Release/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/Release/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/build/default/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/compiled/13.10.1/freebsd/x64/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/addon-build/release/install-root/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/addon-build/debug/install-root/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/addon-build/default/install-root/spdlog.node
 → /usr/local/share/.config/yarn/global/node_modules/code-server/lib/vscode/node_modules/spdlog/lib/binding/node-v79-freebsd-x64/spdlog.node
[...]

Issue #2349, opened 2020-11-21
FreeBSD 11.3-RELEASE-p14 install script throws errors


irgipaulius:
- Web Browser: not applicable
- Local OS: not applicable
- Remote OS: FreeBSD 11.3-RELEASE-p14. (a jail inside a FreeNas server)
- Remote Architecture: x86
- `code-server --version`: 3.7.2 a0db6723c1ef22b40d7ac0df5ac1bb9a92e4fe46

<hr />

Problem with: installation script

<hr />
## How to replicate

1. create a new jail in FreeNas. SSH into it with sudo permissions
2. `pkg upgrade`
3. `curl -fsSL https://code-server.dev/install.sh | sh -s -- --dry-run`
  3.1. it finishes successfully, everything seems fine
4. `curl -fsSL https://code-server.dev/install.sh | sh`
  4.1 installs `code-server` successfully
  4.2 but it spews discouraging errors
  4.3 and does not print out info on how to run and start using code-server after it's finished, like it's supposed to do according to the [documentation](https://github.com/cdr/code-server/blob/v3.7.2/doc/install.md#installsh).

<hr />

Logs go into what seems like an endless loop of these 12 error lines over and over again:
```
Done in 0.07s.
extensions/xml: installing dependencies
yarn install v1.22.10
warning ../package.json: No license field
info No lockfile found.
[1/5] Validating package.json...
warning xml@1.0.0: The engine "vscode" appears to be invalid.
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
[5/5] Building fresh packages...
Done in 0.07s.
```

And upon scrolling a bit higher, you can see this error:

![image](https://user-images.githubusercontent.com/20173109/99888207-05bfc780-2c5c-11eb-95a0-fb22800227ba.png)

<hr />

## Consequences
It does not influence my accessibility to the product, but it did influence the **experience** of the installation process. I panicked that something is wrong with my system and I started debugging until 20 minutes later I had an idea to run `code-server --version` and it proved that it installed successfully. Wasted time I'm not getting back.

code-asher:
Thank you for reporting. VS Code uses rimraf in their post-install to
remove some files from some Node modules although I'm not entirely sure
why as it still works without doing that.

I'm not sure how rimraf is missing though; it should be in the
package.json dependencies. I have sometimes seen dependency downloads
time out and instead of erroring yarn/npm seem to just keep chugging
along which results in problems later when those dependencies are used
[...]

Issue #2320, opened 2020-11-17
could you supply bsd*(freebsd/netbsd/openbsd) release binary package?


jimying:
As title.

nhooyr:
Hoping to have a FreeBSD binary in the next release!

nhooyr:
Done in #2466

Only thing remaining here is we need a static node v12 binary to ship with the release. Seems like I'm going to have to modify the port somehow.

nhooyr:
And need to modify the install script!

nhooyr:
Modified the port and got it down to the following:

```diff
diff --git a/two b/one
index 8d2ff16..8bd6e29 100644
--- a/two
+++ b/one
@@ -1,20 +1,15 @@
-	libz.so.6 => /lib/libz.so.6 (0x80224e000)
-	libuv.so.1 => /usr/local/lib/libuv.so.1 (0x80226a000)
-	libcares.so.2 => /usr/local/lib/libcares.so.2 (0x802299000)
-	libnghttp2.so.14 => /usr/local/lib/libnghttp2.so.14 (0x8022b9000)
-	libcrypto.so.111 => /lib/libcrypto.so.111 (0x8022e6000)
-	libssl.so.111 => /usr/lib/libssl.so.111 (0x8025d8000)
-	libicui18n.so.67 => /usr/local/lib/libicui18n.so.67 (0x802670000)
-	libicuuc.so.67 => /usr/local/lib/libicuuc.so.67 (0x80299c000)
-	libicudata.so.67 => /usr/local/lib/libicudata.so.67 (0x802b9c000)
-	libelf.so.2 => /lib/libelf.so.2 (0x802b9e000)
-	libutil.so.9 => /lib/libutil.so.9 (0x802bb9000)
-	libkvm.so.7 => /lib/libkvm.so.7 (0x802bd1000)
-	libm.so.5 => /lib/libm.so.5 (0x802be4000)
-	libexecinfo.so.1 => /usr/lib/libexecinfo.so.1 (0x802c16000)
-	libc++.so.1 => /usr/lib/libc++.so.1 (0x802c1c000)
-	libcxxrt.so.1 => /lib/libcxxrt.so.1 (0x802ce9000)
-	libthr.so.3 => /lib/libthr.so.3 (0x802d0b000)
-	libc.so.7 => /lib/libc.so.7 (0x802d38000)
-	libdl.so.1 => /usr/lib/libdl.so.1 (0x80312e000
[...]

Issue #1880, opened 2020-07-14
npm install fails on FreeBSD 12.1


mike-pt:
- Local OS: FreeBSD
- Remote OS: FreeBSD
- Remote Architecture: amd64
- `code-server --version`: n/a

I can't seem to install due to missing deps its seems but I also found no clue how to install those:
```
npm WARN @coder/logger@1.1.11 requires a peer of @google-cloud/logging@^4.5.2 but none is installed. You must install peer dependencies yourself.
npm WARN ws@7.3.1 requires a peer of bufferutil@^4.0.1 but none is installed. You must install peer dependencies yourself.
npm WARN ws@7.3.1 requires a peer of utf-8-validate@^5.0.2 but none is installed. You must install peer dependencies yourself.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! code-server@3.4.1 postinstall: `./postinstall.sh`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the code-server@3.4.1 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-07-14T12_30_33_064Z-debug.log
``

**Comment by code-asher** [comment]
Is there more to the output? I think the cause of the error might be further up. I don't think the peer dependency warnings would cause the script to fail.

**Comment by mike-pt** [comment]
Not that I could see but it was over tmux it's possible it's cutoff indeed, I was wondering if it was indeed related to dependencies missing.

I also tried yarn after and it worked but then It keeps failing with connection issues.

Im on mobile now but I'll check later if there's more output

Thanks

**Comment by nhooyr** [comment]
Please upload the debug log file `/root/.npm/_logs/2020-07-14T12_30_33_064Z-debug.log`.

**Comment by mike-pt** [comment]
This is a azure VM so python3 is installed (not 2 though not sure if that matters)

but I do have:
```
pkg-config libsecret-devel libX11-devel libxkbfile-devel
```

nhooyr:
Got it. Can't comment further without the full log file.

mike-pt:
Dang it I though I uploaded it in the previous post:

[2020-07-14T12_30_33_064Z-debug.log](https://github.com/cdr/code-server/files/4956293/2020-07-14T12_30_33_064Z-debug.log)

That's the one from when I run it originally

mike-pt:
[...]

PR #2466, opened 2020-12-14
ci: Add support for FreeBSD to the build process

nhooyr:
Updates #1826
Updates #1880
Updates #2320
Updates #2349

### DIFF
--- a/lib/vscode/package.json
+++ b/lib/vscode/package.json
@@ -116,7 +116,6 @@
     "css-loader": "^3.2.0",
     "debounce": "^1.0.0",
     "deemon": "^1.4.0",
-    "electron": "9.3.3",
     "eslint": "6.8.0",
     "eslint-plugin-jsdoc": "^19.1.0",
     "eslint-plugin-mocha": "8.0.0",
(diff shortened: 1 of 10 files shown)


nhooyr:
Need to fix some other scripts rn as well.

nhooyr:
Now closes #1826, #1880, #2349

I'm going to leave #2320 for now as it needs further investigation.

lib/vscode/package.json
```diff
@@ -116,7 +116,6 @@
     "css-loader": "^3.2.0",
     "debounce": "^1.0.0",
     "deemon": "^1.4.0",
-    "electron": "9.3.3",
```
jsjoeio, line 119: this was intentional, right?

jsjoeio (approved):
All looks good as usual! 🌮 

Only question was about removing `electron` as a package. If that was intentional, should we also remove `lib/vscode/src/typings/electron.d.ts`?

lib/vscode/package.json
nhooyr: @jsjoeio see the per commit breakdown!

nhooyr:
@jsjoeio See commit 3

lib/vscode/package.json
jsjoeio: Ah sorry! Should have looked there first. Got it 👍🏼

jsjoeio:
@nhooyr got it! 

Looks good! 🚀 

(side note: would be curious to hear why we should keep the types for my own learning?)

nhooyr:
> (side note: would be curious to hear why we should keep the types for my own learning?)

VS Code won't compile without it. We still compile the electron parts for some reason. I asked @code-asher and he said the easiest way to avoid problems is just to include the types to quiet `tsc`.

PR #2473, opened 2020-12-15
feat(local development): add vscode:clean script

jsjoeio:
This PR adds a new script to the root `package.json` called `vscode:clean` which resets `lib/vscode`. 

## Changes
- add script to `./ci/dev/clean-vscode.sh`
- add script `vscode:clean` to `package.json` at root

## Checklist
- [x] tested locally
- [ ] added new dependencies
- [ ] updated the docs
- [ ] added a test

Fixes N/A (no accompanying issue)

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "vscode": "./ci/dev/vscode.sh",
     "vscode:patch": "./ci/dev/patch-vscode.sh",
     "vscode:diff": "./ci/dev/diff-vscode.sh",
+    "vscode:reset": "./ci/dev/reset-vscode.sh",
     "build": "./ci/build/build-code-server.sh",
     "build:vscode": "./ci/build/build-vscode.sh",
     "release": "./ci/build/build-release.sh",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,6 +13,7 @@
     "vscode": "./ci/dev/vscode.sh",
     "vscode:patch": "./ci/dev/patch-vscode.sh",
     "vscode:diff": "./ci/dev/diff-vscode.sh",
+    "vscode:clean": "./ci/dev/clean-vscode.sh",
```
nhooyr, line 16: Let's call this `reset`. `clean` would delete all build artifacts and caches which this doesn't do. it just resets the source.

nhooyr:
🔥 this is going to be really helpful.

package.json
jsjoeio: ah gotcha! good point. I'll update now.

nhooyr (approved):
🔥

nhooyr:
@jsjoeio make sure to squash that second commit into the first!

Highly recommend using `git commit --fixup` to make it clear that's the intent.

jsjoeio:
> @jsjoeio make sure to squash that second commit into the first!
> 
> Highly recommend using `git commit --fixup` to make it clear that's the intent.

😮 Thanks for the tip! I haven't used that before. I'll make those changes before I merge.

PR #2476, opened 2020-12-15
refactor (vscode): migrate from submodule to subtree

### LINKED ISSUE #1587: Switch to git subtree for lib/vscode instead of a submodule

jsjoeio:
This PR changes our current approach of using `vscode` as a [submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules) to using it as a [subtree](https://www.atlassian.com/git/tutorials/git-subtree).

## Changes
- remove submodule
- add subtree 
- apply patch
- remove old files and scripts
  - `ci/dev/vscode.patch`
  - `ci/dev/patch-vscode.sh` and `vscode:patch` from `package.json`
  -  `ci/dev/diff-vscode.sh` and `vscode:diff` from `package.json`
- remove `vscode.sh` and add `postinstall` script to `package.json`

### Remaining Todos
- [x] update `ci/dev` scripts
- [x] update `ci/build` scripts
- [x] update docs
- [x] update contributing docs
- [x] address feedback
- [x] add `| grep -v "lib/vscode"` to all the `git ls-files` invocations

## Checklist
- [x] tested locally
- [x] updated the docs

Fixes #1587 

Many thanks to @SPGoding for testing out this approach and outlining the steps to start this effort! 🎉


jsjoeio:
Okay made some initial progress. The migration piece was the easy part. The harder part is understanding how to update our `vscode.sh` script.

How do we keep our subtree up to date with vscode?

Thinking a bit out loud here...
- when someone is making a change to our codebase, they don't need to run any funky commands because the `vscode` code is already there (thanks to the subtree)
- when we (maintainers) need to update the subtree (as in update `vscode` from `1.5.1` to `1.5.2`, how do we do it?

If it were a normal repository, like a fork I'd probably fetch the latest changes then rebase my changes on top of those. But subtree's don't support rebase, only merge strategies. So we have to figure that out.

Resources that may help:
- [How to use the subtree merge strategy](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/howto/using-merge-subtree.html)
- [Why doesn't Git subtree use rebase?](https://stackoverflow.com/questions/22447376/why-doesnt-git-subtree-use-rebase)
- bash script: [git rebase to subtree](https://github.com/eviweb/bash-scripts/blob/master/git/git-rebasetosubtree.sh)

jsjoeio:
Maybe something like this? 

```sh
  # Add vscode as a new remote
  # TODO only do once
  git remote add -f vscode https://github.com/microsoft/vscode.git

  # Prep to merge with version we're currently on
  git merge -s ours --no-commit --allow-unrelated-histories vscode/release/1.51

  # Read changes to our sub directory
  git read-tree --prefix=lib/vscode -u vscode/release/1.51

  # Commit changes
  # git commit -m "chore: update vscode"

  # git pull -s subtree Bproject master
```

code-asher:
I think all we need to do in `vscode.sh` is install VS Code's Node modules. Actually maybe we should just move that to `postinstall` in `package.json` so you don't need to run a separate `yarn vscode`.

That way development would just be:

```
yarn
yarn watch
```

As for updating VS Code once the remote is added I think all we need is something like:

```
git subtree pull --prefix lib/vscode vscode release/1.52 --squash --message "Update VS Code to 1.52.1"
```

I'm not sure we need a script for that, maybe we just add it to the contributing docs.

jsjoeio:
> Actually maybe we should just move that to postinstall in package.json so you don't need to run a separate yarn vscode

Great idea! Didn't think of that. 

> I'm not sure we need a script for that, maybe we just add it to the contributing docs.

Ah, if it's only one line, I agree, we probably don't need that. I'll add to contributing. Thanks @code-asher

jsjoeio:
Alright a few errors I'm seeing.

## Error in fmt script
In our `ci / fmt` workflow, it's upset with the `let` declaration inside the `for` loop. Maybe when that gets called `$1` is undefined?
```
lib/vscode/src/vs/base/node/cpuUsage.sh:9:11: = must follow a name
```

The code there reads:
```bash
#!/bin/bash

function get_total_cpu_time() {
  # Read the first line of /proc/stat and remove the cpu prefix
  CPU=(`sed -n 's/^cpu\s//p' /proc/stat`)

  # Sum all of the values in CPU to get total time
  for VALUE in "${CPU[@]}"; do
    let $1=$1+$VALUE
  done
}
```
(see file on [remote](https://github.com/cdr/code-server/blob/41bee49d077e7b557ffd57b91893c2c3b90fde9b/lib%2Fvscode%2Fsrc%2Fvs%2Fbase%2Fnode%2FcpuUsage.sh#L1-L11))

## Error in lint script

It seems like it can't find our `.eslint.yaml` or rather the `node_modules` where these rule definitions come from.
```
/github/workspace/lib/vscode/test/automation/src/explorer.ts
  1:1  error  Definition for rule 'code-no-unused-expressions' was not found        code-no-unused-expressions
  1:1  error  Definition for rule 'code-translation-remind' was not found           code-translation-remind
  1:1  error  Definition for rule 'code-no-nls-in-standalone-editor' was not found  code-no-nls-in-standalone-editor
  1:1  error  Definition for rule 'code-no-standalone-editor' was not found         code-no-standalone-editor
  1:1  error  Definition for rule 'code-no-unexternalized-strings' was not found    code-no-unexternalized-strings
```

doc/CONTRIBUTING.md
```diff
@@ -127,10 +122,9 @@ and exposed an API to the front-end for file access and all UI needs.
...
 the front-end open source, but not the server. As such, code-server v2 (and later) uses
-the VS Code front-end and implements the server. You can find this in
-[./ci/dev/vscode.patch](../ci/dev/vscode.patch) under the path `src/vs/server`.
+the VS Code front-end and implements the server. We use a git subtree to be able to pull in VS Code's front-end. This code lives under [./lib/vscode](./lib/vscode).
 
-Other notable changes in our patch include:
+Some noteworthy changes in our codebase:
 
 - Adding our build file, which includes our code and VS Code's web code
 - Allowing multiple extension directories (both user and built-in)
```
jsjoeio, line 130: Looking for feedback on the wording here

doc/CONTRIBUTING.md
```diff
@@ -146,12 +140,10 @@ Other notable changes in our patch include:
 
 As the web portion of VS Code matures, we'll be able to shrink and possibly
 eliminate our patch. In the meantime, upgrading the VS Code version requires
-us to ensure that the patch is applied and works as intended. In the future,
+us to ensure that our changes are applied and work as intended. In the future,
 we'd like to run VS Code unit tests against our builds to ensure that features
 work as expected.
```
jsjoeio, line 145: And here

.tours/start-development.tour
```diff
@@ -5,7 +5,7 @@
     {
       "file": "package.json",
       "line": 31,
-      "description": "## Commands\n\nTo start developing, make sure you have Node 12+ and the [required dependencies](https://github.com/Microsoft/vscode/wiki/How-to-Contribute#prerequisites) installed. Then, run the following commands:\n\n1. Install dependencies:\n>> yarn\n\n2. Clone and patch VS Code and install its dependencies:\n>> yarn vscode\n\n3. Start development mode (and watch for changes):\n>> yarn watch"
+      "description": "## Commands\n\nTo start developing, make sure you have Node 12+ and the [required dependencies](https://github.com/Microsoft/vscode/wiki/How-to-Contribute#prerequisites) installed. Then, run the following commands:\n\n1. Install dependencies:\n>> yarn\n\n3. Start development mode (and watch for changes):\n>> yarn watch"
```
code-asher, line 8: Looks like we need to update step `3` to be step `2`.

doc/CONTRIBUTING.md
```diff
@@ -2,13 +2,14 @@
...
-- [Structure](#structure)
-  - [VS Code Patch](#vs-code-patch)
-  - [Currently Known Issues](#currently-known-issues)
+- [Contributing](#contributing)
+  - [Pull Requests](#pull-requests)
+  - [Requirements](#requirements)
+  - [Development Workflow](#development-workflow)
+  - [Build](#build)
+  - [Structure](#structure)
+    - [VS Code Patch](#vs-code-patch)
```
code-asher, line 11: What do you think about calling this `VS Code Modifications` or `VS Code Changes` or something? Or `Modifications to VS Code`.

code-asher (changes requested):
Very exciting! A couple other areas we should update:

- `tour/contributing.tour`: Change `"file": "ci/dev/vscode.patch"` to `"directory": "lib/vscode"` and update the wording to not reference the patch.
- There are some references to applying the patch in `ci/README.md` which we should remove.

I haven't looked at the lint/fmt errors or the space issue yet.

doc/CONTRIBUTING.md
code-asher: I think this makes sense as-is. Although, maybe `our changes are still applied`? Since they'll already be applied, the trick when updating will be to make sure they're still there and still work.

I think we might want to change the `eliminate our patch` part to something like `eliminate our modifications`.

doc/CONTRIBUTING.md
code-asher: Maybe "Some noteworthy changes in our version of VS Code"? 

I keep wanting to say "fork of VS Code" but I'm not sure if that would be confusing since we have it as a subtree. 

For example "changes in our fork of VS Code" or "we do this by using a git subtree to fork and modify VS Code".

jsjoeio:
Re: Feedback in Slack from @code-asher 
> we might need to use something other than `git ls-files` for determining what to run our formatter and linter on
> we can do this by adding `| grep -v "lib/vscode"` to all the `git ls-files` invocations

.tours/start-development.tour
jsjoeio: Good call! I'll update that

doc/CONTRIBUTING.md
jsjoeio: Yeah, we should rename that! I like `Modification to VS Code`

doc/CONTRIBUTING.md
jsjoeio: Updated!

doc/CONTRIBUTING.md
jsjoeio: Updated!

ci/dev/fmt.sh
```diff
@@ -30,7 +30,7 @@ main() {
   doctoc --title '# Contributing' doc/CONTRIBUTING.md > /dev/null
   doctoc --title '# iPad' doc/ipad.md > /dev/null
 
-  if [[ ${CI-} && $(git ls-files --other --modified --exclude-standard) ]]; then
+  if [[ ${CI-} && $(git ls-files --other --modified --exclude-standard | grep -v "lib/vscode") ]]; then
```
code-asher, line 33: To satisfy the linter we could do:
```
  if [[ ${CI-} ]] && git ls-files --other --modified --exclude-standard | grep -q -v "lib/vscode"; then
```

Although I think in this case we could actually leave off the `grep -v "lib/vscode"` since this is just checking if any files were modified by `fmt` and nothing in `lib/vscode` will be modified since we excluded those paths above.

code-asher:
For the space issue it seems to run out when copying to the GCP directory (`build-packages.sh`). I think we just remove that for now since we don't currently upload to GCP anyway and I think it might make sense to add a post-publish script that uploads the releases from GitHub to GCP instead of doing it this way anyway.

ci/dev/fmt.sh
jsjoeio: Ah gotcha! As you can probably tell, I haven't written a lot of shell scripts 😅 

I'll leave it off here!

jsjoeio:
> For the space issue it seems to run out when copying to the GCP directory (`build-packages.sh`). I think we just remove that for now since we don't currently upload automatically to GCP anyway 

Good call. I removed it in 96576bd

> and I think it might make sense to add a separate step that uploads the releases from GitHub to GCP instead of doing it this way anyway.

Do you want me to open an issue for us to do that in the future?

jsjoeio:
>  $ ./ci/build/build-packages.sh
> done (release-packages/code-server-3.7.4-linux-arm64)
> guessing packager from target file extension...
> using deb packager...
> nfpm: error: cannot add data.tar.gz to deb: write release-packages/code-server_3.7.4_arm64.deb: no space left on device

That's interesting...It's still trying to add it. I must have removed the wrong step.

code-asher:
> Do you want me to open an issue for us to do that in the future?

No need, we're tracking it internally.

> That's interesting...It's still trying to add it. I must have removed the wrong step.

Nah you got it right, it's just running out of space later down the line now.

code-asher (approved):

nhooyr:
May submodules burn in hell!

PR #2482, opened 2020-12-17
Symlink node_modules.asar to node_modules in lib/vscode

### LINKED ISSUE #2197: Package with node_modules.asar
Currently we patch `require` to use `node_modules` but this breaks if an extension explicitly tries to look in `node_modules.asar` like the bracket pair colorizer plugin:

```
Activating extension 'CoenraadS.bracket-pair-colorizer-2' failed: ENOENT: no such file or directory, open '/usr/lib/code-server/lib/vscode/node_modules.asar/vscode-oniguruma/release/onig.wasm'. _logMessageInConsole @ abstractExtensionService.ts:416
```

We should package `code-server` with that directory.

code-asher:
Closes #2197.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ release-packages/
 release-gcp/
 release-images/
 node_modules
+node_modules.asar
 node-*
 /plugins
 /lib/coder-cloud-agent
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ release-packages/
 release-gcp/
 release-images/
 node_modules
-node_modules.asar
+/lib/vscode/node_modules.asar
 node-*
 /plugins
 /lib/coder-cloud-agent
--- a/ci/build/npm-postinstall.sh
+++ b/ci/build/npm-postinstall.sh
@@ -41,6 +41,19 @@ main() {
 vscode_yarn() {
   cd lib/vscode
   yarn --production --frozen-lockfile
+
+  # VS Code needs a node_modules.asar but that's just a duplicate of stuff we
+  # already have in node_modules.
+  if [ ! -e node_modules.asar ]; then
+    if [ "${WINDIR-}" ]; then
+      # mklink takes the link name first.
+      mklink /J node_modules.asar node_modules
+    else
+      # ln takes the link name second.
+      ln -s node_modules node_modules.asar
+    fi
+  fi
+
   cd extensions
   yarn --production --frozen-lockfile
   for ext in */; do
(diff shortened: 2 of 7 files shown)


ci/lib.sh
```diff
@@ -93,3 +94,17 @@ export OS
...
 # Defaults to release
 RELEASE_PATH="${RELEASE_PATH-release}"
+
+# Symlink node_modules.asar to node_modules. VS Code needs a node_modules.asar
+# but that's just a duplicate of stuff we already have in node_modules.
+symlink_asar() {
+  if [ ! -e node_modules.asar ]; then
+    if [ "${WINDIR-}" ]; then
+      # mklink takes the link name first.
+      mklink /J node_modules.asar node_modules
```
nhooyr, line 104: God bless you for trying to support windows @code-asher 

haha

Only benefit to writing scripts in javascript is that they'll work on windows too. Maybe we should consider that or adding something to our scripts to pull in something like `cygwin` to map standard commands for us easily. In fact, given the interpreter here is `bash`, they'd need `cygwin` anyway and so have a working `ln` binary for windows.

ci/build/npm-postinstall.sh
```diff
@@ -41,6 +41,19 @@ main() {
 vscode_yarn() {
   cd lib/vscode
   yarn --production --frozen-lockfile
+
+  # VS Code needs a node_modules.asar but that's just a duplicate of stuff we
```
nhooyr, line 45: Let's document what `asar` means/why VS code expects it.

.gitignore
```diff
@@ -9,6 +9,7 @@ release-packages/
 release-gcp/
 release-images/
 node_modules
+node_modules.asar
```
nhooyr, line 12:
```suggestion
/lib/vscode/node_modules.asar
```

ci/build/npm-postinstall.sh
nhooyr: Also should document that this is a copy of the function `symlink_asar` in `lib.sh`

nhooyr (commented):
🔥

ci/lib.sh
code-asher: The weird thing about `ln -s` in Windows (using MINGW64) is that when I tested it seems to just copy the directory rather than make a link. Changes to either end don't reflect on the other side.

nhooyr: weird. does installing `MINGW64` also install `cygwin`?

code-asher: I think `MINGW64` is a slim fork of `cygwin` but I'm not sure. :thinking: 

What I do is use the `git` installer and it comes with `MINGW64` with `bash` and a bunch of other things.

nhooyr (approved):
🚀

PR #2503, opened 2020-12-22
Update FAQ for proper hashed-password generation

confusedguy9:
Get rid of the automatically inserted trailing newline.


jsjoeio:
Thanks for the doc fix @confusedguy9! Do you mind changing your base branch to `master`? Then I can approve and merge this. Thank you!

confusedguy9:
I've changed the base branch as requested.

nhooyr:
Let's use `printf` here instead of `echo`! That way the same example should work correctly across all unix based OSes like the BSDs as well.

code-asher:
Thanks again for the PR! I went ahead and merged https://github.com/cdr/code-server/pull/2533 which has the `printf` fix.

PR #2532, opened 2020-12-30
Fix close folder/workspace

### LINKED ISSUE #2337: Folder in Explorer can not be closed
- Web Browser:chrome
- Local OS:mac os11
- Remote OS:Debian 10.1
- Remote Architecture:x86
- `code-server --version`:3.6.2

When I click "File" - "Close folder",but the folder still can not be closed in Explorer

rdbeach:
When you try to close the current folder, it doesn't close properly because the lastVisited setting is still used. This checks the Empty Window (ew) parameter and ignores lastVisited if true.

Fixes #2337


code-asher:
Awesome, thanks for fixing this!

Do you think we should clear `lastVisited` from the settings file as well? Otherwise if you close the folder then go back later (without the `ew` parameter) the last visited directory would open again. I'm not sure which is the expected behavior though.

rdbeach:
Yes. I didn't think about that.

nhooyr:
Thanks for contributing @rdbeach 

I'm going to merge your PR into `rdbeach` and then make some changes into that branch to get this ready for merge.

PR #2543, opened 2021-01-06
Vscode 1.52.1

### LINKED ISSUE #2319: VIM Extension is not working
- Web Browser: NAVER Whale 2.8.108.15
- Local OS: macOS Catalina Version 10.15.7
- Remote OS: CentOS Linux release 7.4.1708
- Remote Architecture: x86_64
- `code-server --version`: 3.7.1 2bf91ff6a6d1f29a8a93f97684177ccf5388cf34

It works in code-server 3.6.0, but after I upgraded it 3.7.1, keymaps of VIM are not working. I tried other browsers like safari or firefox, they don't work as well. I tried 3.6.2, 3.6.1, none of them are working.. vim extension is only working at 3.6.0 . The version of my vim extension is 1.6, it is the latest and I tried to run code-server with --log trace option to get an information of extensions, there are nothing worth.

please let me know how to fix it?

### LINKED ISSUE #2502: Upgrade Visual Studio Code to V1.52.0
Right now, there is a problem with installing the C++ extension on a raspberry pi due to the incompatibility with the ARM64 architecture and the fixed extension for ARM64 architecture is released but will not be compatible with the current Vs Code version(V1.51.1).

code-asher:
- [X] Update to 1.52.1
- [x] Fix web extensions
- [x] Test everything else

Closes #2502 
Fixes #2319 

I'm also not sure how to generate a diff showing the conflicts I had to resolve.


code-asher:
Working on updating our marketplace to support web extensions then this should be good to go. Big props to @jsjoeio for discovering that vscodevim has native web extension support now which allowed us to remove our entire browser Node shim.

code-asher:
With the marketplace changes deployed this is good to go.

I committed the conflict markers and resolved them in subsequent commits because it doesn't seem possible to view the changes I had to make to resolve conflicts; all I could get was the entire changeset which is obviously not helpful for review.

The original branch is here in case someone wants to take a whack at deriving conflict resolution diffs from it: https://github.com/cdr/code-server/tree/vscode-1.52.1-conflict-test

jsjoeio:
Wow, this PR has 1,363 files changed. @code-asher is anything in particular you want us to look at?

code-asher:
Yeah, almost all of those will be from the squashed update. Check out
each commit individually after the "Update VS Code" commit.

code-asher:
To provide a little more context, most of the changes I made were just resolving conflicts caused by the update, so basically the goal there is to make sure that the functionality we added/modified still works with the new code.

Some of these changes I think aren't well documented (or documented at all) as to why we made them, so if you have any questions definitely hit me with them and I'll update the commit messages and/or add comments.

lib/vscode/src/vs/workbench/services/extensionManagement/common/extensionManagementService.ts
```diff
@@ -304,17 +305,13 @@ export class ExtensionManagementService extends Disposable implements IWorkbench
 			}
 		}
 
-<<<<<<< HEAD
-		// NOTE@coder: Fall back to installing on the remote server.
-		if (this.extensionManagementServerService.remoteExtensionManagementServer) {
+		// NOTE@coder: Fall back to installing on the remote server on web.
```
jsjoeio, line 308: Sweet!

jsjoeio (commented):
> Remove GoHomeContributor
Can you confirm that the "Go Home" button is still in the menu?

Other than that, looks good to me!

nhooyr (approved):
To the moon! 🚀 

Echoing @jsjoeio's comment though to confirm that the go home button is still registered.

code-asher:
The home button shows up above the sidebar hamburger menu icon but they
removed it from the dropdown menu, so we'll need to update the e2e test
to use the button instead of the menu.

jsjoeio:
> The home button shows up above the sidebar hamburger menu icon but they
removed it from the dropdown menu, so we'll need to update the e2e test
to use the button instead of the menu.

Got it! Don't worry about updating the test - that can be on me once this gets merged. Lmk if you need anything else to help move this forward!

code-asher:
Awesome, thanks! I'll go ahead and resolve the conflict and merge this
up.

PR #2563, opened 2021-01-11
pathProxy.ts: Implement --proxy-path-passthrough

### LINKED ISSUE #2222: code-server's reverse proxy seems to always return index.html for all resources
- Web Browser: Chrome 86
- Local OS: Mac OS X
- Remote OS: Ubuntu 20.04 (using docker)
- Remote Architecture:
- `code-server --version`: 3.5.0 de41646fc402b968ca6d555fdf2da7de9554d28a

I'm running a vanilla react project created via `create-react-app`. I'm trying to use code-server's built-in proxy to view my application. I set my sites 'website' field as expected by react to be `http://<code-server-machine-ip>:8443/proxy/3000`. However, when I go to that address, every one of the assets loaded (all .js files, the favicon) seem to be the index.html. I get an error like this in the console: 

Uncaught SyntaxError: Unexpected token '<' bundle.js:1 

Screen Shot 2020-10-21 at 9 36 54 PM

This feels like a bug in the proxy layer of code-server, but I'm hoping I just have something configured incorrectly. I'm running code-server via docker image if that makes a difference.

nhooyr:
Closes #2222


nhooyr:
Even with all of this it doesn't work perfectly...

the `sockjs-node` connection from webpack doesn't use create-react-app's homepage and so is still absolute. I think I'll implement this and also the Referrer solution I referred to in #2222. It's the only way. Very unfortunate.

doc/FAQ.md
```diff
@@ -201,6 +201,36 @@ code-server --proxy-domain <domain>
 Now you can browse to `<port>.<domain>`. Note that this uses the host header so
 ensure your reverse proxy forwards that information if you are using one.
 
+## Why does the code-server proxy strip `/proxy/<port>` from the request path?
```
jsjoeio, line 204: Thanks for adding a detailed explanation for this!

src/node/routes/pathProxy.ts
```diff
@@ -8,12 +8,12 @@ import { Router as WsRouter } from "../wsRouter"
...
-  if (rewrite) {
-    const query = qs.stringify(req.query)
-    return `http://0.0.0.0:${req.params.port}/${req.params[0] || ""}${query ? `?${query}` : ""}`
+const getProxyTarget = (req: Request, passthroughPath: boolean): string => {
+  if (passthroughPath) {
+    return `http://0.0.0.0:${req.params.port}/${req.originalUrl}`
   }
-  return `http://0.0.0.0:${req.params.port}/${req.originalUrl}`
+  const query = qs.stringify(req.query)
+  return `http://0.0.0.0:${req.params.port}/${req.params[0] || ""}${query ? `?${query}` : ""}`
```
jsjoeio, line 16: Should we add a test for this?

jsjoeio: I.e. for the function, like a unit test

src/node/cli.ts
```diff
@@ -50,6 +50,7 @@ export interface Args extends VsArgs {
   "show-versions"?: boolean
   "uninstall-extension"?: string[]
   "proxy-domain"?: string[]
+  "proxy-path-passthrough"?: boolean
```
jsjoeio, line 53: Should we add a test to `cli.test.ts`?

jsjoeio (commented):
Looks good to me! Great work 🎉 

Do we want to write tests for this?

jsjoeio:
I tested locally (following the docs) and it works! 🌮

nhooyr:
Unfortunately Referrer isn't set on websocket requests and not always set in general so not something we can rely on. `create-react-app` will have to fix that to have webpack-dev-server properly work through our proxy. Super unfortunate.

src/node/routes/pathProxy.ts
nhooyr: Will do!

src/node/cli.ts
nhooyr: Will do!

src/node/routes/pathProxy.ts
nhooyr: nvm, I think in this case it's super simple, there's really no room for error. the existing cli tests test more involved logic and complex flags.

nhooyr: Whoops, wrong comment, meant to comment to below.

src/node/cli.ts
nhooyr: nvm, I think in this case it's super simple, there's really no room for error. the existing cli tests test more involved logic and complex flags.

src/node/routes/pathProxy.ts
jsjoeio: Sounds good!

jsjoeio (approved):

jsjoeio:
> Unfortunately Referrer isn't set on websocket requests and not always set in general so not something we can rely on. `create-react-app` will have to fix that to have webpack-dev-server properly work through our proxy. Super unfortunate.

Do you plan to open up an issue on the `create-react-app`? (If there isn’t one already)

nhooyr:
@jsjoeio yup see #2565!

also remember don't review until your review is requested again to avoid reviewing WIP changes.
I still have to write the test you requested.

nhooyr:
I also converted back to a draft, didn't know I could do that.

nhooyr:
Tests nearly done! but there's some bizarre fd leak somewhere I need to fix still :(

nhooyr:
This should be good now @jsjoeio!

nhooyr:
omg, i see now, it was `leaked-handles`. confusing me! updating the commit message.

ci/dev/gen_icons.sh
```diff
@@ -14,10 +14,31 @@ main() {
...
+}
+</style>"
+  # See https://stackoverflow.com/a/22901380/4283659
+  # This escapes all newlines so that sed will accept them.
+  favicon_dark_style="$(printf "%s\n" "$favicon_dark_style" | sed -e ':a' -e 'N' -e '$!ba' -e 's/\n/\\n/g')"
+  sed "$(
+    cat -n << EOF
+s%<rect id="favicon"%$favicon_dark_style<rect id="favicon"%
+EOF
+  )" favicon.svg > favicon-dark-support.svg
```
jsjoeio, line 41: I remember seeing something about you updating the favicon - are you doing those changes in this PR as well?

src/common/util.ts
```diff
@@ -112,3 +112,11 @@ export const getFirstString = (value: string | string[] | object | undefined): s
 
   return typeof value === "string" ? value : undefined
 }
+
+export function logError(prefix: string, err: any): void {
```
jsjoeio, line 116: What type should `err` be here? `Error | string`?

src/node/app.ts
```diff
@@ -22,8 +23,21 @@ export const createApp = async (args: DefaultedArgs): Promise<[Express, Express,
...
+    }
+    server.on("error", (err) => {
+      if (!resolved) {
+        reject(err)
+      } else {
+        // Promise resolved earlier so this is an unrelated error.
+        util.logError("http server error", err)
+      }
+    })
+
```
jsjoeio, line 40: 👏🏼

src/node/routes/index.ts
```diff
@@ -55,6 +55,9 @@ export const register = async (
       })
     })
   })
+  server.on("close", () => {
+    heart.dispose()
+  })
```
jsjoeio, line 60: Is this related to this PR? I thought I saw something about the heart beat error - guessing this helps with that?

src/node/routes/pathProxy.ts
```diff
@@ -8,12 +8,12 @@ import { Router as WsRouter } from "../wsRouter"
 
 export const router = Router()
 
-const getProxyTarget = (req: Request, rewrite: boolean): string => {
-  if (rewrite) {
-    const query = qs.stringify(req.query)
-    return `http://0.0.0.0:${req.params.port}/${req.params[0] || ""}${query ? `?${query}` : ""}`
+const getProxyTarget = (req: Request, passthroughPath: boolean): string => {
+  if (passthroughPath) {
```
jsjoeio, line 12: 👏🏼

test/proxy.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    assert.equal(await resp.json(), "asher is the best")
+  })
+
+  it("should not rewrite the base path", async () => {
+    ;[, , codeServer] = await integration.setup(["--auth=none", "--proxy-path-passthrough=true"], "")
+    const resp = await codeServer.fetch(proxyPath)
+    assert.equal(resp.status, 200)
+    assert.equal(await resp.json(), "joe is the best")
+  })
+})
```
jsjoeio, line 47: Beautiful tests! 

LOL at the response 😂

jsjoeio:
@nhooyr great work and great tests! Left one minor nit/question about an `any` TS type but non-blocking.

Is this PR intentionally fixing other issues? I see stuff related the the dark mode favicon and the heartbeat error that looks like you fixed. Just wanted to make sure! Otherwise, I can approve

code-asher:
Wooo this is fabulous; the tests are 🔥 

For some history on `leaked-handles`: if I recall correctly I had issues with code-server or one of the child processes not being killed with a `SIGTERM` or something due to the socket hanging open, I think. So I added this, tracked down the leaked handles, and it resolved the issue.

Or actually it might have been keeping the tests themselves from finishing? I don't quite recall. Something was hanging somewhere.

With all that said I feel a little more confident if it stayed in but I don't feel super strongly either way.

ci/dev/gen_icons.sh
nhooyr: Nope, something's wrong with my branch. Git is too hard :(

src/common/util.ts
nhooyr: So where this is used err is any as it's logging anything thrown.

src/node/routes/index.ts
nhooyr: See the commit this was introduced in. Always review by commit otherwise you'll miss the context behind changes. This change is from `heart.ts: Fix leak when server closes`.

test/proxy.test.ts
nhooyr: Why thank you sir! <3

nhooyr:
> Is this PR intentionally fixing other issues? I see stuff related the the dark mode favicon and the heartbeat error that looks like you fixed. Just wanted to make sure! Otherwise, I can approve

Yea I screwed up this branch somehow. The favicon commits are unrelated but see my comment re the heartbeat changes.

nhooyr:
> With all that said I feel a little more confident if it stayed in but I don't feel super strongly either way.

Happy to leave it in if we can modify the message to indicate it's from that package. Otherwise, it's just confusing if the leak is elsewhere.

src/node/routes/index.ts
nhooyr: These three lines fixed the leak preventing the tests from returning by disabling the heartbeat timer.

nhooyr:
Fixed the branch so the favicon changes aren't part of the diff anymore! Was based on a very old version of master sorry!

nhooyr:
> Happy to leave it in if we can modify the message to indicate it's from that package. Otherwise, it's just confusing if the leak is elsewhere.

Looks like we can't :(

And https://github.com/myndzi/wtfnode is the preferred utility anyway and has much better output. Opening a PR to switch.

nhooyr:
Also unfortunate and confusing thing about leaked-handles is that it's usage is global as it begins on import instead of letting you call a method to enable. I guess you could require as needed but dynamic requires are a little more annoying. wtfnode just has a global function you call to begin the dumping.

nhooyr:
Oo and we can configure the output further if ever needed! https://github.com/myndzi/wtfnode#configuring-logging

src/common/util.ts
nhooyr: Nvm, my bad I was wrong, will change to `Error` as that's all that seems necessary for now. Good catch!

nhooyr: Though tbh, technically it does accept `any`, there's no reason it has to be `Error` or `string`. Anything else will just be stringified.

nhooyr: Let me know if you still want it changed and I can make a follow up PR.

nhooyr:
Awesome, thanks for the reviews guys!

Issue #2550, opened 2021-01-08
Migrate from Mocha to Jest


jsjoeio:
As discussed with @code-asher and @nhooyr, we've decided to migrate from Mocha to Jest as our testing framework of choice for the following reasons:
- Jest has a larger community, more active and more popular in the JS ecosystem
- more members of the team are familiar with it
- built-in support for code coverage

This issue is for transparency and to track the migration.

## TODOS
- [x] Add `jest` and make sure it works ([`ts-jest`](https://github.com/kulshekhar/ts-jest) Is one of the recommended approaches, easy setup)
- [x] Use codemods to migrate mocha tests over ([helpful resource](https://lirantal.medium.com/migrating-a-mocha-project-to-jest-test-framework-76d13d76685))
- [x] Remove mocha fully
- [x] Add code coverage (including shield to `README`)

jsjoeio:
Closing via #2564

PR #2564, opened 2021-01-11
refactor(tests): migrate from mocha to jest

jsjoeio:
This migrates the testing solution from mocha to jest. We were using some deprecated mocha methods in our tests so those are fixed now.

#2621 was merged into this and includes code-coverage

## Screenshot
B36025DD-780B-43BA-83A2-A402CAC8212A
 
TODOS
- [x] try out @nhooyr suggestion "...a better option might be to ignore the .d.ts file of mocha from being type checked at all. iirc there's some exclude fields in tsconfig right" to see if that's a viable solution
- [x] fix CI issues
- [x] sign old commits and clean up

Related:
- #2550 
- #2621


jsjoeio:
Hmm 🤔 Looks like TS is unhappy with duplicate identifiers. I think it’s because mocha is in the subtree and jest is in our repo and both use the same identifiers in their type definitions. I’ll see what workarounds there may be. 

D9A3100C-C869-4850-9960-C0FB3FEC8A13

jsjoeio:
Okay from quick research seems like there are some things we can try:
- adding `skipLibCheck` to the `tsconfig` (Probably in `lib/vscode`)
- using `jest-without-globals` in our repo ([link](https://www.npmjs.com/package/jest-without-globals))

nhooyr:
This PR is on the wrong base @jsjoeio

jsjoeio:
> This PR is on the wrong base @jsjoeio
Oops, fixed.

lib/vscode/build/tsconfig.json
```diff
@@ -15,7 +15,8 @@
 		"strict": true,
 		"noUnusedLocals": true,
 		"noUnusedParameters": true,
-		"newLine": "lf"
+		"newLine": "lf",
+		"skipLibCheck": true
```
jsjoeio, line 19: I added this to prevent TypeScript from checking the declaration files at build time. This resolves the @types issue with mocha and jest using the same “test” namespace in their declaration files. According to the handbook:
> Skip type checking of declaration files.
> This can save time during compilation at the expense of type-system accuracy

The recommended setting is `true`. I think this is safe to change but appreciate input from y’all @nhooyr @code-asher 

Suggested reading:
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/43912
- [TSConfig Reference: skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck)

nhooyr: Doesn't this mean we have no type checking now or does it mean we only get type checking when linting?

nhooyr: Ah sorry I see. Yea I don't think we can do this, we rely on declaration files throughout and I'm sure Microsoft does as well.

nhooyr: Has this been fixed @jsjoeio?

jsjoeio: Discussed on Slack! I will revert to draft, fix CI, then mark ready and re-request a review

lib/vscode/build/lib/test/i18n.test.ts
```diff
@@ -6,7 +6,7 @@
 import assert = require('assert');
 import i18n = require('../i18n');
 
-suite('XLF Parser Tests', () => {
+describe('XLF Parser Tests', () => {
```
jsjoeio, line 9: After removing `mocha` I had to convert this to `describe` because jest doesn't have `suite` as a function.

jsjoeio:
Once/if CI passes, I will sign older commits and clean them up.

lib/vscode/build/tsconfig.json
nhooyr: Discussed again and it looks like we can't do anything better than skipLibCheck until https://github.com/microsoft/TypeScript/issues/37053 is resolved by the tsc team.

jsjoeio:
Seeing some weird errors in release.

image

Possible solution: [stackoverflow](https://stackoverflow.com/questions/63499090/error-node-modules-jest-diff-build-difflines-d-ts8-13-error-ts1005-expec)

test/socket.test.ts
```diff
@@ -110,10 +111,12 @@ describe("SocketProxyProvider", () => {
     provider.stop() // We don't need more proxies.
 
     proxy.write("server proxy->client")
-    assert.equal(await getData(fromServerToClient), "server proxy->client")
+    const dataFromServerToClient = await (await getData(fromServerToClient)).toString()
+    expect(dataFromServerToClient).toBe("server proxy->client")
     client.write("client->server proxy")
-    assert.equal(await getData(fromClientToProxy), "client->server proxy")
-    assert.equal(errors, 0)
```
nhooyr, line 116: Why were these changes required? Is there any advantage to using jest's assertions over builtin `assert`?

jsjoeio: Not required, more so for consistency (use `expect` + Jest throughout). Also when someone is looking up documentation, they only have to look in one place (jest docs vs jest docs + assert docs).

nhooyr: Sounds good.

nhooyr:
That's bizarre, why is the release process trying to build jest? cc @code-asher

ci/dev/test.sh
```diff
@@ -6,8 +6,9 @@ main() {
 
   cd test/test-plugin
   make -s out/index.js
-  cd "$OLDPWD"
-  mocha -r ts-node/register ./test/*.test.ts "$@"
+  cd "$OLDPWD/test"
+  yarn
+  yarn jest "$@"
```
jsjoeio, line 11:
```suggestion
  yarn test "$@"
```

lib/vscode/extensions/typescript-language-features/src/utils/platform.ts
```diff
@@ -7,5 +7,6 @@ import * as vscode from 'vscode';
 
 export function isWeb(): boolean {
 	// NOTE@coder: Remove unused ts-expect-error directive which causes tsc to error.
```
jsjoeio, line 9:
```suggestion
```

test/update.test.ts
```diff
@@ -101,53 +100,54 @@ describe.skip("update", () => {
...
     update = await p.getUpdate(true)
-    assert.equal(p.isLatestVersion(update), true)
+    expect(p.isLatestVersion(update)).toBe(true)
   })
 
   it("should not reject if unable to fetch", async () => {
+    expect.assertions(2)
     let provider = new UpdateProvider("invalid", settings)
-    await assert.doesNotReject(() => provider.getUpdate(true))
+    await expect(() => provider.getUpdate(true)).resolves.toBe(undefined)
```
code-asher, line 148: Does Jest automatically handle if this is a function or does it need to be `expect(provider.getUpdate(true))`?

code-asher (approved):
Looks good!

test/update.test.ts
jsjoeio: Good question. Let me take a closer look!

test/socket.test.ts
```diff
@@ -93,14 +92,16 @@ describe("SocketProxyProvider", () => {
 
   it("should work without a proxy", async () => {
     server.write("server->client")
-    assert.equal(await getData(fromServerToClient), "server->client")
+    const dataFromServerToClient = await (await getData(fromServerToClient)).toString()
```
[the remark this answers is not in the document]
jsjoeio: Ah! Probably not. Good catch! I'll take a closer look

test/update.test.ts
jsjoeio: Two things:
1. this whole test suite is `skip`ed - not sure when that was, but I will eventually look into adding it back in.
2. Because `provider.getUpdate` returns a promise, we have to pass `expect` a function than assert with `.resolves` so this is correct

source: https://jestjs.io/docs/en/tutorial-async#asyncawait

test/socket.test.ts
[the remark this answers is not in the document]
jsjoeio: Updated the types and removed the unnecessary `await` - good catch! 

Also, seems I have to convert `Buffer` to `string` or I get back characters? 🤷‍♂️ I'll leave that part as is..

image

test/update.test.ts
code-asher: I thought the same thing from a long time ago when I used Jest once but in that link and in their docs they have examples like:

```
 await expect(user.getUserName(5)).resolves.toEqual('Paul');
```

Which doesn't pass a function.

We have several promise-based `expect` calls in other tests that don't use a function as well. For example in the cli tests we have:

```
    await expect(shouldOpenInExistingInstance(args)).resolves.toStrictEqual("test")
```

Maybe both methods work? Or one of them is wrong lol, idk

test/socket.test.ts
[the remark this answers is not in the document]
code-asher: Yup! You'll still need the `toString`. I guess the `assert` library automatically does that conversion for some reason. Or maybe it's because it wasn't a strict equals.

test/update.test.ts
jsjoeio: OH! I see what you mean. I wonder if I did that because the promise resolves to undefined which throws an error? Maybe not! I'll write a note down to go through and look at that again. Sorry I didn't understand what you meant at first. Good eye!

PR #2604, opened 2021-01-20
test: Switch from leaked-handles to wtfnode

nhooyr:
See my comments at
https://github.com/cdr/code-server/pull/2563#issuecomment-763394741


code-asher (approved):
This is much better!

PR #2620, opened 2021-01-22
Allow use of multiple extension galleries

oxy:
This PR should implement support for using multiple extension galleries by default.
There's quite a few things that may be slightly hacky about it, since VSCode's extension gallery design is confusingly broken:

- `galleryUrl` is used for `/extensionquery` and uninstall statistics
- `itemUrl` is supposedly used for linking to item pages but is not essential for functionality (and our gallery doesn't support it) so I just left it out
- `controlUrl` appears to be used for reporting malicious extensions, neither OpenVSX nor our gallery supports this
- `recommendationsUrl` provides recommendations but neither OpenVSX nor our gallery supports this

Awkward thing here is that `galleryUrl` on first glance appears to be the base URL for the whole API, and internally `/extensionquery` is appended to it, but then they go ahead and include separate URLs for other extension gallery functionality - which is quite confusing.

So, I just changed the type of `galleryUrl` from `string` to `string[]` and patched the methods that read from it.

It might also be possible to change the type of `extensionGallery` to a list of objects instead of an object with a list of `galleryUrl` - but given that the other variables aren't really used by either of the two galleries we plan to primarily support (ours and OpenVSX's), so I tried to minimize engineering time there.

I've tested the result by installing extensions from both OpenVSX and our repo - and it works fine!

Also, right now, I don't actually resort results to merge the two lists; it's just all of OpenVSX's first page, with all of Coder's (with duplicates filtered) after that.

### DIFF
--- a/lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
+++ b/lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
@@ -343,7 +343,7 @@ export class ExtensionGalleryService implements IExtensionGalleryService {
 
 	declare readonly _serviceBrand: undefined;
 
-	private extensionsGalleryUrl: string | undefined;
+	private extensionsGalleryUrl: string[] | undefined;
 	private extensionsControlUrl: string | undefined;
 
 	private readonly commonHeadersPromise: Promise<{ [key: string]: string; }>;
@@ -363,8 +363,13 @@ export class ExtensionGalleryService implements IExtensionGalleryService {
 		this.commonHeadersPromise = resolveMarketplaceHeaders(productService.version, this.environmentService, this.fileService, storageService);
 	}
 
-	private api(path = ''): string {
-		return `${this.extensionsGalleryUrl}${path}`;
+	private api(path = ''): string[] {
+		if (!!this.extensionsGalleryUrl) {
+			return this.extensionsGalleryUrl?.map(
+				(galleryUrl) => `${galleryUrl}${path}`
+			);
+		}
+		return [];
 	}
 
 	isEnabled(): boolean {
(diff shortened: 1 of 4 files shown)


jsjoeio:
@oxy You should be able to fix the `navigator` error with this: https://github.com/cdr/code-server/blob/883dd13850cebc83531dfb79da6d15c2f89841a7/lib/vscode/extensions/typescript-language-features/src/utils/platform.ts#L9

jsjoeio:
> Also, right now, I don't actually resort results to merge the two lists; it's just all of OpenVSX's first page, with all of Coder's (with duplicates filtered) after that.

Just to clarify here - are only the first page of OpenVSX's extensions available with this change? Or are all of them available?

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
```diff
@@ -363,8 +363,13 @@ export class ExtensionGalleryService implements IExtensionGalleryService {
 		this.commonHeadersPromise = resolveMarketplaceHeaders(productService.version, this.environmentService, this.fileService, storageService);
 	}
 
-	private api(path = ''): string {
-		return `${this.extensionsGalleryUrl}${path}`;
+	private api(path = ''): string[] {
+		if (!!this.extensionsGalleryUrl) {
```
jsjoeio, line 367: Is the double `!` to convert it to a boolean? Does it have a performance advantage over the [Boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean) function? Only asking/suggesting if it might improve readability - obviously your call :D 
```suggestion
		if (Boolean(this.extensionsGalleryUrl)) {
```

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
```diff
@@ -543,43 +548,53 @@ export class ExtensionGalleryService implements IExtensionGalleryService {
...
+					}
+				}
+			}
 		}
-		return { galleryExtensions: [], total: 0 };
+		return { galleryExtensions, total: galleryExtensions.length };
 	}
 
 	async reportStatistic(publisher: string, name: string, version: string, type: StatisticType): Promise<void> {
+		// TODO: investigate further - currently we just send stats everywhere
```
jsjoeio, line 581: I know folks have GitLens and tools to know who left a TODO comment but at least this puts your name right there so we don't forget. 

```suggestion
		// TODO(oxy): investigate further - currently we just send stats everywhere
```

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
```diff
@@ -543,43 +548,53 @@ export class ExtensionGalleryService implements IExtensionGalleryService {
...
+		// this is only used in one place - uninstall tracking - but
```
jsjoeio, line 582: Unfinished comment? Was there more to this sentence?

jsjoeio (approved):
Great work @oxy 🎉 I'm so excited for this. I left a few minor non-blocking suggestions and questions. Might be good to get an eye from someone else on the team just as extra assurance.

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
oxy: Oops, there was!
was supposed to say `but unsure if MS will start using this elsewhere`

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
oxy: Perhaps I should just use `this.isEnabled()` instead - does the same check in practice but is probably ideal since the code already exists.

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
oxy: :+1: will fix!

lib/vscode/src/vs/platform/extensionManagement/common/extensionGalleryService.ts
jsjoeio: Good thinking!

oxy:
We've decided on an alternate approach for now; see #2659 and #2672.

PR #2622, opened 2021-01-22
Plugin additions

code-asher:
The main thing here was adding support for websocket routes and for disposing. The rest was mostly exposing some things so plugins don't have to recreate or reimport them.


src/node/plugin.ts
```diff
@@ -225,12 +256,28 @@ export class PluginAPI {
...
+  public async dispose(): Promise<void> {
+    await Promise.all(
+      Array.from(this.plugins.values()).map(async (p) => {
+        if (!p.deinit) {
+          return
+        }
+        try {
+          await p.deinit()
+        } catch (error) {
+          this.logger.error("plugin failed to deinit", field("name", p.name), field("error", error.message))
```
nhooyr, line 276: Is it safe to assume `.message` exists on `error`?

typings/pluginapi.d.ts
```diff
@@ -78,6 +82,75 @@ import * as express from "express"
...
+ * to avoid potential version differences or issues caused by running separate
+ * instances.
+ */
+export { express }
+/**
+ * Use to add a field to a log.
+ *
+ * Re-exported so plugins don't have to import duplicate copies of the logger.
+ */
+export { field, Level, Logger }
```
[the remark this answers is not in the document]
nhooyr: Or a util module?

nhooyr (changes requested):

typings/pluginapi.d.ts
```diff
@@ -78,6 +82,75 @@ import * as express from "express"
...
  */
 
+export enum HttpCode {
+  Ok = 200,
+  Redirect = 302,
+  NotFound = 404,
+  BadRequest = 400,
+  Unauthorized = 401,
+  LargePayload = 413,
+  ServerError = 500,
```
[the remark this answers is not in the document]
code-asher: Without these the types don't work from the plugin side. :(

typings/pluginapi.d.ts
```diff
@@ -156,6 +243,11 @@ export interface PluginConfig {
    * All plugin logs should be logged via this logger.
    */
   readonly logger: Logger
+
+  /**
+   * Plugins should default to this directory when applicable.
```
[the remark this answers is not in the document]
code-asher: I wasn't sure how to word this better, definitely open to suggestions. 

For example, for the Jupyter plugin this means launching Jupyter in this directory so when it opens that's the directory that shows up.

For other plugins it might mean something different though.

The difference from `process.cwd` is that it can be set by the user (`code-server working/directory`).

typings/pluginapi.d.ts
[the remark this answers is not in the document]
code-asher: I'm down to organize this differently. The only upshot to leaving it that I can think of is that the usage is a bit shorter, i.e. `cs.field` instead of `cs.util.field` if you have `import * as cs from "code-server"`.

Oh definitely down to do something like `import * as util from "code-server/util"`, is that what you mean?

src/node/plugin.ts
code-asher: Nope lol

I'll run it through `logError` instead.

I do wish we could enforce the throw type with TypeScript though.

typings/pluginapi.d.ts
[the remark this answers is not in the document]
code-asher: I suppose having it all on the one import might be more ergonomic though? In which case I'm down for the `util` object or namespace.

jsjoeio:
Dropping my plugin thoughts here.

## Overall

We want to make the code-server plugin ecosystem as intuitive and easy as possible for plugin authors. Sub-modules are a headache so we shouldn't go that route.

I think this [new architecture](https://github.com/cdr/code-server-example-plugin/pull/2/files) (repo is private) looks like the right direction for us to go in. 


### Education
We probably want to work with @bpmct to create some education around code-server plugins. We'll want it to live in both the `/docs` but also maybe create content on the Coder blog. 
- What are they? 
- How do they compare to VSCode Extensions?
- How do you create one? 

### Further Reading
As we continue to iterate on top of the plugin ecosystem, we should take inspiration from other projects. 
- ["Zero Config First, Full Config Later" Ejectable Defaults](https://www.swyx.io/ejectable-defaults)

.eslintrc.yaml
```diff
@@ -35,7 +35,7 @@ rules:
     [error, { alphabetize: { order: "asc" }, groups: [["builtin", "external", "internal"], "parent", "sibling"] }]
   no-async-promise-executor: off
   # This isn't a real module, just types, which apparently doesn't resolve.
-  import/no-unresolved: [error, { ignore: ["express-serve-static-core"] }]
+  import/no-unresolved: [error, { ignore: ["express-serve-static-core", "code-server"] }]
```
jsjoeio, line 38: [asking to learn/understand]: why is this needed?

src/node/plugin.ts
```diff
@@ -1,13 +1,40 @@
...
+import { Router as WsRouter, WebsocketRouter, wss } from "./wsRouter"
 const fsp = fs.promises
 
+/**
+ * Inject code-server when `require`d. This is required because the API provides
+ * more than just types so these need to be provided at run-time.
+ */
+const originalLoad = require("module")._load
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+require("module")._load = function (request: string, parent: object, isMain: boolean): any {
```
jsjoeio, line 21: What does this actually return? Would it be safer to add that as the type instead of `any`? Or maybe we can create something like:
```typescript
type FixMeLater = any
```

src/node/plugin.ts
jsjoeio: > I do wish we could enforce the throw type with TypeScript though.

Is this possible?

src/node/routes/index.ts
```diff
@@ -127,10 +119,12 @@ export const register = async (
   app.use("/static", _static.router)
   app.use("/update", update.router)
 
-  const papi = new PluginAPI(logger, process.env.CS_PLUGIN, process.env.CS_PLUGIN_PATH)
+  const workingDir = args._ && args._.length > 0 ? path.resolve(args._[args._.length - 1]) : undefined
+  const papi = new PluginAPI(logger, process.env.CS_PLUGIN, process.env.CS_PLUGIN_PATH, workingDir)
```
jsjoeio, line 123: nit: I know this isn't a new addition from your PR but wish we could stick to actual names, like calling this `pluginAPI` instead of `papi` lol

src/node/routes/index.ts
```diff
@@ -170,7 +164,7 @@ export const register = async (
 
   const wsErrorHandler: express.ErrorRequestHandler = async (err, req, res, next) => {
     logger.error(`${err.message} ${err.stack}`)
-    ;(req as WebsocketRequest).ws.end()
+    ;(req as pluginapi.WebsocketRequest).ws.end()
```
jsjoeio, line 167: question: why do we have to assert this? (previous work, but maybe we could add a comment explaining why)

test/httpserver.ts
```diff
@@ -62,6 +72,13 @@ export class HttpServer {
     return nodeFetch.default(`${ensureAddress(this.hs)}${requestPath}`, opts)
   }
 
+  /**
+   * Open a websocket against the requset path.
```
jsjoeio, line 76:
```suggestion
   * Open a websocket against the request path.
```

jsjoeio: [asking to learn/understand]: why do we open a new websocket against the request path?

test/plugin.test.ts
```diff
@@ -70,4 +75,18 @@ describe("plugin", () => {
...
+      ws.once("message", (message) => resolve(message))
+    })
+    ws.terminate()
+    assert.strictEqual(message, "hello")
+  })
+
+  it("/test-plugin/error", async () => {
+    const resp = await s.fetch("/test-plugin/error")
+    assert.equal(HttpCode.LargePayload, resp.status)
+  })
```
jsjoeio, line 91: yay for testing! 👏

typings/pluginapi.d.ts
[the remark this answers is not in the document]
jsjoeio: Then it's probably fine to leave these here

typings/pluginapi.d.ts
```diff
@@ -78,6 +82,75 @@ import * as express from "express"
...
+ * The websocket server used by code-server.
+ */
+export const wss: Websocket.Server
+
+/**
+ * The Express import used by code-server.
+ *
+ * Re-exported so plugins don't have to import duplicate copies of Express and
+ * to avoid potential version differences or issues caused by running separate
+ * instances.
```
jsjoeio, line 130: 👏

typings/pluginapi.d.ts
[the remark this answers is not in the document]
jsjoeio: Is there any large reason (performance or otherwise) to move it? 

Otherwise, probably makes for a nice API for the plugin authors calling `cs.field` etc.

typings/pluginapi.d.ts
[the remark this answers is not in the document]
jsjoeio: Mind adding this comment ^ to your code comment? I think doing that will provide enough explanation. We can iterate on it later as needed.

jsjoeio (approved):
Overall, looks great 🎉 I left a bunch of questions, comments but all are non-blocking.

bpmct:
@jsjoeio thanks for the heads up. We'll sync on this on Wednesday :)

src/node/plugin.ts
code-asher: Nope, not yet at least. https://github.com/microsoft/TypeScript/issues/13219

code-asher: And since anything anywhere can throw whatever it wants I'm not sure it'll even help that much lol

.eslintrc.yaml
code-asher: eslint can't find the types because `code-server` is an alias so I just ignored it because `tsc` resolves it just fine. But the real solution is to add the alias to eslint, so I'll do that.

src/node/plugin.ts
code-asher: `_load` is used by `require` to load a module which can actually be anything so unfortunately we can't narrow down the type.

code-asher: But yeah definitely down to use a type alias for the `any`.

src/node/routes/index.ts
code-asher: Hahaha I'm down for that.

src/node/routes/index.ts
code-asher: The `ErrorRequestHandler` type marks `req` as `Request` so we have to cast it as a websocket request.

But really what we should do is create our own `WsErrorRequestHandler` type so we don't have to do this.

test/httpserver.ts
code-asher: This is to test that the plugin was able to correctly register a websocket route at that path and respond.

src/node/routes/index.ts
code-asher: Creating a custom error handler type is turning out to be weirdly difficult so I'm going to leave this for now.

code-asher:
I rebased on master and resolved comments as well as fixed a problem with proxying inside a plugin (see the last five commits).

However, for some reason in Jest the `module._load` override doesn't work. I've tried a few things but no luck. It's like they have their own `require` mechanism for some insane reason.

test/plugin.test.ts
```diff
@@ -3,11 +3,14 @@ import * as express from "express"
...
 import * as path from "path"
 import { HttpCode } from "../src/common/http"
-import { PluginAPI } from "../src/node/plugin"
+import { codeServer, PluginAPI } from "../src/node/plugin"
 import * as apps from "../src/node/routes/apps"
 import * as httpserver from "./httpserver"
 const fsp = fs.promises
 
+// Jest overrides `require` so our usual override doesn't work.
+jest.mock("code-server", () => codeServer)
```
jsjoeio, line 12: the magic line!

test/plugin.test.ts
```diff
@@ -9,7 +9,7 @@ import * as httpserver from "./httpserver"
 const fsp = fs.promises
 
 // Jest overrides `require` so our usual override doesn't work.
-jest.mock("code-server", () => codeServer)
+jest.mock("code-server", () => codeServer, { virtual: true })
```
jsjoeio, line 12: > virtual mocks – mocks of modules that don't exist anywhere in the system

🤯 TIL what a `virtual` mock is

[Source](https://jest-bot.github.io/jest/docs/jest-object.html#jestmockmodulename-factory-options)

code-asher: Haha yeah idk why it was passing locally without this for me though. :shrug:

PR #2643, opened 2021-01-26
feat(testing): add playwright 

jsjoeio:
This PR adds [playwright](https://playwright.dev/) for end-to-end testing. In ci, we now download the code-server artifact, run code-server in ci and then run tests against it.

## Changes
- add `playwright` to `/test/package.json`
- add basic test for login page
- update `ci.yml` `test` job to work with `e2e`

## Additional Notes

Right now, `yarn test` runs both the unit and the e2e tests. This means it will fail locally if you do not have `code-server` running. In the future, we should consider separating the tests. We could have `test:unit` and `test:e2e` and then have separate jobs for each in ci.

Follow-up issue: #2646

## Todos
- [x] remove `pkill node` in ci job
- [x] clean up commit messages


.github/workflows/ci.yaml
```diff
@@ -29,6 +29,13 @@ jobs:
         uses: ./ci/images/debian10
         with:
           args: ./ci/steps/test.sh
+  e2e:
```
jsjoeio, line 32: note to self: it may be redudant to have a step for testing and end to end. since we'll use jest as the test runner for e2e tests, we can probably just make this work in one job

.github/workflows/ci.yaml
```diff
@@ -22,13 +22,30 @@ jobs:
...
+        uses: actions/download-artifact@v2
         with:
-          args: ./ci/steps/test.sh
+          name: release-packages
+          path: ./release-packages
+      - name: See working directory and files
+        run: |
+          echo "This better print"
+          pwd
+          ls
```
jsjoeio, line 38: This was only there for testing purposes 😅
```suggestion
```

.github/workflows/ci.yaml
```diff
@@ -22,13 +22,30 @@ jobs:
...
+      - name: Untar code-server file
+        run: |
+          tar -xzf release-packages/code-server*-linux-amd64.tar.gz
+      - uses: microsoft/playwright-github-action@v1
+      - name: Install dependencies and run tests
+        run: |
+          node code-server*-linux-amd64 &
```
jsjoeio, line 45: The last `&` is required to run the node process in the background

ci/images/playwright/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# Use the playwright Docker image
+FROM mcr.microsoft.com/playwright:bionic
+
+## NOTES
+# Somehow get code-server artifact
+# Run it locally on port 8080
+# so that it can be accessed in our tests
+
+# It should build the project
+# Start code server and then we should run the tests
```
jsjoeio, line 10: We no longer need this file.

test/e2e.test.ts
```diff
@@ -0,0 +1,22 @@
+import { chromium, Page, Browser } from "playwright"
+
+let browser: Browser
+let page: Page
```
jsjoeio, line 4:
```suggestion
let browser: Browser
let page: Page

```

test/e2e.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  await browser.close()
+})
+beforeEach(async () => {
+  page = await browser.newPage()
+})
+afterEach(async () => {
+  await page.close()
+})
+
+it("should work", async () => {
```
jsjoeio, line 18:
```suggestion
it("should see the code-server login page", async () => {
```

test/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up",
   "#": "the types and generates conflicts with mocha.",
```
jsjoeio, line 2:
```suggestion
  "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up",
  "#": "the types and generates conflicts with mocha.",
```

test/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up",
```
jsjoeio, line 2: Remove by accident

.gitignore
```diff
@@ -15,3 +15,4 @@ node-*
 /lib/coder-cloud-agent
 .home
 coverage
+code-server-*
```
jsjoeio, line 18: Added because if you detar the release package, you get code-server...and we don't want to accidentally commit that.

jsjoeio:
The macos error looks like it was a network connection error: 
image

.gitignore
oxy: We're missing a newline at the end of the file here!

oxy (commented):
Looks alright to me! Haven't worked much on testing so far - so you might want a second opinion.

.gitignore
code-asher: Maybe we should `cd release-packages && tar -xzf code-server*-linux-amd64.tar.gz` to keep the extraction inside the `release-packages` dir instead of in the root.

jsjoeio: Good idea! I will revert this and do that instead

code-asher (dismissed):
This is quite exciting. 😄

.github/workflows/ci.yaml
```diff
@@ -22,13 +22,25 @@ jobs:
...
-          args: ./ci/steps/test.sh
+          name: release-packages
+          path: ./release-packages
+      - name: Untar code-server file
+        run: |
+          cd release-packages && tar -xzf release-packages/code-server*-linux-amd64.tar.gz
+      - uses: microsoft/playwright-github-action@v1
+      - name: Install dependencies and run tests
+        run: |
+          cd release-packages && node code-server*-linux-amd64 &
```
code-asher, line 40:
```suggestion
          node ./release-packages/code-server*-linux-amd64 &
```

Otherwise the following `yarn` won't be run in the root. Could also do a `cd ..`.

jsjoeio: Good call!

jsjoeio:
Seeing strange errors in the macos release
image

image
image

jsjoeio:
Re-running tests

.github/workflows/ci.yaml
```diff
@@ -22,13 +22,25 @@ jobs:
...
+      - name: Untar code-server file
+        run: |
+          cd release-packages && tar -xzf code-server*-linux-amd64.tar.gz
+      - uses: microsoft/playwright-github-action@v1
+      - name: Install dependencies and run tests
+        run: |
+          node ./release-packages/code-server*-linux-amd64 &
+          yarn --frozen-lockfile
+          yarn test
+          pkill node
```
nhooyr, line 43: Don't need this, jobs that weren't disowned are automatically killed when the shell exits.

nhooyr (approved):
Very nice! :)

.github/workflows/ci.yaml
jsjoeio: Ah we don't need the `pkill node`? Cool! I'll remove it!

PR #2648, opened 2021-01-28
feat(testing): add e2e test for 'Go Home' button

jsjoeio:
This PR adds two e2e tests and makes a few modifications for testing.

## Changes
- adds a test for the login page
- adds a test for the 'Go Home' button 
- update jest config testTimeout

## Screenrecording

https://user-images.githubusercontent.com/3806031/107680878-a5896200-6c5b-11eb-8884-1e0e8512dbf7.mp4

## Todos
- [x] fix Ci to not use container `node`
- [x] make test deterministic 
- [x] make requested changes
- [x] update ignore paths for jest
- [x] fix `.gitignore` conflicts
- [x] fix `test.sh` script
- [x] try out setting cookie directly
- [x] double-check that we don't need to be setting `--home`
- [x] fix conflicts
- [x] update test after https://github.com/cdr/code-server/pull/2543 is merged, see [comment](https://github.com/cdr/code-server/pull/2543#issuecomment-775347594)
- [x] refactor to promises and remove done (see comment from Anmol below)
- [x] address PR feedback
- [x] rebase
- [x] fix CI failures (check locally)
  - [x] lint
  - [x] fmt
  - [x] release


.github/workflows/ci.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       - uses: microsoft/playwright-github-action@v1
       - name: Install dependencies and run tests
         run: |
-          cd release-packages && node code-server*-linux-amd64 &
+          cd release-packages && node PASSWORD=e45432jklfdsab code-server*-linux-amd64 &
```
jsjoeio, line 40: Open to alternatives if this isn't the best approach. I considered reading from `config.yaml` but wasn't sure if that would be difficult to read since I think the path is different locally vs ci.

package.json
```diff
@@ -136,6 +136,7 @@
       "global": {
         "lines": 40
       }
-    }
+    },
+    "testTimeout": 30000
```
jsjoeio, line 140: Needed for e2e. Eventually when we split up e2e and unit, we can keep this in e2e only. We may need to increase as our e2e tests get more complicated and take longer.

test/goHome.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    expect(await page.isVisible(goHomeButton))
+    // Click it and navigate to coder.com
+    await page.click(goHomeButton)
+
+    // If there are unsaved changes it will show a dialog
+    // asking if you're sure you want to leave
+    page.on("dialog", (dialog) => dialog.accept())
+
+    await page.waitForTimeout(5000)
+    expect(await page.url()).toBe("https://coder.com/")
```
jsjoeio, line 53: I wonder if this would fail if the site (coder.com) was down 🤔 

I think it would pass but not sure. I wouldn't want this test to be flaky.

I wonder if it's fine, or if I need to look into manipulating [network events](https://playwright.dev/docs/network?_highlight=netw#network-events)

.github/workflows/ci.yaml
code-asher: I think this is perfect.

test/goHome.test.ts
code-asher: Maybe we can start code-server with `--home localhost:8008/test-home` or some other random path. It'll 404 but we'll be guaranteed it'll load.

test/goHome.test.ts
```diff
@@ -0,0 +1,55 @@
+import { chromium, Page, Browser, BrowserContext } from "playwright"
+
+// NOTE: this is hard-coded and passed as an environment variable
+// See the test job in ci.yml
+const PASSWORD = "e45432jklfdsab"
```
code-asher, line 5: Maybe we should use `PASSWORD=e45432jklfdsab yarn test` in CI and then use `process.env.PASSWORD` here, so if a dev runs the test they can set the password used in their instance.

code-asher: Actually, even better, in the `ci.yaml` adding something like:

```
env:
  PASSWORD: e45432jklfdsab
```

Should set the env var for both the node spawn and the yarn test.

test/goHome.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  afterEach(async () => {
+    await page.close()
+    // Remove password from local storage
+    await context.clearCookies()
+  })
+
+  it("should see a 'Go Home' button in the Application Menu that goes to coder.com", async () => {
+    await page.goto("http://localhost:8080")
+    // In case the page takes a long time to load
+    await page.waitForTimeout(2000)
```
code-asher, line 35: Playwright doesn't wait for the page to finish loading?  Maybe we need something like https://playwright.dev/docs/navigations/#custom-wait

test/goHome.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Type in password
+    await page.fill(".password", PASSWORD)
+    // Click the submit button and login
+    await page.click(".submit")
+    // Click the Applicaiton menu
```
code-asher, line 40:
```suggestion
    // Click the Application menu
```

test/goHome.test.ts
```diff
@@ -0,0 +1,55 @@
...
+    const goHomeButton = "a.action-menu-item span[aria-label='Go Home']"
+    expect(await page.isVisible(goHomeButton))
+    // Click it and navigate to coder.com
+    await page.click(goHomeButton)
+
+    // If there are unsaved changes it will show a dialog
+    // asking if you're sure you want to leave
+    page.on("dialog", (dialog) => dialog.accept())
+
+    await page.waitForTimeout(5000)
```
code-asher, line 52: Do we need the timeout? It looks like the click is supposed to wait until page navigation completes. https://playwright.dev/docs/navigations/#auto-wait-1

If it does work we might need to move the dialog code above the click? If it doesn't then no need.

If it doesn't work maybe we could wait for an element on the page instead (if we do the `--home` thing we could look for the 404 title or something).

test/login.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  })
+
+  afterEach(async () => {
+    await page.close()
+    // Remove password from local storage
+    await context.clearCookies()
+  })
+
+  it("should be able to login with the password from config.yml", async () => {
+    await page.goto("http://localhost:8080")
```
code-asher, line 32: Let's do something like:

```suggestion
    await page.goto(process.env.CODE_SERVER_ADDRESS)
```
And then set that in the `ci.yaml`. We could still default to `http://localhost:8080` though. We could add a little `getCodeServerAddress()` utility function or something since it'll be used by multiple tests.

code-asher (changes requested):

test/goHome.test.ts
code-asher: Oh we could use `http://localhost:8080/healthz` if we need a path that exists.

test/goHome.test.ts
jsjoeio: Genius! I'll add that now.

.github/workflows/ci.yaml
jsjoeio: Going with the "set `PASSWORD` as an env var in `ci` approach so removing this

test/goHome.test.ts
jsjoeio: I like the `/test-home` approach:
- guaranteed
- look for 404

test/goHome.test.ts
jsjoeio: That worked like a charm! Good find

test/goHome.test.ts
jsjoeio: The auto-wait worked perfectly - thanks man!

jsjoeio:
Something wrong in test. Also, if I modify the `yarn test` script and add the env there, then i might not need to in ci 🤔

image

jsjoeio:
tests are passing locally but no in CI? so weird 🤔 
image

jsjoeio:
I'm running into strange errors 🤔

image

test/goHome.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    page = await context.newPage()
+  })
+
+  afterEach(async () => {
+    await page.close()
+    // Remove password from local storage
+    await context.clearCookies()
+  })
+
+  it("should see a 'Go Home' button in the Application Menu that goes to coder.com", async () => {
```
nhooyr, line 28:
```suggestion
  it("should see a 'Go Home' button in the Application Menu that goes to health check", async () => {
```

test/goHome.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    // Type in password
+    await page.fill(".password", process.env.PASSWORD || "password")
+    // Click the submit button and login
+    await page.click(".submit")
+    // Click the Application menu
+    await page.click(".menubar-menu-button[title='Application Menu']")
+    // See the Go Home button
+    const goHomeButton = "a.action-menu-item span[aria-label='Go Home']"
+    expect(await page.isVisible(goHomeButton))
+    // Click it and navigate to coder.com
```
nhooyr, line 52: outdated.

test/goHome.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    // NOTE: ran into issues of it failing intermittently
+    // without having button: "middle"
+    await page.click(goHomeButton, { button: "middle" })
+
+    // If there are unsaved changes it will show a dialog
+    // asking if you're sure you want to leave
+    page.on("dialog", (dialog) => dialog.accept())
+
+    // We make sure to wait on a request to the GO_HOME_URL
+    await page.waitForRequest(GO_HOME_URL)
```
nhooyr, line 62: This means we don't need the above `requestedGoHomeUrl` stuff right?

test/login.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    page = await context.newPage()
+  })
+
+  afterEach(async () => {
+    await page.close()
+    // Remove password from local storage
+    await context.clearCookies()
+  })
+
+  it("should be able to login with the password from config.yml", async () => {
```
nhooyr, line 27: Is this a WIP? There's nothing being read from `config.yml`.

nhooyr (changes requested):

test/login.test.ts
jsjoeio: yes, sorry need to update! it was working but only intermittently

.github/workflows/ci.yaml
```diff
@@ -37,9 +40,14 @@ jobs:
...
         run: |
-          node ./release-packages/code-server*-linux-amd64 &
+          ./release-packages/code-server*-linux-amd64/bin/code-server --home $CODE_SERVER_ADDRESS/healthz &
           yarn --frozen-lockfile
           yarn test
+      - name: Upload test artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: test-videos
+          path: ./test/videos
```
jsjoeio, line 50: This can help with debugging: after the job ends, you can download a zip with the videos recorded during the e2e tests.

ci/dev/test.sh
```diff
@@ -9,7 +9,8 @@ main() {
   # information. We must also run it from the root otherwise coverage will not
   # include our source files.
   cd "$OLDPWD"
-  ./test/node_modules/.bin/jest "$@"
+  # We use the same environment variables set in ci.yml in the test job
+  PASSWORD=e45432jklfdsab CODE_SERVER_ADDRESS=http://localhost:8080 ./test/node_modules/.bin/jest "$@"
```
code-asher, line 13: We don't need to set these because they're already set from CI right? And locally we wouldn't want to set these either I think since devs might have their code-server on a different port or with a different password.

test/goHome.test.ts
```diff
@@ -0,0 +1,79 @@
...
+    // so we listen if it comes, we accept it
+    page.on("dialog", (dialog) => dialog.accept())
+
+    // waitUntil: "domcontentloaded"
+    // In case the page takes a long time to load
+    await page.goto(process.env.CODE_SERVER_ADDRESS || "http://localhost:8080", { waitUntil: "domcontentloaded" })
+
+    // For some odd reason, the login method used in globalSetup.ts
+    // I don't know if it's on playwright clearing our cookies by accident
+    // or if it's our cookies disappearing.
```
code-asher, line 57: One thing we could try is setting the cookie using the password directly. So something like (untested, the syntax might be a bit off and I'm not sure if we'd need to set some other params like `url`, `domain`, etc):

```
import { hash } from "../src/node/util"

browser.newContext({ storageState: { cookies: [{ name: "key", value: hash(process.env.PASSWORD) }]} })
```

Still, definitely weird it's not working.

ci/dev/test.sh
code-asher: If they aren't set we'll probably want to error so the dev knows what to set to make the tests work.

jsjoeio: > And locally we wouldn't want to set these either I think since devs might have their code-server on a different port or with a different password.

Ah good point. I didn't think of that.

jsjoeio: I'll modify that and add print an error message telling them to set those.

Eventually when e2e is separate, it will be more clear.

test/goHome.test.ts
jsjoeio: I'll try this out!

ci/dev/test.sh
jsjoeio: image

test/goHome.test.ts
jsjoeio: I think that worked! I ran the test locally and it passed lol

src/common/util.ts
```diff
@@ -120,3 +120,39 @@ export function logError(prefix: string, err: any): void {
...
+  secure: boolean
+  sameSite: "Strict" | "Lax" | "None"
+}
+
+/**
+ * Checks if a cookie exists in array of cookies
+ */
+export function checkForCookie(cookies: Array<Cookie>, key: string): boolean {
+  // Check for at least one cookie where the name is equal to key
+  return cookies.filter((cookie) => cookie.name === key).length > 0
```
nhooyr, line 144: Use `cookies.find`.

test/goHome.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  let browser: Browser
+  let page: Page
+  let context: BrowserContext
+
+  beforeAll(async (done) => {
+    browser = await chromium.launch()
+    // Create a new context with the saved storage state
+    const storageState = JSON.parse(process.env.STORAGE || "")
+
+    //
```
nhooyr, line 15: missing comment

test/goHome.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    page = await context.newPage()
+    done()
+  })
+
+  // NOTE: this test will fail if you do not run code-server with --home $CODE_SERVER_ADDRESS/healthz
+  it("should see a 'Go Home' button in the Application Menu that goes to /healthz", async (done) => {
+    // Ideally, this test should pass and finish before the timeout set in the Jest config
+    // However, if it doesn't, we don't want a memory leak so we set this backup timeout
+    // Otherwise Jest may throw this error
+    // "Jest did not exit one second after the test run has completed.
```
nhooyr, line 67: Jest's error is correct. By setting this backup timeout, we're hiding that the test failed as the go home url wasn't requested.

This should be rewritten to use purely promises instead of the done callback. I'm guessing we can't combine them and so jest gets confused when you're returning a promise but also calling done later.

nhooyr (changes requested):

test/goHome.test.ts
jsjoeio: > By setting this backup timeout, we're hiding that the test failed as the go home url wasn't requested.

Ahhh - now I see what you're saying. Oversight on my part. 

It's failing because we wait indefinitely until the request to the go home url has been made. This causes the memory leak because we never finish. I think a way forward to solve this is to add an expect call in the `backupTimeout` before `done` so we know that we've waited long enough for the request to happen to the go home url but that it was unsuccessful.

nhooyr (approved):
To the moon 🌔

test/goHome.test.ts
jsjoeio: Correction: I think you're right about Jest being confused using both promises and the done callback. I don't see the two used together [in the docs](https://jestjs.io/docs/en/asynchronous#callbacks) and I'm still getting the same leak error. I'll rewrite to use promises as suggested.

test/goHome.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    // or if it's our cookies disappearing.
+    // This means we need an additional check to make sure we're logged in.
+    // We do this by manually adding the cookie to the browser environment
+    // if it's not there at the time the test starts
+    const cookies: Cookie[] = storageState.cookies || []
+    // If the cookie exists in cookies then
+    // this will return the cookies with no changes
+    // otherwise if it doesn't exist, it will create it
+    // hence the name maybeUpdatedCookies
+    const maybeUpdatedCookies = createCookieIfDoesntExist(cookies, cookieToStore)
```
code-asher, line 45: The playwright storage thing sometimes works and sometimes doesn't right? Or does it never work?

Wondering if we should add an optional test here just to keep an eye on how often it fails.

code-asher: Doesn't look like there are optional tests so maybe we can just log.

test/goHome.test.ts
```diff
@@ -0,0 +1,102 @@
...
+      }
+    })
+    // Sometimes a dialog shows up when you navigate
+    // asking if you're sure you want to leave
+    // so we listen if it comes, we accept it
+    page.on("dialog", (dialog) => dialog.accept())
+
+    // waitUntil: "domcontentloaded"
+    // In case the page takes a long time to load
+    await page.goto(process.env.CODE_SERVER_ADDRESS || "http://localhost:8080", { waitUntil: "domcontentloaded" })
```
code-asher, line 86: Not a big deal but I think maybe we create a `test/constants.ts` with the code-server address so we don't have to have `|| "http://localhost:8080"` in a bunch of places. We could do the same with the password.

We also need to update the address used in `e2e.test.ts`.

test/goHome.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    const goHomeButton = "a.action-menu-item span[aria-label='Go Home']"
+    expect(await page.isVisible(goHomeButton))
+
+    // Click it and navigate to /healthz
+    // NOTE: ran into issues of it failing intermittently
+    // without having button: "middle"
+    await page.click(goHomeButton, { button: "middle" })
+    // Give it 3 seconds for request to be sent and completed and for our value to update
+    await setTimeoutPromise(3000)
+    expect(requestedGoHomeUrl).toBe(true)
```
code-asher, line 100: Sorry to keep visiting this haha but it'd be awesome to get this working without timeouts. What about this?

```suggestion
    await Promise.all([ 
      page.waitForNavigation(), 
      page.click(goHomeButton, { button: "middle" }) 
    ])
    expect(page.url()).toBe(GO_HOME_URL)
```

Source: https://playwright.dev/docs/navigations/#asynchronous-navigation

code-asher (changes requested):
I ran this against my dev build and it worked like a dream. Very exciting that I can just run e2e tests as I dev now.

test/goHome.test.ts
```diff
@@ -0,0 +1,102 @@
...
+
+async function setTimeoutPromise(milliseconds: number): Promise<void> {
+  return new Promise((resolve, _) => {
+    setTimeout(() => {
+      resolve()
+    }, milliseconds)
+  })
+}
+
+describe("login", () => {
```
code-asher, line 13:
```suggestion
describe("go home", () => {
```

test/goHome.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  let browser: Browser
+  let page: Page
+  let context: BrowserContext
+
+  beforeAll(async () => {
+    browser = await chromium.launch()
+    // Create a new context with the saved storage state
+    const storageState = JSON.parse(process.env.STORAGE || "")
```
code-asher, line 21: I think `JSON.parse("")` will error.

```suggestion
    const storageState = JSON.parse(process.env.STORAGE || "{}")
```

ci/dev/test.sh
```diff
@@ -9,6 +9,19 @@ main() {
   # information. We must also run it from the root otherwise coverage will not
   # include our source files.
   cd "$OLDPWD"
+  # We use the same environment variables set in ci.yml in the test job
+  if [[ -z ${PASSWORD+x} ]] || [[ -z ${CODE_SERVER_ADDRESS+x} ]]; then
```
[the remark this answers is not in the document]
jsjoeio: Genius! Good point!

src/common/util.ts
```diff
@@ -120,3 +120,39 @@ export function logError(prefix: string, err: any): void {
     logger.error(`${prefix}: ${err}`)
   }
 }
+
+// Borrowed from playwright
+export interface Cookie {
```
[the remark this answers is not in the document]
jsjoeio: Yes, that's right. Good point! I'll create a `helpers.ts` inside `/test`

test/globalSetup.ts
```diff
@@ -0,0 +1,28 @@
...
+// so that it authenticates us into code-server
+// ensuring that we're logged in before we run any tests
+import { chromium, Page, Browser, BrowserContext } from "playwright"
+
+module.exports = async () => {
+  console.log("🚨 Running Global Setup for Jest Tests")
+  console.log("  Please hang tight...")
+  const browser: Browser = await chromium.launch()
+  const context: BrowserContext = await browser.newContext()
+  const page: Page = await context.newPage()
```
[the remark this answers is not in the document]
jsjoeio: I thought it was complaining when it didn't, you're probably right though, i'll double check!

test/goHome.test.ts
jsjoeio: yeah no optional tests :( should I make a note to add more login tests (testing different strategies) at a later time (different PR)?

test/goHome.test.ts
jsjoeio: agreed! lol let's give it a shot!

test/goHome.test.ts
jsjoeio: great suggestion! I agree, good idea

test/goHome.test.ts
code-asher: Yeah let's just smack down a TODO to figure this out.

jsjoeio:
Not sure why release is failing. Any ideas on how to fix this? @code-asher 

```
$ ./ci/build/build-code-server.sh
grep: out/node/entry.js: No such file or directory
sed: can't read out/node/entry.js: No such file or directory
chmod: cannot access 'out/node/entry.js': No such file or directory
```

test/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const CODE_SERVER_ADDRESS = process.env.CODE_SERVER_ADDRESS || "http://localhost:8080"
+export const PASSWORD = process.env.PASSWORD || "e45432jklfdsab"
+export const STORAGE = process.env.STORAGE || ""
```
code-asher, line 3: I think here we'll need to tweak this a bit.

The reason is that at the top of the global setup file we import this file which causes `STORAGE` to get set to `process.env.STORAGE` (which will be undefined at that point so we'll end up setting it to a blank string).

Then later we set the storage to `process.env.STORAGE` but this won't update the `STORAGE` constant since we already set it to a blank string.

Some ideas:
1. Make this a `let` and set it directly instead of `process.env.STORAGE`
2. Make this a function so it grabs the current `process.env.STORAGE`

code-asher (changes requested):
Looks good to me after this change!

test/constants.ts
code-asher: Or, wait, am I wrong about how this works? If I was right then the `JSON.parse` should have errored since it can't parse a blank string. But the tests passed just fine.

code-asher: Oh maybe it's Jest clearing the module cache before/after each test or something? In which case this is completely fine.

code-asher: Yup confirmed that's what's going on.

code-asher (approved):

PR #2656, opened 2021-02-02
echo without newline to hash password

timreinke:
Otherwise the input to sha256sum will include the newline character


timreinke:
Ah, didn't realize this was fixed in master!

PR #2671, opened 2021-02-03
feat(testing): add unit tests for common/util

jsjoeio:
This PR adds several unit tests for the `src/common/util` functions.

## Screenshot

image

It increases our code coverage (Lines) from 46.88% to 49.53% (↑ 2.65%)
🟩 100% coverage for `src/common/util`
image

## TODOS
- [x] write tests for `logError`
- [x] write tests for `getOptions`
- [x] write tests for `generateUuid`
- [x] update screenshot and percentage

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -1,11 +1,12 @@
 {
-  "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up",
   "#": "the types and generates conflicts with mocha.",
   "devDependencies": {
     "@types/jest": "^26.0.20",
+    "@types/jsdom": "^16.2.6",
     "@types/node-fetch": "^2.5.8",
     "@types/supertest": "^2.0.10",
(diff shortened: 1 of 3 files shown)


test/package.json
```diff
@@ -1,11 +1,12 @@
 {
-  "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up",
```
oxy, line 2: Half the comment went missing! I suspect this may be because both of them used the `"#"` key.

oxy (changes requested):
PR looks good! Might have to combine the two comments in `test/package.json` into a single line to avoid future issues, though.

test/package.json
jsjoeio: Weird, good catch!

oxy (approved):
LGTM!

Issue #1430, opened 2020-03-19
Symbol not found in alpine


yujinqiu:
Hi there, 
code-server2.1698-vsc1.41.1-alpine-x86_64  unable to run in Alpine Linux 3.9 with the following error

```
~ # ./code-server -h
Error relocating ./code-server: _ZNSt7__cxx1119basic_ostringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found
Error relocating ./code-server: _ZNSt19_Sp_make_shared_tag5_S_eqERKSt9type_info: symbol not found
Error relocating ./code-server: _ZNSt7__cxx1118basic_stringstreamIcSt11char_traitsIcESaIcEEC1Ev: symbol not found
```
code-server2.1692-vsc1.39.2-alpine-x86_64  works well. 

```
~ # code-server --version
info  2.1692-vsc1.39.2
info  e14362f3224c101585a26bf7ee5fa6e3bf425046
info  x64
```

kylecarbs:
Weird weird. Any ideas @code-asher ?

code-asher:
We were using Drone for that release; we must have broken the Alpine step in the process. I'm unsure what the best way to fix it is since we no longer have Alpine builds; the current recommendation is to install `libc-dev` so maybe we should just remove the broken Alpine builds from that release?

mjgallag:
I can’t get libc-dev to work in alpine for 3.0.0. I get similar errors. If I take the node binary from official node alpine image it works just fine.

nhooyr:
Looks like there's been confusion regarding libc-dev. I don't think that installs glibc or would help with this issue. I think that was my bad. I think the best approach here is to solve #1396 instead of adding an alpine release but whatever you think is best @code-asher

code-asher:
We'd need an Alpine container on CI either way, I think? In order to build the native Node modules. I suppose that means we'd end up with an Alpine release during deploy anyway.

nhooyr:
We wouldn't include the native node modules, they'd be built by node when fetched.

code-asher:
Oh right, of course. Yeah that sounds good to me.

nhooyr:
Fixed, please use the npm package to install on alpine. The erroneous alpine entry in the FAQ has been removed.

See https://github.com/cdr/code-server#npm

anonhostpi:
If this is an issue on Alpine, it is also likely an issue on OpenWRT as they both use musl instead of glibc

PR #2696, opened 2021-02-08
docs: clarify Alpine install path in install.md

bpmct:
From install.md, the path to install code-server on Linux Alpine is unclear.

#1430 recommends using the npm script, since the install script doesn't work.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -2,7 +2,7 @@
 set -eu
 
 # code-server's automatic install script.
-# See https://github.com/cdr/code-server/blob/master/docs/install.md
+# See https://github.com/cdr/code-server/blob/v3.8.1/docs/install.md
 
 usage() {
   arg0="$0"
@@ -67,7 +67,7 @@ Usage:
 
 It will cache all downloaded assets into ~/.cache/code-server
 
-More installation docs are at https://github.com/cdr/code-server/blob/master/docs/install.md
+More installation docs are at https://github.com/cdr/code-server/blob/v3.8.1/docs/install.md
 EOF
 }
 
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -419,7 +419,7 @@ install_npm() {
   echoh
   echoerr "Please install npm or yarn to install code-server!"
   echoerr "You will need at least node v12 and a few C dependencies."
-  echoerr "See the docs https://github.com/cdr/code-server#yarn-npm"
+  echoerr "See the docs https://github.com/cdr/code-server/blob/master/docs/install.md#yarn-npm"
```
jsjoeio, line 422: This should probably point to the version branch, right? That way it always matches the latest released version of code-server. 

(And when we update release, we update links like these so they're always up-to-date!)
```suggestion
  echoerr "See the docs https://github.com/cdr/code-server/blob/v3.8.1/docs/install.md#yarn-npm"
```

jsjoeio (changes requested):
Looks great! Thanks for adding this, Ben! 🙌

One small change

install.sh
bpmct: I noticed two other links in this file reference the master doc. Should we make all of them reflect the current version? Also, would it be possible to use the `$VERSION` var in this so we don't have to manually change them?

Or are the links auto-updated with some job?

jsjoeio: Oh good catch!

> Should we make all of them reflect the current version

Yes!

> Also, would it be possible to use the $VERSION var in this so we don't have to manually change them?

I think so? But I'm not sure. I wasn't sure if that was defined globally or on a script-by-script basis. I'll defer to you though - whichever option you think is best. 

> Or are the links auto-updated with some job?

Right now it's manual with a "Find and Replace all" as part of the release process, but I do want to automate it.

jsjoeio (approved):
image

jsjoeio:
@bpmct just realized - can you sign your commits so that they're verified and we can merge this? 

If not, happy to pull down, rebase and sign them for you!

bpmct:
@jsjoeio done, sorry about that!

PR #2702, opened 2021-02-09
feat(testing): add unit tests for src/common/http

jsjoeio:
This PR adds a few tests for `src/common/http`. Note: intentionally selected `add-unit-tests-constants` as the base since I branched off it.

## Screenshot

image

It increases our code coverage (Lines) from 49.83% to 50.09% (↑ 0.0.26%).
🟩 100% Line coverage (last 100) for `src/node/http.ts`
image

### DIFF
--- a/test/http.test.ts
+++ b/test/http.test.ts
@@ -1,0 +1,35 @@
+import { HttpCode, HttpError } from "../src/common/http"
+
+describe("http", () => {
+  describe("HttpCode", () => {
+    it("should return the correct HTTP codes", () => {
+      expect(HttpCode.Ok).toBe(200)
+      expect(HttpCode.Redirect).toBe(302)
+      expect(HttpCode.NotFound).toBe(404)
+      expect(HttpCode.BadRequest).toBe(400)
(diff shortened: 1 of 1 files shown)


test/http.test.ts
```diff
@@ -0,0 +1,36 @@
+import { HttpCode, HttpError } from "../src/common/http"
+
+describe("http", () => {
+  describe("HttpCode", () => {
+    it("should return the correct HTTP codes", () => {
+      expect(HttpCode.Ok).toBe(200)
+      expect(HttpCode.Redirect).toBe(302)
+      expect(HttpCode.NotFound).toBe(404)
+      expect(HttpCode.BadRequest).toBe(400)
+      expect(HttpCode.BadRequest).toBe(400)
```
oxy, line 10: `BadRequest` appears twice!

oxy (changes requested):
I think BadRequest should only need to appear once.

test/http.test.ts
jsjoeio: Definitely a typo - thanks for catching that!

oxy (approved):
LGTM!

PR #2715, opened 2021-02-10
feat(testing): add unit tests for emitter

jsjoeio:
This PR adds a few tests for `src/common/emitter.ts`.

## Screenshot

image

It increases our code coverage (Lines) from 50.09% to 50.26% (↑ 0.17%).
🟩 100% Line coverage (last 100) for `src/common/emitter.ts` and also 100% coverage for `src/common`

image

## TODOS
- [x] fix duplicate describe

### DIFF
--- a/test/emitter.test.ts
+++ b/test/emitter.test.ts
@@ -1,0 +1,52 @@
+import { Emitter } from "../src/common/emitter"
+
+describe("emitter", () => {
+  describe("Emitter", () => {
+    it("should return an Emitter", async () => {
+      const HELLO_WORLD = "HELLO_WORLD"
+      const GOODBYE_WORLD = "GOODBYE_WORLD"
+      const mockCallback = jest.fn(() => "Mock function called")
+      const mockSecondCallback = jest.fn(() => undefined)
+
+      const emitter = new Emitter<{ event: string; callback: () => void }>()
+
+      const onHelloWorld = ({ event, callback }: { event: string; callback: () => void }): void => {
+        if (event === HELLO_WORLD) {
+          callback()
+        }
+      }
+
+      const onGoodbyeWorld = ({ event, callback }: { event: string; callback: () => void }): void => {
+        if (event === GOODBYE_WORLD) {
+          callback()
+        }
+      }
+
+      // Register the onHelloWorld listener
+      // and the onGoodbyeWorld
+      emitter.event(onHelloWorld)
+      emitter.event(onGoodbyeWorld)
+
+      await emitter.emit({ event: HELLO_WORLD, callback: mockCallback })
+
+      // Double-check that our callback is called only once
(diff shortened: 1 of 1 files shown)


test/emitter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    afterEach(() => {
+      jest.clearAllMocks()
+    })
+
+    afterAll(() => {
+      jest.restoreAllMocks()
+    })
+
+    it("should return an Emitter", async () => {
```
oxy, line 23: I think "should run the correct callbacks" would be a more apt description for the test? Not sure :sweat_smile:

jsjoeio: good idea! let's change it to that!

jsjoeio, line 23:
```suggestion
    it("should run the correct callbacks", async () => {
```

test/emitter.test.ts
```diff
@@ -0,0 +1,88 @@
+// Note: we need to import logger from the root
+// because this is the logger used in logError in ../src/common/util
+import { logger } from "../node_modules/@coder/logger"
```
code-asher, line 3: If we `import { logger } from "@coder/logger"` it should pull the same module right?

test/emitter.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { Emitter } from "../src/common/emitter"
+
+describe("emitter", () => {
+  describe("Emitter", () => {
```
code-asher, line 8: Why double `describe`?

code-asher (approved):

test/emitter.test.ts
jsjoeio: hmm I don't believe so. based on my understanding, I think because we have two separate package.json's, it will look for `@coder/logger` in `/test` which is not the one used inside `/src/common/emitter` so we have to import from above to ensure we're mocking the one used by functions located in `/src`

test/emitter.test.ts
jsjoeio: eh i was trying to follow a pattern used in other places 😂 i'll clean this up

test/emitter.test.ts
code-asher: There is no `@coder/logger` in `/test/node_modules` is there? It should then traverse upward and find it in `../node_modules`.

jsjoeio: Right, and I think it does! But when I'm spying on a module, I have to tell Jest where it is, which is why I have to say, "look up at the node_modules at the root!"

code-asher: Huh, weird! Sounds good to me then.

jsjoeio: I know, Jest weirdness strikes again lol

Issue #2480, opened 2020-12-17
Fix CI release step for linux-arm64


jsjoeio:
As discussed briefly with @code-asher, we'd like to make some improvements to the CI workflow, specifically in the release step for `linux-arm64`. 

## Problem
The ci / linux-arm64 workflow is failing at the “Run ./ci/steps/release-packages.sh build step.

## Reason
The machine has no space left.

## Potential Solutions
- doing a shallow clone with the repo in the checkout step
- delete the asset we download (if it build successfully)
- deleting the `node_modules` directory on machine

## Relevant
- [GitHub Actions failing with no space left on device](https://github.community/t/github-actions-failing-with-errno-28-no-space-left-on-device/18164/7)
- #2476

code-asher:
We worked out this was caused by an accumulation of Docker images which was resolved by deleting them. Will need some kind of cleanup or a way to prevent this accumulation.

jsjoeio:
> last time i think i just added a docker rm $(docker ps -aq) to ci and ran that once then removed it lol
> and a docker rmi $(docker images -q) as well i think

From Asher.

vapurrmaid:
Fixed in d48a74c317f59d4a467ba004dcbe38abfefbf73f

PR #2719, opened 2021-02-12
feat(testing): add unit tests for register

jsjoeio:
This PR adds a few tests for `src/browser/register.ts`.

## Screenshot

image

It increases our code coverage (Lines) from 50.09% to 51.13% (↑ 1.04%).
🟩 100% Line coverage (last 100) for `src/browser/register.ts` 

image

TODOs
- [x] rebase branch onto `master`
- [x] address feedback
- [x] refactor based on Asher's suggestion
- [x] add test for else statement in register


oxy (approved):
Looks good! Your choice on if you want a second opinion from asher on this one; I'm not too confident in my TS just yet :sweat_smile:

jsjoeio:
I'll play it safe since we're releasing today - @code-asher would love an 👀 on this one!

test/util.test.ts
```diff
@@ -17,7 +17,6 @@ import {
 
 const dom = new JSDOM()
 global.document = dom.window.document
-// global.window = (dom.window as unknown) as Window & typeof globalThis
```
jsjoeio, line 20: intentionally removed (it was an old comment from a previous PR)

src/browser/register.ts
```diff
@@ -1,18 +1,38 @@
...
+}: HandleServiceWorkerRegistration): void {
+  if (typeof navigator !== "undefined" && "serviceWorker" in navigator) {
+    const options = getOptions()
+    const path = normalize(`${options.csStaticBase}/dist/serviceWorker.js`)
+    registerServiceWorker(navigator, path, options)
+  }
+}
+
+// Written this way so that it's easier to test
+handleServiceWorkerRegistration({ getOptions, normalize, registerServiceWorker })
```
jsjoeio, line 38: I don't _love_ this function signature (taking in an object with these three keys), but it makes it super easy to test 🤷‍♂️

src/browser/register.ts
```diff
@@ -1,18 +1,38 @@
...
+  try {
+    await navigator.serviceWorker.register(path, {
       scope: (options.base ?? "") + "/",
     })
-    .then(() => {
-      console.log("[Service Worker] registered")
-    })
+    console.log("[Service Worker] registered")
+  } catch (error) {
+    logger.error(`[Service Worker] failed to register: ${error.message}`)
```
jsjoeio, line 15: Figured it would be a good idea to capture this error and send to our logger

package.json
```diff
@@ -139,6 +139,9 @@
       "global": {
         "lines": 40
       }
+    },
+    "moduleNameMapper": {
+      "^.+\\.(css|less)$": "<rootDir>/test/cssStub.ts"
```
jsjoeio, line 144: Jest doesn't know how to handle css files natively so this was needed because of these lines in `browser/register.ts`

```

import "./pages/error.css"
import "./pages/global.css"
import "./pages/login.css"
```

src/browser/register.ts
code-asher: It does feel a little strange to me as well. Could we export a single `registerServiceWorker` that takes no parameters?

Then we polyfill `navigator`, `document.getElementById`, and `getAttribute` (the last two are used in `getOptions` so we can use that to test different base values). That would make the test most closely mirror how the code runs in the browser, I think.

code-asher (changes requested):

src/browser/register.ts
code-asher: We could also do no export and `require` the module each time and clear the module cache after each test.

test/emitter.test.ts
```diff
@@ -0,0 +1,88 @@
+// Note: we need to import logger from the root
+// because this is the logger used in logError in ../src/common/util
+import { logger } from "../node_modules/@coder/logger"
```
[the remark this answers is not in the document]
jsjoeio: oops! could be the wrong base, i'll check

test/register.test.ts
```diff
@@ -0,0 +1,134 @@
...
+            return "hello"
+          },
+        },
+      })
+    })
+
+    beforeEach(() => {
+      loggerSpy = jest.spyOn(logger, "error")
+      spy = jest.fn()
+      global.navigator.serviceWorker.register = spy
```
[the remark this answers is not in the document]
jsjoeio: Nope, just me being inconsistent 😅 thanks for catching that!

test/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+export function createCookieIfDoesntExist(cookies: Array<Cookie>, cookieToStore: Cookie): Array<Cookie> {
+  const cookieName = cookieToStore.name
+  const doesCookieExist = checkForCookie(cookies, cookieName)
+  if (!doesCookieExist) {
+    const updatedCookies = [...cookies, cookieToStore]
+    return updatedCookies
+  }
+  return cookies
+}
```
jsjoeio, line 35: Technically this code comes from #2648 but I cherry-picked the file. I just needed the `helpers.ts` file to put the `loggerModule` export.

test/register.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      Object.defineProperty(global.navigator, "serviceWorker", {
+        value: {
+          register: mockRegisterFn,
+        },
+      })
+    })
+
+    beforeEach(() => {
+      jest.mock("@coder/logger", () => loggerModule)
+    })
```
jsjoeio, line 24: I hope it's okay that this test mocks here in the `beforeEach`.  When I move it outside the `describe` blocks and use the `require` method like you suggested for `constants` etc. @code-asher - it doesn't work :( 

I couldn't figure out why, but this way works!

code-asher: Yup this is perfect. In this case since we `require` within each test and the mocks are cleared after each test we also have to mock it before each test.

For the other test the require (import) happens at the top of the file, so we had to mock at the top of the file as well.

jsjoeio: Oh! I understand now. Thanks for explaining that! Let me know if there's anything else you need from me to move this forward!

test/emitter.test.ts
```diff
@@ -59,6 +60,25 @@ describe("Emitter", () => {
     emitter.dispose()
   })
 
+  it("should log an error if something goes wrong", async () => {
```
code-asher, line 63: Should this test check that the logger was called? It doesn't seem to make any assertions right now.

code-asher (changes requested):

test/emitter.test.ts
jsjoeio: woah weird, maybe it got sucked away when i rebased. good catch!

jsjoeio: 🤦‍♂️ rebase gone wrong. this is a duplicate test lol

test/register.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      global.window = (undefined as unknown) as Window & typeof globalThis
+      global.document = (undefined as unknown) as Document & typeof globalThis
+      global.navigator = (undefined as unknown) as Navigator & typeof globalThis
+      global.location = (undefined as unknown) as Location & typeof globalThis
+    })
+
+    it("should register a ServiceWorker", () => {
+      // Load service worker like you would in the browser
+      require("../src/browser/register")
+      // Load service worker like you would in the browser
```
code-asher, line 44: Misplaced comment.

code-asher (approved):
Looks good!!

test/register.test.ts
jsjoeio: fixing!

jsjoeio:
CI only failed due to #2480

PR #2723, opened 2021-02-12
Fix terminal process leak when closing the window.

outermeasure:
Calling `this.process.dispose();` before `this.process.shutdown(true);` sets a flag `_isDisposed` to true.
This causes the `_kill()` procedure to exit early essentially making `this.process.shutdown(true)` a no op.

It's necessary to await on the shutdown because _kill() is an async function. Otherwise  `_isDisposed` will be set to true before actually killing the process regardless of the call order of `dispose` and `shutdown`.

Impact: Closing the session in the browser should now kill long running processes such as filesystem watchers or services started in the terminal.

### DIFF
--- a/lib/vscode/src/vs/server/node/channel.ts
+++ b/lib/vscode/src/vs/server/node/channel.ts
@@ -575,17 +575,17 @@ class Terminal {
 		// type: 'orphan?';
 	}
 
-	public dispose() {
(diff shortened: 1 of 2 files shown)


code-asher (approved):
Thank you for the fix and the detailed explanation!

PR #2744, opened 2021-02-18
feat(testing): add tests for serviceWorker.ts

jsjoeio:
This PR adds a few tests for `src/browser/serviceWorker.ts`.

## Screenshot

image

It increases our code coverage (Lines) from 51.31% to 51.74% (↑ 0.43%).
🟩 100% Line coverage (last 100) for `src/browser/serviceWorker.ts` 

image

This also completes Lines coverage for `src/browser` 🎉
image

## TODOS
- [x] investigate unhandled promise rejections
- [x] figure out how to write a patch for a node_module
- [x] rebase
- [x] fix inconsistency
- [x] try mocking ourselves


ci/dev/lint.sh
```diff
@@ -6,7 +6,7 @@ main() {
 
   eslint --max-warnings=0 --fix $(git ls-files "*.ts" "*.tsx" "*.js" | grep -v "lib/vscode")
   stylelint $(git ls-files "*.css" | grep -v "lib/vscode")
-  tsc --noEmit
+  tsc --noEmit --skipLibCheck
```
jsjoeio, line 9: Note: added this because TS was complaining about type name conflicts in `dom` and the `webworker` lib. I _think_ this is okay because there's no need for us to check library declaration files when we're linting/running tsc IMO.

image

test/patches/service-worker-mock-response.patch
```diff
@@ -0,0 +1,9 @@
+--- node_modules/service-worker-mock/models/Response.js	2021-02-18 13:47:55.000000000 -0700
++++ patches/service-worker-mock-response.js	2021-02-18 15:57:12.000000000 -0700
+@@ -1,5 +1,6 @@
+ // stubs https://developer.mozilla.org/en-US/docs/Web/API/Response
+ const Body = require('./Body');
++const Blob = require('./Blob')
```
oxy, line 6: this line is missing a `;` but other lines have them; minor consistency thing

oxy (approved):
LGTM! Your call on the `;` in one of the patches we apply; I think internal consistency with the patched file may be the best.

test/patches/service-worker-mock-response.patch
jsjoeio: I agree with you! Consistency is 🔑

test/package.json
```diff
@@ -1,14 +1,19 @@
 {
   "#": "We must put jest in a sub-directory otherwise VS Code somehow picks up the types and generates conflicts with mocha.",
+  "scripts": {
+    "postinstall": "./scripts/patch.sh"
+  },
   "devDependencies": {
     "@types/jest": "^26.0.20",
     "@types/jsdom": "^16.2.6",
     "@types/node-fetch": "^2.5.8",
+    "@types/service-worker-mock": "^2.0.1",
```
code-asher, line 10: Should we try mocking it ourselves to avoid the patching?

Something like this would work:

```typescript
interface Listener {
  event: string
  cb: (event?: { claim: () => void }) => void
}

const listeners: Listener[] = []

Object.assign(global, {
  self: global,
  addEventListener: (event: string, cb: () => void) => {
    listeners.push({ event, cb })
  },
  clients: {
    claim: () => undefined // Or a Jest spy
  },
})

function emit(event: string) {
  listeners.filter((l) => l.event === event).forEach((l) => {
    switch (event) {
      case "activate":
        l.cb({
          waitUntil: () => undefined // Or a Jest spy
        })
        break
      default:
        l.cb()
    }
  })
}
```

jsjoeio: Sure, it's worth trying! Oh perfect. I'll try this out.

jsjoeio: Wow...dude I didn't realize it was that easy to mock 😂 that worked like a charm ✨

jsjoeio:
Weird error in the macos build. I'll rerun

```
$ ./ci/build/test-standalone-release.sh
Testing standalone release.
[2021-02-24T18:49:55.609Z] info  Wrote default config file to ~/.config/code-server/config.yaml
Unexpected output from listing extensions:

Error: Process completed with exit code 1.
```

PR #2800, opened 2021-03-01
Remove default home

code-asher:


jsjoeio (approved):
These are my favorite kind of PRs: short and sweet

- [x] tested by setting `home` in config. still works!
- [x] tested if home icon is removed when `home` isn't set

image

reggiezhang:
"--home" no longer works on v3.9.2

jsjoeio:
@reggiezhang apologies! That should have been mentioned in the release notes (updating now).

VS Code removed that option so we had to remove it from code-server here: https://github.com/cdr/code-server/pull/2862

PR #2804, opened 2021-03-01
fix(ci): add step to remove docker images on linux-arm64 

### LINKED ISSUE #2480: Fix CI release step for linux-arm64

jsjoeio:
This PR fixes #2480 .

- add step to remove docker images on `linux-arm64` ci job (thanks to @code-asher for the tip, @vapurrmaid for the refactor!)
- add step to remove release-packages and test/videos on the `test` job 
- changes the extension used in `test:standalone-release`

I suggest changing the extension away from the Python because it's notorious for breaking in code-server. A theme extension changes less frequently and is much more reliable for testing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,9 +105,7 @@ jobs:
           name: release-packages
           path: ./release-packages
       - name: Remove docker images
-        run: |
-          docker rm $(docker ps -aq)
-          docker rmi $(docker images -q)
+        run: docker system prune -af
 
   macos-amd64:
     needs: release
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -49,8 +49,7 @@ jobs:
           name: test-videos
           path: ./test/videos
       - name: Remove release packages and test artifacts
-          run: |
-            rm -rf ./release-packages ./test/videos
+        run: rm -rf ./release-packages ./test/videos
 
   release:
     runs-on: ubuntu-latest
@@ -105,6 +104,10 @@ jobs:
         with:
           name: release-packages
           path: ./release-packages
+      - name: Remove docker images
+        run: |
+          docker rm $(docker ps -aq)
+          docker rmi $(docker images -q)
 
   macos-amd64:
     needs: release
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -48,6 +48,9 @@ jobs:
         with:
           name: test-videos
           path: ./test/videos
+      - name: Remove release packages and test artifacts
+          run: |
+            rm -rf ./release-packages ./test/videos
 
   release:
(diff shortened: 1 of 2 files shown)


jsjoeio:
image

.github/workflows/ci.yaml
```diff
@@ -102,6 +104,10 @@ jobs:
         with:
           name: release-packages
           path: ./release-packages
+      - name: Remove docker images
+        run: |
+          docker rm $(docker ps -aq)
+          docker rmi $(docker images -q)
```
vapurrmaid, line 110: Might be cleaner to just use `docker system prune` <https://docs.docker.com/engine/reference/commandline/system_prune/>

jsjoeio: TIL! Thanks for the tip @vapurrmaid 🙌

.github/workflows/ci.yaml
```diff
@@ -102,6 +106,8 @@ jobs:
         with:
           name: release-packages
           path: ./release-packages
+      - name: Remove docker images
+        run: docker system prune -af
```
vapurrmaid, line 110: Just curious why it needs to run twice - not saying it's wrong or anything, but maybe a comment would help

jsjoeio: It shouldn't! (I hope not at least). But I think I need to run it at the beginning once to fix the issue, then I can remove that line and I'll keep the one that runs at the end. Don't know if there's an easier way to do it - just the first one that came to mind 😂

.github/workflows/ci.yaml
```diff
@@ -87,6 +89,8 @@ jobs:
     needs: release
     runs-on: ubuntu-arm64-latest
     steps:
+      - name: Prune docker images
+        run: docker system prune -af
```
jsjoeio, line 93: I'll remove this after it runs once. I think it just needs to do a quick cleanup and then in general, it should be enough to run it at the end!

vapurrmaid: oh makes sense!

.github/workflows/ci.yaml
vapurrmaid: makes  sense! ty!

vapurrmaid (approved):

PR #2813, opened 2021-03-02
docs(README): add release shield

jsjoeio:
This PR adds a release shield to the `README`

## Screenshot

image


vapurrmaid (approved):
This is great! One potential issue I could see is if a patch to prior release is newer by date than the latest release. We could get around that through using `GitHub release (latest SemVer)` from https://shields.io/category/version, but I don't suspect we'll face this problem atm.

Thanks for adding this! :tada:

jsjoeio:
Good call! I didn't think of that. I switched it to use the `GitHub release (latest SemVer)` as you suggested! 🙌

Issue #2559, opened 2021-01-11
Make CI work on forks


nhooyr:
Unfortunately github actions CI will only work on forks if there are no custom secrets.

So the npm package and docker image release process will have to be ran manually instead of via CI. Those are also in a separate workflow and github actions still won't run the workflow that does not depend on any secrets.

It's a worthy tradeoff to make it easier for contributors to get feedback on their changes.

jawnsy:
This might be a new feature, but I think GitHub Actions supports running builds in forks with secrets, requiring approval before doing so (to avoid people trying to steal secrets using echo $ENV_VAR or whatever): https://docs.github.com/en/actions/reference/environments#environment-secrets

PR #2830, opened 2021-03-06
chore: use dependabot to manage dependencies

jawnsy:
Use dependabot to manage the dependencies defined in package.json and
GitHub Actions workflows, so that we can proactively update versions.

Outdated versions of third-party dependencies frequently have known
security vulnerabilities with CVEs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,13 +11,8 @@ updates:
     - "jawnsy"
   ignore:
     # GitHub always delivers the latest versions for each major
-    # release tag, so ignore minor version tags
-    - dependency-name: "actions/cache"
-      versions:
-        - 2.x
-    - dependency-name: "actions/checkout"
-      versions:
-        - 2.x
+    # release tag, so handle updates manually
+    - dependency-name: "actions/*"
 
 - package-ecosystem: "npm"
   directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,25 +1,25 @@
 version: 2
 updates:
-- package-ecosystem: "github-actions"
-  directory: "/"
-  schedule:
-    interval: "daily"
-    time: "11:00"
-  assignees:
-    - "jawnsy"
-  reviewers:
-    - "jawnsy"
-  ignore:
-    # GitHub always delivers the latest versions for each major
-    # release tag, so handle updates manually
-    - dependency-name: "actions/*"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "11:00"
+    assignees:
+      - "jawnsy"
+    reviewers:
+      - "jawnsy"
+    ignore:
+      # GitHub always delivers the latest versions for each major
+      # release tag, so handle updates manually
+      - dependency-name: "actions/*"
 
-- package-ecosystem: "npm"
-  directory: "/"
-  schedule:
-    interval: "daily"
-    time: "11:00"
-  assignees:
-    - "jawnsy"
-  reviewers:
-    - "jawnsy"
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "11:00"
+    assignees:
+      - "jawnsy"
+    reviewers:
+      - "jawnsy"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,15 +5,26 @@ updates:
   schedule:
     interval: "daily"
     time: "11:00"
-  open-pull-requests-limit: 10
   assignees:
-    - "dependabot"
+    - "jawnsy"
+  reviewers:
+    - "jawnsy"
+  ignore:
+    # GitHub always delivers the latest versions for each major
+    # release tag, so ignore minor version tags
+    - dependency-name: "actions/cache"
+      versions:
+        - 2.x
+    - dependency-name: "actions/checkout"
+      versions:
+        - 2.x
 
-- package-ecosystem: npm
+- package-ecosystem: "npm"
   directory: "/"
   schedule:
-    interval: daily
+    interval: "daily"
     time: "11:00"
-  open-pull-requests-limit: 10
   assignees:
-    - "dependabot"
+    - "jawnsy"
+  reviewers:
+    - "jawnsy"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"
+    time: "11:00"
+  open-pull-requests-limit: 10
+  assignees:
+    - "dependabot"
+
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
+    time: "11:00"
+  open-pull-requests-limit: 10
+  assignees:
+    - "dependabot"


.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"
```
jsjoeio, line 6:
```suggestion
    interval: "weekly"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    time: "11:00"
+  open-pull-requests-limit: 10
+  assignees:
+    - "dependabot"
+
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: daily
```
jsjoeio, line 15:
```suggestion
    interval: weekly
```

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+    time: "11:00"
+  open-pull-requests-limit: 10
```
jsjoeio, line 17:
```suggestion
  open-pull-requests-limit: 5
```

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "daily"
+    time: "11:00"
+  open-pull-requests-limit: 10
```
jsjoeio, line 8:
```suggestion
  open-pull-requests-limit: 5
```

jsjoeio (changes requested):
Thanks for the contribution, @jawnsy! 🎉

In the past when I've used `dependabot` to manage dependencies, it tends to be pretty noisy and can be overwhelming to keep up with. 

Just took a look at the Security tab of our repo and it looks like we can set a lot of this up through there, including dependabot alerts.

image

Do you know the difference between using that vs adding this in like you've done?

jsjoeio:
bump @jawnsy

.github/dependabot.yml
jsjoeio: Thanks for the thorough response!

> Just listing out the tradeoffs we're making here as we evaluate this decision. However, since this is an easily reversible decision, regardless of which way we go, I would suggest we make a quick decision and iterate - we can always change the config later.

Sure, I'm all for iterating quickly. Ideally things like this start off as a discussion in an issue rather than a drive-by PR which gives us (maintainers) a chance to ask questions and discuss whether this makes sense for the project before moving to the PR stage. 

> Daily/Weekly schedule interval

I agree with most of the points you're making here but think this fails to recognize an important point: time. Like I said before, it gets noisy and it's hard to keep up with. 

You're coming in here proposing this change, which has value, but are you willing to be the one to maintain it? As you can see from the backlog of issues, we have a lot of other work to be done and merging this comes with a high cost.

> We will be informed about new versions more quickly, which can be important for security releases

I like the sound of that, but do we have evidence to say this is something we need? Probably not?

> Dependabot's PR Workflow vs Console

Thanks for explaining the differences! Sounds like this approach is preferable.

Okay just going to summarize my final thoughts here:
- next time, will you open up an issue so we discuss this before you open a PR?
- we can approve this and merge if you're willing to take ownership on reviewing and merging these PRs

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
```
jsjoeio, line 4:
```suggestion
  directory: "/"
  default_assignees:
      - "jawnsy"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+    interval: "daily"
+    time: "11:00"
+  open-pull-requests-limit: 10
+  assignees:
+    - "dependabot"
+
+- package-ecosystem: npm
+  directory: "/"
```
jsjoeio, line 13:
```suggestion
  directory: "/"
  default_assignees:
    - "jawnsy"
```

jsjoeio (changes requested):

.github/dependabot.yml
jawnsy: Sure, I'd be happy to write an issue next time. I didn't bother doing that because I didn't think this change would be objectionable :smile: 

> I like the sound of that, but do we have evidence to say this is something we need? Probably not?

As software maintainers, we're responsible for the supply chain and making sure that we have a clean bill of health with respect to both our code, as well as third-party dependencies that we depend on.

Here's some evidence, the outstanding advisories shown by `yarn audit` run against the latest `main` branch; it's possible that these aren't real issues, but it seems like good hygiene to stay up-to-date, provided that the risk to upgrade isn't significant (and you're in a better position to assess that than I am):

```
[coder@jawnsy-m code-server]$ yarn audit
yarn audit v1.22.10
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ low           │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ ini                                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >1.3.6                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ stylelint                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ stylelint > global-modules > global-prefix > ini             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1589                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ moderate      │ Use of a Broken or Risky Cryptographic Algorithm             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ elliptic                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=6.5.4                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ parcel-bundler                                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ parcel-bundler > node-libs-browser > crypto-browserify >     │
│               │ browserify-sign > elliptic                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1648                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ moderate      │ Use of a Broken or Risky Cryptographic Algorithm             │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ elliptic                                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=6.5.4                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ parcel-bundler                                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ parcel-bundler > node-libs-browser > crypto-browserify >     │
│               │ create-ecdh > elliptic                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1648                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Prototype Pollution in node-forge                            │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ node-forge                                                   │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >= 0.10.0                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ parcel-bundler                                               │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ parcel-bundler > node-forge                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1561                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
4 vulnerabilities found - Packages audited: 1216
Severity: 1 Low | 2 Moderate | 1 High
Done in 1.44s.
```

.github/dependabot.yml
jawnsy: I'd be happy to be the default assignee, but I'm not sure that's actually what you want, because I don't know this project as well as you do, and so wouldn't know whether upgrades are likely to break things. I would need some guidance from you regarding what is needed when deciding whether to approve/merge - what sort of manual testing do I need to do? Or can I just take a quick look at the changes and approve/merge provided that tests are passing?

jsjoeio: > Or can I just take a quick look at the changes and approve/merge provided that tests are passing?

If CI is passing, then it should be okay for you to merge. If it breaks something than I will take responsibility for fixing it and then it will for us to add more tests in order to continue being able to rely on CI and our testing suites.

.github/dependabot.yml
jsjoeio: And this one? 5 seems like a great place to start. Like you said, easily reversible so we can increase later if needed.

.github/dependabot.yml
jawnsy: lmao, living dangerously, I like it :joy:

jsjoeio: The only way to live 😉

jsjoeio: Okay feel free to make the changes and then we can approve and merge!

jawnsy: Applying this change, but I think you're using the old dependabot syntax (Dependabot Preview), and not the new one, since they were acquired/integrated into GitHub: https://docs.github.com/en/github/administering-a-repository/customizing-dependency-updates

I think we'd want to set me as reviewer, then?

.github/dependabot.yml
jawnsy: Looks like 5 is the default, according to the docs, so we can just remove this

> To keep pull requests manageable and easy to review, Dependabot raises a maximum of five pull requests to start bringing dependencies up to the latest version.

From the docs: https://docs.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#checking-the-status-of-version-updates

.github/dependabot.yml
jawnsy, line 15: @jsjoeio Are you OK sticking with daily? I have a preference for daily, but if you feel strongly about it, then I'm happy to go with weekly

jsjoeio: Well they're being assigned and reviewed by you so if you think you can handle daily, all you 😛

.github/dependabot.yml
jsjoeio: Oh huh I just pulled that from their docs actually so maybe it's outdated. https://dependabot.com/docs/config-file/#default_assignees

You're right though, we want you as the reviewer (and the assignee, probably). Good catch!

code-asher:
I'm down to own these if that's helpful.

.github/dependabot.yml
jawnsy: Yeah, that's for "Dependabot Preview", then they were acquired by GitHub and it became "Dependabot Native" or whatever... at least that's what I was able to gather based on the docs lol

.github/dependabot.yml
jawnsy: Putting this here so we can have a threaded conversation. cc @code-asher 

So, I think this is sorta blocked by #2559. I've learned recently that [Dependabot PR runs in Actions won't receive secrets](https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-dependabot#about-dependabot-version-updates-for-actions):

> **Note**: Pull requests for Dependabot version updates will trigger workflow runs with a read-only `GITHUB_TOKEN`. These workflow runs will not be granted access to any secrets.

However, we found in `cdr/coder-cli` that a workaround is to manually re-run the Dependabot PR builds, in which case the run will get secrets since it seems to use the credentials of the person triggering the re-run. This seems like sensible - though non-obvious - behavior.

So I don't know if this is still worth pursuing given the required workaround, or not. What are your thoughts?

.github/dependabot.yml
jsjoeio: Ah, that's confusing. Totally missed the yellow box saying that at the top lol

.github/dependabot.yml
code-asher: Personally if it was me reviewing the dependabot PRs I wouldn't mind re-running manually for now until we make the required CI fixes. If anything I'd consider it motivation to get CI fixed. :wink:

jsjoeio (dismissed):
LGTM!

@jawnsy 
good luck han solo gif

.github/dependabot.yml
```diff
@@ -0,0 +1,25 @@
...
+      interval: "daily"
+      time: "11:00"
+    assignees:
+      - "jawnsy"
+    reviewers:
+      - "jawnsy"
+    ignore:
+      # GitHub always delivers the latest versions for each major
+      # release tag, so handle updates manually
+      - dependency-name: "actions/*"
```
jawnsy, line 15: @cdr/code-server-reviewers FYI, I made two minor changes to this since it was last reviewed:

1. Ignore all of the official GitHub Actions actions, we'll need to manage these manually
2. Format dependabot.yml with Prettier as CI needs that

jsjoeio: Sweet! Thanks for the update!

jawnsy: @jsjoeio I'll need your review again, it seems your personal reviews don't count :raised_eyebrow: 

image

jsjoeio: Ah weird! Maybe it's cause you dismissed the stale one? I'll reapprove!
image

jsjoeio (approved):

.github/dependabot.yml
jawnsy: interesting, I wonder if it did that automatically because I added commits, that'd make sense I guess!

jawnsy: Thanks for the approval, excited to see how this goes... :crossed_fingers:

PR #2843, opened 2021-03-08
add instructions for deploying to a cloud provider 

bpmct:
I have been working on the [cdr/deploy-code-server](https://github.com/cdr/deploy-code-server) project which includes one-click buttons and guides to deploy code-server to a cloud provider, such as Linode or Heroku.

deploy-code-server repo

@jsjoeio and I have been discussing including parts of this in the code-server repo, as it is a great way to get started using it. It is only a few steps to get an environment that can be accessed from anywhere!

However, I think that we cannot simply add all of the docs and guides to this repo. Here's why:

- Certain providers (such as Heroku and Railway) require that some info lives in the base directory of the repository, such as the [heroku.yml](https://github.com/cdr/deploy-code-server/blob/main/heroku.yml) flle. This would mess up code-server's folder structure.
- Since the latest code-server version is the default branch, users will not see updates for deploy options until the next version. We may add support for other cloud providers or add fixes before then
- Providers like Railway and Heroku involve the user "forking" the repo to their account so they can modify their environment. It wouldn't be ideal for the user to fork the whole code-server codebase 

However, it may be valuable to add links to deploy to certain providers instead of linking to the entire project. We could include a partial table to README or install.md to include this info?

I think the source-of-truth for the docs and guides should live on the separate [deploy-code-server](https://github.com/cdr/deploy-code-server) repo though

Currently, our [install.md](https://github.com/cdr/code-server/blob/main/docs/install.md#app-engines-azure-heroku) currently has (now outdated) instructions for deploying. We could also add there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,11 @@ You can use whatever linux distribution floats your boat but in our [guide](./do
 
 ## Getting Started
 
-There are two ways to get started:
+There are three ways you can get started:
 
 1. Using the [install script](./install.sh), which automates most of the process. The script uses the system package manager (if possible)
 2. Manually installing code-server; see [Installation](./docs/install.md) for instructions applicable to most use cases
+3. Use our one-click buttons and guides to [deploy code-server to a popular cloud provider](https://github.com/cdr/deploy-code-server)
 
 If you choose to use the install script, you can preview what occurs during the install process:
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ There are three ways you can get started:
 
 1. Using the [install script](./install.sh), which automates most of the process. The script uses the system package manager (if possible)
 2. Manually installing code-server; see [Installation](./docs/install.md) for instructions applicable to most use cases
-3. Use our one-click buttons and guides to [deploy code-server to a popular cloud provider](https://github.com/cdr/deploy-code-server)
+3. Use our one-click buttons and guides to [deploy code-server to a popular cloud provider](https://github.com/cdr/deploy-code-server) ⚡
 
 If you choose to use the install script, you can preview what occurs during the install process:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -24,10 +24,11 @@ You can use whatever linux distribution floats your boat but in our [guide](./do
 
 ## Getting Started
 
-There are two ways to get started:
+There are three ways you can get started:
 
 1. Using the [install script](./install.sh), which automates most of the process. The script uses the system package manager (if possible)
 2. Manually installing code-server; see [Installation](./docs/install.md) for instructions applicable to most use cases
+3. Use our one-click buttons and guides to [deploy code-server to a popular cloud provider](https://github.com/cdr/deploy-code-server)
```
jsjoeio, line 31: I think this is perfect!

jsjoeio (changes requested):
I like what you added to the `README.md`. 

> I think the source-of-truth for the docs and guides should live on the separate deploy-code-server repo though

Agreed! I like the approach you've taken - add a short text snippet and link out to that repo.

I think we should update the `install.md` as well like you said then merge this

bpmct:
So, currently the install script takes up a large % of the README, which I think is good. However, are there potentially other things we could do to draw attention to this "quick deploy" option?

A few ideas:
- put a badge like this somewhere (lol): [deploy-code-server](https://github.com/bpmct/deploy-code-server) 
- add a small table/button to deploy to ONE cloud provider, which links to the guide on the external repo
- do nothing, step 3 will be visible enough for more people :)

jsjoeio:
I could be on board with any of those options! I'll defer to you on which you think is best :D

bpmct:
Ok just added an emoji lol

PR #2852, opened 2021-03-09
dev(testing): separate unit and e2e tests

### LINKED ISSUE #2646: [dev]: separate unit and end-to-end testing
> In the future, we should consider separating the tests. We could have test:unit and test:e2e and then have separate jobs for each in ci.

When #2643 gets merged, running `yarn test` will only work if you have `code-server` running. Not ideal. 

Instead, we should separate them and make it so both unit tests and e2e tests can be run locally and separately in ci.

jsjoeio:
This PR fixes #2646 by separating unit and e2e testing. Hopefully this speeds things up with CI.


.eslintrc.yaml
```diff
@@ -32,7 +32,13 @@ rules:
...
   eqeqeq: error
   import/order:
-    [error, { alphabetize: { order: "asc" }, groups: [["builtin", "external", "internal"], "parent", "sibling"] }]
+    [
+      error,
+      {
+        alphabetize: { order: "asc" },
+        groups: [["builtin", "external", "internal"], "parent", "sibling"],
+      },
+    ]
```
jsjoeio, line 41: 💄

.eslintrc.yaml
```diff
@@ -23,7 +23,7 @@ rules:
   # if they are unused.
   "@typescript-eslint/no-unused-vars": ["error", { "args": "none" }]
   # For overloads.
-  no-dupe-class-members: off
+  no-dupe-class-members: 0
```
jsjoeio, line 26: 0 === off (my yaml linter complained about this so I fixed it)

test/unit/test-plugin/src/index.ts
```diff
@@ -1,22 +1,25 @@
+// @ts-ignore: this is a virtual module which we mock in the test so it doesn't exist when tsc runs
```
jsjoeio, line 1: When this line isn't here, tsc says "can't find module code-server" and then it doesn't compile and the tests fail. Any ideas on a better fix? @code-asher

jsjoeio: Nevermind, I figured it out

.eslintrc.yaml
oxy: Um, it complained about this `off` but not all the other `off`s below it, confusingly?
I think `off` makes sense here

oxy (commented):
Not sure about the `off` -> `0` thing, LGTM otherwise!

.eslintrc.yaml
jsjoeio: fixed!

oxy (approved):
LGTM!

PR #2866, opened 2021-03-12
chore: run builds on pull requests

jawnsy:
* Run builds when changes are pushed to the main branch, or for pull
  requests opened against the main branch. This is a subtle change
  in behavior because, previously, builds would've run on pushes to
  any branch, regardless if a pull request is open yet.
* Update to GitHub actions/checkout v2, which performs a shallow
  clone by default, which should speed up the build.

Closes: #2559 (:crossed_fingers:)

--

I believe that this should work, since only one job receives secrets according to the workflow definition (unless there's an administrative setting somewhere for global secrets passed using environment variables). However, I think the only way to be certain is to merge this and then open a pull request from a fork. If this breaks things or doesn't work correctly, we can always revert and try again.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,12 +1,18 @@
 name: ci
 
-on: [push]
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
 
 jobs:
   fmt:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


code-asher (approved):

jsjoeio (approved):
Awesome work @jawnsy 🎉 Thanks for the detailed PR description too! 

Feel free to merge whenever you want!

PR #2889, opened 2021-03-15
Remove fs-extra

code-asher:
There's more detail in the commits.


src/node/cli.ts
```diff
@@ -510,13 +509,20 @@ export async function readConfigFile(configPath?: string): Promise<ConfigArgs> {
...
+  } catch (error) {
+    // EEXIST is fine; we don't want to overwrite existing configurations.
+    if (error.code !== "EEXIST") {
+      throw error
+    }
   }
 
-  const configFile = await fs.readFile(configPath)
-  return parseConfigFile(configFile.toString(), configPath)
+  const configFile = await fs.readFile(configPath, "utf8")
```
jawnsy, line 524: nice fix :+1: 

is there a way to configure linters to enforce usage of utf8?

package.json
```diff
@@ -25,7 +25,7 @@
     "_____": "",
     "fmt": "./ci/dev/fmt.sh",
     "lint": "./ci/dev/lint.sh",
-    "test": "./ci/dev/test.sh",
+    "test": "echo 'Run yarn test:unit or yarn test:e2e' && exit 1",
```
jawnsy, line 28: Will this affect anything in CI?

jawnsy (approved):

package.json
jsjoeio: It shouldn't, unless I forgot to change something in https://github.com/cdr/code-server/pull/2852

code-asher: The file it was running doesn't exist so if it did I think we'd have test failures. :smile:

code-asher:
Realized I didn't need `unlinkf` and `rmdir` with recursive works on files so I just pushed a change to use rmdir instead. Although semantically it feels a bit weird.

src/node/cli.ts
code-asher: The linter does recognize this, I just personally prefer to pass "utf8" rather than call `toString()` afterward haha. I'm not sure if there's a performance benefit or anything though.

code-asher: Oh wait I might have misread. Like enforce using "utf8" instead of `toString()`? I'd love that.

jawnsy: makes sense, I think configuring utf8 on the stream makes things easier to use... sometimes these encoding things can be non-obvious in code reviews and lead to bugs that only happen sometimes (i.e. not with English/Latin-script text), and unfortunately older languages like JavaScript and Java have bad defaults, since they began in a pre-Unicode era

jsjoeio (dismissed):
image

PR #2891, opened 2021-03-15
feat: add prettier-plugin-sh

jsjoeio:
This PR adds prettier-plugin-sh and removes shfmt.

_Note: it makes some minor modifications to the way our `.sh` files looked before. I tried setting some of the parser options based on [the docs](https://www.npmjs.com/package/prettier-plugin-sh#parser-options) but nothing seemed to work so left it as is. It uses the default options based on [mvdan/sh](https://github.com/mvdan/sh)_


ci/images/debian10/Dockerfile
```diff
@@ -42,7 +42,6 @@ ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
 
 # Install Go dependencies
 ENV GO111MODULE=on
-RUN go get mvdan.cc/sh/v3/cmd/shfmt
 RUN go get github.com/goreleaser/nfpm/cmd/nfpm@v1.9.0
```
jawnsy, line 45: If we're able to drop `nfpm` next, does this mean we can drop the dependency on go?

jawnsy (dismissed):
LGTM

ci/images/debian10/Dockerfile
jsjoeio: I hope so! 🤞 tbh, I don't know what `nfpm` is used for, but hopefully we can drop that and Go soon! I'll add a note to the Clean Up issue to investigate

jawnsy: seems to be something release-related, I think we should clean up all of those images, particularly in light of the fact that CentOS changed their model recently with CentOS Stream

ci/lib.sh
```diff
@@ -75,7 +75,7 @@ download_artifact() {
   local tmp_file
   tmp_file="$(mktemp)"
 
-  curl -fsSL "$(get_artifact_url "$artifact_name")" > "$tmp_file"
+  curl -fsSL "$(get_artifact_url "$artifact_name")" >"$tmp_file"
```
jawnsy, line 78: just a personal preference, I like to use the long forms of the flags to make it clearer what's happening

ci/lib.sh
```diff
@@ -2,11 +2,11 @@
 set -euo pipefail
 
 pushd() {
-  builtin pushd "$@" > /dev/null
+  builtin pushd "$@" >/dev/null
```
jawnsy, line 5: I just might steal this code...

ci/images/debian10/Dockerfile
jsjoeio: Good call. I'll add a note for that as well.

ci/lib.sh
jsjoeio: do it! who knows who wrote it 😂

jawnsy (approved):

ci/lib.sh
jsjoeio: I didn't write this 🤷‍♂️ but agree, I generally like things to be extra clear

SuperSandro2000:
Whats the motivation behind this?

PR #2906, opened 2021-03-17
rename references of "coder could"

bpmct:
Corresponds with https://github.com/cdr/cloud-agent/pull/12

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ Consider running code-server with the beta flag `--link` if you don't want to wo
 
 ```bash
 $ code-server --link
-Proxying code-server to Coder Cloud, you can access your IDE at https://valmar-jon.cdr.co
+Proxying code-server, you can access your IDE at https://valmar-jon.cdr.co
 ```
 
 ## FAQ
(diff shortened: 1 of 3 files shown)


jawnsy:
Unrelated to your change, but under Hiring there's a link to a job posting which is still visible, but not shown on the [main listing page](https://jobs.lever.co/coder), so I'm not sure if it was taken down - in any case I'd suggest linking to the main page rather than a specific posting, and making the language more generic (something like "Coder is hiring! See our jobs page") so that it's not referring to a specific position, as the open positions change over time

Maybe @cami333 has thoughts here

cami333:
> Unrelated to your change, but under Hiring there's a link to a job posting which is still visible, but not shown on the [main listing page](https://jobs.lever.co/coder), so I'm not sure if it was taken down - in any case I'd suggest linking to the main page rather than a specific posting, and making the language more generic (something like "Coder is hiring! See our jobs page") so that it's not referring to a specific position, as the open positions change over time
> 
> Maybe @cami333 has thoughts here

@jawnsy I would prefer to just put the careers page link as you suggested. :)

Issue #2913, opened 2021-03-17
Make Accounts icon visible when signed in with GitHub


jsjoeio:
Right now, it sometimes works and sometimes doesn't and it's confusing (see: https://github.com/cdr/code-server/discussions/2568)

It works when you're signed in with GitHub so maybe only show it then.

Suggested by @@AurangAsif here: https://github.com/cdr/code-server/discussions/2568#discussioncomment-494759

atn:
+1 to this, the user icon that doesn't work could mislead people into thinking that there is something wrong with their instance (happened to me and i stumbled upon this issue)

PR #2922, opened 2021-03-17
feat(lib/vscode): add log out to application menu

### LINKED ISSUE #778: Log out button
## Description

I couldn’t find a way to logout of the code server once logged in to the code-server

The following docker command was used to host the code-server

```bash
docker run -it -p 127.0.0.1:8443:8443 -v "${PWD}:/home/coder/project" codercom/code-server --allow-http --no-auth
```

This functionality is kind of important to me as sometimes I have to use different machines temporarily. Is there by chance a way for me to logout? Thanks a lot.

jsjoeio:
This adds a new option to the Application Menu called Log out. It deletes the code-server cookie and logs a user out.

## TODOs
- [x] use vscode logger
- [x] refactor `Cookie` to be in `lib/vscode` so we can use it
- [x] write e2e test
- [x] refactor `Cookie` and don't share between `lib/vscode`

Fixes #778

### DIFF
--- a/lib/vscode/src/vs/server/common/cookie.ts
+++ b/lib/vscode/src/vs/server/common/cookie.ts
@@ -1,3 +1,3 @@
 export enum Cookie {
-  Key = "key",
+  Key = 'key',
 }
--- a/lib/vscode/src/vs/server/common/cookie.ts
+++ b/lib/vscode/src/vs/server/common/cookie.ts
@@ -1,0 +1,3 @@
+export enum Cookie {
+  Key = "key",
+}
--- a/lib/vscode/src/vs/workbench/browser/parts/titlebar/menubarControl.ts
+++ b/lib/vscode/src/vs/workbench/browser/parts/titlebar/menubarControl.ts
@@ -40,6 +40,7 @@ import { KeybindingWeight } from 'vs/platform/keybinding/common/keybindingsRegis
 import { IsWebContext } from 'vs/platform/contextkey/common/contextkeys';
 import { ICommandService } from 'vs/platform/commands/common/commands';
 import { ILogService } from 'vs/platform/log/common/log';
+import { Cookie } from 'vs/server/common/cookie';
 
 export abstract class MenubarControl extends Disposable {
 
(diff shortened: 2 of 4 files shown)


lib/vscode/src/vs/server/common/cookie.ts
```diff
@@ -0,0 +1,3 @@
+export enum Cookie {
+  Key = 'key',
+}
```
oxy, line 3: Missing a newline at the end of the file.

oxy (changes requested):
LGTM! Just a newline at the end of `lib/vscode/src/vs/server/common/cookie.ts` should be all that's needed.

oxy:
So I was wondering why `tsc` was "failing" to use `outDir`...

Turns out it was, but since we import from `lib/vscode` in `src` now, what used to be `out/node/entry.js` has now become `out/src/node/entry.js` (and there's an `out/lib/vscode/.../cookie.js`).

Since our build script, `./ci/steps/build-code-server.sh` looks for `out/node/entry.js`, it fails.

test/unit/util.test.ts
```diff
@@ -1,4 +1,5 @@
 import { JSDOM } from "jsdom"
+import { Cookie as CookieEnum } from "../../lib/vscode/src/vs/server/common/cookie"
```
oxy, line 2: Importing this from `lib/vscode` breaks the expected directory structure in `out`, see https://github.com/cdr/code-server/pull/2922#issuecomment-801746756

jsjoeio:
> I think it might be best to just duplicate the key cookie.

Ah okay. Yeah it doesn't seem worth it to have to fix everything else for one thing to not be duplicated 😂 thanks for looking into this! I'll convert to draft and fix!

oxy (approved):
LGTM!

bpmct:
Love this!

binaryfire:
Nice! Just a thought - can you also add it to the profile icon at the bottom of the left sidebar? That seems like the most intuitive place to look for it.

jsjoeio:
@binaryfire thanks for the suggestion! There's some issues with the profile icon (see #2913) but when we work on that, we can consider it! Feel free to leave more thoughts there

PR #2952, opened 2021-03-23
dev: add release:prep script

jsjoeio:
This PR adds a new script called `release:prep` to help with the release process by asking for the new version and then automating the following steps:
- Update in `package.json`
- Update in [./docs/install.md](../docs/install.md)
- Update in [./ci/helm-chart/README.md](../ci/helm-chart/README.md)
- Remember to update the chart version as well on top of appVersion in `Chart.yaml`.
- Update all values mentioning the version
- Update the code coverage badge (see [here](#updating-code-coverage-in-readme) for instructions)
- Update the docs badge in [./README.md](../README.md)

## Demo

https://user-images.githubusercontent.com/3806031/112201249-4826f000-8bcd-11eb-8354-dde5c1816d1a.mov

## TODOS
- [x] check if dry_run works with `CMD="echo"`
- [x] fix DRY_RUN initialized at top
- [x] refactor git check branch and push upstream 
- [x] fix `git add` mistake


ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+
+main() {
+  if [ "${DRY_RUN-1}" -eq 0 ]; then
+    echo "Performing a dry run..."
+    # Disabled because shellcheck wants me to do this
+    # SC2209: Use var=$(command) to assign output (or quote to assign string).
+    # But neither of those match my needs
+    # We do this so that the dry run echos the commands instead of running them
+    # shellcheck disable=SC2209
+    CMD=echo
```
code-asher, line 20: Quotes should work! (I think.)

```suggestion
    CMD="echo"
```

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+# 1. Check that you have a $GITHUB_TOKEN set and hub installed
+# 2. Update the version of code-server (package.json, docs, etc.)
+# 3. Update the code coverage badge in the README
+# 4. Open a draft PR using the release_template.md and view in browser
+# If you want to perform a dry run of this script run DRY_RUN=0 yarn release:prep
+
+set -euo pipefail
+
+main() {
+  if [ "${DRY_RUN-1}" -eq 0 ]; then
```
code-asher, line 13: Thoughts on comparing to one instead? Setting `DRY_RUN` to zero feels to me like it should be falsy. 

We also might want to use `=` in case `DRY_RUN` is blank or some other non-integer.

```suggestion
  if [ "${DRY_RUN-}" = 1 ]; then
```

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+    # shellcheck disable=SC2209
+    CMD=echo
+  else
+    CMD=''
+  fi
+
+  cd "$(dirname "$0")/../.."
+
+  # Check that $GITHUB_TOKEN is set
+  if [[ -z "${GITHUB_TOKEN}" ]]; then
```
code-asher, line 28: In case it's not set we can add `-`. Also fun fact, with the double bracket syntax you don't need the double quotes on the left side of expressions (but totally fine if you want to leave them for consistency).

```suggestion
  if [[ -z ${GITHUB_TOKEN-} ]]; then
```

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+    echo "sd could not be found."
+    echo "We use this when updating files across the codebase."
+    echo -e "See docs here: https://github.com/chmln/sd#installation"
+    exit
+  fi
+
+  # Check that they have node installed
+  if ! command -v node &>/dev/null; then
+    echo "node could not be found."
+    echo "That's surprising..."
```
code-asher, line 70: Ahahahahaha true true :laughing:

oxy (changes requested):
In line 36 of the script, you're checking for `hub` but towards the end of the script, you're using `gh` - they're not the same tool!

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+  # Check that $GITHUB_TOKEN is set
+  if [[ -z "${GITHUB_TOKEN}" ]]; then
+    echo "We couldn't find an environment variable under GITHUB_TOKEN."
+    echo "This is needed for our scripts that use hub."
+    echo -e "See docs regarding GITHUB_TOKEN here under 'GitHub OAuth authentication': https://hub.github.com/hub.1.html"
+    exit
+  fi
+
+  # Check that hub is installed
+  if ! command -v hub &>/dev/null; then
```
oxy, line 36: check for `gh` instead of `hub` here

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+  $CMD yarn badges
+  # Updates the svg to be green for the badge
+  $CMD sd "red.svg" "green.svg" ../../README.md
+
+  $CMD git add . && $CMD git commit -am "chore(release): bump version to $CODE_SERVER_VERSION_TO_UPDATE"
+
+  CURRENT_BRANCH=$(git branch --show-current)
+  # Note: we need to set upstream as well or the gh pr create step will fail
+  # See: https://github.com/cli/cli/issues/575
+  $CMD git push -u origin "$CURRENT_BRANCH"
```
code-asher, line 104: This could be sketchy, most people probably use `origin` as the name for upstream but it's no guarantee.

Also my version of git doesn't seem to recognize `--show-current`? (I'm on 2.17.1.)

Maybe we can do a check if upstream is set and error if it isn't using something like:

```
git rev-parse --abbrev-ref --symbolic-full-name @{u}
```

code-asher (changes requested):
Awesome to see this automated!

ci/build/release-prep.sh
code-asher: Or maybe it's sufficient to just `git push`?

ci/build/release-prep.sh
```diff
@@ -0,0 +1,117 @@
...
+
+  # Ensure the tests are passing and code coverage is up-to-date
+  echo -e "Running unit tests and updating code coverage...\n"
+  $CMD yarn test:unit
+  # Updates the Lines badge in the README
+  $CMD yarn badges
+  # Updates the svg to be green for the badge
+  $CMD sd "red.svg" "green.svg" ../../README.md
+
+  $CMD git add . && $CMD git commit -am "chore(release): bump version to $CODE_SERVER_VERSION_TO_UPDATE"
```
code-asher, line 99: I think `-a` stages files so we might not need the `git add`.

ci/build/release-prep.sh
jsjoeio: > they're not the same tool!
I may be a noob at writing scripts but I'm not THAT much of a noob.
I know they're not the same tool 😂 That's intentional! 

`gh` is `hub`'s successor, hence why I'm using it here.  Eventually, the plan is to remove the use of `hub` and only use `gh` but now, the maintainer releasing should have both because it's used [here](https://github.com/cdr/code-server/blob/main/ci/build/release-github-assets.sh#L18) and [here](https://github.com/cdr/code-server/blob/main/ci/build/release-github-draft.sh#L10). 

I'll make a follow-up issue to remove the use of `hub` and replace with `gh`.

jsjoeio: https://github.com/cdr/code-server/issues/2960

ci/build/release-prep.sh
jsjoeio: ooo i didn't even think to try! okay i'll test to verify. good call! (then I can remove shellcheck supression)

ci/build/release-prep.sh
jsjoeio: Doesn't matter to me - I guess `0` is falsey in the JS world.

Makes sense though and good call! I'll fix. 

(note: I probably still need a number after the `-` in `${DRY_RUN-}` right? otherwise it will be undefined and exit the script?)

jsjoeio: Actually, I see your comment below. TIL!

ci/build/release-prep.sh
jsjoeio: TIL two things:
1. thought you had to put something after the `-` to set it to _something_ but looks like you don't.
2. didn't know that about the double quotes 🎉

ci/build/release-prep.sh
jsjoeio: I should put a link to that Halo 3 image 🤣

ci/build/release-prep.sh
jsjoeio: > Also my version of git doesn't seem to recognize --show-current? (I'm on 2.17.1.)

Time to update 😛 jk, looks like it's new as of [2.22](Also my version of git doesn't seem to recognize --show-current? (I'm on 2.17.1.)).

I should use something more backwards-compatible. 

> This could be sketchy, most people probably use origin as the name for upstream but it's no guarantee.

Good call! 

> git rev-parse --abbrev-ref --symbolic-full-name @{u}

🤯 time to go look up what this does.

> git push -u ${u}

oo that could work too. Either way, this is great feedback. I'll refactor.

ci/build/release-prep.sh
jsjoeio: noob mistake 🤦‍♂️

ci/build/release-prep.sh
jsjoeio: it worked! 🎉

ci/build/release-prep.sh
jsjoeio: Had to go with something different. I'll push up and would be curious to get your feedback

ci/build/release-prep.sh
```diff
@@ -0,0 +1,119 @@
...
+  CODE_SERVER_CURRENT_VERSION=$(node -pe "require('./package.json').version")
+  # Ask which version we should update to
+  # In the future, we'll automate this and determine the latest version automatically
+  echo "Current version: ${CODE_SERVER_CURRENT_VERSION}"
+  # The $'\n' adds a line break. See: https://stackoverflow.com/a/39581815/3015595
+  read -r -p "What version of code-server do you want to update to?"$'\n' CODE_SERVER_VERSION_TO_UPDATE
+
+  echo -e "Great! We'll prep a PR for updating to $CODE_SERVER_VERSION_TO_UPDATE\n"
+  # I can't tell you why but
+  # when searching with rg, the version needs to in this format: '3\.7\.5'
```
code-asher, line 82: I think by default `rg` uses regular expressions so the `.` matches any character and needs to be escaped to match a literal `.`.

One way around this would be to use `--fixed-strings` which treats the input as a literal string instead of a regular expression.

ci/build/release-prep.sh
oxy: Do we check for `gh` elsewhere? I'm not too sure >.<

oxy (approved):
LGTM!

code-asher (approved):
:rocket:

ci/build/release-prep.sh
jsjoeio: TIL! I'll add that in and then merge! 🎉

oxy: Ooh, didn't know about fixed-strings!

ci/build/release-prep.sh
jsjoeio: I think in the `update-vscode` script too since it's used there.

PR #2961, opened 2021-03-24
fix(lib/vscode): remove native-keymap and keytar

oxy:
native-keymap is only used in the electron process, so we don't need it.
This allows us to drop our libx11 and libxkbfile build dependencies.


lib/vscode/src/typings/native-keymap.d.ts
```diff
@@ -0,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
```
jsjoeio, line 1: If we don't need it, how come we are adding in the types? Do we need this for some reason?

jsjoeio (changes requested):
LGTM! Just confused about the `native-keymap` declaration file if we're removing the dependency

lib/vscode/src/typings/native-keymap.d.ts
oxy: It's kind of like what we're doing for electron - we need the types for it to compile correctly - but then it never gets loaded, so we don't need the module itself - _just_ the types so tsc won't complain.

oxy:
Update: I spent the day investigating `keytar`, which initially was a much scarier dep to remove, but also allows us to drop our dependency on `libsecret-dev`, which means CI no longer needs to pull in any desktop-specific Linux packages! :tada:

I put that change in this PR too, since it felt like the logical thing to do (they touch nearly the same lines in the Dockerfiles, which means that if I made two PRs, one would have to be rebased and have conflicts resolved).

There's some scarier stuff going on with how `keytar` is used in vscode, but it's not an issue to drop it:

`keytar` is provided for extensions to be able to use the operating system keyring (when running natively) to store password secrets. To make this work, VSCode actually has a hook on require that replaces `require('keytar')` in an extension with a RPC-like system that shares the same API, but is proxied elsewhere deep inside.

For the web, VSCode already includes `class LocalStorageCredentialsProvider implements ICredentialsProvider` which stores them in a JSON in local storage, and provides a keytar-esque API.

I've tested this by writing a custom VSCode extension that uses "keytar" injected by vscode to set credentials, and it successfully stores them in LocalStorage. I also tested against GH Codespaces, which shows the same behavior.

I've also added notes to the `lib/vscode/src/typings` files to indicate why I pulled them in.

lib/vscode/package.json
```diff
@@ -106,7 +104,6 @@
     "@types/graceful-fs": "4.1.2",
     "@types/gulp-postcss": "^8.0.0",
     "@types/http-proxy-agent": "^2.0.1",
-    "@types/keytar": "^4.4.0",
```
oxy, line 109: Note for review: `@types/keytar` is deprecated since `keytar` includes types - I removed both and added a `typings` file.

jsjoeio (approved):
Woo! Great work and thanks for adding great comments for future us!

Ship it!

image

Issue #2903, opened 2021-03-16
[dev]: split up test & release into separate workflows


jsjoeio:
Right now, we have one main workflow: [ci.yaml](https://github.com/cdr/code-server/blob/main/.github/workflows/ci.yaml)

When you have an issue with `fmt` or `lint`, you have to push up and run the entire workflow again. We'd like to split up the workflows into two:
1. main workflow - lint, fmt, test:unit
2. release + e2e workflow

We also don't need to run the docker or arm64 steps on every workflow either. 

We're hoping this + smart caching could speed up the workflow.

PR #2966, opened 2021-03-25
dev(ci): complete refactor

oxy:
Slowly reorganize CI steps and switch to using GH actions instead of docker images where possible. Initial test only replaces the `fmt` runner.


ci/steps/nfpm.sh
```diff
@@ -0,0 +1,386 @@
+#!/bin/sh
+set -e
+# Code generated by godownloader on 2019-12-25T12:47:14Z. DO NOT EDIT.
+# NOTE@coder: edited anyway - did not support arm64 despite builds existing on GH
+# switch back when upstream fixes!
```
oxy, line 5: https://github.com/goreleaser/nfpm/issues/314

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
     branches:
       - main
 
+# Note: if: success() is used in several jobs -
+# this ensures that it only executes if all previous jobs succeeded.
```
jsjoeio, line 12: 👏

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+
+# if: steps.cache-yarn.outputs.cache-hit != 'true'
+# will skip running `yarn install` if it successfully fetched from cache
```
jsjoeio, line 15: 👏

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+    name: Pre-build checks
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: Run ./ci/steps/fmt.sh
-        uses: ./ci/images/debian10
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Install node.js v12
```
jsjoeio, line 25:
```suggestion
      - name: Install Node.js v12
```

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+
+      - name: Install dependencies
+        if: steps.cache-yarn.outputs.cache-hit != 'true'
+        run: yarn --frozen-lockfile
+
+      - name: Build code-server
+        run: yarn build
+
+      # Parse the hash of the latest commit inside lib/vscode
+      # use this to avoid rebuilding it if nothing changed
```
jsjoeio, line 85: Genius! 🔥

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+      - name: Get latest lib/vscode rev
+        id: vscode-rev
```
jsjoeio, line 87: What's this for?

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
-          name: npm-package
-          path: ./release-npm-package
-      - name: Run ./ci/steps/release-packages.sh
-        uses: ./ci/images/centos7
-        with:
-          args: ./ci/steps/release-packages.sh
-      - name: Upload release artifacts
-        uses: actions/upload-artifact@v2
+
+      - name: Install node.js v12
```
jsjoeio, line 144:
```suggestion
      - name: Install Node.js v12
```

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+        uses: actions/setup-node@v2
         with:
-          name: release-packages
-          path: ./release-packages
+          node-version: "12"
+
+      - name: Install nfpm
```
jsjoeio, line 149: What's this used for? I had a note to myself to see if we could remove it

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
-          args: ./ci/steps/release-packages.sh
+
+      - name: Decompress npm package
+        run: tar -xzf package.tar.gz
+
+      - name: Build standalone release
+        run: yarn release:standalone
+
+      - name: Sanity test standalone release
+        run: yarn test:standalone-release
```
jsjoeio, line 166: I wonder if this should be grouped with the testing workflow

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+      AR: aarch64-linux-gnu-ar
+      CC: aarch64-linux-gnu-gcc
+      CXX: aarch64-linux-gnu-g++
+      LINK: aarch64-linux-gnu-g++
+      NPM_CONFIG_ARCH: arm64
+
     steps:
       - uses: actions/checkout@v2
+
+      - name: Install node.js v12
```
jsjoeio, line 191:
```suggestion
      - name: Install Node.js v12
```

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
-          # For whatever reason only MacOS needs it.
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Decompress npm package
+        run: tar -xzf package.tar.gz
+
+      - name: Build standalone release
+        run: yarn release:standalone
+
+      - name: Replace node with arm64 equivalent
```
jsjoeio, line 215: How come we do this? (might be helpful to add a comment above)

.github/workflows/ci.yaml
```diff
@@ -8,177 +8,303 @@ on:
...
+      - name: Untar code-server file
+        run: |
+          cd release-packages && tar -xzf code-server*-linux-amd64.tar.gz
+
+      - name: Install dependencies
+        if: steps.cache-yarn.outputs.cache-hit != 'true'
+        run: yarn --frozen-lockfile
+
+      # HACK: this shouldn't need to exist, but put it here anyway
+      # in an attempt to solve Playwright cache failures.
```
jsjoeio, line 270: OOO interesting.. bug in playwright?

package.json
```diff
@@ -65,6 +65,7 @@
     "parcel-bundler": "^1.12.4",
     "prettier": "^2.2.1",
     "prettier-plugin-sh": "^0.6.0",
+    "shellcheck": "^1.0.0",
```
jsjoeio, line 68: How come we're adding this in?

jsjoeio (changes requested):
Great work! Initial pass looks 🔥

A couple things:
- thank you for adding comments! 
- nit: can we use spell Node with a capital "N"? 😛
- can we split up workflows for main, release, testing? https://github.com/cdr/code-server/issues/2903 that way, if a test fails, we only have to re-run those jobs

oxy:
> * thank you for adding comments!

:smile: 

> * nit: can we use spell Node with a capital "N"? :stuck_out_tongue:

Sure! patching that soon.

> * can we split up workflows for main, release, testing? #2903 that way, if a test fails, we only have to re-run those jobs

I'm hoping that since the CI is much faster in general (lint/fmt now only takes 3min), and since we cache wherever possible right now (currently yarn build dependencies, VSCode build - will refactor all the scripts in `ci/` in a future PR once this lands, and also cache the `package` steps), we shouldn't strictly need to split it into multiple workflows. In addition, I'm not sure if its possible for items in one workflow to depend on another.

.github/workflows/ci.yaml
oxy: "bug"-ish - it sometimes writes browser binaries to `~/.cache/`, sometimes doesn't, and sometimes it looks for it in there and can't find it.

Wasn't sure about adding home-relative directories to our cache, so I just threw that in there.

package.json
oxy: `shellcheck` is also offered as an npm package that downloads the latest release from GitHub - see https://github.com/gunar/shellcheck/blob/master/download.js - and I figured that this would make it easier for writing CI scripts, and also for contributors, since `shellcheck` would just automatically be pulled and live in `node_modules` instead of having to install it to the entire system.

.github/workflows/ci.yaml
oxy: It's all the magic above - I set `AR`/`CC`/`CXX`/`LINK` to their `aarch64` cross compile variants in the job's `env`, and then yarn would use those to build; however qemu-user-static doesn't correctly run node aarch64 on 16.04 (tested locally), so I just sub in the "native"/x86-64 node with the arm64 node before we package.

.github/workflows/ci.yaml
oxy: I'm not sure if we should move this to the e2e workflow - this test essentially downloads and installs a vscode theme extension, just as a very quick test to see if `code-server` can run correctly. What do you think we should do?

.github/workflows/ci.yaml
oxy: `nfpm` is what we're using right now to build `.deb`/`.rpm` files - it may be possible to move off nfpm, but I think that's better suited for another PR down the line :p

.github/workflows/ci.yaml
oxy: this - the hash of the last commit that touched a file inside `lib/vscode` - is the "key" that I'm using to cache the vscode build - it takes about 5-10 minutes in CI, so if we've built it once before and no files inside changed, just pull the build from the cache and then package it up for the next step.

to check history, I needed to set `fetch-depth: 0` in `actions/checkout`, but a full clone instead of a shallow clone only costs us about three to five seconds right now.

jsjoeio: Ahh, I see now. That makes sense! Might be worth adding what you wrote as part of your comment so Future us won't forget!

.github/workflows/ci.yaml
jsjoeio: Sounds good to me!

.github/workflows/ci.yaml
jsjoeio: I'll lean towards your judgement! If you think it's fine here, let's leave it. We can always move it later if needed!

.github/workflows/ci.yaml
jsjoeio: Alrighty then! Might be worth adding a comment here just summarizing that.

.github/workflows/ci.yaml
jsjoeio: Argh that's a bummer! Well we can always look into it later. Glad you found a workaround!

package.json
jsjoeio: Ah gotcha. Sounds good to me!

jsjoeio:
> I'm hoping that since the CI is much faster in general (lint/fmt now only takes 3min), and since we cache wherever possible right now (currently yarn build dependencies, VSCode build - will refactor all the scripts in ci/ in a future PR once this lands, and also cache the package steps), we shouldn't strictly need to split it into multiple workflows. In addition, I'm not sure if its possible for items in one workflow to depend on another.

Ah, I see what you mean. Yeah not sure if that's possible either. Okay no worries then!

code-asher (dismissed):
I haven't been able to do a full review yet, just dropping this here so we remember to get the libc dependencies back down.

.github/workflows/ci.yaml
jsjoeio: Thoughts on adding a comment here?

jsjoeio (commented):
Nice work 🔥

Left one comment asking you to document a couple lines in the CI file. I'll let @code-asher give the final approval

.github/workflows/ci.yaml
oxy: I thought that it made the most sense to explain the situation above the whole workflow, before I set the env vars etc., so I put it here - https://github.com/cdr/code-server/pull/2966/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR181-R187 (you might need to click `Load diff` for the lines to get highlighted)

jsjoeio: Ah, I missed that! LGTM!

jsjoeio (approved):
Wait for second approval but LGTM!

oxy:
Hi there! This has been stalling for a while, mostly because building against an old version of glibc for CentOS 7 on a newer version of Ubuntu turns out to be not an easy process.

I haven't pushed these experiments, but I know they don't work:
- [glibc_version_header](https://github.com/wheybags/glibc_version_header) - repo says it doesn't work with libstdc++ easily
- `-static-libgcc -static-libstdc++ -static` - does generate `.o` files instead of `.so` files but then `node-gyp` creates a shared object after the fact anyway :/
- `statifer` on the `.node` led to weird bugs on my setup
- I also looked at the `node-gyp` source code to see if I can force static `.node` files, but I felt lost in there so I bailed after a few hours.

Over the next days, I'll be looking at docker buildx, and at "dropping" compatibility for CentOS 7 for per-pull request CI and moving the slower CentOS 7 build to a tag/draft release workflow.

I really hope we can get new CI soon - I've been fighting this for weeks x.x

jsjoeio:
No worries at all! Thanks for the update.

> Over the next days, I'll be looking at docker buildx, and at "dropping" compatibility for CentOS 7 for per-pull request CI and moving the slower CentOS 7 build to a tag/draft release workflow.

Is it possible to break up this work? i.e. get this PR to a stable state, review, merge and then do the remaining work in a follow-up PR? Just wondering if that would help at all.

oxy:
We definitely could do that - the one thing I'm worried about is that we land this (and temporarily drop CentOS 7 support), and then we end up stalling on a follow-up that refactors our release workflow and gets CentOS 7 working again, and having to delay a release/manually build+upload binaries - given our v3.9.2 incident, I don't really want to repeat that >.<

jsjoeio:
> given our v3.9.2 incident, I don't really want to repeat that >.<

Me neither! Sounds good to me then holding off.

jsjoeio:
@oxy  did you address this?

> just dropping this here so we remember to get the libc dependencies back down.

PR #2967, opened 2021-03-25
feat(ranger): add action for extension-request

jsjoeio:
This PR helps automate how we handle extension request issues. It also updates the docs for installing OpenVSX extensions via the command line.

TODOS
- [x] refactor docs based on feedback

### DIFF
--- a/.github/ranger.yml
+++ b/.github/ranger.yml
@@ -8,6 +8,8 @@ default:
 
     # Default comment to post when an issue is first marked with a closing label
     comment: "⚠️ This issue has been marked $LABEL and will be closed in $DELAY."
+  close-faster:
+    delay: "1 minute"
 
 labels:
   duplicate: close
@@ -23,3 +25,15 @@ labels:
     action: comment
     delay: 5s
     message: "Thanks for making your first contribution! :slightly_smiling_face:"
+  extension-request:
+    action: close-faster
+    delay: 5s
+    message: >
+      Thanks for opening an extension request!
+      We are currently in the process of switching extension
+      marketplaces and transitioning over to [Open VSX](https://open-vsx.org/).
+      Once https://github.com/eclipse/openvsx/issues/249 is implemented, we
+      can fully make this transition. Therefore, we are no longer accepting
+      new requests for extension requests. We suggest installing the VSIX
+      file and then installing into code-server as a temporary workaround.
+      See [docs](https://github.com/cdr/code-server/blob/main/docs/FAQ.md#installing-vsix-extensions-via-the-command-line) for more info."
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -1,36 +1,39 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
 # FAQ
 
-- [Questions?](#questions)
-- [iPad Status?](#ipad-status)
-- [Community Projects (awesome-code-server)](#community-projects-awesome-code-server)
-- [How can I reuse my VS Code configuration?](#how-can-i-reuse-my-vs-code-configuration)
-- [Differences compared to VS Code?](#differences-compared-to-vs-code)
-- [How can I request a missing extension?](#how-can-i-request-a-missing-extension)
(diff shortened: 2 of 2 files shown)


.github/ranger.yml
```diff
@@ -23,3 +25,7 @@ labels:
     action: comment
     delay: 5s
     message: "Thanks for making your first contribution! :slightly_smiling_face:"
+  extension-request:
+    action: close-faster
+    delay: 5s
+    message: "Thanks for opening an extension request! We are currently in the process of switching extension marketplaces and transitioning over to [Open VSX](https://open-vsx.org/). Once https://github.com/eclipse/openvsx/issues/249 is implemented, we can fully make this transition. Therefore, we are no longer accepting new requests for extension requests. We suggest installing the VSIX file and then installing into code-server as a temporary workaround. See [docs](https://github.com/cdr/code-server/blob/main/docs/FAQ.md#installing-vsix-extensions-via-the-command-line) for more info."
```
jsjoeio, line 31: Is there a way to wrap this? And does the link look right?

jawnsy: Yes! You can do long lines in YAML using `|` or `>` https://en.wikipedia.org/wiki/YAML#Basic_components

Try:

```yaml
extension-request:
  message: >
    Thanks for opening an extension request!
    We are currently in the process of
```

This is if you want it to be a single paragraph, you can also use `|` to preserve the newlines if you want (which of course might not make a difference if this message is Markdown)

jsjoeio: TIL!
You're the man! Thanks @jawnsy

docs/FAQ.md
```diff
@@ -132,6 +133,21 @@ code-server --install-extension ms-python.python
 code-server --install-extension downloaded-ms-python.python.vsix
 ```
 
+### Installing VSIX Extensions via the Command Line
```
code-asher, line 136: I think this overlaps a little with the previous paragraph that lists how to install extensions on the CLI.

What if we move that into this heading, then create another heading for this content with something like "Installing OpenVSX extensions via the command line" for this content?

We could also mention they can switch early to OpenVSX by setting `SERVICE_URL` and `ITEM_URL` (we already have a section so maybe we just link it from here).

code-asher: Actually maybe we should put this under "How can I request a missing extension?"

jsjoeio: Agreed.

Yes! Let's do that!

code-asher: Or just remove that section.

jsjoeio: I opted to add it under "How can I request a missing extension" - I say we move remove the extension request section once we fully switch to OpenVSX

code-asher: My thinking is that the flow from a user might look like:

"How do I install an extension?" 
"Oh, it's not available, how do I request it?"
"Ah, they aren't accepting requests, how do I install it then?"
"OK, I can either do it manually or download from Open VSX."

jsjoeio: Ahh yeah good point about that. Okay I'll make these changes then re-request a review! Thanks for adding the structure/flow too! That helps a ton!!

jsjoeio: Embarrassingly, I didn't know you could do it this way, hence the `curl` suggestion:
```
SERVICE_URL=https://open-vsx.org/vscode/gallery ITEM_URL=https://open-vsx.org/vscode/item code-server --install-extension stylelint.vscode-stylelint
```

docs/FAQ.md
```diff
@@ -82,15 +84,43 @@ These are the closed source extensions presently unavailable:
...
-See below for installing an extension from the cli.
+If an extension is not available from the marketplace or does not work, you can
+grab its VSIX from its GitHub releases or build it yourself.
+
+Once you have downloaded the VSIX to the remote machine you can either:
+
+- Run the `Extensions: Install from VSIX` command in the Command Palette.
+- Use `code-server --install-extension <path to vsix>`
+
+If the extension exists on another marketplace, you can also download it from there using the same command. For instance, downloading off OpenVSX can be done like this:
```
jsjoeio, line 119:
```suggestion
You can also download extensions from the command line. For instance, downloading off OpenVSX can be done like this:
```

jsjoeio:
bump @code-asher ready for another review

code-asher (approved):

PR #2953, opened 2021-03-23
release: 3.9.2

jsjoeio:
This PR is to generate a new release of `code-server` at `3.9.2`

## Screenshot

https://user-images.githubusercontent.com/3806031/112514732-d8427200-8d52-11eb-8f25-39f0d389174e.mov

## TODOs

- [x] test locally
- [x] update release notes
- [x] upload assets to draft release
- [x] double-check github release tag is the commit with artifacts
- [x] open issue for macOS CI issue #2970
- [x] build macOS package locally and manually upload
- [x] test packages
- [x] update release template (based on asher's suggestions) (separate PR) #2969
- [x] update `lib.sh` (separate PR) #2968
- [x] publish release
- [ ] merge PR
- [ ] update the homebrew package
- [ ] update the AUR package

PR #2968, opened 2021-03-25
refactor: add check for artifacts url in lib.sh

jsjoeio:
This PR refactors `lib.sh` to catch and report an error getting the get_artifacts_url.

It was noticed while working on a release in #2953.

### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -57,7 +57,17 @@ curl() {
 # This will contain the artifacts we want.
 # https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs
 get_artifacts_url() {
-  curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=push' | jq -r ".workflow_runs[] | select(.head_sha == \"$(git rev-parse HEAD)\") | .artifacts_url" | head -n 1
+  local head_sha
+  local artifacts_url
+  head_sha=$(git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


ci/lib.sh
```diff
@@ -57,7 +57,17 @@ curl() {
...
 # https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs
 get_artifacts_url() {
-  curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=push' | jq -r ".workflow_runs[] | select(.head_sha == \"$(git rev-parse HEAD)\") | .artifacts_url" | head -n 1
+  local head_sha
+  local artifacts_url
+  head_sha=$(git rev-parse HEAD)
+  artifacts_url=$(curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request' | jq -r ".workflow_runs[] | select(.head_sha == \"$head_sha\") | .artifacts_url" | head -n 1)
+  if [[ -z "$artifacts_url" ]]; then
+    echo "ERROR: artifacts_url came back empty"
+    echo "Check the get_artifacts_url function"
```
code-asher, line 66: I think this will not show up because the stdout of this function is used directly in the curl command. We could log it to stderr.

```suggestion
    >&2 echo "ERROR: artifacts_url came back empty"
    >&2 echo "Check the get_artifacts_url function"
```

code-asher (changes requested):

ci/lib.sh
jsjoeio: Nice catch! TIL!

code-asher (approved):

PR #2981, opened 2021-03-26
refactor: change get_artifacts_url in lib.sh

jsjoeio:
This PR refactors get_artifacts_url in lib.sh to work of the version branch instead of the head sha.

### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -57,13 +57,16 @@ curl() {
 # This will contain the artifacts we want.
 # https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs
 get_artifacts_url() {
-  local head_sha
   local artifacts_url
-  head_sha=$(git rev-parse HEAD)
-  artifacts_url=$(curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request' | jq -r ".workflow_runs[] | select(.head_sha == \"$head_sha\") | .artifacts_url" | head -n 1)
+  local workflow_runs_url="https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request"
+  # For releases, we look for run based on the branch name v$code_server_version
+  # example: v3.9.2
+  local version_branch="v$VERSION"
+  artifacts_url=$(curl -fsSL "$workflow_runs_url" | jq -r ".workflow_runs[] | select(.head_branch == \"$version_branch\") | .artifacts_url" | head -n 1)
   if [[ -z "$artifacts_url" ]]; then
     echo >&2 "ERROR: artifacts_url came back empty"
-    echo >&2 "Check the get_artifacts_url function"
+    echo >&2 "We looked for a successful run triggered by a pull_request with for code-server version: $code_server_version and a branch named $version_branch"
+    echo >&2 "URL used for curl call: $workflow_runs_url"
     exit 1
   fi
 


ci/lib.sh
```diff
@@ -57,13 +57,20 @@ curl() {
...
 get_artifacts_url() {
-  local head_sha
   local artifacts_url
-  head_sha=$(git rev-parse HEAD)
-  artifacts_url=$(curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request' | jq -r ".workflow_runs[] | select(.head_sha == \"$head_sha\") | .artifacts_url" | head -n 1)
+  local code_server_version
+  local version_branch
+  local workflow_runs_url
+  workflow_runs_url="https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request"
+  code_server_version=$(node -pe "require('./package.json').version")
```
code-asher, line 65: Looks like we can use `$VERSION` (set below).

ci/lib.sh
```diff
@@ -57,13 +57,20 @@ curl() {
...
 # https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs
 get_artifacts_url() {
-  local head_sha
   local artifacts_url
-  head_sha=$(git rev-parse HEAD)
-  artifacts_url=$(curl -fsSL 'https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request' | jq -r ".workflow_runs[] | select(.head_sha == \"$head_sha\") | .artifacts_url" | head -n 1)
+  local code_server_version
+  local version_branch
+  local workflow_runs_url
+  workflow_runs_url="https://api.github.com/repos/cdr/code-server/actions/workflows/ci.yaml/runs?status=success&event=pull_request"
```
code-asher, line 64: Completely a nitpick, but this could be one line (the separate declaration is to avoid masking return values which in this case there are none since it's just a string.

ci/lib.sh
```diff
@@ -57,13 +57,20 @@ curl() {
...
+  # For releases, we look for run based on the branch name v$code_server_version
+  # example: v3.9.2
+  version_branch="v$code_server_version"
+  artifacts_url=$(curl -fsSL "$workflow_runs_url" | jq -r ".workflow_runs[] | select(.head_branch == \"$version_branch\") | .artifacts_url" | head -n 1)
   if [[ -z "$artifacts_url" ]]; then
     echo >&2 "ERROR: artifacts_url came back empty"
-    echo >&2 "Check the get_artifacts_url function"
+    echo >&2 "We looked for a successful run triggered by a pull_request with for code-server version: $code_server_version and a branch named $version_branch"
+    echo >&2 "URL used for curl call: $workflow_runs_url"
     exit 1
```
code-asher, line 74: This is weird, apparently this doesn't actually exit the script due to the way we call this function (in a subshell with `$()`). Not a big deal since `curl` will fail later anyway and exit the script though.

code-asher (approved):
Beautiful.

ci/lib.sh
jsjoeio: LOL 🤦‍♂️ didn't think to scroll down in my editor 😂 Will change

ci/lib.sh
jsjoeio: TIL! Yeah shellcheck (editor extension) yelled at me before. I'll change those too

ci/lib.sh
jsjoeio: 😮 Makes sense! TIL. I'll look into it and if I can't do a quick fix, I'll leave for now

jsjoeio: Yeah doesn't look easy to fix, and I don't want to switch to Python or Perl 😂
https://unix.stackexchange.com/a/48537/363304

Leaving for now

code-asher: Agreed! I experimented and it looks like if we assign it to a variable instead of using it directly in the `curl` command it might work. But definitely something we can do later.

Issue #2975, opened 2021-03-26
New 3.9.2 not work on iPadOS


springhack:
## OS/Web Information

- Web Browser: iPad Safari
- Local OS: iPadOS 14.4.1
- Remote OS: Ubuntu Debian GNU/Linux 9.13 (stretch)
- Remote Architecture: amd64
- `code-server --version`: 3.9.2 109d2ce3247869eaeab67aa7e5423503ec9eb859

## Steps to Reproduce

1. Open safari, navigate to server url

## Expected

1. Show UI after login

## Actual

1. Just blank after login

## Notes

Seems cause by `permessage-deflate` extension, maybe add a option via command line is better?

```javascript
[Log] [Service Worker] registered (register.js, line 14)
[Error] WebSocket connection to 'wss://ide.dosk.win/?reconnectionToken=4ab75c9c-1248-4a08-80ec-491ff6d51b5e&reconnection=false&skipWebSocketFrames=false' failed: Received unexpected extension: permessage-deflate
[Log]   ERR – "[remote-connection][Management   ][4ab75…][initial][ide.dosk.win:80] socketFactory.connect() failed or timed out. Error:" (workbench.web.api.js, line 626)
[Log]   ERR (workbench.web.api.js, line 626)
Error: WebSocket close with status code 1006
[Log]   ERR – "[remote-connection] An error occurred in the very first connect attempt, it will be treated as a permanent error! Error:" (workbench.web.api.js, line 626)
[Log]   ERR – Error: WebSocket close with status code 1006 (workbench.web.api.js, line 626)
Error: WebSocket close with status code 1006
[Log]   ERR – Error: WebSocket close with status code 1006 (workbench.web.api.js, line 626)
Error: WebSocket close with status code 1006
[Log]  WARN – "Ignoring the error while validating workspace folder vscode-remote://ide.dosk.win/data00/home/chenfuchun/Public/Codes/native/aha …" (workbench.web.api.js, line 626)
"Ignoring the error while validating workspace folder vscode-remote://ide.dosk.win/data00/home/chenfuchun/Public/Codes/native/aha - WebSocket close with status code 1006"
[Warning] The web worker extension host is started without an iframe sandbox! (workbench.web.api.js, line 1713)
[Log]   ERR – "WebSocket close with status code 1006: M@https://ide.dosk.win/static/109d2ce3247869eaeab67aa7e5423503ec9eb859/usr/lib/code-server/lib/vscode…" (workbench.web.api.js, line 626)
[...]

PR #2993, opened 2021-03-29
Don't send permessage-deflate header if not supported

code-asher:
I have no idea if this fixes https://github.com/cdr/code-server/issues/2975 since I don't have a way to test.

TODOS 
- [ ] create follow-up issue to enable e2d testing on Safari to catch this in the future (@jsjoeio)

### DIFF
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -210,7 +210,9 @@ wsRouter.ws("/", ensureAuthenticated, async (req) => {
   // TODO: Parse this header properly.
   const extensions = req.headers["sec-websocket-extensions"]
   const permessageDeflate = extensions ? extensions.includes("permessage-deflate") : false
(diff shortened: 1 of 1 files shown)


code-asher:
The `sed` examples in that issue set `permessageDeflate` to false and comment out the response header.

However, I'm not sure why `permessageDeflate` is set to false. Is Safari reporting that it supports `permessage-deflate` when it doesn't really?

jsjoeio:
Seeing if I can help out by testing locally on my iPad:

1. `gh pr checkout 2993`
2. `yarn && yarn watch --link` - run with code-server link to be able to access from iPad
3. open in Safari on iPad

jsjoeio:
## Before this change

Reproducing on `main` i.e. `3.9.2`
todo - add video

## After this change

Pulled down this branch and ran fine without issues


https://user-images.githubusercontent.com/3806031/112877827-15bd3d00-907c-11eb-8fb1-d8c826d0dfab.MP4

jsjoeio (approved):

PR #2999, opened 2021-03-29
fix(dev-deps): update ts-node and spawn

jsjoeio:
This PR fixes how we spawn a child_process and bumps `ts-node`.

(H/T to @cspotcode for identifying this after we opened an issue in `ts-node`)

Related:
- https://github.com/TypeStrong/ts-node/issues/1283#issuecomment-807704228
- Reverts https://github.com/cdr/code-server/pull/2962
- https://github.com/cdr/code-server/pull/2965

image

### DIFF
--- a/src/node/wrapper.ts
+++ b/src/node/wrapper.ts
@@ -314,7 +314,7 @@ export class ParentProcess extends Process {
         CODE_SERVER_PARENT_PID: process.pid.toString(),
         NODE_OPTIONS: `--max-old-space-size=2048 ${process.env.NODE_OPTIONS || ""}`,
       },
-      stdio: ["ipc"],
+      stdio: ["inherit", "pipe", "pipe", "ipc"],
     })
   }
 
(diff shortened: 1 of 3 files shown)


src/node/wrapper.ts
```diff
@@ -314,7 +314,7 @@ export class ParentProcess extends Process {
         CODE_SERVER_PARENT_PID: process.pid.toString(),
         NODE_OPTIONS: `--max-old-space-size=2048 ${process.env.NODE_OPTIONS || ""}`,
       },
-      stdio: ["ipc"],
+      stdio: ["inherit", "pipe", "pipe", "ipc"],
```
code-asher, line 317: I'm thinking we could use `inherit` so we can get rid of the manual `pipe` calls below.

```suggestion
      stdio: ["inherit", "inherit", "inherit", "ipc"],
```

code-asher (changes requested):
Thank you for taking care of this!

src/node/wrapper.ts
code-asher: Uh, so, this is embarrassing...I was looking at the wrong file. We could still use inherit but we'd need to remove the `pipe` calls to `process.stdout` and `process.stderr` and pipe them to the logger instead of the child. 

I say we just leave these at `pipe` for now haha. Sorry for the noise. :sob:

jsjoeio: LOL no worries at all! Glad you caught it before we merged. I'm less familiar with all this `pipe` stuff so kinda following your lead. Okay will do!

jsjoeio:
bump @code-asher - ready for a second review

code-asher (approved):

PR #3006, opened 2021-03-30
chore(dev-deps): add ignore block to dependabot config for @types/node

jsjoeio:
This PR tells dependabot to ignore version updates for @types/node 14.x. 

Related:
- https://github.com/cdr/code-server/pull/2989#issuecomment-809551319

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,6 @@ updates:
     schedule:
       interval: "daily"
       time: "11:00"
+    ignore:
+      - dependency-name: "@types/node"
+        versions: ["14.x"]


jawnsy (approved):
LGTM

.github/dependabot.yml
```diff
@@ -15,3 +15,6 @@ updates:
     schedule:
       interval: "daily"
       time: "11:00"
+    ignore:
+      - dependency-name: "@types/node"
```
jawnsy, line 19: According to the docs, dependabot is supposed to only update the lockfile by default (so not modify the major versions), but it might have something to do with whether it considers our project an application or a library: https://docs.github.com/en/code-security/supply-chain-security/configuration-options-for-dependency-updates#versioning-strategy

This seems like a decent solution, but if you want something more generic, you can also use `versioning-strategy: lockfile-only`, which is supposedly supposed to prevent dependabot from modifying packages.json (but then we have the same problem where we have to keep an eye on the latest available versions)

jsjoeio: Ah, good to know. Thanks for looking into that and adding those notes.

I think it makes sense for it to increase version like it's been doing since that's probably what we want, right? So I vote to leave as is? Thoughts?

jawnsy: Yeah, it's good as-is :+1:

PR #3012, opened 2021-03-31
chore(dev-deps): bump @types/node to 12.20.7

jsjoeio:
This PR bumps `@types/node` to `12.20.7`

Tested locally using `node 12.20.2` and everything works as expected.

Related:
- https://github.com/cdr/code-server/pull/3011

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@types/express": "^4.17.8",
     "@types/http-proxy": "^1.17.4",
     "@types/js-yaml": "^4.0.0",
-    "@types/node": "^12.12.7",
+    "@types/node": "~12.20.7",
     "@types/parcel-bundler": "^1.12.1",
     "@types/pem": "^1.9.5",
     "@types/proxy-from-env": "^1.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "@types/express": "^4.17.8",
     "@types/http-proxy": "^1.17.4",
     "@types/js-yaml": "^4.0.0",
-    "@types/node": "^12.12.7",
+    "@types/node": "~12.20.7",
```
oxy, line 44: Not sure if `~` is intentional here to keep us on the 12.20.x series instead of the 12.x series (which would be behavior for `^`) - @jsjoeio

jsjoeio: It was made intentionally! I was thinking we should stay on `12.20` but open to changing it if you think that's the wrong move!

oxy: I don't know enough about `@types/node` to know for sure, I think I'd defer to your judgements, though feel free to pull in asher for another opinion!

oxy (approved):

package.json
jsjoeio: I think it's okay then. I think we make it strict like this with `~` and if that's the wrong move, then we reverse it. Easily reversible!

code-asher: Since we support a range of Node versions I'm not really sure what versions of the types we should use.

I suppose we should keep it in sync with whatever we use to build in CI? So if we use Node 12.20.x then we use 12.20.x types, etc.

jsjoeio: I'm in agreement! Sounds good to me!

Issue #2963, opened 2021-03-25
[dev]: add e2e test for dev workflow (yarn watch)


jsjoeio:
After hours of debugging and a false positive with a dependabot PR, we've come to the conclusion that we should have testing in place for our dev workflow: `yarn watch`. 

This will prevent issues like https://github.com/cdr/code-server/pull/2962 from happening again.

One way to achieve this is by:
- adding a new script called `yarn test:watch` which is similar to `yarn test:e2e`
- add new CI step/job for running the e2e tests against code-server running with `yarn watch`

We'd have to add some timeout or sleep though since running with `yarn watch` takes time to finish.

Related:
- https://github.com/cdr/code-server/pull/2962
- https://github.com/TypeStrong/ts-node/issues/1283

PR #3015, opened 2021-04-01
Fix tsc watch continually restarting during vscode build

code-asher:
See the commits for more information.

Note that I only moved ipc.d.ts and added a comment to the top but it doesn't show as a rename because the indentation changed from tabs to spaces and git thinks it's a different file.


jsjoeio (approved):
Nice!! This will make local development much more enjoyable. 

Non-blocking: wish we had an e2e test to make sure this doesn't cause any regressions (like the ts-node stuff): https://github.com/cdr/code-server/issues/2963

Issue #2998, opened 2021-03-29
[dev]: enable end-to-end test suite on Safari and Firefox


jsjoeio:
Currently, we only run end-to-end tests on Chrome. However, [playwright](https://playwright.dev/) supports all modern browsers:

> Single API to automate Chromium, Firefox and WebKit.

After shipping a feature that causes code-server to not work in Safari, we think it's time to enable this. 

Related: 
- https://github.com/cdr/code-server/issues/2975
- https://github.com/microsoft/playwright/discussions/5991
- waiting on next release of `playwright` which should include a fix for: https://github.com/microsoft/playwright/issues/6009

PR #3016, opened 2021-04-01
dev(testing): add jest-playwright and reduce flakiness of e2e tests

jsjoeio:
This PR adds `jest-playwright` and refactors the e2e tests to hopefully reduce the flakiness.

## Changes
- add `jest-playwright` in preparation of #2998 
  - will add support for Firefox after https://github.com/microsoft/playwright/issues/6020 is fixed
- refactors `globalSetup` to manually create cookie for auth reuse
- refactors other tests based on `jest-playwright`
- removes old code
- modifies tests to run with `--runInBand` (opposite of in parallel)
- fixes flakiness

## Screenshots
image

image

## Checklist
- [x] tested locally
- [x] added new dependencies
- [x] added a test
- [x] fix flaky `logout` test
- [x] remove unused code `maybeUpdatedCookies`
- [x] get tests passing locally

Related:
- lays the foundation for #2998


ci/steps/test-e2e.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 main() {
   cd "$(dirname "$0")/../.."
 
-  "./release-packages/code-server*-linux-amd64/bin/code-server" --home "$CODE_SERVER_ADDRESS"/healthz &
+  "./release-packages/code-server*-linux-amd64/bin/code-server" &
```
jsjoeio, line 7: Should have been removed before.

ci/dev/test-e2e.sh
```diff
@@ -15,7 +15,7 @@ main() {
     echo -e "\n"
     exit 1
   fi
-  CS_DISABLE_PLUGINS=true ./test/node_modules/.bin/jest "$@" --config ./test/jest.e2e.config.ts
+  CS_DISABLE_PLUGINS=true ./test/node_modules/.bin/jest "$@" --config ./test/jest.e2e.config.ts --runInBand
```
jsjoeio, line 18: > Run all tests serially in the current process, rather than creating a worker pool of child processes that run tests

[Jest Docs](https://jestjs.io/docs/cli#--runinband)

This reduces flakiness of tests.

.eslintrc.yaml
```diff
@@ -16,6 +16,9 @@ extends:
   - plugin:import/typescript
   - plugin:prettier/recommended
   - prettier # Removes eslint rules that conflict with prettier.
+  # Recommended by jest-playwright
+  # https://github.com/playwright-community/jest-playwright#globals
+  - plugin:jest-playwright/recommended
```
oxy, line 21: I'm not sure if this has to go at the end - because usually Prettier recommends their ruleset to be the last one, afaik

ci/dev/test-e2e.sh
oxy: Surprised that running it out of band is the default! Multiple instances of `code-server` logging in and out at the same time doesn't sound fun in retrospect.

ci/steps/test-e2e.sh
oxy: Good catch!

test/package.json
```diff
@@ -12,5 +12,11 @@
     "playwright": "^1.8.0",
     "supertest": "^6.1.1",
     "ts-jest": "^26.4.4"
+  },
+  "dependencies": {
+    "jest-playwright-preset": "^1.5.1",
+    "playwright-chromium": "^1.10.0",
+    "playwright-firefox": "^1.10.0",
+    "playwright-webkit": "^1.10.0"
```
oxy, line 20: I think it may make sense to move the others to devDependencies too

oxy (changes requested):
I think we need to move the `playwright-*` packages to devDependencies, and am not sure about `.eslintrc` order but looks good otherwise!

.eslintrc.yaml
jsjoeio: Ah, got it. I'll reorder these!

test/package.json
jsjoeio: Yes! Good catch

oxy (approved):
Looks good!

dgozman:
Hi @jsjoeio! Would you consider trying [`playwright-test`](https://github.com/microsoft/playwright-test) instead of `jest-playwright`? We at Playwright team believe that with the focused test runner it is possible to deliver better experience for e2e tests. If you are open to this, I can guide/help with the migration, as well as address the feedback on the test runner side 😄  Let me know if you are interested!

jsjoeio:
Hey @dgozman! Thanks for getting in touch. Yes, we would, esp. if the Playwright team recommends it :) 

I've opened a new issue for us to discuss + to track us making the switch: #3115

PR #3041, opened 2021-04-05
fix(deps): update and fix vulnerabilities

jsjoeio:
This PR fixes a few vulnerabilities in code-server.

H/T to @PatrickDerichs for pointing them out 🎉

## Changes
- update `yarn.lock` using `yarn upgrade parcel-bundler` and `yarn upgrade proxy-agent`

## Screenshots

image

image

## Checklist
- [x] tested locally
- [x] updated `yarn.lock`

Fixes #2964

Issue #2964, opened 2021-03-25
Vulnerabilitiies in code-server


PatrickDerichs:
## OS/Web Information

- Web Browser: Chrome
- Local OS: Ubuntu 20.04
- Remote OS: Ubuntu 20.04
- Remote Architecture: 64-bit
- `code-server --version`:

## Steps to Reproduce

1. Build an image with the latest code-server
2. Run Trivy with said image

## Expected

No vulnerabilities found

## Actual

Trivy finds the following vulnerabilities:

```
 usr/lib/code-server/lib/vscode/yarn.lock
 ========================================
 Total: 3 (HIGH: 3, CRITICAL: 0)
 +---------+------------------+----------+-------------------+---------------+---------------------------------------+
 | LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
 +---------+------------------+----------+-------------------+---------------+---------------------------------------+
 | is-svg  | CVE-2021-28092   | HIGH     | 3.0.0             | 4.2.2         | nodejs-is-svg: ReDoS                  |
 |         |                  |          |                   |               | via malicious string                  |
 |         |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-28092 |
 +---------+------------------+          +-------------------+---------------+---------------------------------------+
 | ssri    | CVE-2021-27290   |          | 6.0.1             | 8.0.1         | nodejs-ssri: Regular                  |
 |         |                  |          |                   |               | expression DoS when parsing           |
 |         |                  |          |                   |               | malicious SRI in strict mode          |
 |         |                  |          |                   |               | -->avd.aquasec.com/nvd/cve-2021-27290 |
 +         +                  +          +-------------------+               +                                       +
 |         |                  |          | 8.0.0             |               |                                       |
 |         |                  |          |                   |               |                                       |
 |         |                  |          |                   |               |                                       |
 |         |                  |          |                   |               |                                       |
 +---------+------------------+----------+-------------------+---------------+---------------------------------------+
 usr/lib/code-server/yarn.lock
 =============================
 Total: 1 (HIGH: 1, CRITICAL: 0)
 +---------+------------------+----------+-------------------+---------------+---------------------------------------+
 | LIBRARY | VULNERABILITY ID | SEVERITY | INSTALLED VERSION | FIXED VERSION |                 TITLE                 |
 +---------+------------------+----------+-------------------+---------------+---------------------------------------+
[...]

PR #3042, opened 2021-04-05
feat(ci): add audit job for security

jsjoeio:
This PR adds a new CI job using [`audit-ci`](https://github.com/IBM/audit-ci) to check for vulnerabilities using `yarn audit`.

Inspired by @PatrickDerichs and #2964 

NOTE: I don't think `audit` will pass until https://github.com/cdr/code-server/pull/3041 is merged and I rebase. 

TODOS
- [x] refactor based on jawnsy's feedback (add to `package.json`, add as `devDependency` and add script to run in CI)


.github/workflows/ci.yaml
```diff
@@ -26,6 +26,14 @@ jobs:
...
         with:
           args: ./ci/steps/lint.sh
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Audit for vulnerabilities
+        # Prevents integration with moderate or higher vulnerabilities
+        # Docs: https://github.com/IBM/audit-ci#options
+        run: npx audit-ci --moderate
```
jawnsy, line 36: is it possible to put this in a config file, so that we can run `npx audit-ci` locally to reproduce results?

also, is it useful to add audit-ci to the packages.json so that it's version-controlled?

jawnsy (approved):
LGTM! :+1:

I'm curious generally where this tool gets its data from, and why Dependabot seems to have missed some of the flagged issues?

I think we'll need several of these tools because they scan different things and in different ways (this covers us for vulns in code-server itself, but not for other stuff that we include in our code-server docker images, for example)

.github/workflows/ci.yaml
jsjoeio: Good call! I could add a new script to the root `package.json` called `audit` which runs `npx audit-ci`? 

Or we could install as a `devDependency`? Either works. 

> useful to add audit-ci to the packages.json so that it's version-controlled

We could do that! I'd probably need to add a script then and update the CI job to install dependencies and then run `audit-ci --moderate`.

jawnsy: I dunno... go with whatever's easy :smile: I think devDependency makes sense, it's the traditional way to add dev tools right? and they would be found by `yarn run` or whatever (I am sure my JavaScript ignorance is showing here :flushed:)

I think @vapurrmaid did that for cdr/docs for stuff like prettier and markdownlint, and then dependabot handles updating those too (in a way that will notify us when there's updates/won't risk things suddenly breaking if there's an accidental breaking change upstream) https://github.com/cdr/docs/blob/master/package.json

jsjoeio: Yeah, that seems like the best approach to take. Plus, dependabot can update it too (whereas it can't when we use `npx`). Sounds like a plan! I'll make these changes tomorrow.

vapurrmaid: Sounds good!

PR #3065, opened 2021-04-07
fix: update ranger file

jsjoeio:
This PR fixes a couple syntax issues in the `ranger.yml` configuration and adjusts the close time to 2 days instead of 3.

I thought we could define our own custom actions but I don't think that's the case after re-reading [the docs](https://www.notion.so/Documentation-8d7627bb1f3c42b7b1820e8d6f157a57) and asking the maintainer why it wasn't working 😂.

### DIFF
--- a/.github/ranger.yml
+++ b/.github/ranger.yml
@@ -4,12 +4,10 @@ default:
   close:
     # Default time to wait before closing the label. Can either be a number in milliseconds
     # or a string specified by the `ms` package (https://www.npmjs.com/package/ms)
-    delay: "3 days"
+    delay: "2 days"
 
     # Default comment to post when an issue is first marked with a closing label
     comment: "⚠️ This issue has been marked $LABEL and will be closed in $DELAY."
-  close-faster:
-    delay: "1 minute"
 
 labels:
   duplicate: close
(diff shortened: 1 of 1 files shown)


.github/ranger.yml
```diff
@@ -26,7 +24,7 @@ labels:
     delay: 5s
     message: "Thanks for making your first contribution! :slightly_smiling_face:"
   extension-request:
-    action: close-faster
+    action: close
```
jawnsy, line 27: @jsjoeio Ah that's too bad that there's no schema validator for these, I would expect some sort of warning if fields are unknown

jawnsy (approved):

.github/ranger.yml
jsjoeio: Right? Well based on what I read and saw in the config, I just assumed I could create my own custom actions like this: https://github.com/cdr/code-server/pull/3065/files#diff-15592ca299f8c76c553cc34e528b303896afa722a32d8f983cdcf67dfa7b38f1L11-L12

But turns out that is not the case 😂

oxy (approved):
Looks good!

PR #3076, opened 2021-04-08
fix: use curl to install nfpm

jsjoeio:
This PR uses `curl` to install `nfpm` instead of Go. 

See failed publish workflow: https://github.com/cdr/code-server/runs/2300619746?check_suite_focus=true#step:3:1110

NOTE: follow-up PR.
- [ ] remove Go
- [ ] use @jawnsy's suggestion

### DIFF
--- a/ci/images/debian10/Dockerfile
+++ b/ci/images/debian10/Dockerfile
@@ -37,9 +37,8 @@ ENV GOPATH=/gopath
 RUN mkdir -p $GOPATH && chmod -R 777 $GOPATH
 ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
 
-# Install Go dependencies
-ENV GO111MODULE=on
-RUN go get github.com/goreleaser/nfpm/cmd/nfpm@v2.3.1
+# More stable than go get
+RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/nfpm.sh | sh
 
 RUN VERSION="$(curl -fsSL https://storage.googleapis.com/kubernetes-release/release/stable.txt)" && \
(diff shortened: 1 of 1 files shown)


code-asher (approved):

ci/images/debian10/Dockerfile
```diff
@@ -37,9 +37,8 @@ ENV GOPATH=/gopath
 RUN mkdir -p $GOPATH && chmod -R 777 $GOPATH
 ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
 
-# Install Go dependencies
-ENV GO111MODULE=on
-RUN go get github.com/goreleaser/nfpm/cmd/nfpm@v2.3.1
+# More stable than go get
+RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/nfpm.sh | sh
```
jawnsy, line 41: I think this means you can also skip the previous step (i.e. delete lines 32-38) because we no longer depend on Go? :tada: 

I think curl'ing from the GitHub release URL is better, otherwise we depend on both install.goreleaser.com as well as GitHub, and if either service has problems, then we'll get a failure again

So I'm proposing to do something like this instead:

```
RUN curl -sSL https://github.com/goreleaser/nfpm/releases/download/v2.3.1/nfpm_2.3.1_Linux_x86_64.tar.gz | tar -C /tmp -zxv nfpm
```

This command puts nfpm into /tmp, but we can change it to install to /usr/local or whatever insetad

jawnsy (approved):

ci/images/debian10/Dockerfile
jsjoeio: YES! Thanks for double-confirming. Asher said the same thing earlier but we weren't 100% sure. 

Ahh...I like the sound of that.

I will make both of those changes in the follow-up PR. Thanks a bunch @jawnsy 🎉

PR #3077, opened 2021-04-08
fix(ci): remove Go and refactor install nfpm line

jsjoeio:
Follow-up PR from https://github.com/cdr/code-server/pull/3076#discussion_r610111979
- [x] remove Go
- [x] use @jawnsy's suggestion

### DIFF
--- a/ci/images/debian10/Dockerfile
+++ b/ci/images/debian10/Dockerfile
@@ -29,8 +29,8 @@ RUN curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.7.1/s
   && mv shellcheck*/shellcheck /usr/local/bin \
   && rm -R shellcheck*
 
-# More stable than go get
-RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/nfpm.sh | sh
+# Using GitHub release URL is better because now only rely on one service instead of GitHub and install.goreleaser.com
+RUN curl -sSL https://github.com/goreleaser/nfpm/releases/download/v2.3.1/nfpm_2.3.1_Linux_x86_64.tar.gz | tar -C /tmp -zxv nfpm
 
 RUN VERSION="$(curl -fsSL https://storage.googleapis.com/kubernetes-release/release/stable.txt)" \
   && curl -fsSL "https://storage.googleapis.com/kubernetes-release/release/$VERSION/bin/linux/amd64/kubectl" >/usr/local/bin/kubectl \
--- a/ci/images/debian10/Dockerfile
+++ b/ci/images/debian10/Dockerfile
@@ -6,13 +6,13 @@ RUN apt-get update
 RUN apt-get install -y curl gnupg
 
 # Installs node.
-RUN curl -fsSL https://deb.nodesource.com/setup_12.x | bash - && \
-    apt-get install -y nodejs
+RUN curl -fsSL https://deb.nodesource.com/setup_12.x | bash - \
+  && apt-get install -y nodejs
 
 # Installs yarn.
-RUN curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
-    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update && apt-get install -y yarn
+RUN curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+  && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+  && apt-get update && apt-get install -y yarn
 
 # Installs VS Code build deps.
 RUN apt-get install -y build-essential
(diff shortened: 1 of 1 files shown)


ci/images/debian10/Dockerfile
```diff
@@ -24,25 +24,17 @@ RUN apt-get install -y gettext-base
...
-RUN mkdir -p $GOPATH && chmod -R 777 $GOPATH
-ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
-
-# More stable than go get
-RUN curl -sfL https://install.goreleaser.com/github.com/goreleaser/nfpm.sh | sh
-
-RUN VERSION="$(curl -fsSL https://storage.googleapis.com/kubernetes-release/release/stable.txt)" && \
-    curl -fsSL "https://storage.googleapis.com/kubernetes-release/release/$VERSION/bin/linux/amd64/kubectl" > /usr/local/bin/kubectl \
-    && chmod +x /usr/local/bin/kubectl
+RUN curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.$(uname -m).tar.xz \
```
jawnsy, line 27: I think there's a Debian package for shellcheck, so you can apt-get install shellcheck here? The current stable release is a little old though: https://packages.debian.org/buster/shellcheck unless you also enable backports: https://packages.debian.org/buster-backports/shellcheck

No worries though, this is fine as it is :)

ci/images/debian10/Dockerfile
```diff
@@ -6,13 +6,13 @@ RUN apt-get update
...
 
 # Installs yarn.
-RUN curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
-    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
-    apt-get update && apt-get install -y yarn
+RUN curl -fsSL https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+  && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+  && apt-get update && apt-get install -y yarn
 
 # Installs VS Code build deps.
```
jawnsy, line 17: In general I think we should condense these RUN commands, each one creates a separate copy-on-write layer and this can lead to "wasted" space. Not something to do now, but just something to keep in mind for the future :)

When you install packages or run an update, apt creates cache files in `/var/cache/apt` and some other places (it's downloaded packages and the package list), and these consume space in the container image unless you remove them in the same layer (using `apt-get clean`). There's a [magic hook in the Debian docker image](https://github.com/debuerreotype/debuerreotype/blob/d29dd5e030525d9a5d9bd925030d1c11a163380c/scripts/debuerreotype-minimizing-config#L85-L100) that does this automatically, but it's usually better to be explicit (update, install, and clean all in the same step) rather than rely on that magic.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

https://dev.to/cloudx/analyzing-the-docker-layers-with-dive-5e7o

jawnsy (approved):

ci/images/debian10/Dockerfile
oxy: We now actually grab shellcheck from npm now, which uses a script that directly downloads/installs it to `node_modules` - see https://github.com/cdr/code-server/blob/main/package.json#L71 - I landed this in the CI rewrite to make building easier

jawnsy: nice! so we can remove the curl line?

I was confused that shellcheck in package.json is v1 while the latest release in GitHub is 0.7.1, and it seems that the npm package just downloads the latest release? This would make the build less reproducible, since shellcheck can can change (and break) without any changes on our part

code-asher: Oh huh yeah good point. I think we wanted to use the `npm` version to reduce the number of dependencies users have to manually install so maybe we can fork it, make the version static, and publish `@coder/shellcheck` or something.

jawnsy: Could we convince the shellcheck people to release versioned releases that correspond with the upstream versions? e.g. shellcheck@0.71-1 or something (upstream version 0.71, first version of the npm package) - that would be useful for other people too and means we don't have to maintain our own thing

code-asher: Maybe we can do the same for `nfpm`.

code-asher: That would definitely be better if we could convince upstream! Seems the maintainer is open to proposals: https://github.com/gunar/shellcheck/issues/24#issuecomment-684924783

code-asher: My main concern was that since 1.0.0 was already published it seems a bit sketchy.

ci/images/debian10/Dockerfile
```diff
@@ -24,25 +24,17 @@ RUN apt-get install -y gettext-base
...
-RUN VERSION="$(curl -fsSL https://storage.googleapis.com/kubernetes-release/release/stable.txt)" && \
-    curl -fsSL "https://storage.googleapis.com/kubernetes-release/release/$VERSION/bin/linux/amd64/kubectl" > /usr/local/bin/kubectl \
-    && chmod +x /usr/local/bin/kubectl
+RUN curl -fsSL https://github.com/koalaman/shellcheck/releases/download/v0.7.1/shellcheck-v0.7.1.linux.$(uname -m).tar.xz \
+  | tar -xJ \
+  && mv shellcheck*/shellcheck /usr/local/bin \
+  && rm -R shellcheck*
+
+# Using GitHub release URL is better because now only rely on one service instead of GitHub and install.goreleaser.com
+RUN curl -sSL https://github.com/goreleaser/nfpm/releases/download/v2.3.1/nfpm_2.3.1_Linux_x86_64.tar.gz | tar -C /tmp -zxv nfpm
```
code-asher, line 33: We might want to use the same architecture logic we used for fetching the `go` binary.

ci/images/debian10/Dockerfile
```diff
@@ -24,25 +24,17 @@ RUN apt-get install -y gettext-base
...
+
+RUN VERSION="$(curl -fsSL https://storage.googleapis.com/kubernetes-release/release/stable.txt)" \
+  && curl -fsSL "https://storage.googleapis.com/kubernetes-release/release/$VERSION/bin/linux/amd64/kubectl" >/usr/local/bin/kubectl \
```
code-asher, line 36: This wasn't introduced by your PR but we should use the detected architecture here as well.

code-asher (changes requested):
Now that CI has been refactored the only place this is used is in `run.sh` right? Does anyone here actually use the Docker container for development? I wonder if it even makes sense to maintain.

jsjoeio:
> Now that CI has been refactored the only place this is used is in run.sh right? Does anyone here actually use the Docker container for development? I wonder if it even makes sense to maintain.

Oh! I didn't even realize it but you're right. I don't use it for development. Not sure if others do, but I would be in favor of closing this PR and removing `ci/images/debian10/Dockerfile`. 

Not sure if `run.sh` is even used?

jsjoeio:
Closing in favor of #3134

Issue #2647, opened 2021-01-28
Login rate limiter should not count against successful logins


jsjoeio:
When running e2e tests against the login page, I'm hitting the login rate limit even though I'm using the correct password

<img width="738" alt="image" src="https://user-images.githubusercontent.com/3806031/106203322-6e558400-6178-11eb-86cc-310f60824567.png">

Relevent code:
https://github.com/cdr/code-server/commit/5accf3fe5fe8f3430cf06f95adff18c4107756f7

sr229:
this looks like the limiter couldn't distinguish from successful logins to non-successful logins. Will give this a look later.

PR #3133, opened 2021-04-14
refactor(testing): migrate to playwright-test from jest-playwright

### LINKED ISSUE #3115: [dev]: switch from jest-playwright to playwright test
The Playwright team has reached out and suggested we use `playwright-test` instead of `jest-playright` as our test runner. The APIs look similar but they offered to help/guide the migration.

> Hi @jsjoeio! Would you consider trying playwright-test instead of jest-playwright? We at Playwright team believe that with the focused test runner it is possible to deliver better experience for e2e tests. If you are open to this, I can guide/help with the migration, as well as address the feedback on the test runner side 😄 Let me know if you are interested!

Originally posted: https://github.com/cdr/code-server/pull/3016#issuecomment-816329571

Note: I'll probably start on this sometime this week and post thoughts and ask questions as I go here.

jsjoeio:
This PR refactors our test runner from [`jest-playwright`](https://github.com/playwright-community/jest-playwright) to [`playwright-test`](https://github.com/microsoft/playwright-test) at the recommendation of the Playwright Team.

Other handy features:
- use `--repeat-each` to repeat the test suite multiple times

Major changes:
- only videos for failed tests are saved
- `playwright-test` automatically retries failed tests (3 times locally, 2 times in CI)
- now runs tests in all three browsers (before Chromium. now WebKit and Firefox)

It should allow for a more flexible configuration and help us identify flaky tests faster.

## Screenshot
image

## TODOs
- [x] fix flaky tests
- [x] fix failing tests
- [x] fix logs to not say jest in globalSetup

## Notes

`logout` is flaky/failing because of the login rate limiter (#2647) :( See video:

https://user-images.githubusercontent.com/3806031/114786113-79a95c00-9d32-11eb-8338-a0cab7868a29.mp4

Fixes #3115 #2998


test/config.ts
```diff
@@ -0,0 +1,71 @@
...
+  PlaywrightOptions,
+  Config,
+  globalSetup,
+} from "@playwright/test"
+import * as crypto from "crypto"
+import path from "path"
+import { PASSWORD } from "./utils/constants"
+import * as wtfnode from "./utils/wtfnode"
+
+// Playwright doesn't like that ../src/node/util has an enum in it
```
jawnsy, line 16: :-1:  that you had to do this, but :+1:  that you added a comment... I could certainly be better about doing this myself, it's a great habit :smile:

test/config.ts
```diff
@@ -0,0 +1,71 @@
...
+})
+
+const config: Config = {
+  testDir: path.join(__dirname, "e2e"), // Search for tests in this directory.
+  timeout: 30000, // Each test is given 30 seconds.
+  retries: 3, // Retry failing tests 2 times
+}
+
+if (process.env.CI) {
+  config.retries = 2
```
jawnsy, line 58: I guess this is to handle flakiness in CI environments?

test/e2e/openHelpAbout.test.ts
```diff
@@ -1,38 +1,46 @@
...
+      // Make sure the editor actually loaded
+      expect(await page.isVisible("div.monaco-workbench"))
 
-    // Click the Application menu
-    await page.click("[aria-label='Application Menu']")
-    // See the Help button
-    const helpButton = "a.action-menu-item span[aria-label='Help']"
-    expect(await page.isVisible(helpButton))
+      // Click the Application menu
+      await page.click("[aria-label='Application Menu']")
```
jawnsy, line 27: do people still do [page objects](https://martinfowler.com/bliki/PageObject.html) or is that not cool anymore? also, is it possible to select by IDs instead of other attributes, or is this the recommended approach with Playwright?

jawnsy (approved):

test/package.json
```diff
@@ -2,19 +2,19 @@
...
-    "playwright": "^1.8.0",
-    "playwright-chromium": "^1.10.0",
-    "playwright-firefox": "^1.10.0",
-    "playwright-webkit": "^1.10.0",
+    "playwright": "^1.11.0-next-alpha-apr-13-2021",
     "supertest": "^6.1.1",
     "ts-jest": "^26.4.4"
+  },
+  "resolutions": {
+    "@playwright/test/playwright": "^1.11.0-next-alpha-apr-13-2021"
```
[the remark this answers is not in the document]
jsjoeio: Possibly! I will take responsibility if something breaks.

I had to do this because I was running into bugs that don't occur in the newest (alpha) version of Playwright, specifically this one: https://github.com/microsoft/playwright/issues/6020

I'll keep an eye on this and remove once the next playwright release comes out.

ci/dev/test-e2e.sh
```diff
@@ -3,19 +3,8 @@ set -euo pipefail
...
-    echo -e "\n"
-    echo "Please set the following environment variables locally:"
-    echo "  \$PASSWORD"
-    echo "  \$CODE_SERVER_ADDRESS"
-    echo -e "\n"
-    exit 1
-  fi
-  CS_DISABLE_PLUGINS=true ./test/node_modules/.bin/jest "$@" --config ./test/jest.e2e.config.ts --runInBand
+  cd test
+  PASSWORD=e45432jklfdsab CODE_SERVER_ADDRESS=http://localhost:8080 yarn folio --config=config.ts --reporter=list "$@"
```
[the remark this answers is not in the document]
jsjoeio: > useful

Probably not. We need to password to be set as an environment variable for the tests in order to log in. I added a comment just now to explain that

test/config.ts
jsjoeio: Yup! I thought the comment on L54 would suffice:
https://github.com/cdr/code-server/pull/3133/files/3b1802c2357c7707e487287391d16e93fb9eb756#diff-8c108cf05702abdfb2e259659b8823f7c917fced47193febd0ab7f4a39134294R54

But I'll add another

jawnsy: that comment describes _what_ the setting does but not _why_ we need it, though :) IMO comments should be about the _why_ because it gives context

test/e2e/openHelpAbout.test.ts
jsjoeio: Lol not sure about page objects. Unfortunately the way the HTML is structured, there isn't an ID to use as a selector. Otherwise, we would:
image

```html
<div class="menubar-menu-button" role="menuitem" tabindex="0" aria-label="Application Menu" title="Application Menu" aria-haspopup="true" style="visibility: visible;"><div class="menubar-menu-title toolbar-toggle-more codicon codicon-menubar-more" role="none" aria-hidden="true"></div></div>
```

And this HTML comes from VS Code. We could add our own IDs, but there's a chance they get overwritten during a git subtree update to `lib/vscode`. 

This means that if the VS Code Team changes the HTML (i.e. removes the aria-label) then our tests break. This can be annoying for us, but it's the best approach. Plus, aria-labels are less likely to change as frequently as something like a class so at least this attribute is a bit more stable.

test/config.ts
jsjoeio: Oops...I guess I thought the flakiness covered the _why_ but maybe not. Working on that 😅

Issue #2818, opened 2021-03-04
☂️ [Roadmap]: Clean Up


jsjoeio:
This is an umbrella issue to related to one of our priorities for code-server: clean up.

They're chores. Think of this as cleaning up the repository and making things easier both for maintainers and contributors.

### 1. Easier for Maintainers

code-server is a big project. The easier to maintain, the faster we can fix bugs, ship features and help the community.

- ✅ separate unit and e2e testing #2646 
- ⏳ improve updating vscode process
  - ⏳ improve existing script #2787 
  - figure out a way to pull in their changes on a regular cadence (so updating is easier)
- ⏳ pre-install script that checks dependencies #2807 
- 🚧 automate release process #2817 
  - ✅ automate first step of process #2952 
  - TODO automate attaching artifacts to release
- ✅ add `homebrew` job to `publish.yaml` #2841
- ⏳ add `aur` job to `publish.yaml` #2845 
- TODO investigate dropping `nfpm` see [comment](https://github.com/cdr/code-server/pull/2891#discussion_r594724298)
- TODO clean up images [see comment](https://github.com/cdr/code-server/pull/2891#discussion_r594731273)
  > particularly in light of the fact that CentOS changed their model recently with CentOS Stream
- ⏳ split up github workflows #2903 
- ⏳ try changing `esModuleInterop` to false

### 2. Easier for Contributors 

If we want to encourage contributions from the community (and even Coder employees), we need to make it as simple as possible to contribute. 

- ✅ Remove `shfmt` as a dependency #2772 
- ⏳  "...maybe we could make an image to do code-server development inside code-server, and also a variant for code-server development inside coder." —@jawnsy https://github.com/cdr/code-server/pull/3134#pullrequestreview-636157886

### 3. Remove Dead Code

We need to do an analysis to see how much unused code exists in the code-server codebase. This could slim down the size, maybe help performance but also help with maintenance long-term. 

- ✅ remove pages/updates.css #2823 
- TODO add more tasks here

### 4. Improve Security Practices

We want to make sure we're staying on top of security vulnerabilities in code-server by following best practices.

-⏳ utilize GitHub security tools #2839 

<br>
<br>

```
LEGEND

⏳ = to do
🚧 = in progress
✅ = done
```
_Note: this is part of an experiment to be more transparent about our priorities and our roadmap. See everything related to this effort in the [project board](https://github.com/cdr/code-server/projects/3)_

vapurrmaid:
[...]

PR #3134, opened 2021-04-14
refactor: remove debian10/Dockerfile and update CONTRIBUTING

jsjoeio:
This PR removed `debian10/Dockerfile` which was introduced orignally by moog in https://github.com/cdr/code-server/pull/1499 for Docker development.

It also updates the `CONTRIBUTING.md` file with global module/packages required for developing code-server. 

Originally discussed here noting that we no longer use or need to maintain this.
https://github.com/cdr/code-server/pull/3077#pullrequestreview-633775205


jawnsy (approved):
I like having docker images to make development environments reproducible/easier to set up for new contributors, though an unmaintained image is worse than not having one, so this makes sense to me :smile: I think for the future we should revisit this, maybe we could make an image to do code-server development inside code-server, and also a variant for code-server development inside coder.

jsjoeio:
Love that idea! Noted for future in https://github.com/cdr/code-server/issues/2818 :D

oxy:
Uh oh - we still use the debian10 image in our release workflow, and our custom centos7 image is actually unused; I'll work on removing the dependency on the debian10 image in our workflow before we make a release :sweat_smile:

PR #3141, opened 2021-04-15
jsjoeio/fix-login-rate-limiter

### LINKED ISSUE #2647: Login rate limiter should not count against successful logins

jsjoeio:
This PR fixes the login rate limiter to not count successful logins.

## Changes

- adds unit test for `RateLimiter`
- moves `loginPage.test.ts` into `login.test.ts`
- adds e2e tests for missing password, wrong password (and rate limiter)

Fixes #2647


test/config.ts
```diff
@@ -50,7 +50,7 @@ globalSetup(async () => {
 
 const config: Config = {
   testDir: path.join(__dirname, "e2e"), // Search for tests in this directory.
-  timeout: 30000, // Each test is given 30 seconds.
+  timeout: 60000, // Each test is given 60 seconds.
```
jsjoeio, line 53: A couple of new the new tests are a bit longer so I increased the timeout.

test/config.ts
```diff
@@ -64,7 +64,7 @@ setConfig(config)
 
 const options: PlaywrightOptions = {
   headless: true, // Run tests in headless browsers.
-  video: "retain-on-failure",
+  video: "retry-with-video",
```
jsjoeio, line 67: I'm actually not sure if we want `retain-on-failure` or `retry-with-video`. Here is the difference:

> `retain-on-failure` - Record video for each test, but remove all videos from successful test runs.
> `retry-with-video` - Record video only when retrying a test.

If we do the `retain-on-failure` option, we only have videos for failures. If we do the other option, we keep the videos from test retries. And a test is only retried if it fails. Hmm..thinking out loud, I think we want to keep it on `retain-on-failure`. Open to other ideas though!

Source: [`playwright-test` docs](https://github.com/microsoft/playwright-test#modify-options)

test/e2e/login.test.ts
```diff
@@ -10,6 +10,12 @@ test.describe("login", () => {
     },
   }
 
+  test("should see the login page", options, async ({ page }) => {
+    await page.goto(CODE_SERVER_ADDRESS, { waitUntil: "networkidle" })
+    // It should send us to the login page
+    expect(await page.title()).toBe("code-server login")
+  })
+
```
jsjoeio, line 18: This was moved from `loginPage.test.ts`. I figured it made sense to put this here instead of having a separate file for that test.

test/config.ts
jsjoeio: Also, since we retry tests, they may run a bit longer, hence why we need to increase the timeout.

test/e2e/login.test.ts
```diff
@@ -20,4 +26,70 @@ test.describe("login", () => {
...
+    }
+
+    // The 15th should fail
+    await page.click(".submit")
+    await page.waitForLoadState("networkidle")
+    expect(await page.isVisible("text=Login rate limited!"))
+  })
+
+  // This test takes 8mins to run and is probably not worth adding to our e2e suite
+  // test.only("should not count successful logins against the rate limiter", options, async ({ page }) => {
```
jsjoeio, line 69: See comment. I originally wrote the test. I think it was a bit flaky but hard to tell. It adds 8mins to the e2e test suite for little value (making sure successful logins aren't counted against the rate limiter).

I don't think we should include this test (or rewrite it to be better).

The other option is to write an intergration level test (maybe just testing against the `/login` route) in a follow-up PR.

src/node/routes/login.ts
```diff
@@ -59,10 +59,6 @@ router.get("/", async (req, res) => {
 
 router.post("/", async (req, res) => {
   try {
-    if (!limiter.try()) {
-      throw new Error("Login rate limited!")
-    }
-
```
oxy, line 65: Moving this from the top to the bottom is not a good idea!

This is because we'd be trying to check if the login is valid, and then authenticate the user, and _then_ check if we're ratelimited. Instead, I'd suggest adding something like;

```ts
  public canTry(): boolean {
    return this.minuteLimiter.getTokensRemaining() > 0 && this.hourLimiter.getTokensRemaining() > 0
  }
```

to our RateLimiter, and then check `canTry` in the beginning of the route, and then use `.try()` to withdraw an attempt from the limiter if it failed.

With the current code, what happens is that you can still brute-force past the ratelimit and the logins can go through; which is not something that we want to happen.

src/node/routes/login.ts
```diff
@@ -12,8 +12,8 @@ export enum Cookie {
 }
 
 // RateLimiter wraps around the limiter library for logins.
-// It allows 2 logins every minute and 12 logins every hour.
-class RateLimiter {
+// It allows 2 logins every minute plus 12 logins every hour.
```
oxy, line 15: I think that the expected behavior for a ratelimiter would match the former comment; instead of changing the comment, it may make more sense to change the behavior to match, eg.

```ts
  public try(): boolean {
    if (this.hourLimiter.getTokensRemaining() > 0) {
      return this.minuteLimiter.tryRemoveTokens(1) && this.hourLimiter.tryRemoveTokens(1)
    }
    return false;
  }
```

test/unit/routes/login.test.ts
```diff
@@ -0,0 +1,23 @@
...
+      const limiter = new RateLimiter()
+
+      // The limiter allows 2 tries per minute plus 12 per hour
+      // so if we run it 15 times, 14 should return true and the last
+      // should return false
+      for (let i = 1; i <= 14; i++) {
+        expect(limiter.try()).toBe(true)
+      }
+
+      expect(limiter.try()).toBe(false)
```
oxy, line 20: May need to update the tests if you decide that this PR is a good place to change the behavior.

oxy (changes requested):
Ratelimiter needs adjustment - see individual review comments.

src/node/routes/login.ts
oxy: With the current behavior, this would be `||`; with the behavior I suggested in the next comment, it would be `&&`

test/config.ts
jawnsy: Is there a way to just get it to always keep videos, regardless of pass/fail/retry? They are useful to make sure our tests are doing what we think they're doing -- otherwise, we can have tests that pass but are not testing what we think they're testing (I've certainly written my share of tests that were not testing what I thought I was testing! and the passing tests gave me a false sense of security).

I know that disk usage can be a concern, but we can address that through other means, such as a shorter retention duration for old tests?

src/node/routes/login.ts
jsjoeio: > This is because we'd be trying to check if the login is valid, and then authenticate the user, and then check if we're ratelimited

But that's intentional! I shouldn't be ratelimited for successful logins, right? 

If a user types in the correct password and it's valid, we should authenticate them.

Otherwise, as a user, I am penalized even when I know my password. This also causes issues with the e2e tests, which I'm hoping to fix. 

> to our RateLimiter, and then check canTry in the beginning of the route, and then use .try() to withdraw an attempt from the limiter if it failed.

This sounds better. With your suggested change, tokens would only be withdrawn for failed attempts, right? Way better than the current approach.

src/node/routes/login.ts
jsjoeio: Ah, okay. @code-asher and I discussed this on Slack and were a bit confused. So the comment behavior was right, but the implementation was wrong. Got it!

test/unit/routes/login.test.ts
jsjoeio: Yeah, I like the changes you suggested so I'll update the test.

src/node/routes/login.ts
jsjoeio: Okay confirmed that should work! You rock!

src/node/routes/login.ts
code-asher: IMO we should keep the original implementation. Only twelve logins an hour seems overly restrictive to me.

test/config.ts
jsjoeio: Yup! According to the docs we can use `on` here to always record videos.

> I know that disk usage can be a concern, but we can address that through other means, such as a shorter retention duration for old tests

Now you're asking for extra fancy features! That's going to cost ya! 😛

AFAIK, that isn't an option built-in to playwright and would probably be something we run separately (maybe a GitHub Action) on a schedule. I think we can address that if/when it comes up. 

For now, I'll change it to `on`.

jawnsy: Oh yeah I meant GitHub has a setting for retention in our admin panel!

image

If it becomes a problem we can move the builds to Google Cloud and store them in Cloud Storage, too

src/node/routes/login.ts
oxy: We could raise it to 30 logins per hour; the way the logic I suggested works is like;
fail 1, fail 2, 60 second ratelimit, fail 3, fail 4, 60s ratelimit... repeat until n per hour (12, could increase it to 30 for now)

In the future, we should probably ratelimit by incoming IP address so that you can't deny service by just spamming the login endpoint; will create an issue about it.

test/config.ts
jawnsy: Sorry, it would've been better if I linked you to it lol https://github.com/cdr/code-server/settings/actions

jsjoeio: Ohhh!!!!!!!!! 

TIL. 

You already know how to fix/address then 😅 We're good then.

src/node/routes/login.ts
code-asher: Ah I think maybe we're talking about different rate limit algorithms.

These are token buckets so if I understand correctly the current implementation should get you:

fail 1..14, minute timeout, fail 1, fail 2, minute timeout, etc

With the `&&` you could get locked out for an hour if you use up all the tokens in the bucket. Or in other words:
fail 1...14, locked out for an hour

code-asher: I got the `&&` wrong, you'd get locked out for an hour after six minutes (if you tried twice per minute).

jsjoeio: So where did we land on this? Do I implement with @oxy's changes, or increase to 30 logins per hour?

code-asher: I think I would prefer the original implementation since it feels a little less abusable (a service denial attack would require you to constantly log in every minute rather than just up to the hourly amount).

I think the idea behind the current implementation is to allow a steady but small stream of attempts but also support an occasional burst (the twelve).

code-asher: I think the ideal way to implement that is actually a single limiter that accrues two tokens per minute up to a maximum but it doesn't look like this API supports that, thus the two separate limiters.

jsjoeio: Sounds good to me! I'll go with that and then have you two double-check when ready for a second review.

src/node/routes/login.ts
```diff
@@ -12,16 +12,20 @@ export enum Cookie {
...
   private readonly hourLimiter = new Limiter(12, "hour")
 
+  public canTry(): boolean {
+    return this.minuteLimiter.getTokensRemaining() > 0 || this.hourLimiter.getTokensRemaining() > 0
+  }
+
   public try(): boolean {
-    if (this.minuteLimiter.tryRemoveTokens(1)) {
-      return true
+    if (this.canTry()) {
```
code-asher, line 25: Nit: we don't need the `canTry` here I think? Should be the same behavior without it.

src/node/routes/login.ts
```diff
@@ -84,6 +89,12 @@ router.post("/", async (req, res) => {
       return redirect(req, res, to, { to: undefined })
     }
 
+    // Note: successful logins should not count against the RateLimiter
+    // which is why this logic must come after the successful login logic
+    if (!limiter.try()) {
```
code-asher, line 94: Should we move this after the `console.error` so we can log the failed attempt?

code-asher: Actually maybe we don't want to throw here because the actual error is that the login attempt failed. Then on the next attempt it'll be rate limited.

src/node/routes/login.ts
oxy: Oh hmm, yeah - in this case this is true; with `&&` you might end up with removing a token from one limiter but not the other, but with `||` there's no need for it.

test/unit/routes/login.test.ts
```diff
@@ -0,0 +1,36 @@
...
```
oxy, line 33: might want to also test `expect(limiter.canTry()).toBe(false)` here

src/node/routes/login.ts
jsjoeio: nice catch! i'll fix

oxy (dismissed):
Just a few nits; looking in good shape otherwise!

test/unit/routes/login.test.ts
jsjoeio: Good call!

test/e2e/login.test.ts
```diff
@@ -20,4 +26,46 @@ test.describe("login", () => {
...
+    expect(await page.isVisible("text=Incorrect password"))
+  })
+
+  test("should hit the rate limiter for too many unsuccessful logins", options, async ({ page }) => {
+    await page.goto(CODE_SERVER_ADDRESS, { waitUntil: "networkidle" })
+    // Type in password
+    await page.fill(".password", "password123")
+    // Click the submit button and login
+    // The current RateLimiter allows 2 logins per minute plus
+    // 12 logins per hour for a total of 14
```
code-asher, line 55: We might want to mention it's `-1` because of a previous incorrect password test (or remove that one).

code-asher (approved):

test/e2e/login.test.ts
jsjoeio: No you're right, this should be 14

PR #3150, opened 2021-04-16
Add builtin extension

code-asher:
This will let us add features without needing to modify VS Code.

To start I'm seeing if we can implement the logout feature using this.


jsjoeio:
Oh snap, this will make life SO much easier if this approach works 🎉

code-asher:
Closing this for now. The logout option ended up not being a good fit for an extension at the moment since we can't add top-level menu items.

I'll re-open this later when doing the port stuff most likely.

PR #3167, opened 2021-04-19
chore(deps): deduplicate deps in yarn.lock

oxy:
Probably a bit of a sledgehammer approach to the whole process; but should reduce duplicate dependencies and maybe save some small amount of space!


jawnsy (approved):
If the tests pass this seems good to me. Any thoughts on using that yarn deduplicate thing? Maybe a separate change though.

oxy:
At the moment, it should be good enough to run the tool manually whenever things pile up (for eg. from dependabot); we can look into integrating this with our workflow in the future.

oxy:
macOS build is still hitting `ESOCKETTIMEOUT` even when retrying downloads with yarn...

PR #3169, opened 2021-04-19
feat(testing): add e2e tests for code-server and terminal

jsjoeio:
This PR adds a test for the integrated terminal in code-server.

## Changes
- add `CodeServer` class for simplifying testing (thank you @jawnsy for the [POM](https://playwright.dev/docs/pom/) tip!)
- add test for `CodeServer`
- add test for terminal
- fix `expect` in a couple e2e tests (I forgot `.toBe(true)`)

## Screenshots
image

https://user-images.githubusercontent.com/3806031/115481505-97c6fe80-a201-11eb-9cba-4f4db09a28f8.mp4

## Checklist
- [x] tested locally
- [x] added a test
- [x] refactor and use temp folder for terminal test
- [x] use descriptive name like `terminal_e2e_test.txt`
- [x] fix `CodeServer.focusTerminal()` method
- [x] remove all TODOs for this PR
- [x] refactor `tmpdir`
- [x] clean up commits
- [x] try removing `waitForTimeout`s in `terminal.test.ts`
- [x] ~~try changing `esModuleInterop` to false~~ I don't want to mess with this. If someone wants to create a ticket to check out later, go for it. Otherwise, I'll add a note to the "Clean up" issue
- [ ] address feedback

### DIFF
--- a/src/node/constants.ts
+++ b/src/node/constants.ts
@@ -1,5 +1,6 @@
 import { logger } from "@coder/logger"
 import { JSONSchemaForNPMPackageJsonFiles } from "@schemastore/package"
+import * as os from "os"
 import * as path from "path"
 
 export function getPackageJson(relativePath: string): JSONSchemaForNPMPackageJsonFiles {
@@ -18,3 +19,4 @@ const pkg = getPackageJson("../../package.json")
 export const version = pkg.version || "development"
 export const commit = pkg.commit || "development"
 export const rootPath = path.resolve(__dirname, "../..")
+export const tmpdir = path.join(os.tmpdir(), "code-server")
--- a/test/e2e/browser.test.ts
+++ b/test/e2e/browser.test.ts
@@ -1,15 +1,23 @@
 import { test, expect } from "@playwright/test"
-import { CODE_SERVER_ADDRESS } from "../utils/constants"
+import { CodeServer } from "./models/CodeServer"
 
 // This is a "gut-check" test to make sure playwright is working as expected
-test("browser should display correct userAgent", async ({ page, browserName }) => {
-  const displayNames = {
-    chromium: "Chrome",
-    firefox: "Firefox",
-    webkit: "Safari",
-  }
-  await page.goto(CODE_SERVER_ADDRESS, { waitUntil: "networkidle" })
-  const userAgent = await page.evaluate("navigator.userAgent")
+test.describe("browser", () => {
+  let codeServer: CodeServer
 
-  expect(userAgent).toContain(displayNames[browserName])
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
+
+  test("browser should display correct userAgent", async ({ page, browserName }) => {
+    const displayNames = {
+      chromium: "Chrome",
+      firefox: "Firefox",
+      webkit: "Safari",
+    }
+    const userAgent = await page.evaluate("navigator.userAgent")
+
+    expect(userAgent).toContain(displayNames[browserName])
+  })
 })
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -20,10 +20,10 @@ test.describe("CodeServer", () => {
 
   test.beforeEach(async ({ page }) => {
     codeServer = new CodeServer(page)
-    await codeServer.navigate()
+    await codeServer.setup()
   })
 
-  test("should navigate to the CODE_SERVER_ADDRESS", options, async ({ page }) => {
+  test(`should navigate to ${CODE_SERVER_ADDRESS}`, options, async ({ page }) => {
     // We navigate codeServer before each test
     // and we start the test with a storage state
     // which means we should be logged in
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -1,0 +1,46 @@
+import { test, expect } from "@playwright/test"
+import { STORAGE } from "../utils/constants"
+import { CodeServer } from "./models/CodeServer"
+
+test.describe("CodeServer", () => {
+  // Create a new context with the saved storage state
+  // so we don't have to logged in
+  const options: any = {}
+  let codeServer: CodeServer
+
+  // TODO@jsjoeio
+  // Fix this once https://github.com/microsoft/playwright-test/issues/240
+  // is fixed
+  if (STORAGE) {
+    const storageState = JSON.parse(STORAGE) || {}
+    options.contextOptions = {
+      storageState,
+    }
+  }
+
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
+
+  test("should open the default folder if not open", options, async ({ page }) => {
+    await codeServer.openFolder()
+
+    // find workspaceStorage in the Explorer menu, which would be open in the User folder
+    // which is the default folder that opens
+    expect(await page.isVisible("text=workspaceStorage")).toBe(true)
+  })
+
+  test("should show the Integrated Terminal", options, async ({ page }) => {
+    await codeServer.viewTerminal()
+    expect(await page.isVisible("#terminal")).toBe(true)
+  })
+
+  test("should open a file with quickOpen", options, async ({ page }) => {
+    await codeServer.openFolder()
+    await codeServer.quickOpen("extensions.json")
+    // If the file is open, we will see an empty array
+    // assuming no extensions are installed
+    expect(await page.isVisible("text=[]"))
+  })
+})
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -1,5 +1,5 @@
 import { test, expect } from "@playwright/test"
-import { STORAGE } from "../utils/constants"
+import { CODE_SERVER_ADDRESS, STORAGE } from "../utils/constants"
 import { CodeServer } from "./models/CodeServer"
 
 test.describe("CodeServer", () => {
@@ -23,24 +23,23 @@ test.describe("CodeServer", () => {
     await codeServer.navigate()
   })
 
-  test("should open the default folder if not open", options, async ({ page }) => {
-    await codeServer.openFolder()
+  test("should navigate to the CODE_SERVER_ADDRESS", options, async ({ page }) => {
+    // We navigate codeServer before each test
+    // and we start the test with a storage state
+    // which means we should be logged in
+    // so it should be on the address
+    const url = page.url()
+    // We use match because there may be a / at the end
+    // so we don't want it to fail if we expect http://localhost:8080 to match http://localhost:8080/
+    expect(url).toMatch(CODE_SERVER_ADDRESS)
+  })
 
-    // find workspaceStorage in the Explorer menu, which would be open in the User folder
-    // which is the default folder that opens
-    expect(await page.isVisible("text=workspaceStorage")).toBe(true)
+  test("should always see the code-server editor", options, async ({ page }) => {
+    expect(await codeServer.isEditorVisible()).toBe(true)
   })
 
   test("should show the Integrated Terminal", options, async ({ page }) => {
-    await codeServer.viewTerminal()
+    await codeServer.focusTerminal()
     expect(await page.isVisible("#terminal")).toBe(true)
-  })
-
-  test("should open a file with quickOpen", options, async ({ page }) => {
-    await codeServer.openFolder()
-    await codeServer.quickOpen("extensions.json")
-    // If the file is open, we will see an empty array
-    // assuming no extensions are installed
-    expect(await page.isVisible("text=[]"))
   })
 })
--- a/test/e2e/globalSetup.test.ts
+++ b/test/e2e/globalSetup.test.ts
@@ -20,6 +20,6 @@ test.describe("globalSetup", () => {
   test("should keep us logged in using the storageState", options, async ({ page }) => {
     await page.goto(CODE_SERVER_ADDRESS, { waitUntil: "networkidle" })
     // Make sure the editor actually loaded
-    expect(await page.isVisible("div.monaco-workbench"))
+    expect(await page.isVisible("div.monaco-workbench")).toBe(true)
   })
 })
--- a/test/e2e/globalSetup.test.ts
+++ b/test/e2e/globalSetup.test.ts
@@ -1,5 +1,6 @@
 import { test, expect } from "@playwright/test"
-import { CODE_SERVER_ADDRESS, STORAGE } from "../utils/constants"
+import { STORAGE } from "../utils/constants"
+import { CodeServer } from "./models/CodeServer"
 
 // This test is to make sure the globalSetup works as expected
 // meaning globalSetup ran and stored the storageState in STORAGE
@@ -7,6 +8,7 @@ test.describe("globalSetup", () => {
   // Create a new context with the saved storage state
   // so we don't have to logged in
   const options: any = {}
+  let codeServer: CodeServer
 
   // TODO@jsjoeio
   // Fix this once https://github.com/microsoft/playwright-test/issues/240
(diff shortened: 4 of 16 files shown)


test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      storageState,
+    }
+  }
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
+
+  test("should echo a string to a file", options, async ({ page }) => {
+    const tmpFolderPath = fs.mkdtempSync(path.join(tmpdir(), "code-server-test"))
```
jsjoeio, line 32: Note: I tried to use `tmpdir` from `src/node/util` but ran into this error. I think it's because Playwright requires the test files and something doesn't support enums 

image

Leaving as is.

jawnsy: haha you might want to add a comment with the stack trace etc, so people don't spend time trying the same thing later :stuck_out_tongue:

code-asher: Maybe we can move it to `src/node/constants.ts`. Or if that fails we can duplicate it for now in `test/utils/constants.ts`.

jsjoeio: Good idea! I'll try that.

jsjoeio: I moved it to `constants` but new error :( 

I think Playwright uses CommonJS and we're using ES Modules so anything that isn't done the "CommonJS" way is going to cause Playwright to throw.

image

I think the refactor still makes sense, so I'll leave it, but I'll leave the test how it is and leave a comment.

jsjoeio: Possibly related? https://github.com/microsoft/playwright/issues/2269#issue-619578701

code-asher: Maybe the `"esModuleInterop": true` option in our `tsconfig.json` is allowing ES Module-style imports and we should remove that.

jsjoeio: Ah, maybe you're right: [docs](https://www.typescriptlang.org/tsconfig#esModuleInterop).

I wonder if turning that off will break anything.

src/node/constants.ts
```diff
@@ -18,3 +19,4 @@ const pkg = getPackageJson("../../package.json")
 export const version = pkg.version || "development"
 export const commit = pkg.commit || "development"
 export const rootPath = path.resolve(__dirname, "../..")
+export const tmpdir = path.join(os.tmpdir(), "code-server")
```
jsjoeio, line 22: 💄 Moved here for aesthetics. It is a constant and never changes

test/e2e/browser.test.ts
```diff
@@ -1,15 +1,23 @@
...
-  const userAgent = await page.evaluate("navigator.userAgent")
-
-  expect(userAgent).toContain(displayNames[browserName])
+test.describe("browser", () => {
+  let codeServer: CodeServer
+
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
```
jsjoeio, line 11: 💄 Refactored this to use the new `CodeServer` model

test/e2e/codeServer.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const storageState = JSON.parse(STORAGE) || {}
+    options.contextOptions = {
+      storageState,
+    }
+  }
+
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.setup()
+  })
```
jsjoeio, line 24: 💄 Refactored this to use the new `CodeServer` model

test/e2e/globalSetup.test.ts
```diff
@@ -17,9 +19,12 @@ test.describe("globalSetup", () => {
       storageState,
     }
   }
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.setup()
+  })
```
jsjoeio, line 25: 💄 Refactored this to use the new `CodeServer` model

test/e2e/login.test.ts
```diff
@@ -9,26 +10,32 @@ test.describe("login", () => {
       storageState: {},
     },
   }
+  let codeServer: CodeServer
+
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
```
jsjoeio, line 18: 💄 Refactored this to use the new `CodeServer` model

test/e2e/logout.test.ts
```diff
@@ -9,22 +10,31 @@ test.describe("logout", () => {
       storageState: {},
     },
   }
+  let codeServer: CodeServer
+
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.navigate()
+  })
```
jsjoeio, line 18: 💄 Refactored this to use the new `CodeServer` model

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
+import { Page } from "playwright"
+import { CODE_SERVER_ADDRESS } from "../../utils/constants"
+// This is a Page Object Model
+// We use these to simplify e2e test authoring
+// See Playwright docs: https://playwright.dev/docs/pom/
```
jsjoeio, line 5: Thanks @jawnsy for the tip! 🙌

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
...
+   * and reloads until it is
+   */
+  async reloadUntilEditorIsVisible() {
+    const editorIsVisible = await this.isEditorVisible()
+    let reloadCount = 0
+
+    // Occassionally code-server timeouts in Firefox
+    // we're not sure why
+    // but usually a reload or two fixes it
+    // TODO@jsjoeio @oxy look into Firefox reconnection/timeout issues
```
jsjoeio, line 31: Heads-up @oxy - tagged you and me in this todo. Something for us to eventually look into.

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
...
+    while (!editorIsVisible) {
+      reloadCount += 1
+      if (await this.isEditorVisible()) {
+        console.log(`    Editor became visible after ${reloadCount} reloads`)
+        break
+      }
+      // When a reload happens, we want to wait for all resources to be
+      // loaded completely. Hence why we use that instead of DOMContentLoaded
+      // Read more: https://thisthat.dev/dom-content-loaded-vs-load/
+      await this.page.reload({ waitUntil: "load" })
```
jsjoeio, line 41: The fact that we have to do this is probably a bug — either in code-server or something with Firefox. Probably us though.

jsjoeio: I noticed, after introducing this logic, the flakiness for tests went completely away. Most times when tests were failing before, it was because code-server didn't load.

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
...
+    await this.page.hover("text=View")
+    await this.page.click("text=View")
+
+    // Click text=Terminal
+    await this.page.hover("text=Terminal")
+    await this.page.click("text=Terminal")
+
+    // Wait for terminal to receive focus
+    // Sometimes the terminal reloads once or twice
+    // which is why we wati for it to have the focus class
```
jsjoeio, line 87:
```suggestion
    // which is why we wait for it to have the focus class
```

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * Navigates to CODE_SERVER_ADDRESS
+   * and reloads until the editor is visible
+   *
+   * Helpful for running before tests
+   */
+  async setup() {
+    await this.navigate()
+    await this.reloadUntilEditorIsVisible()
+  }
```
jsjoeio, line 103: Super handy to use in the `beforeEach` for e2e tests!

test/e2e/openHelpAbout.test.ts
```diff
@@ -15,32 +17,30 @@ test.describe("Open Help > About", () => {
     }
   }
 
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.setup()
+  })
+
```
jsjoeio, line 24: 💄 Refactored this to use the new `CodeServer` model

jsjoeio:
Womp. 404. Network issue. I wonder if we could automatically retry if it 404s? cc @oxy 

image

oxy:
@jsjoeio sadly there's no "please re-run the same step" within an action on GitHub - and we don't have control over `@actions/download-artifact`'s behavior either.

I think I can address this in a future PR where we split up all the things to upload to different artifacts, eg. `release-package-macos` instead of just everything being in `release-packages` - or maybe we should file a bug/support request with the actions team? looks to me kind of like a race condition

(file starts uploading from one action (eg. `package-macos`), and then the other action (eg. `docker-arm64`) tries to download it but can't because it isn't done uploading yet)

jsjoeio:
> address this in a future PR

Okay, sounds good!

> maybe we should file a bug/support request with the actions team?

Yeah, that could be a good idea too! I'll leave that to you if you want.

test/e2e/models/CodeServer.ts
```diff
@@ -0,0 +1,104 @@
...
+      await this.page.reload({ waitUntil: "load" })
+    }
+  }
+
+  /**
+   * Checks if the editor is visible
+   */
+  async isEditorVisible() {
+    // Make sure the editor actually loaded
+    // If it's not visible after 2 seconds, something is wrong
```
code-asher, line 50: Out of sync with the code (2 vs 5).

src/node/constants.ts
jawnsy: Always love your use of emoji, this put a smile on my face Joe!

Based on a quick read of the docs, `os.tmpdir()` returns an a-priori known directory (e.g. /tmp) so there are potential security issues, since this effectively convinces the test to write to a directory that we don't directly control. Probably not a problem since these are tests though. Would it make sense to add an unpredictable part to this, e.g. `path.join(os.tmpdir(), "code-server" + current pid)`?

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+  test.beforeEach(async ({ page }) => {
+    codeServer = new CodeServer(page)
+    await codeServer.setup()
+    // NOTE@jsjoeio
+    // We're not using tmpdir from src/node/constants
+    // because Playwright doesn't fully support ES modules from
+    // the erorrs I'm seeing
+    tmpFolderPath = fs.mkdtempSync(path.join(tmpdir(), "code-server-test"))
```
code-asher, line 36: Definitely a nit but i'd love if we use the same path format as the other tests.

We may also need to create the parent directory since I'm not sure `mkdtemp` is recursive. 

I think we should stick to the async pattern to be consistent as well.

Should we make a `tmpdir` constant in the test utils for now since we can't import the other one?

```suggestion
    const tmpdir = path.join(tmpdir(), "code-server/tests")
    const testdir = path.join(tmpdir, "terminal")
    await fs.mkdir(testdir, { recursive: true }) // this could be in beforeAll so it's only done once
    
    tmpFolderPath = await fs.mkdtemp(path.join(testdir, "temp-"))
```

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    tmpFile = `${tmpFolderPath}${path.sep}${testFileName}`
```
code-asher, line 37:
```suggestion
    tmpFile = path.join(tmpFolderPath, testFileName)
```

jawnsy (approved):

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+import util from "util"
+import * as cp from "child_process"
+import { STORAGE } from "../utils/constants"
+import { CodeServer } from "./models/CodeServer"
+
+test.describe("Integrated Terminal", () => {
+  // Create a new context with the saved storage state
+  // so we don't have to logged in
+  const options: any = {}
+  const testFileName = "test.txt"
```
code-asher, line 14: Inconsequential, but this matches a bit better with the switch to the pipe method:

```suggestion
  const testFileName = "pipe"
```

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    // We're not using tmpdir from src/node/constants
+    // because Playwright doesn't fully support ES modules from
+    // the erorrs I'm seeing
+    tmpFolderPath = fs.mkdtempSync(path.join(tmpdir(), "code-server-test"))
+    tmpFile = `${tmpFolderPath}${path.sep}${testFileName}`
+  })
+
+  test.afterEach(async () => {
+    // Ensure directory was removed
+    fs.rmdirSync(tmpFolderPath, { recursive: true })
```
code-asher, line 42: If we decide to make the path change above we''ll want to remove the directory for this test only:

```suggestion
    await fs.rmdir(testdir, { recursive: true })
```

Also, any thoughts on moving this to `beforeAll`? My thinking is that it could be useful for debugging in some scenarios to see the files generated by the tests.

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const { stdout } = await output
+    expect(stdout).toMatch(testString)
+
+    // .access checks if the file exists without opening it
+    // it doesn't return anything hence why we expect it to
+    // resolve to undefined
+    // If the promise rejects (i.e. the file doesn't exist)
+    // then the assertion will fail
+    await expect(fs.promises.access(tmpFile)).resolves.toBeUndefined()
```
code-asher, line 65: We already know the command worked so I think we can skip this.

test/e2e/terminal.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    await fs.promises.rmdir(tmpFolderPath, { recursive: true })
+    // Make sure neither file nor folder exist
+    // Note: We have to use ts-ignore because of an upstream typing error
+    // See: https://github.com/microsoft/folio/issues/230#event-4621948411
+    /* eslint-disable @typescript-eslint/ban-ts-comment */
+    // @ts-ignore
+    expect(fs.promises.access(tmpFile)).rejects.toThrowError(/no such file or directory/)
+    // @ts-ignore
+    expect(fs.promises.access(tmpFolderPath)).rejects.toThrowError(/no such file or directory/)
```
code-asher, line 75: We `rmdir` in `afterEach` so I think we can skip this too.

code-asher (changes requested):
I like the new model!

src/node/constants.ts
code-asher: What about something like this for tests (and we move it into `test/utils/constants.ts`):

```
export async function tmpdir(testName: string): Promise<string> {
  const dir = path.join(os.tmpdir(), "code-server")
  await fs.mkdir(dir, { recursive: true })
  return fs.mkdtemp(path.join(dir, `test-${name}-`))
}
```

`tmpdir` is used in one place outside of tests when creating the socket proxy but I think maybe we should use the data directory instead of using tmp and we should randomize it. Would fix https://github.com/cdr/code-server/issues/2200#issuecomment-709460909

test/e2e/codeServer.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  test("should navigate to the CODE_SERVER_ADDRESS", options, async ({ page }) => {
```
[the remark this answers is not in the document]
jsjoeio: Good call!

src/node/constants.ts
jsjoeio: TIL about the potential security issues!

Totally I will add that to `constants`.

Re: changing to `data` directory feels out of scope of this PR, but will make a note and do in a followup PR

test/e2e/models/CodeServer.ts
jsjoeio: Oops! Nice catch -  I went back and forth between the two 😂

test/e2e/terminal.test.ts
jsjoeio: yes! I'll refactor to use the suggestion for `tmp` in constants that you added above!

test/e2e/terminal.test.ts
jsjoeio: Good call!

test/e2e/terminal.test.ts
jsjoeio: Good idea! Will move

test/e2e/terminal.test.ts
jsjoeio: sounds good!

jsjoeio:
> No coverage uploaded for pull request base (main@72ca12c)

I think this is cause I started this PR before Codecov was added. We'll see though in the next PR on code-server.

code-asher (approved):

PR #3178, opened 2021-04-20
Minor connections refactor

code-asher:
More details in the commit messages.

### DIFF
--- a/lib/vscode/src/vs/server/node/connection.ts
+++ b/lib/vscode/src/vs/server/node/connection.ts
@@ -3,30 +3,49 @@ import * as cp from 'child_process';
 import { VSBuffer } from 'vs/base/common/buffer';
 import { Emitter } from 'vs/base/common/event';
 import { FileAccess } from 'vs/base/common/network';
-import { ISocket } from 'vs/base/parts/ipc/common/ipc.net';
 import { INativeEnvironmentService } from 'vs/platform/environment/common/environment';
+import { IRemoteExtensionHostStartParams } from 'vs/platform/remote/common/remoteAgentConnection';
 import { getNlsConfiguration } from 'vs/server/node/nls';
 import { Protocol } from 'vs/server/node/protocol';
 import { IExtHostReadyMessage } from 'vs/workbench/services/extensions/common/extensionHostProtocol';
 
 export abstract class Connection {
 	private readonly _onClose = new Emitter<void>();
+	/**
+	 * Fire when the connection is closed (not just disconnected). This should
+	 * only happen when the connection is offline and old or has an error.
+	 */
 	public readonly onClose = this._onClose.event;
 	private disposed = false;
 	private _offline: number | undefined;
 
-	public constructor(protected protocol: Protocol, public readonly token: string) {}
+	protected readonly logger: Logger;
+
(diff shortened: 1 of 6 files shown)


lib/vscode/src/vs/server/node/connection.ts
```diff
@@ -181,20 +212,20 @@ export class ExtensionHostConnection extends Connection {
...
 					break;
 				case 'VSCODE_EXTHOST_DISCONNECTED':
-					this.logger.trace('Going offline');
+					this.logger.debug('Got disconnected message');
 					this.setOffline();
 					break;
 				case 'VSCODE_EXTHOST_IPC_READY':
-					this.logger.trace('Got ready message');
-					this.sendInitMessage(buffer);
+					this.logger.debug('Handshake completed');
```
jsjoeio, line 219: Nice addition 🎉

lib/vscode/src/vs/server/node/protocol.ts
```diff
@@ -1,21 +1,31 @@
...
 export interface SocketOptions {
+	/** The token is how we identify and connect to existing sessions. */
 	readonly reconnectionToken: string;
+	/** Specifies that the client is trying to reconnect. */
 	readonly reconnection: boolean;
+	/** If true assume this is not a web socket (always false for code-server). */
 	readonly skipWebSocketFrames: boolean;
+	/** Whether to support compression (web socket only). */
 	readonly permessageDeflate?: boolean;
+	/**
```
jsjoeio, line 17: awesome docs! thanks for adding 👏

jsjoeio (approved):
This is awesome! Love the code comments too ♥️

P.S. - was this the thing we were discussing where I said I would learn from the code and write a test?

code-asher:
Fixed up the double/single quote errors.

Nope, this is mostly just to set the stage so when it happens again we'll have some better logs to figure out what's going on.

lib/vscode/src/vs/server/node/connection.ts
```diff
@@ -27,7 +27,7 @@ export abstract class Connection {
 	) {
 		this.logger = logger.named(
 			this.name,
-			field("token", this.protocol.options.reconnectionToken),
+			field('token', this.protocol.options.reconnectionToken),
```
jawnsy, line 30: what's the difference between `"` and `'` in JavaScript/TypeScript?

code-asher: None but VS Code's linting rules enforce `'` for "unlocalized" strings so the only time they are allowed is in calls like:

```
localize('projectQuickPick.help', "TypeScript help")
```

jawnsy: ah very interesting pattern, so they're using something that extracts all `"`-quoted strings into resource bundles?

code-asher: I think they might be doing something like that, although if they can already detect and report to the dev when a string should be using `"` or `'` I'm not sure why they can't just use that knowledge to extract it and don't bother the dev. :stuck_out_tongue:

jsjoeio:
Nice work! Merge whenever 🚢

PR #3200, opened 2021-04-23
feat(testing): add tests for registerServiceWorker

jsjoeio:
This PR adds two more tests to cover the last line for `src/browser/register.ts` which should be at 100% now. 

image

Interestingly - the output from Jest and what Codecov picks up are a little different 🤷‍♂️

### DIFF
--- a/src/browser/register.ts
+++ b/src/browser/register.ts
@@ -4,12 +4,12 @@ import "./pages/error.css"
 import "./pages/global.css"
 import "./pages/login.css"
 
-async function registerServiceWorker(): Promise<void> {
+export async function registerServiceWorker(): Promise<void> {
   const options = getOptions()
   const path = normalize(`${options.csStaticBase}/dist/serviceWorker.js`)
   try {
     await navigator.serviceWorker.register(path, {
-      scope: (options.base ?? "") + "/",
+      scope: options.base + "/",
     })
     console.log("[Service Worker] registered")
   } catch (error) {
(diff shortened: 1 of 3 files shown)


src/browser/register.ts
```diff
@@ -4,12 +4,12 @@ import "./pages/error.css"
...
 import "./pages/login.css"
 
-async function registerServiceWorker(): Promise<void> {
+export async function registerServiceWorker(): Promise<void> {
   const options = getOptions()
   const path = normalize(`${options.csStaticBase}/dist/serviceWorker.js`)
   try {
     await navigator.serviceWorker.register(path, {
-      scope: (options.base ?? "") + "/",
+      scope: options.base + "/",
```
jsjoeio, line 12: TLDR: `options.base` will always return a string. No need to use [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)
See comment message for explanation: https://github.com/cdr/code-server/pull/3200/commits/67dbb36732afe76c26c4d7db793c5912a1dfafcb

test/unit/register.test.ts
```diff
@@ -84,4 +92,90 @@ describe("register", () => {
...
+      serviceWorkerPath = path
+      serviceWorkerScope = options.scope
+      return undefined
+    })
+
+    beforeAll(() => {
+      const location: LocationLike = {
+        pathname: "",
+        origin: "http://localhost:8080",
+        // search: "?environmentId=600e0187-0909d8a00cb0a394720d4dce",
```
jawnsy, line 108: leftover comment?

jawnsy (approved):
LGTM!

test/unit/register.test.ts
jsjoeio: Oops! Good catch. I'll remove!

PR #3217, opened 2021-04-26
docs(guide.md): add `caddy` example for serving from sub-path

catthehacker:
Simple example of serving from sub-path in Caddy


jsjoeio (approved):
This is a fantastic addition to our docs! 👏

Thanks so much @catthehacker 🎉

PR #3223, opened 2021-04-26
chore(deps): update dependencies with CVEs

oxy:
Update `devDependencies` with known CVEs.
They only process trusted data, so there is no need for us to release a security advisory; but cherry-pick updates for them nonetheless to satisfy tools like Trivy.


jsjoeio (approved):
Thanks for taking care of these 🙌

PR #3261, opened 2021-04-29
feat(ci): add trivy job for security

### LINKED ISSUE #3177: Scan built docker images using trivy or grype
We use AquaSec Trivy in some of our other projects for scanning our source code as well as built containers for possible security issues related to our third-party dependencies (e.g. packages installed with `apt-get` or `yarn`), and should consider the same for code-server

This would complement the audit-ci tool that we have already integrated.

jsjoeio:
This PR adds a new job to our CI called `trivy` which uses [`trivy-action`](https://github.com/aquasecurity/trivy-action) to scan our code and upload the Trivy scan results to the GitHub Security tab.

Fixes #3177

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -406,3 +406,61 @@ jobs:
         with:
           name: release-images
           path: ./release-images
+
+  trivy-scan-image:
+    runs-on: ubuntu-20.04
+    needs: docker-amd64
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Download release images
+        uses: actions/download-artifact@v2
+        with:
+          name: release-images
+          path: ./release-images
+
+      - name: Run Trivy vulnerability scanner in image mode
+        # Commit SHA for v0.0.14
+        uses: aquasecurity/trivy-action@b38389f8efef9798810fe0c5b5096ac198cffd54
+        with:
+          input: "./release-images/code-server-amd64-*.tar"
+          scan-type: "image"
+          ignore-unfixed: true
+          format: "template"
+          template: "@/contrib/sarif.tpl"
+          output: "trivy-image-results.sarif"
+          severity: "HIGH,CRITICAL"
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: "trivy-image-results.sarif"
+
+  # We have to use two trivy jobs
+  # because GitHub only allows
+  # codeql/upload-sarif action per job
+  trivy-scan-repo:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        # Commit SHA for v0.0.14
+        uses: aquasecurity/trivy-action@b38389f8efef9798810fe0c5b5096ac198cffd54
+        with:
+          scan-type: "fs"
+          scan-ref: "."
+          ignore-unfixed: true
+          format: "template"
+          template: "@/contrib/sarif.tpl"
+          output: "trivy-repo-results.sarif"
+          severity: "HIGH,CRITICAL"
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: "trivy-repo-results.sarif"


.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
...
+    steps:
+      - name: Checkout code
+      - uses: actions/checkout@v2
+
+      - name: Build an image from Dockerfile
+        run: |
+          docker build -t codercom/code-server:${{ github.sha }} ci/release-image/Dockerfile
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
```
jawnsy, line 422: you may want to be careful with using `@master` because it can break unexpectedly, I'd suggest using the latest current tag version and letting dependabot open PRs for updates

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
         with:
           name: release-images
           path: ./release-images
+
+  trivy:
+    runs-on: ubuntu-18.04
```
jsjoeio, line 411: ❓ I don't know if it can use `ubuntu-latest` - their `README.md` was updated recently and uses this version so I stuck with it

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
...
+    needs: prebuild
```
jsjoeio, line 412: 🛑 I did this so we only run this job if all the prebuild checks pass

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
...
+    steps:
+      - name: Checkout code
+      - uses: actions/checkout@v2
+
+      - name: Build an image from Dockerfile
+        run: |
+          docker build -t codercom/code-server:${{ github.sha }} ci/release-image/Dockerfile
```
jawnsy, line 419: why is building the image needed if we're just doing a filesystem scan?

alternatively, maybe we could also do an image scan? because the images contain other stuff (e.g. OS dependencies) that would not be caught in a filesystem scan

jsjoeio, line 419: ❓it's been a while since I've actually used `docker build`. I think this is right but not sure if I need `./` in front of the path?

jawnsy (approved):
Nice work Joe! :heart:

.github/workflows/ci.yaml
jsjoeio: Actually. Now that I'm looking a bit closer. I wonder if I should instead run `sh ./ci/release-image/build.sh`? 🤔

https://github.com/cdr/code-server/blob/main/ci/release-image/build.sh#L8

And I also think my `docker build` syntax is off - missing `.` at end?

jsjoeio: And I think I need `-f` since I'm pointing to a file? 

Source: [Docker build docs](https://docs.docker.com/engine/reference/commandline/build/#options)

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
...
+
+      - name: Run Trivy vulnerability scanner in repo mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'fs'
+          format: 'template'
+          template: '@/contrib/sarif.tpl'
+          output: 'trivy-results.sarif'
+          severity: 'CRITICAL'
```
jsjoeio, line 428: Explanation of options:

- `fs` scan files instead of `image`
- `template` because we use the `sarif` template to upload to GitHub Security tab
- `CRITICAL` type of severity. other options: `UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL`


Source: [trivy-action inputs](https://github.com/aquasecurity/trivy-action#inputs)

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,28 @@ jobs:
...
+
+      - name: Upload Trivy scan results to GitHub Security tab
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: 'trivy-results.sarif'
```
jsjoeio, line 433: Suggested in docs. Let's us view vulnerabilities in [Security tab](https://github.com/cdr/code-server/security).

.github/workflows/ci.yaml
jsjoeio: I almost messaged you to ask which we should do. 

What would you recommend?

.github/workflows/ci.yaml
jsjoeio: Ah, good point! 

Would that be changing it to `aquasecurity/trivy-action@v0.0.14` if that's the [latest release](https://github.com/aquasecurity/trivy-action/releases/tag/0.0.14)?

.github/workflows/ci.yaml
jawnsy: I'm not going to make specific _recommendations_ because you have more context about the code-server project than me, but will just share some experience and ideas for you to think about, not in any particular order:

* In the monorepo, we do both a filesystem scan and image scan.
  * The filesystem scan is nice because it runs quickly and can give feedback rapidly, and we can do it on every PR/merge to master without adding much time to the build duration. The downside is that you will either need to ignore errors for the check using `continue-on-error`, or else accept that sometimes the build will start failing until you have a chance to remediate the failing checks. If using `continue-on-error`, you'll need to remember to manually check the tab to see if things were OK, or look out for warnings that GitHub shows (it will show that there was a failure as a warning, but it won't appear red).
  * The image scan is useful because it can detect problems with installed packages (via `apt-get`) which source code scans cannot detect, and ultimately we're responsible for the full supply chain (both our code, third-party dependencies we import with package.json or submodules, upstream container images that we build `FROM`, and packages that we install via `apt-get` or similar)
* In the monorepo, we put our security scanning in a separate workflow and run it on a scheduled basis (nightly).
  * This means that it's not part of the critical path and won't impact PRs. This may be mitigated depending on the test parallelism we're using, if the other builds take longer than these scans and everything happens in parallel, then these checks would not be in the critical path so it's a moot point.
  * We have a central scan workflow that handles both filesystem and image scans. We do this both on commits to our default branch as well as on a nightly basis, this is to catch problems that show up even if we aren't changing the code (i.e. a security issue is discovered in a package we installed with apt-get, so even though we didn't modify any source code, we suddenly went from clean to vulnerable).
  * I broke the workflow so it was convenient that I could turn it off without affecting other stuff

Other thoughts/ideas:

* In the future, we might want to add other scan tools as well (such as Anchore Grype, Twistlock, OpenSCAP) and having these in a separate workflow means they're outside the critical path and won't slow down your normal development. The downside of the separate workflow is that it may take a day to discover vulnerabilities, but I don't think that's a big problem as long as we fix the issues quickly.
* Generating a Software Bill of Materials (i.e. the full list of "ingredients" of a code-server image, including system packages, third-party dependencies brought in via package.json, etc) might be useful in a similar step. This can be done using tools like [Anchore Syft](https://github.com/anchore/syft).
* We may want to upload reports in a different way (e.g. an HTML/CSS report page) to an S3 bucket or something, just to make these results easier for us to process.

I really like the integration with CodeQL that you discovered! Looks like it'll help make things easier to navigate, regardless of how we implement the workflow :)

.github/workflows/ci.yaml
jawnsy: Actions generally are just small Node.js/JavaScript programs. GitHub's Actions Runner includes Node.js 12, so these actions just do something like run commands to download trivy and run the scan. So my guess is that using a more recent version of Ubuntu should be fine, though I would suggest pinning it to `ubuntu-20.04` because otherwise GitHub issues a warning about ubuntu-latest changing from ubuntu-18.04 to ubuntu-20.04

It just generally seems better for us to control versions of things so that we can decide when to upgrade, and avoid breakages

.github/workflows/ci.yaml
jawnsy: Ah, looking at their code, it looks like it runs trivy inside a Docker container and just runs the latest version: https://github.com/aquasecurity/trivy-action/blob/0.0.14/Dockerfile - so pinning the Action doesn't pin the dependency on `trivy` :frowning: 

We can just accept the risk of stuff breaking suddenly, I just have... some scars from that :joy:

jsjoeio: Ahh, okay. I will take on the burden so you don't have to relive the scars if this breaks 😂

.github/workflows/ci.yaml
jsjoeio: Ahh, okay that makes more sense. I'll update to `ubuntu-20.04` then. Thanks for the explanation! 🙌

.github/workflows/ci.yaml
jsjoeio: Wow, thank you for the detailed explanation! I think based on all of that, I'm going to make some significant changes:
- move to a separate workflow
- add `continue-on-error` so it doesn't fail PR checks
- scan both repo and image

I'll make those then re-request your reivew!

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: "fs"
+          ignore-unfixed: true
+          format: "template"
+          template: "@/contrib/sarif.tpl"
+          output: "trivy-repo-results.sarif"
+          severity: "CRITICAL"
+
+      - name: Upload Trivy scan results to GitHub Security tab
```
jawnsy, line 31: do you know if CodeQL merges these correctly, since we have two uploads happening in the same PR or build, or does it do something like overwrite one set of results with the other?

asking because I had those issues with Coveralls, so now try to think more carefully :joy:

jawnsy (approved):
LGTM! Just some minor thoughts for your consideration

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,61 @@
+name: "Trivy - Scan for vulnerabilities"
```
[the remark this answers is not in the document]
jsjoeio: TIL! I'll change that

.github/workflows/trivy.yml
jsjoeio: honestly, i have no idea...is it bad to test after merging? 😂

jawnsy: nope, sometimes with CI it's a lot of testing in prod lol :joy: good luck!!

jsjoeio: image

jawnsy: stay reckless, my friends! :cowboy_hat_face:

jsjoeio:
> 2021-04-29T21:24:34.963Z	FATAL	scan error: image scan failed: failed analysis: walk dir: failed to analyze file: analyze file (.): unable to open a file (.): unable to read file: read .: is a directory

Hmm time to figure out why this failed 🤔 [link to line in logs](https://github.com/cdr/code-server/runs/2470769790?check_suite_focus=true#step:4:16)

jsjoeio:
> Step 7/14 : COPY release-packages/code-server*.deb /tmp/

Turns out the `build-docker-image` script relies on the `release-packages` being present 😅 so I refactored it back into `ci.yaml` since they get built in that workflow.

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,52 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Download release package
+        uses: actions/download-artifact@v2
+        with:
+          name: release-packages
+          path: ./release-packages
+
+      - name: Build Docker image
+        run: ./ci/steps/build-docker-image.sh
```
oxy, line 425: It might make more sense to fetch the docker image we build in `docker-amd64` instead of building them again in the Trivy workflow, I think

jsjoeio: Ooo didn't think of that. I'll see if I can do that!

jsjoeio: I'm going to get it working and then I'll refactor to fetch the docker image we build

ci/release-image/build.sh
```diff
@@ -5,7 +5,7 @@ main() {
   cd "$(dirname "$0")/../.."
   source ./ci/lib.sh
 
-  docker build -t "codercom/code-server-$ARCH:$VERSION" -f ./ci/release-image/Dockerfile .
+  docker build -t "codercom/code-server-$ARCH:$VERSION" -t "codercom/code-server:$COMMIT_SHA" -f ./ci/release-image/Dockerfile .
```
jawnsy, line 8: nice, using the commit hash as a version tag is a great way to handle this :+1:

jsjoeio:
Okay I finally got  it all working thanks to help from @code-asher 🎉

Though it says no vulnerabilities were found 🤔 

https://github.com/cdr/code-server/security/code-scanning?query=ref%3Arefs%2Fpull%2F3261%2Fmerge+tool%3ATrivy

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,60 @@ jobs:
...
+      - name: Run Trivy vulnerability scanner in image mode
+        uses: aquasecurity/trivy-action@master
+        with:
+          input: "./release-images/code-server-amd64-*.tar"
+          scan-type: "image"
+          ignore-unfixed: true
+          format: "template"
+          template: "@/contrib/sarif.tpl"
+          output: "trivy-image-results.sarif"
+          severity: "CRITICAL"
```
jawnsy, line 433: I suggest we use HIGH,CRITICAL here as those are usually the ones that people expect fixed quickly

jsjoeio:
image

Thought maybe the findings were not new? https://github.com/cdr/code-server/pull/3261/checks?check_run_id=2496406312

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,60 @@ jobs:
...
+        uses: actions/checkout@v2
+
+      - name: Download release images
+        uses: actions/download-artifact@v2
+        with:
+          name: release-images
+          path: ./release-images
+
+      - name: Run Trivy vulnerability scanner in image mode
+        uses: aquasecurity/trivy-action@master
```
jawnsy, line 425: suggest using the version tag here, even though we're always running with the latest version of trivy, we can at least avoid breakages in the Action itself by pinning

jawnsy (approved):
Nice stuff! It looks so easy when it's done :)

jsjoeio:
> Nice stuff! It looks so easy when it's done :)

😂 I know right? With Asher's help, it was a lot easier.

.github/workflows/ci.yaml
jsjoeio: Good call! I'll change that

.github/workflows/ci.yaml
jsjoeio: Oh right! I always forget. Nice catch!

.github/workflows/ci.yaml
```diff
@@ -406,3 +406,60 @@ jobs:
...
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: "trivy-image-results.sarif"
+
+  # We have to use two trivy jobs
+  # because GitHub only allows
+  # codeql/upload-sarif action per job
+  trivy-scan-repo:
+    runs-on: ubuntu-20.04
+    needs: docker-amd64
```
oxy, line 445: I don't think we need the Docker image in this step;
also question: does using trivy for our scans kind of obsolete "audit-ci"?

jawnsy: No it doesn't, these tools all use different feeds and work in different ways, so they are complementary - I've seen that Dependabot sometimes detects things that Trivy misses, and vice-versa, and sometimes `yarn audit` shows other stuff too. I am not too sure why this happens, but it seems worthwhile to have multiple scans, especially given that the cost of running these additional scans is low

oxy: Ah, that makes sense - though I'd definitely prefer one tool that does it all rather than many of them that do same but not exactly the same things from a maintainability perspective.

I guess having multiple tools works, at least in the short term!

jsjoeio: Oh yes! When I split up the two jobs, I forgot to remove the "needs" step. You're right. Since it's just checking the repo code, it doesn't need the docker step. Will update!

jsjoeio:
Note: for the trivy-action version I went with the commit SHA. 

> Using the commit SHA of a released action version is the safest for stability and security.

[GitHub Docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepsuses)

PR #3260, opened 2021-04-29
fix(deps): remove eslint-plugin-jest-playwright

jsjoeio:
This PR addresses forgotten work from when we migrated from `jest-playwright` to `playwright-test`.

PR #3263, opened 2021-04-29
fix(testing): reduce flakiness of terminal.test.ts and use 1 worker for e2e tests

jsjoeio:
This PR reduces the flakiness of the `terminal.test.ts` and testing overall by using 1 worker. It doesn't eliminate all flakiness but should reduce it.

## Changes
- removes quotes from `echo` command in `terminal.test.ts`
- use 1 worker when running tests
- update `yarn.lock` (forgot to do so in https://github.com/cdr/code-server/pull/3260)

## Screenshots

Running the `terminal.test.ts` 30 times locally (10 for each major browser) and it only failed once -> in chromium. 
image

It's failing because it's either not receiving the full `echo` command or hitting enter before it receives the full command. 

https://user-images.githubusercontent.com/3806031/116630124-8b305d80-a907-11eb-98bc-52ec97b808a9.mp4

## Checklist
- [x] tested locally

I'm going to open up a separate PR to add an additional check for e2e tests to check the connection. https://github.com/cdr/code-server/issues/3270

### DIFF
--- a/test/config.ts
+++ b/test/config.ts
@@ -52,6 +52,7 @@ const config: Config = {
   testDir: path.join(__dirname, "e2e"), // Search for tests in this directory.
   timeout: 60000, // Each test is given 60 seconds.
   retries: 3, // Retry failing tests 2 times
+  workers: 1,
 }
 
 if (process.env.CI) {
--- a/test/e2e/terminal.test.ts
+++ b/test/e2e/terminal.test.ts
@@ -50,7 +50,7 @@ test.describe("Integrated Terminal", () => {
     await codeServer.focusTerminal()
 
     await page.waitForLoadState("load")
-    await page.keyboard.type(`echo '${testString}' > '${tmpFile}'`)
+    await page.keyboard.type(`echo ${testString} > ${tmpFile}`)
     await page.keyboard.press("Enter")
     // It may take a second to process
     await page.waitForTimeout(1000)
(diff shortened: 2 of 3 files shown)


test/config.ts
```diff
@@ -52,6 +52,7 @@ const config: Config = {
   testDir: path.join(__dirname, "e2e"), // Search for tests in this directory.
   timeout: 60000, // Each test is given 60 seconds.
   retries: 3, // Retry failing tests 2 times
+  workers: 1,
```
jawnsy, line 55: does this make tests slower?

jawnsy (approved):
sweet!

test/config.ts
jsjoeio: > does this make tests slower?

Yes 😢 Unfortunately, it seems having these tests run in parallel against the same code-server application leads to flakiness (i.e. weird things happen). It's probably the way code-server is built and/or having concurrent connections to the same instance. 

More about workers: 
> Each worker process creates a new environment to run tests. Different environments always run in different workers. By default, Folio reuses the worker as much as it can to make testing faster, but it will create a new worker when retrying tests, after any test failure, to initialize a new environment, or just to speed up test execution if the worker limit is not reached.
— `folio` [docs](https://github.com/microsoft/folio#workers) (playwright-test uses folio under the hood)

PR #3271, opened 2021-04-30
feat(testing): add isConnected check

### LINKED ISSUE #3270: [dev]: add check to e2e tests for connection
## Problem
We've noticed the occasional timeout in e2e tests. In an effort to reduce flakiness due to connection timeouts, we think it would be beneficial to add a check for the connection.

## Solution

Add method to `CodeServer` e2e model to check connection by asserting that the `CODE_SERVER_ADDRESS` shows up in the bottom left-corner. If it doesn't, reload.

jsjoeio:
This PR adds a new check for e2e tests to ensure we have a connection.

## Changes
- add `isConnected` method to `CodeServer`
- add e2e test for `isConnected`

## Screenshots
todo

## Checklist
- [x] tested locally
- [x] added a test

Fixes #3270


test/e2e/models/CodeServer.ts
```diff
@@ -56,6 +59,21 @@ export class CodeServer {
...
 
+  /**
+   * Checks if the editor is visible
+   */
+  async isConnected() {
+    await this.page.waitForLoadState("networkidle")
+
+    // Slices http://localhost:8080
+    // to localhost:8080
+    const hostWithoutHttp = CODE_SERVER_ADDRESS.slice(7)
```
jawnsy, line 70: you could also do: `new URL("https://localhost:8080").host` so that http/https both work

```suggestion
    const hostWithoutHttp = new URL(CODE_SERVER_ADDRESS).host
```

jawnsy (approved):
Nice!

test/e2e/models/CodeServer.ts
jsjoeio: TIL! even better 🔥

Issue #3014, opened 2021-04-01
zlib error with websocket compression enabled


sweet-mentat:
## OS/Web Information

- Web Browser: Chrome Version 89.0.4389.90 (Official Build) (32-bit)
- Local OS: Windows Server 2016 Datacenter
- Remote OS: Ubuntu 18.04.5
- Remote Architecture: Amazon EC2
- `code-server --version`: v3.9.2

config.yaml:

## Steps to Reproduce

1. systemctl --user start code-server
2. log into it from browser
3. config.yaml at ~/.config/code-server/
bind-addr: 127.0.0.1:9001
auth: none
cert: /usr/lib/code-server/cert.pem
cert-key: /usr/lib/code-server/key.pem

## Expected
journalctl:
Apr 01 15:16:16 ip-10-*-*-* systemd[11951]: Started code-server.
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.691Z] info  code-server 3.8.1 41ad0c0c4c14df78e1bfa3a44d8b633ed2fc26e5
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.694Z] info  Using user-data-dir ~/.local/share/code-server
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.734Z] info  Using config file ~/.config/code-server/config.yaml
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.734Z] info  HTTP server listening on http://127.0.0.1:9001
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.734Z] info    - Authentication is disabled
Apr 01 15:16:18 ip-10-*-*-* code-server[13218]: [2021-04-01T15:16:18.734Z] info    - Using certificate for HTTPS: /usr/lib/code-server/cert.pem

## Actual
journalctl:
Mar 31 21:12:22 ip-10-*-*-* systemd[8572]: Started code-server.
Mar 31 21:12:23 ip-10-*-*-* code-server[26055]: [2021-03-31T21:12:23.935Z] info  code-server 3.9.2 109d2ce3247869eaeab67aa7e5423503ec9eb859
Mar 31 21:12:23 ip-10-*-*-* code-server[26055]: [2021-03-31T21:12:23.937Z] info  Using user-data-dir ~/.local/share/code-server
[...]

PR #3286, opened 2021-05-04
Add flag for toggling permessage-deflate

code-asher:
This is a workaround for https://github.com/cdr/code-server/issues/3014 until we figure it out (if it is indeed related which it might not be; the zlib errors could merely be side effects of a connection issue).

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -6,6 +6,11 @@ import * as path from "path"
 import { Args as VsArgs } from "../../typings/ipc"
 import { canConnect, generateCertificate, generatePassword, humanPath, paths } from "./util"
 
+export enum Feature {
+  /** Web socket compression. */
+  PermessageDeflate = "permessage-deflate",
+}
+
 export enum AuthType {
   Password = "password",
   None = "none",
@@ -35,6 +40,7 @@ export interface Args extends VsArgs {
   "cert-key"?: string
   "disable-telemetry"?: boolean
   "disable-update-check"?: boolean
+  enable?: string[]
   help?: boolean
   host?: string
   json?: boolean
@@ -128,6 +134,9 @@ const options: Options<Required<Args>> = {
       "Disable update check. Without this flag, code-server checks every 6 hours against the latest github release and \n" +
       "then notifies you once every week that a new release is available.",
   },
+  // --enable can be used to enable experimental features. These features
+  // provide no guarantees.
+  enable: { type: "string[]" },
   help: { type: "boolean", short: "h", description: "Show this output." },
   json: { type: "boolean" },
   open: { type: "boolean", description: "Open in browser on startup. Does not work remotely." },
--- a/test/unit/health.test.ts
+++ b/test/unit/health.test.ts
@@ -12,7 +12,7 @@ describe("health", () => {
   })
 
   it("/healthz", async () => {
-    ;[, , codeServer] = await integration.setup(["--auth=none"], "")
+    codeServer = await integration.setup(["--auth=none"], "")
     const resp = await codeServer.fetch("/healthz")
     expect(resp.status).toBe(200)
     const json = await resp.json()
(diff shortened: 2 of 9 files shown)


jsjoeio:
> increase coverage by 0.12%
👏 You + Codecov helping us build a habit of increasing code coverage!

src/node/cli.ts
```diff
@@ -128,6 +129,7 @@ const options: Options<Required<Args>> = {
       "Disable update check. Without this flag, code-server checks every 6 hours against the latest github release and \n" +
       "then notifies you once every week that a new release is available.",
   },
+  enable: { type: "string[]" }, // There is no support for this flag.
```
jsjoeio, line 132: ❓ How come we say "There is no support for this flag?" is that to imply that it's not supported for code-server users?

src/node/feature.ts
```diff
@@ -0,0 +1,29 @@
...
+  return false
+}
+
+/**
+ * Return true if the feature is enabled. Features are always enabled in
+ * development mode to ensure we are testing them.
+ */
+export function isFeatureEnabled(feature: Feature) {
+  return enabledFeatures.includes(feature)
+}
```
jsjoeio, line 29: This looks great and ++ for the documentation! 🎉 

nit: we may want to modify these functions to take `enabledFeatures` as an arg. That way, they're not relying and a global value, which could lead to unintended consequences later. Plus, it helps with testing.

jsjoeio (changes requested):
Overall, great work! 🎉 Just a couple nits about function being pure

src/node/cli.ts
code-asher: Ah actually idk if we should keep this comment. I guess what I mean is actually "no guarantee these features will work".

jsjoeio: Ahhh gotcha. Maybe change it to that instead? your call!

src/node/feature.ts
code-asher: Hmmm, yeah if we made `enabledFeatures`  immutable then we'd get purity.

Buuuut having to pass it around everywhere we need it seems a bit tiresome. :stuck_out_tongue: 

It also seems to create a bit of a mismatch. Feature flags are global settings but if you can pass them in it creates the illusion that they aren't. So the use and the design don't quite match up.

What kind of unintended consequences you were thinking of?

code-asher: Actually...this is all entirely unnecessary lol. I'll just leave the enum then a caller can do `req.args.enable.includes(Feature.MyFeature)`.

src/node/cli.ts
jawnsy: Maybe we could change the "Enabling features" info message to say something like:

> Enabling the following experimental features:

and at the bottom, something like:

> The code-server project does not provide stability guarantees or commit to fixing bugs relating to these experimental features. When filing bug reports, please ensure that you can reproduce the bug with all experimental features turned off.

code-asher: That's some premium copy :+1: Thanks! Will add.

code-asher:
Refactored; it's not any more pure but it's simpler. I think we could have an interesting discussion on whether we should pass `args` (and other variables) around instead of using a global access pattern at some point.

Coverage went down because of the code added in entry so I split some of those functions out in order to test them. One of them is now used for all the integration tests so those should reflect reality more closely now.

test/unit/health.test.ts
```diff
@@ -12,15 +12,15 @@ describe("health", () => {
   })
 
   it("/healthz", async () => {
-    ;[, , codeServer] = await integration.setup(["--auth=none"], "")
+    codeServer = await integration.setup(["--auth=none"], "")
```
jsjoeio, line 15: this looks way cleaner/easier to read 👏

jsjoeio (approved):
Great work 👏

jsjoeio:
> I think we could have an interesting discussion on whether we should pass args (and other variables) around instead of using a global access pattern at some point

Agreed! I prefer passing args because it's easier to mock/test things. But global access does come in handy when writing code. It would be a good discussion though.

jsjoeio:
I'm going to fix the trivy scan in a PR soon so you can ignore for now

test/unit/health.test.ts
jawnsy: :+1: yea, the code before was :raised_eyebrow: kinda inscrutable, but then again, I'm not a JavaScript/TypeScript person

PR #3288, opened 2021-05-04
docs(maintaining): add versioning

jsjoeio:
This PR updates `MAINTAINING.md` with a section for "Versioning."

### DIFF
--- a/docs/MAINTAINING.md
+++ b/docs/MAINTAINING.md
@@ -7,6 +7,7 @@
     - [Milestones](#milestones)
     - [Triage](#triage)
     - [Project Boards](#project-boards)
+  - [Versioning](#versioning)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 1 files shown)


docs/MAINTAINING.md
```diff
@@ -56,3 +57,14 @@ We use project boards for projects or goals that span multiple milestones.
...
+## Versioning
+
+`<major.minor.patch>`
+
+Because code-server is an application and not a library or framework, traditional [semver](https://semver.org/) doesn't really make sense for the project.
+
+Instead, we default to patch updates. We only use minor updates when:
+
+- VS Code bumps minor
+- we update our plugin API
```
jsjoeio, line 70:
```suggestion
- We update our plugin API
```

docs/MAINTAINING.md
```diff
@@ -56,3 +57,14 @@ We use project boards for projects or goals that span multiple milestones.
 Think of this as a place to put miscellaneous things (like testing, clean up stuff, etc). As a maintainer, random todos may come up here and there. This gives you a place to add notes temporarily before opening a new issue. Given that our release milestones function off of issues, we believe tasks should have dedicated issues.
 
 It also gives us a way to separate the issue triage from bigger-picture, long-term work.
+
+## Versioning
+
+`<major.minor.patch>`
+
+Because code-server is an application and not a library or framework, traditional [semver](https://semver.org/) doesn't really make sense for the project.
```
jawnsy, line 65: The first line doesn't really seem necessary to me; we're still following semantic versioning, just in a way that corresponds to Visual Studio Code upstream, right? I would propose a phrasing like:

> The code-server project follows traditional [semantic versioning](https://semver.org), with the objective of minimizing major changes that break backward compatibility. We increment the patch level for all releases, except when the upstream Visual Studio Code project increments its minor version or we change the plugin API in a backward-compatible manner. In those cases, we increment the minor version rather than the patch level.

jawnsy (approved):

docs/MAINTAINING.md
jsjoeio: You have a way with words! That sounds more straightforward and elegant 🕺

PR #3290, opened 2021-05-04
feat(testing): add test for src/node/constants.ts

jsjoeio:
This PR adds additional testing for `src/node/constants.ts`, specifically checking that both `commit` and `version` return the correct values when defined in and missing from `package.json`.

- update `coveragePathIgnorePatterns` to `/out` (so that we collect coverage for `src/node/routes`
- add tests for `src/node/constants.ts`

## Screenshot
image

### DIFF
--- a/test/unit/constants.test.ts
+++ b/test/unit/constants.test.ts
@@ -1,5 +1,4 @@
 import * as fs from "fs"
-import { commit, getPackageJson, version } from "../../src/node/constants"
 import { tmpdir } from "../../test/utils/constants"
 import { loggerModule } from "../utils/helpers"
 
@@ -8,12 +7,14 @@ jest.mock("@coder/logger", () => require("../utils/helpers").loggerModule)
 
 describe("constants", () => {
   describe("getPackageJson", () => {
+    const { getPackageJson } = require("../../src/node/constants")
     afterEach(() => {
       jest.clearAllMocks()
     })
 
     afterAll(() => {
       jest.restoreAllMocks()
+      jest.resetModules()
     })
 
     it("should log a warning if package.json not found", () => {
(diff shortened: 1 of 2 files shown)


jsjoeio:
Codecov: "Merging #3290 (027106a) into main (af5a1c9) will not change coverage."

Me: 
Alt Text

test/unit/constants.test.ts
```diff
@@ -36,20 +37,95 @@ describe("constants", () => {
...
-      expect(version).not.toBe(null)
-      expect(isValidSemVar).toBe(true)
+    describe("with package.json.version defined", () => {
+      let mockPackageJson = {
+        name: "mock-code-server",
+        version: "1.0.0",
+      }
+      let version = ""
+
+      beforeEach(() => {
```
code-asher, line 47: NBD but all these `beforeEach`  and `afterEach` blocks are either identical or nearly identical so thoughts on combining them and moving into a common describe block above? Would make adding future tests a bit smoother I think.

To combine the nearly identical ones we could do 
```
constants = require("../../src/node/constants")
```

And then the test uses whichever property it's trying to test.

code-asher (approved):

test/unit/constants.test.ts
jsjoeio: It did feel a bit repetitive 😅 I didn't bother thinking of a better solution while writing it but I like the idea of a common describe block above! Let me give that a try and see how it works 🚀

jsjoeio: Fixed!

test/unit/constants.test.ts
```diff
@@ -1,54 +1,96 @@
...
-      expect(packageJson.repository).toBe("https://github.com/cdr/code-server")
-    })
-  })
-  describe("version", () => {
     it("should return the package.json version", () => {
       // Source: https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39#gistcomment-2896416
       const validSemVar = new RegExp("^(0|[1-9]d*).(0|[1-9]d*).(0|[1-9]d*)")
       const isValidSemVar = validSemVar.test(version)
       expect(version).not.toBe(null)
       expect(isValidSemVar).toBe(true)
```
code-asher, line 46: I think we can drop this because now that we provide `1.0.0` as part of the test we already know it'll pass.

test/unit/constants.test.ts
```diff
@@ -1,54 +1,96 @@
...
+      expect(version).toBe("1.0.0")
```
code-asher, line 47:
```suggestion
      expect(version).toBe(mockPackageJson.version)
```

test/unit/constants.test.ts
```diff
@@ -1,54 +1,96 @@
...
 
-    it("should log a warning if package.json not found", () => {
-      const expectedErrorMessage = "Cannot find module './package.json' from 'src/node/constants.ts'"
-
-      getPackageJson("./package.json")
-
-      expect(loggerModule.logger.warn).toHaveBeenCalled()
-      expect(loggerModule.logger.warn).toHaveBeenCalledWith(expectedErrorMessage)
+    it("should provide the commit", () => {
+      // Source: https://gist.github.com/jhorsman/62eeea161a13b80e39f5249281e17c39#gistcomment-2896416
```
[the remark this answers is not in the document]
jsjoeio: moving!

test/unit/constants.test.ts
code-asher: There are a few places we could do this, I'm not sure if it's better than just writing out the string though.

PR #3297, opened 2021-05-05
Test static route

code-asher:
Mostly just wanted to add a test for static auth.

### DIFF
--- a/test/unit/helpers.test.ts
+++ b/test/unit/helpers.test.ts
@@ -1,0 +1,14 @@
+import { promises as fs } from "fs"
+import { tmpdir } from "../../test/utils/helpers"
+
+/**
+ * This file is for testing test helpers (not core code).
+ */
+describe("test helpers", () => {
+  it("should return a temp directory", async () => {
+    const testName = "temp-dir"
+    const pathToTempDir = await tmpdir(testName)
+    expect(pathToTempDir).toContain(testName)
+    expect(fs.access(pathToTempDir)).resolves.toStrictEqual(undefined)
+  })
+})
--- a/test/unit/routes/static.test.ts
+++ b/test/unit/routes/static.test.ts
@@ -1,0 +1,136 @@
+import { promises as fs } from "fs"
+import * as path from "path"
+import { tmpdir } from "../../utils/helpers"
+import * as httpserver from "../../utils/httpserver"
+import * as integration from "../../utils/integration"
+
+describe("/static", () => {
+  let _codeServer: httpserver.HttpServer | undefined
+  function codeServer(): httpserver.HttpServer {
+    if (!_codeServer) {
+      throw new Error("tried to use code-server before setting it up")
+    }
+    return _codeServer
+  }
+
+  let testFile: string | undefined
+  let testFileContent: string | undefined
+  let nonExistentTestFile: string | undefined
+
+  // The static endpoint expects a commit and then the full path of the file.
+  // The commit is just for cache busting so we can use anything we want. `-`
+  // and `development` are specially recognized in that they will cause the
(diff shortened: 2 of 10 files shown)


test/unit/helpers.test.ts
```diff
@@ -0,0 +1,14 @@
+import { promises as fs } from "fs"
+import { tmpdir } from "../../test/utils/helpers"
+
+/**
+ * This file is for testing test helpers (not core code).
+ */
```
jsjoeio, line 6: I like how you separated this and added a comment 👏

test/unit/routes/static.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    codeServer = await integration.setup(["--auth=none"], "")
+    let resp = await codeServer.fetch("/static")
+    expect(resp.status).toBe(404)
+
+    resp = await codeServer.fetch(`/static/-${testFile}`)
+    expect(resp.status).toBe(200)
+
+    const content = await resp.text()
+    expect(content).toStrictEqual(testFileContent)
+  })
```
jsjoeio, line 36: question: a bit confused here. I know the test is testing the `/static` route when unuathed but why do we expect a `404` when you fetch the route and then a `200` when fetching a file?

And why the `/-file` instead of `/file for `/static/-${testFile}?

test/unit/routes/static.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  it("/static authed", async () => {
```
jsjoeio, line 38: nit: testname should explain what the test is testing for -> 
```suggestion
  it("should return a 401 if authentication is enabled and user requests files", async () => {
```

test/unit/routes/static.test.ts
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  afterEach(async () => {
+    if (codeServer) {
+      await codeServer.close()
+      codeServer = undefined
+    }
+  })
+
+  it("/static unauthed", async () => {
```
jsjoeio, line 26: nit: test should explain what the test is testing for ->
```suggestion
  it("should return a 200 if not authed and requesting files", async () => {
```

test/unit/routes/static.test.ts
```diff
@@ -0,0 +1,47 @@
+import { promises as fs } from "fs"
+import * as path from "path"
+import { tmpdir } from "../../utils/helpers"
+import * as httpserver from "../../utils/httpserver"
+import * as integration from "../../utils/integration"
+
+describe("static", () => {
```
jsjoeio, line 7:
```suggestion
describe("/static", () => {
```

jsjoeio (changes requested):
Looks great! 🚀 

Mostly nits around test names and then a question

test/unit/routes/static.test.ts
code-asher: The static endpoint expects a commit (for cache purposes, in this case I'm just using `-`) and then the full path of the file to serve.

So anything other than that will result in a 404 (since no path means no file). We could also do `/static/-`, `/static/-/some/path/that/probably/doesnot/exist`, etc.

jsjoeio: Oh TIL. I didn't know how it worked so thanks for explaining that!

> We could also do...

Your call!

code-asher:
I need to refactor the logger module mock because it causes errors in the e2e tests but that conflicts with some constant test changes you're making so I'm going to wait for that to get merged then rebase.

jsjoeio:
> I'm going to wait for that to get merged then rebase.

jsjoeio (approved):
Nice work!

PR #3298, opened 2021-05-06
refactor(ci): split audit from prebuild

oxy:
Move dependency audits from prebuild to their own jobs, so that an error does not affect the rest of the build/test process.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,10 +43,6 @@ jobs:
         if: steps.cache-yarn.outputs.cache-hit != 'true'
         run: yarn --frozen-lockfile
 
-      - name: Audit for vulnerabilities
-        run: yarn _audit
-        if: success()
-
       - name: Run yarn fmt
         run: yarn fmt
         if: success()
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -59,10 +55,60 @@ jobs:
...
+        id: cache-yarn
+        uses: actions/cache@v2
+        with:
+          path: "**/node_modules"
+          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install dependencies
+        if: steps.cache-yarn.outputs.cache-hit != 'true'
+        run: yarn --frozen-lockfile
+
```
jsjoeio, line 80: The codecov upload relies on the coverage being generated, which happens when we run the unit tests. So if this is a fresh checkout of the repo, you may need to add a step after install dependencies and before upload code coverage to run the unit tests.

jsjoeio (changes requested):
One minor change (I think) but otherwise, this looks way better 🙌

.github/workflows/ci.yaml
oxy: Ah, I'll revert that change if it depends on the unit tests in the first place.

jsjoeio: Feel free to add a comment too (apologies, I should have done that when I added it!)

oxy: In hindsight it makes perfect sense that you need to run tests to know what coverage for the tests is like - I just didn't notice in a hurry, heh. Pushed before I could add a comment, fingers crossed its fine.

jsjoeio (approved):
Good work!

PR #3330, opened 2021-05-09
Check the logged user instead of $USER

### LINKED ISSUE #2767: Cannot start stopped docker container, user 'coder' does not exist
- Web Browser:
- Local OS: MacOS
- Remote OS: Ubuntu 18.04.5 LTS (GNU/Linux 4.15.0-123-generic x86_64)
- Remote Architecture:
- `code-server --version`: 3.9.0

I set up code-server with Docker as indicated in https://hub.docker.com/r/codercom/code-server

```
mkdir -p ~/.config
docker run -it --name code-server -p 127.0.0.1:8080:8080 \
  -v "$HOME/.config:/home/coder/.config" \
  -v "$PWD:/home/coder/project" \
  -u "$(id -u):$(id -g)" \
  -e "DOCKER_USER=$USER" \
  codercom/code-server:latest
```

The initial run works perfectly. I then stop the container using ctrl+c. When I try to start the container again using `docker start code-server` the container stops immediately. Running `docker logs code-server` outputs:

```
[2021-02-25T14:46:34.514Z] info  code-server 3.9.0 fc6d123da59a4e5a675ac8e080f66e032ba01a1b
[2021-02-25T14:46:34.516Z] info  Using user-data-dir ~/.local/share/code-server
[2021-02-25T14:46:34.537Z] info  Using config file ~/.config/code-server/config.yaml
[2021-02-25T14:46:34.538Z] info  HTTP server listening on http://0.0.0.0:8080 
[2021-02-25T14:46:34.538Z] info    - Authentication is enabled
[2021-02-25T14:46:34.538Z] info      - Using password from ~/.config/code-server/config.yaml
[2021-02-25T14:46:34.538Z] info    - Not serving HTTPS 
^Cusermod: user 'coder' does not exist
```

This issue seems to be related to [#2267](https://github.com/cdr/code-server/issues/2267), but I am still getting the user error message.

videlanicolas:
Given that `sudo usermod --login "$DOCKER_USER" coder` and `sudo groupmod -n "$DOCKER_USER" coder` modify the container's disk it'll persist across restarts, but environment variables will be reset to whatever state they had at the end of `Dockerfile`. In this case, `$USER` is set to `coder`, so this branch will always be true.

By checking with the output of `whoami`, which gets it's information from `/etc/passwd`, we make sure to get the real logged user and not the one defined by $USER.

We also move `USER="$DOCKER_USER"` out of the branch, since we always want this to happen at entry-point. If we don't do this assignment, $USER will contain `coder` upon restart.

Fixes #2767.

### DIFF
--- a/ci/release-image/entrypoint.sh
+++ b/ci/release-image/entrypoint.sh
@@ -5,14 +5,14 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ] && [ "$DOCKER_USER" != "$USER" ]; then
+USER="$DOCKER_USER"
+
+if [ "${DOCKER_USER-}" ] && [ "$DOCKER_USER" != "$(whoami)" ]; then
   echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd >/dev/null
   # Unfortunately we cannot change $HOME as we cannot move any bind mounts
   # nor can we bind mount $HOME into a new home as that requires a privileged container.
   sudo usermod --login "$DOCKER_USER" coder
   sudo groupmod -n "$DOCKER_USER" coder
-
-  USER="$DOCKER_USER"
 
   sudo sed -i "/coder/d" /etc/sudoers.d/nopasswd
 fi
--- a/ci/release-image/entrypoint.sh
+++ b/ci/release-image/entrypoint.sh
@@ -5,16 +5,17 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-USER="$DOCKER_USER"
+if [ "${DOCKER_USER-}" ]; then
+  USER="$DOCKER_USER"
+  if [ "$DOCKER_USER" != "$(whoami)" ]; then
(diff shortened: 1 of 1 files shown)


ci/release-image/entrypoint.sh
```diff
@@ -5,15 +5,15 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ] && [ "$DOCKER_USER" != "$USER" ]; then
+USER="$DOCKER_USER"
```
code-asher, line 8: Should we put this behind an `if [ "${DOCKER_USER-}" ]` in case `DOCKER_USER` is not defined?

code-asher (changes requested):
Thank you for fixing this!! Your PR description was fantastic too.

jsjoeio:
You can ignore the trivy-scan failure. We're aware of that and working on a fix. As for the audit vulnerabilities, we're also aware of a postcss CVE that needs to be fixed. Everything else looks good!

ci/release-image/entrypoint.sh
videlanicolas: We can certainly do so, although entrypoint.sh should assume this env variable is defined when running the container. We should check it just in case it's not.

code-asher (approved):
:tada: :tada: :tada:

PR #3332, opened 2021-05-10
Fixes for Dependabot alerts (xmldom, underscore)

oxy:
Remove `plist`, since we don't use it (and it depends on xmldom, per Dependabot), and modify the resolution for `underscore` to be specific to doctoc.


package.json
```diff
@@ -76,7 +76,7 @@
   },
   "resolutions": {
     "normalize-package-data": "^3.0.0",
-    "underscore": "1.13.1",
+    "doctoc/underscore": "^1.13.1",
```
code-asher, line 79: Cool I didn't know you could do this!

code-asher (approved):

PR #3341, opened 2021-05-11
fix(ci): disable trivy-scan-repo

jsjoeio:
This PR disables `trivy-scan-repo` in CI since it's currently broken due to an upstream issue.


.github/workflows/ci.yaml
```diff
@@ -461,30 +461,29 @@ jobs:
...
-          format: "template"
-          template: "@/contrib/sarif.tpl"
-          output: "trivy-repo-results.sarif"
-          severity: "HIGH,CRITICAL"
-
-      - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@v1
-        with:
-          sarif_file: "trivy-repo-results.sarif"
+  # trivy-scan-repo:
```
jawnsy, line 470: if this was in its own workflow file, then you can disable the whole workflow in the UI :stuck_out_tongue_winking_eye: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

another way you can disable it that doesn't require commenting it all out, is to use `if: 1 == 2` or something in the job definition. We do that sometimes too https://github.com/cdr/m/blob/dba70b6a0c904e360dac0ed930028a0107ef2140/.github/workflows/nightly.yaml#L14

jawnsy (approved):

.github/workflows/ci.yaml
Atom1985: > if this was in its own workflow file, then you can disable the whole workflow in the UI :stuck_out_tongue_winking_eye: https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow
> 
> another way you can disable it that doesn't require commenting it all out, is to use `if: 1 == 2` or something in the job definition. We do that sometimes too https://github.com/cdr/m/blob/dba70b6a0c904e360dac0ed930028a0107ef2140/.github/workflows/nightly.yaml#L14

PR #3351, opened 2021-05-11
fix(lib/vscode): update to 1.56.1

oxy:


oxy:
@jsjoeio I think it might just be a one-time fluke because I've accumulated quite the mess of outdated branches and tags locally - I imagine deleting/recloning might help.

PR #3383, opened 2021-05-13
fix(lib/vscode): fix missing rimraf when running postinstall script

kenve:
Fix the error that `rimraf` is missing when running `yarn --production`.

```bash
$ node ./postinstall
internal/modules/cjs/loader.js:818
  throw err;
  ^

Error: Cannot find module 'rimraf'
```


code-asher (approved):
Thank you for finding and fixing this!

PR #3391, opened 2021-05-14
Add socket type to websocket connection query (regression)

code-asher:
This was added a long time ago but accidentally lost when we reverted
the retry changes to this file after one of the VS Code updates that
added their own retry (we reverted everything not realizing we had other
unrelated changes in here).

This is definitely one downside to the subtree over patches but I added 
a note to help ensure we notice the change in case something similar 
happens again.


jsjoeio (approved):
Looks good to me! Thanks for adding that note too

PR #3393, opened 2021-05-14
Introduce extraInitContainers to helm chart

### LINKED ISSUE #3392: Add extraInitContainers to helm chart
extraInitContainers, structured similarly to extraContainers would be a good addition, so that we could (for example) configure a programmatically installed extensions before the main container has started.

strowk:
New extraInitContainers configuration added.
It allows to pass template with a list of containers to execute before
main code-server container started. Main container will only start when
all init containers are  completed (exited with 0 code).

 Additionally changes the way extraContainers is used - instead of
 toYaml use tpl, because this allows to
 reference any values from extraContainers string.

close #3392


jsjoeio:
Thanks for opening an issue and raising a PR @strowk 🎉

This sounds like a great addition but would love to get some extra eyes on it from the other code-server team members. 

In the meantime, can you:
- rebase so your branch is up to date with `main`
- update the `CHANGELOG`

jsjoeio:
cc @alexgorbatchev and @Matthew-Beckett - any chance either of you would be able to help review this?

ci/helm-chart/README.md
```diff
@@ -65,6 +65,7 @@ and their default values.
 | extraArgs | list | `[]` |  |
 | extraConfigmapMounts | list | `[]` |  |
 | extraContainers | string | `""` |  |
+| extraInitContainers | string | `""` |  |
```
Matthew-Beckett, line 68: Should this type be string? All other values with plural names are type list.

ci/helm-chart/README.md
```diff
@@ -65,6 +65,7 @@ and their default values.
 | extraArgs | list | `[]` |  |
 | extraConfigmapMounts | list | `[]` |  |
 | extraContainers | string | `""` |  |
```
Matthew-Beckett, line 67: Again should this be type list? Looking at the whole file, you would need to define a list of objects which contain extra container properties.

ci/helm-chart/README.md
strowk: This is supposed to be a string, because it contains a template, which would be resolved and is supposed to contain yaml describing list of containers

Matthew-Beckett: Are you able to provide an example as a comment in the values.yank? Also please define this in the values.yaml as a null value string for readability sake. Sometimes people just look at the values and not the README so it might not be clear what values are available if not documented in values.yaml

ci/helm-chart/README.md
strowk: I think this is better to be a string, which is used as a template rather than a list of objects. It coudl theoretically be made like for example `extraConfigmapMounts` - as a list of objects. However I would not recommend it, firstly because container is a much more complicated structure than a confgmap mounts. Mounts only have couple of fields, which are unlikely to be much changed in the future. Container have a lot more stuff in there and trying to strictly cover all of stuff possible in there is too complicated (I will not be doing such a PR myself for sure) and would also have to be changed each time k8s introduces new stuff possible in the containers. Keeping this as a template (string, containing templated yaml) is much more flexible.

Matthew-Beckett: I agree, you're right. Can you add an example then we should be good to go?

ci/helm-chart/README.md
strowk: I kinda doubt that example would be readable. Since the value is supposed to be a template of a yaml, it would not look particularly good in an example inside of another yaml... I can make example in the README though.

Matthew-Beckett: I agree, this is a fairly advanced feature the majority of users don't have to extend themselves. A short example in the readme will suffice ☺️

strowk: Added example to [readme](https://github.com/cdr/code-server/pull/3393/files#diff-602747651030aa31de8873535c3f3757416bb752591bb8acd0401650e0891131R120)

strowk:
@jsjoeio , rebased on main and also updated changelog. I guess that next version should be `3.11.0`, since this is kinda new feature (even though only for helm chart)

strowk:
Upd: changed entry in changelog to "Next Version" for now, guess there could be more stuff in there before next release..

Matthew-Beckett:
@jsjoeio can you advise on the CHANGELOG before I approve? Not sure on the rules for adding changes being a part time maintainer. This pull has a milestone associated with it so is @strowk correct to put it in "Next Update"?

jsjoeio:
@strowk thanks for making those changes! Using "Next Version" in the `CHANGELOG` is correct (apologies, it's new and we need to add more documentation around that).

@Matthew-Beckett looks good from our side. Thanks so much for reviewing this so quickly and helping out with maintaining code-server! We really appreciate it.

jsjoeio (approved):

Matthew-Beckett:
@strowk nice work! @jsjoeio it's my pleasure, any time.

PR #3422, opened 2021-05-19
fix: use sufficient computational effort for password hash

### LINKED ISSUE #3381: Fix password hash to use sufficient computational effort
Issue identified by CodeQL in `src/node/util.ts`

See issue explanation and recommendation: https://github.com/cdr/code-server/security/code-scanning/20?query=ref%3Arefs%2Fheads%2Fmain

jsjoeio:
This PR modifies the underlying algorithm used in the `hash` function to use sufficient computational effort.

## Changes
- refactor to use `argon2` instead of `sha256`
- add unit tests for hash and helper functions

## Screenshots

Using a hashed-password (`sha256`)

https://user-images.githubusercontent.com/3806031/120695344-25804580-c460-11eb-9b7b-b948ebd3eda0.mov

Using a regular password (not hashed)

https://user-images.githubusercontent.com/3806031/120695764-9aec1600-c460-11eb-82e5-d428f7d32743.mov

Using a hashed-password (`argon2`)

https://user-images.githubusercontent.com/3806031/120861593-6483c900-c53c-11eb-904a-10016f8449b5.mov

## Checklist
- [x] tested locally
- [x] added new dependencies
- [x] added tests
- [x] updated `CHANGELOG.md`
- [x] ensure we still support hashed_password in the config which is hashed with sha256

Fixes #3381 

Follow-up: https://github.com/cdr/code-server/issues/3432

## Notes

Here is how authentication works in code-server:

image

<small>
  <a href="https://excalidraw.com/#json=6187060784267264,Oi0cJLvO_7QUYxFk-1ZTMQ">
  Link to Excalidraw
  </a>
</small>


ammario (approved):
bcrypt!

jsjoeio:
@oxy do you mind taking a look at this today or tomorrow?

package.json
```diff
@@ -85,6 +86,7 @@
   },
   "dependencies": {
     "@coder/logger": "1.1.16",
+    "bcrypt": "^5.0.1",
```
jawnsy, line 89: I don't think there's anything wrong with bcrypt, but was just curious, have we considered argon2? It's supposed to be a newer & better key derivation function: https://cryptobook.nakov.com/mac-and-key-derivation/argon2

jsjoeio: I went with the first recommendation from CodeQL (but I see argon2 is in there as well).

image

This is not my wheelhouse so easily convinced to switch to `argon2`. There's an npm package for it: https://www.npmjs.com/package/argon2

I'll let @oxy weigh in.

jsjoeio: It might also eliminate (or reduce work required) for https://github.com/cdr/code-server/issues/3432

src/node/routes/login.ts
```diff
@@ -74,12 +74,13 @@ router.post("/", async (req, res) => {
 
     if (
       req.args["hashed-password"]
-        ? safeCompare(hash(req.body.password), req.args["hashed-password"])
+        ? isHashLegacyMatch(req.body.password, req.args["hashed-password"])
         : req.args.password && safeCompare(req.body.password, req.args.password)
     ) {
+      const hashedPassword = req.args["hashed-password"] ? hashLegacy(req.body.password) : hash(req.body.password)
```
oxy, line 80: Don't we also have to use `isHashMatch` here too? I'm a little confused at what you're trying to do with both the if statement and the `const hashedPassword` variable - and we might need to take a second look at auth as a whole :sweat_smile:

oxy (changes requested):
`argon2` is a stronger algorithm but `bcrypt` is also alright - though I'm still a little confused about how you're handling authentication, what `hashedPassword` is used for with the cookie, and how/if we compare against bcrypt passwords in the auth route itself.

CHANGELOG.md
```diff
@@ -48,14 +48,15 @@ VS Code v0.00.0
 ### Documentation
 
 - docs: add Pomerium #3424 @desimone
-- docs: fix confusing sentence in pull requests section #3460 @shiv-tyagi
+- docs: fix confusing sentence in pull requests section #3460 @shiv-tyag
```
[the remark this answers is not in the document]
jsjoeio: nice catch! bad rebase

src/node/routes/login.ts
jsjoeio: IIRC, `safeCompare` was used to compare the hashed-password with the `req.body.password`. We use `isHashLegacyMatch` which uses `safeCompare` inside it I believe. This is for backwards-compatibility.

`const hashedPassword` is needed for backwards-compatibility as well. If they're using the legacy `sha256` hashed-password, then we need to use that as `hashedPassword` which gets stored as a cookie. If they use the new formula then we use `hash`.

Hope that makes sense! Let me know how I can make that more clear.

package.json
jsjoeio: Sounds like we should go with `argon2` then with 2 votes for it. I'm all for making it better and it shouldn't be too hard to switch out `bcrypt` for `argon2`

CHANGELOG.md
```diff
@@ -74,6 +75,10 @@ VS Code v1.56.1
 
 - fix(ci): update brew-bump.sh to update remote first #3438 @jsjoeio
 
+### Documentation
+
+- item
+
```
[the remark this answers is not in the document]
jsjoeio: Fixed

CHANGELOG.md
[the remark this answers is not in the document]
jsjoeio: fixed

package.json
jsjoeio: I spoke to soon 😅 `argon2` only supports async methods. The previous methods were sync. In order to use `argon2` we would have to rewrite a lot of code. I tried just now by making a bunch of functions `async` but it seems to have broken a lot of things 😅

So instead, I think we'll have to stay with `bcrypt` as part of this fix. But we can look into switching to `argon2` if we find `bcrypt` to not fulfill our needs. Sorry for the false hope!

jawnsy: Ahh things are never as easy as they seem lmao. I thought that `const hash = await argon2.hash("password");` would turn it into a sync operation?

jsjoeio:
Notes:

1. merge in bcrypt (this PR)
2. write tests for router (testing one level up from hashing) - make an issue
3. add token exchange -> make an issue
4. rewrite other parts if needed? -> add to 3

package.json
jsjoeio: I don't think the codebase is set up to use top-level `await` which means we have to use it in an `async` function, which means any function that calls it must also use `await` and also be converted to an `async` function unfortunately

jsjoeio: But I am take a second stab at refactoring to use `argon2` 🤞

jawnsy: Ahhhhh. Good luck Joe! And thanks for trying, regardless of the outcome, it seems tough :)

jsjoeio:
@oxy do you mind taking a look at the security warnings/errors flagged by CodeQL? I know we plan to fix most of these in a follow-up PR but want to make sure my code isn't introducing anything you would flag: https://github.com/cdr/code-server/pull/3422/checks?check_run_id=2749748946

src/node/routes/login.ts
```diff
@@ -62,24 +61,31 @@ router.get("/", async (req, res) => {
...
+      hashedPasswordFromArgs,
+      passwordFromRequestBody: password,
+      passwordFromArgs: req.args.password,
+    })
+
+    if (isPasswordValid) {
       // The hash does not add any actual security but we do it for
       // obfuscation purposes (and as a side effect it handles escaping).
-      res.cookie(Cookie.Key, hash(req.body.password), {
+      res.cookie(Cookie.Key, hashedPassword, {
```
jsjoeio, line 88: 👀 Major change

src/node/http.ts
```diff
@@ -57,20 +61,27 @@ export const ensureAuthenticated = (req: express.Request, _?: express.Response,
...
+      const hashedPasswordFromArgs = req.args["hashed-password"]
+      const passwordMethod = getPasswordMethod(hashedPasswordFromArgs)
+      const isCookieValidArgs: IsCookieValidArgs = {
+        passwordMethod,
+        cookieKey: req.cookies.key as string,
+        passwordFromArgs: req.args.password || "",
+        hashedPasswordFromArgs: req.args["hashed-password"],
+      }
+
+      return await isCookieValid(isCookieValidArgs)
```
jsjoeio, line 80: 👀 Major change

test/package.json
```diff
@@ -16,5 +17,6 @@
   },
   "resolutions": {
     "@playwright/test/playwright": "^1.11.0-next-alpha-apr-13-2021"
-  }
+  },
+  "dependencies": {}
```
jsjoeio, line 21: Please remove
```suggestion
```

typings/pluginapi.d.ts
```diff
@@ -145,12 +145,16 @@ export const proxy: ProxyServer
 /**
  * Middleware to ensure the user is authenticated. Throws if they are not.
  */
-export function ensureAuthenticated(req: express.Request, res?: express.Response, next?: express.NextFunction): void
+export function ensureAuthenticated(
+  req: express.Request,
+  res?: express.Response,
+  next?: express.NextFunction,
+): Promise<void>
```
jsjoeio, line 152: Only real change is making it return a `Promise`

nooobcoder:
@jsjoeio impressed to see how good you test code-server. Just awesome, man!

src/node/util.ts
```diff
@@ -115,10 +118,162 @@ export const generatePassword = async (length = 24): Promise<string> => {
...
+ * Used to determine the password method.
+ *
+ * There are three options for the return value:
+ * 1. "SHA256" -> the legacy hashing algorithm
+ * 2. "ARGON2" -> the newest hashing algorithm
+ * 3. "PLAIN_TEXT" -> regular ol' password with no hashing
+ *
+ * @returns {PasswordMethod} "SHA256" | "ARGON2" | "PLAIN_TEXT"
+ */
+export function getPasswordMethod(hashedPassword: string | undefined): PasswordMethod {
```
jawnsy, line 178: I'm not sure where in the code to put it, but it would be nice to issue a warning if someone is using the plain text/cookie/sha256 methods to tell them that their authentication method is insecure, and to consider upgrading to argon2.

Feel free to do that as a separate PR, though!

jawnsy (approved):
Awesome work, Joe!!

src/node/cli.ts
```diff
@@ -270,9 +283,11 @@ export const parse = (
       let key: keyof Args | undefined
       let value: string | undefined
       if (arg.startsWith("--")) {
-        const split = arg.replace(/^--/, "").split("=", 2)
+        const split = splitOnFirstEquals(arg.replace(/^--/, ""))
         key = split[0] as keyof Args
         value = split[1]
+
+        console.log(`Hello key: ${key}, and value: ${value}`)
```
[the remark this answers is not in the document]
jsjoeio: Whoops! Thanks for flagging that :)

src/node/cli.ts
```diff
@@ -240,6 +240,19 @@ export const optionDescriptions = (): string[] => {
...
 
+export function splitOnFirstEquals(str: string): string[] {
+  // we use regex instead of "=" to ensure we split at the first
+  // "=" and return the following substring with it
+  // important for the hashed-password which looks like this
+  // $argon2i$v=19$m=4096,t=3,p=1$0qR/o+0t00hsbJFQCKSfdQ$oFcM4rL6o+B7oxpuA4qlXubypbBPsf+8L531U7P9HYY
+  // 2 means return two items
+  // Source: https://stackoverflow.com/a/4607799/3015595
+  // We use the ? to say the the substr after the = is optional
+  const split = str.split(/=(.+)?/, 2)
```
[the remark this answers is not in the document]
jsjoeio: I am probably misunderstanding what you mean but I don't think that is what we want. (though I agree string splits are hard to read). 

We need the value to be split on the first equals and I don't think using a regex match would give us the same result (based on this quick test). 

image

> If separator is a regular expression with capturing parentheses, then each time separator matches, the results (including any undefined results) of the capturing parentheses are spliced into the output array.

[MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#Capturing_parentheses)

src/node/util.ts
jsjoeio: Oh man, that is a fantastic idea! Then it would inform the ones using the sha256 that a better method exists! I'll do this in a follow-up PR 🙌 Thanks for the suggestion!

jsjoeio: https://github.com/cdr/code-server/issues/3560

oxy (changes requested):
Haven't had the time to test this in much detail, but CI is failing right now since `argon2` uses a Linux build that is too new for CentOS 7 (which we use and support right now);

To fix, we'd need to add `npm_config_build_from_source=true` as an environment variable, likely in `npm-postinstall.sh` and in `build-standalone-release.sh` (both in `ci/build/`). We might also want to add this to the yarn/npm docs, in case people are using `yarn global add` on distros older than Ubuntu 18.04.

There's an issue upstream about RHEL7/CentOS 7 compatibility for the official binary builds, but it appears to be stalled: https://github.com/ranisalt/node-argon2/issues/276

Haven't taken a look at the rest of the PR yet, but will do once we get CI working with CentOS 7 again!

jsjoeio:
Thanks so much for digging into this, finding the root and guiding me towards the solution! ♥️

src/node/cli.ts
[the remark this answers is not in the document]
jawnsy: Ahhh, I see, I misunderstood what this code is doing. I guess it's because we wrote our own argument parser, and now need a way to parse: `--hashed-password=abc=def` into `key` = `hashed-password` and `value` = `abc=def`

Is there some reason we're implementing our own parser rather than [relying on a library like `yargs`](https://nodejs.org/en/knowledge/command-line/how-to-parse-command-line-arguments/)? That's too big of a change for this PR, but maybe we can consider adopting a library in the future.

src/node/util.ts
```diff
@@ -115,10 +118,162 @@ export const generatePassword = async (length = 24): Promise<string> => {
...
+    return ""
+  }
+}
+
+/**
+ * Used to verify if the password matches the hash
+ */
+export const isHashMatch = async (password: string, hash: string) => {
+  try {
+    return await argon2.verify(hash, password)
```
RA80533, line 138: :warning: `password` has gotten quite deep without any actual validation of its shape. It’s was type asserted a few calls up. It could be an object with a prototype pollution if the Express handler didn’t perform any sanitization of that.

src/node/http.ts
```diff
@@ -57,20 +61,27 @@ export const ensureAuthenticated = (req: express.Request, _?: express.Response,
...
-        (req.args["hashed-password"]
-          ? safeCompare(req.cookies.key, req.args["hashed-password"])
-          : req.args.password && safeCompare(req.cookies.key, hash(req.args.password)))
-      )
-    default:
+      const hashedPasswordFromArgs = req.args["hashed-password"]
+      const passwordMethod = getPasswordMethod(hashedPasswordFromArgs)
+      const isCookieValidArgs: IsCookieValidArgs = {
+        passwordMethod,
+        cookieKey: req.cookies.key as string,
```
RA80533, line 75: :exclamation:

src/node/util.ts
jsjoeio: That's a really good point! Thanks for flagging that. I guess I probably want to write a function that sanitizes the password somewhere before we get to this point. I'll look into this (suggestions welcome too).

src/node/http.ts
jsjoeio: There's the type assertion you were referring to right? I guess we'll want to sanitize here as well?

RA80533: This was the one but there might be others. Aside from type assertions being misleading further down the chain, I believe the express handlers are too permissive in their typings.

jsjoeio:
*Writes a bunch of new tests 

Them: image

Me:

RA80533:
Express handlers are typed to be far too permissive towards the data it passes to the consumer. Most sources of data are typed as `any` which is problematic in most cases. Two solutions of note are:

1. Adopt a different web framework. Express is quite old and hasn't seen any new light in a while. Its typings are suboptimal.
2. Stick with Express and do either:
  1. Override the definitions on the types that are too permissive. This can be achieved through a few different ways:
    1. Override as much of the types as possible through the mechanisms with which TypeScript allows
    2. Patch the offending types (e.g., `git diff`, etc.)
  2. Attach some handlers to attempt to sanitize the data being sent through the route handlers down the line. This is where prototype pollution can be taken care of (Express does not implement anything related to this).
  3. Be extra careful. This is not ideal because it requires vigilance on the part of everyone who interacts with the codebase in a way that touches the problematic areas.

src/node/http.ts
jsjoeio: That's fair! And probably something we'll tackle later. I'll make a note in the Roadmap for us to look into it in the future.

src/node/util.ts
jsjoeio: I added some basic sanitization up the chain for when we use `req.body.password` and `req.cookie.key` which should address this.

jsjoeio:
> Express handlers are typed to be far too permissive towards the data it passes to the consumer. Most sources of data are typed as `any` which is problematic in most cases. Two solutions of note are:
> 
> 1. Adopt a different web framework. Express is quite old and hasn't seen any new light in a while. Its typings are suboptimal.
> 2. Stick with Express and do either:
> 3. Override the definitions on the types that are too permissive. This can be achieved through a few different ways:
>    1. Override as much of the types as possible through the mechanisms with which TypeScript allows
>    2. Patch the offending types (e.g., `git diff`, etc.)
> 4. Attach some handlers to attempt to sanitize the data being sent through the route handlers down the line. This is where prototype pollution can be taken care of (Express does not implement anything related to this).
> 5. Be extra careful. This is not ideal because it requires vigilance on the part of everyone who interacts with the codebase in a way that touches the problematic areas.

Those are really good points and definitely something we want to consider moving forward. I made a note about this in the Roadmap. I don't think we're in a place to switch away from Express but something we can consider when we look at cleaning up and refactoring things down the road.

For now, I've added a basic function to sanitize both the password and cookie before they're used by the password hashing functions. Thanks a bunch for your input and sharing your knowledge! We really appreciate it.

jsjoeio:
Fixed security vulnerabilities in separate PR: https://github.com/cdr/code-server/pull/3563

src/node/http.ts
```diff
@@ -57,20 +61,27 @@ export const ensureAuthenticated = (req: express.Request, _?: express.Response,
...
+      const hashedPasswordFromArgs = req.args["hashed-password"]
+      const passwordMethod = getPasswordMethod(hashedPasswordFromArgs)
+      const isCookieValidArgs: IsCookieValidArgs = {
+        passwordMethod,
+        cookieKey: sanitizeString(req.cookies.key),
+        passwordFromArgs: req.args.password || "",
+        hashedPasswordFromArgs: req.args["hashed-password"],
+      }
+
+      return await isCookieValid(isCookieValidArgs)
```
oxy, line 80: Is there a reason we're passing the cookie stuff into a separate function instead of handling it in `authenticated`? All paths seem to lead here, so I don't know why we folded that out into a separate function,

src/node/routes/login.ts
```diff
@@ -62,24 +61,31 @@ router.get("/", async (req, res) => {
...
```
oxy, line 88: While hashing is a major step forward, the problem with this approach is that it still allows attackers who have access to the hash to just submit it as-is and gain access to code-server - effectively not very different from storing it in plaintext.

We should definitely look into replacing this with something more robust so that password hashing isn't just a placebo - now or in a future version is up to you, but IMO we should tackle this before asking users to upgrade to Argon2.

src/node/http.ts
jsjoeio: There are definitely multiple approaches one could take for this.

I went with this approach because it's:
- testable (rather than having to mock requests and test `authenticated`, though we will get there at one point)
- easy to read -> `isCookieValid` makes it very clear what the functionality is and what we're looking for
- maintainable -> should we move to a token approach, we can make changes to the underlying function and test in isolation

I completely understand if your style is to not extract things into separate functions. 

I will take this comment as a `nit` since it doesn't seem like there is much we're discussing besides the style/approach I took to solving the problem (unless I'm missing something).

src/node/routes/login.ts
jsjoeio: > effectively not very different from storing it in plaintext.

I'm really glad you mention that and agree.

> We should definitely look into replacing this with something more robust so that password hashing isn't just a placebo

Also agree! 

> now or in a future version is up to you, but IMO we should tackle this before asking users to upgrade to Argon2

I would argue that it falls out of the scope of this PR. If we close this PR without merging, we're in the same situation we're discussing right now (see [code](https://github.com/cdr/code-server/blob/main/src/node/routes/login.ts#L82-L86)) because we're doing the same thing - hashing the password and storing as a cookie. 

The scope of this PR was to replace the hashing algorithm (not to fix the problem with attackers who have the hash).

Therefore, I think it makes sense to separate the two concerns and tackle them in separate PRs. I will create a new issue to track that though!

jsjoeio: https://github.com/cdr/code-server/issues/3576

src/node/http.ts
code-asher: By the way we have login tests that make real requests so testing that should be pretty easy!

jsjoeio: Oh fantastic! I didn't even realize. That will be easy then!

src/node/util.ts
```diff
@@ -115,10 +118,176 @@ export const generatePassword = async (length = 24): Promise<string> => {
...
+  passwordMethod: PasswordMethod
+  cookieKey: string
+  hashedPasswordFromArgs: string | undefined
+  passwordFromArgs: string | undefined
+}
+
+/** Checks if a req.cookies.key is valid using the PasswordMethod */
+export async function isCookieValid(isCookieValidArgs: IsCookieValidArgs): Promise<boolean> {
+  let isValid = false
+  const { passwordFromArgs = "", cookieKey, hashedPasswordFromArgs = "" } = isCookieValidArgs
```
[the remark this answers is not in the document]
jsjoeio: oh good idea! that might be easier to read too.

> lol I just realized this pattern reminds me a lot of React props

lolol you could say React has left an imprint on me 😅

src/node/cli.ts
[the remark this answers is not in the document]
jsjoeio: > Is there some reason we're implementing our own parser rather than relying on a library like yargs?

Fantastic question! All I can say is "legacy" 😛

But in all seriousness, I agree with you. Too big for this PR but something we want to do. I know @oxy has mentioned it multiple times. Hoping we get to it eventually!

src/node/routes/login.ts
RA80533: Totally get it. Sometimes PRs bring in a lot of heat just because they touch an area that would otherwise have a lot of attention drawn towards. This one specifically has allowed us to become aware of other things in the codebase we want to change.

:pray:

src/node/cli.ts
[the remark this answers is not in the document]
code-asher: https://github.com/cdr/code-server/issues/1908

oxy (approved):
All good! Thanks for sticking through this :heart:

jsjoeio:
> Thanks for sticking through this

Thanks for sharing your security knowledge with me! I feel like I'm slowing becoming more mindful of it thanks to you 😂

PR #3424, opened 2021-05-20
docs: add Pomerium

desimone:
RE: https://github.com/pomerium/pomerium/pull/2223

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -183,7 +183,7 @@ code-server only supports password authentication natively.
 **note**: code-server will rate limit password authentication attempts at 2 a minute and 12 an hour.
 
 If you want to use external authentication (i.e sign in with Google) you should handle this
-with a reverse proxy using something like [oauth2_proxy](https://github.com/pusher/oauth2_proxy)
+with a reverse proxy using something like [Pomerium](https://www.pomerium.io/guides/vs-code-server.html), [oauth2_proxy](https://github.com/pusher/oauth2_proxy)
 or [Cloudflare Access](https://teams.cloudflare.com/access).
 
 For HTTPS, you can use a self signed certificate by passing in just `--cert` or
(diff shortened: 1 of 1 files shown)


docs/FAQ.md
```diff
@@ -183,7 +183,7 @@ code-server only supports password authentication natively.
 **note**: code-server will rate limit password authentication attempts at 2 a minute and 12 an hour.
 
 If you want to use external authentication (i.e sign in with Google) you should handle this
-with a reverse proxy using something like [oauth2_proxy](https://github.com/pusher/oauth2_proxy)
+with a reverse proxy using something like [Pomerium](https://www.pomerium.io/guides/vs-code-server.html), [oauth2_proxy](https://github.com/pusher/oauth2_proxy)
```
jsjoeio, line 186: Will the URL need to change once https://github.com/pomerium/pomerium/pull/2223 is merged?
```suggestion
with a reverse proxy using something like [Pomerium](https://www.pomerium.io/guides/code-server.html), [oauth2_proxy](https://github.com/pusher/oauth2_proxy)
```

jsjoeio (changes requested):
Thanks for adding this! 🙌

One minor issue with URL (I think)

docs/FAQ.md
desimone: Actually -- I think the name of the file would have to change as well. And we should add a redirect so it doesn't break anyone's bookmarks. 

- https://github.com/pomerium/pomerium/blob/master/docs/.vuepress/config.js#L170
- https://github.com/pomerium/pomerium/blob/master/docs/.vuepress/public/_redirects#L33

or we could keep the original uri and just add a redirect. LMK

jsjoeio: Oh! Good point. Okay, I'll make those changes in my PR

jsjoeio: Made the changes! Thanks for the guidance!

jsjoeio (approved):
LGTM! Thanks so much @desimone 👍🏼

jsjoeio:
If you want to update the `CHANGELOG` with your change under "New Version" > "Documentation", then we'll make sure to credit you in the Release Notes for this update. Not required though :)

PR #3466, opened 2021-05-24
fix: remove CHANGELOG from fmt.sh and remove toc from CHANGELOG

jsjoeio:
This PR removes `CHANGELOG` from `fmt.sh` and removes the table of contents from `CHANGELOG`. 

Why?

It causes more headache than value. We don't need a table of contents in the `CHANGELOG` and we shouldn't make contributors run `yarn fmt` after updating the `CHANGELOG`.

## Checklist

- [ ] updated `CHANGELOG.md`


oxy (approved):
Made some changes to merge since Joe is on vacay, merging :eyes:

PR #3468, opened 2021-05-24
fix(publish): update cdrci fork in brew-bump.sh

### LINKED ISSUE #3444: Fix homebrew step in publish workflow
Something went wrong when we tried to update cdrci's fork: https://github.com/cdr/code-server/runs/2643613944?check_suite_focus=true

We need to figure out how to do that correctly. 

This might help: https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token#using-a-token-on-the-command-line

`echo "cdr\n$token" | git push` might work as well.

jsjoeio:
This PR fixes the homebrew `brew-bump.sh` step which runs when we publish a new release.

### Problem

When we publish a new release, we run a script to bump the formula for code-server on homebrew. The script works using `cdrci`'s GitHub token (secret in repo) so that the PR comes from them and not a maintainer.

The problem is `bump-formula-pr` uses the user's `homebrew-core` fork.

This means we need to ensure the fork is up-to-date.

To do that, we need to pull the remote and push the changes to the fork before running `bump-formula-pr`.

We tried to see if it would "just work" by pulling and pushing, but the push command asks for a username and password.

### Solution

You can get around this with `--all`. We found this solution [here](https://serverfault.com/a/912788).

> This has the advantage that the password won't be visible in the process list too.

## Changes
- update `brew-bump.sh`

## Screenshots

Tested locally and this new approach should work. 
image

## Checklist
- [x] add to repo secrets
- [x] tested locally
- [x] updated `CHANGELOG.md`

Fixes #3444

### DIFF
--- a/ci/steps/brew-bump.sh
+++ b/ci/steps/brew-bump.sh
@@ -29,7 +29,11 @@ main() {
   # Source: https://serverfault.com/a/912788
   # shellcheck disable=SC2016,SC2028
   echo '#!/bin/sh\nexec echo "$HOMEBREW_GITHUB_API_TOKEN"' >"$HOME"/.git-askpass.sh
-  GIT_ASKPASS="$HOME/.git-askpass.sh" git push https://cdrci@github.com/cdrci/homebrew-core.git --all
+  # Ensure it's executable since we just created it
+  chmod +x "$HOME/.git-askpass.sh"
+  # GIT_ASKPASS lets us use the password when pushing without revealing it in the process list
+  # See: https://serverfault.com/a/912788
+  GIT_ASKPASS="$HOME/.git-askpass.sh" git push https://cdr-oss@github.com/cdr-oss/homebrew-core.git --all
 
   # Find the docs for bump-formula-pr here
   # https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L18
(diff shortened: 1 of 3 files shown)


ci/steps/brew-bump.sh
```diff
@@ -26,7 +26,10 @@ main() {
   git merge upstream/master
 
   echo "Pushing changes to cdrci/homebrew-core fork on GitHub"
-  git push origin master
+  # Source: https://serverfault.com/a/912788
+  # shellcheck disable=SC2016,SC2028
+  echo '#!/bin/sh\nexec echo "$HOMEBREW_GITHUB_API_TOKEN"' >"$HOME"/.git-askpass.sh
+  GIT_ASKPASS="$HOME/.git-askpass.sh" git push https://cdrci@github.com/cdrci/homebrew-core.git --all
```
jsjoeio, line 32: Does this all look right?

code-asher: Oh I guess we'll need to update the bot username though.

jsjoeio: Oh good call! When I ran it locally it wasn't executable. 

Ah! I changed it locally when testing but not here. Good catch too!

jsjoeio: fixed!

jsjoeio:
@code-asher ready for a second review

code-asher (approved):

PR #3509, opened 2021-05-30
lib/vscode update to 1.56.2; script to patch vscode automatically

Sam0230:


oxy:
Hi there - we don't really want to make any major changes to our workflow right now, and your branch is nigh-impossible to review, since it mixes both toolchain changes and an update; would you be able to split this up?

PR #3516, opened 2021-06-01
feat(ci): add restore-keys for ci speedup

oxy:
See https://github.com/actions/cache/blob/main/examples.md#node---yarn for an example. `restore-keys` will pull from an older set of dependencies in the cache, then update them and store them again, instead of starting from scratch.

This should bring us decent speed benefits for Dependabot PRs that update a few packages at a time.

## Checklist

- [ ] updated `CHANGELOG.md`


code-asher (approved):
:fire:

PR #3544, opened 2021-06-04
feat(vscode): update to version 1.57.0

oxy:
This PR updates vscode to 1.57.0

## TODOS

- [x] test editor locally
- [x] test terminal locally
- [x] redo notes about significant changes in docs/CONTRIBUTING.md#notes-about-changes
- [x] address formatting, etc. feedback
- [x] look at other patches and move over any needed ones
- [x] Investigate broken functionality
  - [x] Cannot install `vscode-icons`
  - [ ] `VSCODE_IPC_HOOK_CLI` unset in terminal
- [ ] update lib/vscode/yarn.lock to fix security issues as needed

## Files with conflicts (fix these)
- [x] lib/vscode/build/package.json
- [x] lib/vscode/build/yarn.lock
- [x] lib/vscode/coder.js
- [x] lib/vscode/package.json
- [x] lib/vscode/src/vs/server/node/channel.ts
- [x] lib/vscode/src/vs/server/node/server.ts
- [x] lib/vscode/yarn.lock


lib/vscode/.gitignore
```diff
@@ -7,7 +7,8 @@ node_modules/
 extensions/**/dist/
 /out*/
 /extensions/**/out/
-src/vs/server
+# NOTE@coder: remove to provide our own server
+#src/vs/server
```
jsjoeio, line 11: nit: insert space here before `#`

lib/vscode/src/vs/server/browser/client.ts
```diff
@@ -0,0 +1,198 @@
...
+	});
+
+	// Add logout to the (web-only) home menu.
+	MenuRegistry.appendMenuItem(MenuId.MenubarHomeMenu, {
+		command: {
+			id: LOGOUT_COMMAND_ID,
+			title: localize('logout', "Log out")
+		},
+		when: ContextKeyExpr.has('code-server.authed')
+	});
```
jsjoeio: Nice! Love the comments too.

jsjoeio (approved):
Looks good to me! Only one nit but nice work!

lib/vscode/src/vs/platform/environment/common/environmentService.ts
```diff
@@ -15,6 +15,24 @@ import { URI } from 'vs/base/common/uri';
...
+	const resolved = resolve(arg);
+
+	if (normalize(arg) === resolved) {
+		return resolved;
+	}
+
+	return resolve(process.env['VSCODE_CWD'] || process.cwd(), arg);
+}
+
+
```
code-asher: If VS Code no longer uses this we can get rid of it as well. I believe it was used because `extensions-dir` was using it so we can match what they do there. Looks like they use `resolve` now.

lib/vscode/src/vs/server/node/channel.ts
```diff
@@ -502,7 +500,7 @@ export class TerminalProviderChannel implements IServerChannel<RemoteAgentConnec
 		};
 
 		const getEnvironment = async (): Promise<platform.IProcessEnvironment> => {
-			const env = await this.terminalProfileResolverService.getEnvironment(remoteAuthority);
+			const env = await this.ptyService.getEnvironment();
 			env.VSCODE_IPC_HOOK_CLI = process.env['VSCODE_IPC_HOOK_CLI']!;
 			return env;
 		};
```
code-asher: We can get rid of this wrapper function and call `this.ptyService.getEnvironment()` directly.

Previously this fetched the parent's environment which is why we had to add `VSCODE_IPC_HOOK_CLI` back but now that it always inherits the environment we don't need to set it manually.

Actually I think maybe we just use `process.env` directly since that's what we had before when `inheritEnv` was optional.

lib/vscode/src/vs/server/node/connection.ts
```diff
@@ -170,7 +171,7 @@ export class ExtensionHostConnection extends Connection {
 			// While not technically necessary, makes it easier to tell which process
 			// bootstrap-fork is executing. Can also do pkill -f extensionHost
 			// Other spawns in the VS Code codebase behave similarly.
-			[ '--type=extensionHost' ],
+			[ '--type=extensionHost', `--uriTransformerPath=${join(FileAccess.asFileUri('bootstrap-fork', require).fsPath, '../../../../out/node/uri_transformer.js')}` ],
```
code-asher, line 174: Should we use the symlinked path here? Then we can use `FileAccess.asFileUri('vs/server/node/uriTransformer', require).fsPath`.

src/node/uri_transformer.ts
```diff
@@ -0,0 +1,44 @@
...
+	path?: string;
+}
+
+interface IRawURITransformer {
+	transformIncoming(uri: UriParts): UriParts;
+	transformOutgoing(uri: UriParts): UriParts;
+	transformOutgoingScheme(scheme: string): string;
+}
+
+// This is deliberate! see src/vs/workbench/services/extensions/node/extensionHostProcessSetup.ts
```
code-asher, line 16: What does this comment mean?

I wonder if we should put an explanation of `export =` since the behavior is obscure.

src/node/uri_transformer.ts
```diff
@@ -0,0 +1,44 @@
+// NOTE: copied over from lib/vscode/src/vs/common/uriIpc.ts
```
code-asher, line 1: I think we should call this `uriTransformer.ts` since that seems to match up better with the other files. I know we have `coder_cloud.ts` and `proxy_agent.ts` but IMO we should rename those too.

lib/vscode/src/vs/server/node/channel.ts
```diff
@@ -397,7 +397,8 @@ export class TerminalProviderChannel implements IServerChannel<RemoteAgentConnec
 			case '$onPtyHostStartEvent': return this.ptyService.onPtyHostStart || Event.None;
 			case '$onPtyHostUnresponsiveEvent': return this.ptyService.onPtyHostUnresponsive || Event.None;
 			case '$onPtyHostResponsiveEvent': return this.ptyService.onPtyHostResponsive || Event.None;
-
+			case '$onPtyHostRequestResolveVariablesEvent': return this.ptyService.onPtyHostRequestResolveVariables || Event.None; 
+
```
code-asher, line 401: Nit: extra blank line and a bit of trailing whitespace. :smile:

lib/vscode/.eslintignore
```diff
@@ -16,3 +16,8 @@
 **/extensions/markdown-language-features/notebook-out/**
 **/extensions/typescript-basics/test/colorize-fixtures/**
 **/extensions/**/dist/**
+# These are code-server code symlinks.
+src/vs/base/node/proxy_agent.ts
+src/vs/ipc.d.ts
```
code-asher, line 21: `src/vs/server/common/ipc.ts` now I think.

lib/vscode/src/vs/server/node/cliProcessMain.ts
```diff
@@ -0,0 +1 @@
+../../code/node/cliProcessMain.ts
```
code-asher, line 1: Should we add this one to the rule instead? I think it should be safe to call `code/node` code from `server/node`.

lib/vscode/.eslintrc.json
```diff
@@ -799,17 +799,40 @@
 				]
 			},
 			{
-				"target": "**/vs/server/**",
+				"target": "**/vs/server/node/**",
```
code-asher, line 802: :fire: really great to have these rules fixed.

lib/vscode/product.json
```diff
@@ -20,9 +20,8 @@
 	"darwinBundleIdentifier": "com.visualstudio.code.oss",
 	"linuxIconName": "com.visualstudio.code.oss",
 	"licenseFileName": "LICENSE.txt",
-	"reportIssueUrl": "https://github.com/microsoft/vscode/issues/new",
+	"reportIssueUrl": "https://github.com/cdr/code-server/issues/new",
```
code-asher: This is probably a task for another time but I wonder if we should change these at build time instead of patching like we have been.

lib/vscode/src/vs/workbench/common/resources.ts
```diff
@@ -75,8 +76,10 @@ export class ResourceContextKey extends Disposable implements IContextKey<URI> {
 		if (!ResourceContextKey._uriEquals(this._resourceKey.get(), value)) {
 			this._contextKeyService.bufferChangeEvents(() => {
 				this._resourceKey.set(value);
-				this._schemeKey.set(value ? value.scheme : null);
-				this._filenameKey.set(value ? basename(value) : null);
+				// NOTE@coder: this is to get Git context actions to show up
+				// See issue #1140 / commit 7e4a73ce2d19eee08ceea25113debefeb8ac27e2
+				// TODO@oxy: Codespaces has this working alright without this patch - investigate why we need this and remove it.
+				this._schemeKey.set(value ? (value.scheme === Schemas.vscodeRemote ? Schemas.file : value.scheme) : null);				this._filenameKey.set(value ? basename(value) : null);
```
code-asher, line 82: Looks like a newline got lost here.

code-asher (changes requested):
First pass, going to check the login stuff now.

lib/vscode/src/vs/platform/environment/common/environmentService.ts
code-asher: Ah yup I followed `resolve` down and they handle `VSCODE_CWD` further down the line so we're good to just use `resolve` directly.

lib/vscode/.eslintignore
code-asher: Actually maybe we don't need this line? The linter didn't seem to complain. Maybe because it's a `.d.ts`? I wonder if we should move it into `typings`.

code-asher:
This resolved those odd differences we had with upstream! Well done.

There are still quite a few patches missing. Here's the script I'm using:

```bash
#!/usr/bin/env bash
set -euo pipefail

main() {
  cd "$(dirname "$0")/../.."

  local upstream=$1 ; shift
  local main=$1 ; shift

  git diff "$upstream" "$main" --name-only \
    | grep -v 'yarn.lock$' \
    | grep -v '^src/typings/' \
    | xargs git diff "$upstream" "$main" --
}

main "$@"
```

Then:

```
./diff.bash vscode/release/1.56 github/main:lib/vscode > old.diff
./diff.bash vscode/release/1.57 github/vscode-1.57:lib/vscode > new.diff
diff old.diff new.diff # Or better yet use side-by-side diff in editor.
```

lib/vscode/src/vs/server/node/connection.ts
oxy: This is actually a bit of an ugly hack; could not figure out how to compile with CommonJS inside here, so I moved the file into our `src/` folder where we (already did) compile with CommonJS, and then just load the version we compiled instead of the version VSCode compiled, heh.

lib/vscode/.eslintrc.json
oxy: Heads up - I restored the rules to the defaults, and moved a bunch of our files to other places inside vscode.
Will update the docs on changes. :eyes:

lib/vscode/.gitignore
oxy: done!

lib/vscode/src/vs/server/browser/client.ts
oxy: I think you'd want to compliment @code-asher on this one; the code got lost in the update script mess and I just brought it back 😅

lib/vscode/src/vs/platform/environment/common/environmentService.ts
oxy: done!

lib/vscode/src/vs/server/node/channel.ts
oxy: Discussed elsewhere, but conclusion is we're keeping it this way:
ptyHostService calls getEnvironment in the ptyHost process it creates, which uses that process's environment. The process spawned doesn't have VSCODE_IPC_HOOK_CLI in its env, so we add it here.

lib/vscode/src/vs/server/node/connection.ts
oxy: Added comments to explain this!

src/node/uri_transformer.ts
oxy: Added comments!

lib/vscode/src/vs/server/node/channel.ts
oxy: Should be fixed now 🤞

lib/vscode/.eslintignore
oxy: Not sure if we should move into `typings`, but I moved it to `src/vs/base/common` (and updated the path.)

lib/vscode/src/vs/server/node/connection.ts
code-asher: Right! But you also symlinked it right? So we could just require the symlinked file.

code-asher:
Some notes:

1. `vscode-icons` works for me, what errors are you seeing?
2. We can tackle the missing hook variable in another PR since this issue exists in the current version.

Some missing patches:

1. Node version and yarnrc. This causes version mismatch issues.
2. `delete data.dependencies`. This causes missing Node modules when installing from npm.
3. Proxy agent in `extensionHostProcessSetup.ts`.
4. Changes to the `workspace` and `folder` query parameters to make them use paths instead of full URIs.

Missing patches that I haven't looked into so I'm not sure if they are an issue:
1. Proposed API patch in browser/environmentService.ts are missing. This may cause the proposed API to not get enabled with the flag.
2. Patch that makes storage service return a promise is missing. I think this caused an issue with rust-analyzer because it would write to storage and then reload the window but wouldn't wait for the write so it got canceled.

oxy:
@code-asher Thanks for the review! ^^

> Some notes:
>     1. `vscode-icons` works for me, what errors are you seeing?
I didn't know what I was getting so I moved to a native Linux install and it was all alright - I'm going to chalk it up to WSL :sweat_smile: 

>     2. We can tackle the missing hook variable in another PR since this issue exists in the current version.
Sure! SGTM. I started going through the version history and trying to find any mentions of it, but its been sending me on a wild goose chase; problem for another time :sweat_smile: 

> Some missing patches:
>     1. Node version and yarnrc. This causes version mismatch issues.
Ah, didn't notice this caused an error; it worked alright in my setup (and looks alright in CI), but I can still restore the patch :smile:

>     2. `delete data.dependencies`. This causes missing Node modules when installing from npm.
You brought this up earlier about how `yarn pack` would destroy anything in `node_modules`, and that stuff from `node_modules` was carried into `release/` AFAIU; can restore this!

>     3. Proxy agent in `extensionHostProcessSetup.ts`.
Good catch! Missed that :sweat_smile: 

>     4. Changes to the `workspace` and `folder` query parameters to make them use paths instead of full URIs.
Another good catch! Missed those patches in those files;

> Missing patches that I haven't looked into so I'm not sure if they are an issue:
>     1. Proposed API patch in `browser/environmentService.ts`. This may cause the proposed API to not get enabled with the flag.
I haven't tested if anything really depends on it right now, but can do so!

>     2. Patch that makes storage service return a promise. I think this caused an issue with rust-analyzer because it would write to storage and then reload the window but wouldn't wait for the write so it got canceled.
I've been using rust-analyzer on a fresh install of 1.57 over the past few days and everything worked okay - do you think we still need to restore this patch?

code-asher:
> Ah, didn't notice this caused an error; it worked alright in my setup (and looks alright in CI), but I can still restore the patch

Fair! I'm not really sure what causes this one to manifest. I only get it when I run on a new Docker container (I do: `git clone`, `yarn`, `yarn watch`, then load VS Code and get errors from spdlog). My guess is that if the Node modules don't exist yet and there's no cache then yarn builds them according to the `.yarnrc` which causes them not to match up with the version of Node you're using (potentially, if you run a different version of Node than the one specified in the `.yarnrc`). Removing the version causes it to just use the version of Node you are building with.

> You brought this up earlier about how `yarn pack` would destroy anything in `node_modules`, and that stuff from `node_modules` was carried into `release/` AFAIU; can restore this!

Exactly! `yarn` seems to remove all `node_modules` before publish and since the deps are deleted from the `package.json` files they don't get re-installed when code-server is installed from npm.

> I haven't tested if anything really depends on it right now, but can do so!

It's definitely possible that the flag is natively supported now! We had to patch it into working previously. Would be great if it worked out of the box.

> I've been using rust-analyzer on a fresh install of 1.57 over the past few days and everything worked okay - do you think we still need to restore this patch?

I'm hesitant unless we can test that reloading the page after writing to storage doesn't cause the storage write to get canceled. This specific bug I think happened when you download the rust binary or when you switch channels or something like that. It tries to write the channel out and reload the page.

code-asher:
The rust-analyzer issue: https://github.com/cdr/code-server/issues/2021

jsjoeio:
Hoping we can move this forward when you two are back next week 🤞 Let me know how I can help!

code-asher (approved):

nooobcoder:
So excited for `v1.57` it will support **Github Copilot** now when release publicly. @oxy @jsjoeio @code-asher @bpmct 
Party 🥳🎉

madawei2699:
> So excited for `v1.57` it will support **Github Copilot** now when released publicly. @oxy @jsjoeio @code-asher @bpmct
> Party 🥳🎉

cannot install Copilot from marketplace, how do you install it?

jsjoeio:
@madawei2699 I don't think Copilot exists outside of the official marketplace so I don't think you can install it but feel free to open a Discussion and maybe the community can chime in!

nooobcoder:
@madawei2699 @jsjoeio FYI, you can install the official VSIX from the vscode marketplace. I do not know if this is legal or not, but is better to know.

When you visit an extension page, look to the right panel, and there is an option to download the extension vsix

image

chengxuncc:
> FYI, you can install the official VSIX from the vscode marketplace. I do not know if this is legal or not, but is better to know.
> 
> When you visit an extension page, look to the right panel, and there is an option to download the extension vsix

Gave it a shot and doesn't work.
```
info  [Service Worker] registered
WARN: The web worker extension host is started without an iframe sandbox!
WARN: loader.js:1362 Duplicate definition of module 'vs/workbench/services/keybinding/browser/keyboardLayouts/layout.contribution.win'
WARN: Extension 'GitHub.copilot' uses PROPOSED API which is subject to change and removal without notice.
Failed to initialize remote Link authority: No remote extension installed to resolve coder-link.
console.ts:137 [Extension Host] [INFO] [auth] [2021-09-28T02:08:43.147Z] Invalid copilot token: missing token
```

jsjoeio:
@chengxuncc feel free to open a Discussion and we'll see if we can help!

madawei2699:
> > FYI, you can install the official VSIX from the vscode marketplace. I do not know if this is legal or not, but is better to know.
> 
> > 
> 
> > When you visit an extension page, look to the right panel, and there is an option to download the extension vsix
> 
> 
> 
> Gave it a shot and doesn't work.
> 
> ```
> 
> info  [Service Worker] registered
> 
> WARN: The web worker extension host is started without an iframe sandbox!
> 
> WARN: loader.js:1362 Duplicate definition of module 'vs/workbench/services/keybinding/browser/keyboardLayouts/layout.contribution.win'
> 
> WARN: Extension 'GitHub.copilot' uses PROPOSED API which is subject to change and removal without notice.
> 
> Failed to initialize remote Link authority: No remote extension installed to resolve coder-link.
> 
> console.ts:137 [Extension Host] [INFO] [auth] [2021-09-28T02:08:43.147Z] Invalid copilot token: missing token
> 
> ```

I try this way that download extension and copy it to code server then install, but it does not work even I use the GitHub copilot token, it seems like the code sever is not compatible with the extension.

jsjoeio:
@madawei2699 Yes, I think you're right, though feel free to [open a Discussion](https://github.com/cdr/code-server/discussions/new) and maybe the community can help!

PR #3563, opened 2021-06-08
fix(security): update css-what, glob-parent, trim-newlines

jsjoeio:
This PR fixes the following security vulnerabilities:
- https://www.npmjs.com/advisories/1753
- https://www.npmjs.com/advisories/1754
- https://www.npmjs.com/advisories/1751

## Checklist

- [ ] updated `CHANGELOG.md`


jawnsy (approved):

oxy:
This broke parcel-bundler :(
We're likely either going to have to ignore the vuln or move to parcel v2, which is *still* in beta months after v1 was deprecated :(

jsjoeio:
> We're likely either going to have to ignore the vuln or move to parcel v2

Shoot :( I wonder how easy/difficult that will be. Do you have a preference on which approach we take?

code-asher:
Could also consider removing Parcel and just compile with plain `tsc` using the UMD target. We would need a loader like require.js to handle the loading on the browser side.

I don't believe we make use of any other Parcel features other than the loading.

jsjoeio:
Closing in favor of https://github.com/cdr/code-server/pull/3586 and https://github.com/cdr/code-server/pull/3578

PR #3582, opened 2021-06-09
chore(docs): add README to the docs

BrunoQuaresma:

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,77 @@
+# code-server &middot; [!["GitHub Discussions"](https://img.shields.io/badge/%20GitHub-%20Discussions-gray.svg?longCache=true&logo=github&colorB=purple)](https://github.com/cdr/code-server/discussions) [!["Join us on Slack"](https://img.shields.io/badge/join-us%20on%20slack-gray.svg?longCache=true&logo=slack&colorB=brightgreen)](https://cdr.co/join-community) [![Twitter Follow](https://img.shields.io/twitter/follow/CoderHQ?label=%40CoderHQ&style=social)](https://twitter.com/coderhq)
+
+[![codecov](https://codecov.io/gh/cdr/code-server/branch/main/graph/badge.svg?token=5iM9farjnC)](https://codecov.io/gh/cdr/code-server)
+[![See latest docs](https://img.shields.io/static/v1?label=Docs&message=see%20latest%20&color=blue)](https://github.com/cdr/code-server/tree/v3.10.2/docs)
+
+Run [VS Code](https://github.com/Microsoft/vscode) on any machine anywhere and access it in the browser.
+
+![Screenshot](./assets/screenshot.png)
+
+## Highlights
+
+- Code on any device with a consistent development environment
+- Use cloud servers to speed up tests, compilations, downloads, and more
+- Preserve battery life when you're on the go; all intensive tasks run on your server
+
+## Requirements
+
+For a good experience, we recommend at least:
+
+- 1 GB of RAM
+- 2 cores
+
+You can use whatever linux distribution floats your boat but in our [guide](./guide.md) we assume Debian on Google Cloud.
+
+## Getting Started
+
+There are three ways you can get started:
+
+1. Using the [install script](https://github.com/cdr/code-server/blob/main/install.sh), which automates most of the process. The script uses the system package manager (if possible)
+2. Manually installing code-server; see [Installation](./install.md) for instructions applicable to most use cases
+3. Use our one-click buttons and guides to [deploy code-server to a popular cloud provider](https://github.com/cdr/deploy-code-server) ⚡
+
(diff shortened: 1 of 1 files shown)


jsjoeio:
We're working on the security fix so you can ignore the audit-ci failure.

For additional context, we're adding this because Bruno is working on deploying the code-server docs to a site and we need a Markdown file to render fo the home page. We'll most like change this later!

jsjoeio:
I had an idea. @BrunoQuaresma instead of adding this file to `/docs` and duplicating `README.md`, could we add a script that just copies the `README.md` at the root to the `/docs` folder when we deploy the docs?

JammSpread:
Does it have to be duplicated, couldn't the README just be moved to the 'docs/' folder since according to [this](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-readmes):
> If you put your README file in your repository's root, docs, or hidden .github directory, GitHub will recognize and automatically surface your README to repository visitors.

Although that may decrease the intuitiveness/simplicity of the repository.

BrunoQuaresma:
Thanks @JammSpread I didn't know about this one. I will try that.

BrunoQuaresma:
The badge can be a problem for what we are trying to achieve with these changes - have a doc site for code-server docs. I'm wondering if makes sense we have something similar to what react-query does on their repo: https://github.com/tannerlinsley/react-query. @jsjoeio @JammSpread thoughts?

jsjoeio:
> I'm wondering if makes sense we have something similar to what react-query does on their repo:

Sorry @BrunoQuaresma - a bit dense this morning. What am I looking at in react-query's repo? I see their badges in the README. Or do you mean how the `README` is very short and then they have the `/docs` repo? 

I do like what @JammSpread suggested. If GitHub surfaces the `README.md` that could be a solid approach.

BrunoQuaresma:
>  Or do you mean how the README is very short and then they have the /docs repo?
Yes, I was talking about that.

> I do like what @JammSpread suggested. If GitHub surfaces the README.md that could be a solid approach.
Me too, but I faced some issues trying to load the badge images with our current markdown parser and I was not able to solve 😞. Because of that, I think the first option can be a good approach for now. 

We can have a README on root with the badges and a brief summary, and a docs/README.md with a more complete introduction. What do you think?

jsjoeio:
> We can have a README on root with the badges and a brief summary, and a docs/README.md with a more complete introduction. What do you think?

I think that could work. As long as we add a short blurb like "look in `/docs` for more documentation` or something, should be fine!

jsjoeio:
> I faced some issues trying to load the badge images with our current markdown parser

Would it be better if the badges were listed using HTML instead of markdown? We have options. See https://shields.io/

image

bpmct:
I would be much more in favor of doing one of these options:

1. Use shields.io (like @jsjoeio suggested)
2. Update our hosted docs detect and to hide the shields from the root-level README
3. Change the codecov link (or whatever one is breaking outside of the repo) to show cdr/code-server stats, regardless where it is hosted.
4. Remove the culprit image

Reasoning being:
- We won't have to manually copy over ./README.md -> ./docs/README.md, or set a prescience for doing it (every release, every commit, etc?)
- docs/README.md can simply inform users of the [hosted site](https://coder.com/docs/code-server/latest)

bpmct:
I would also be in favor of a slimmed down ./README.md in favor of a larger one here. Not a fan of the duplication, but if necessary, we can go that route.

jsjoeio:
@BrunoQuaresma anything you need from us to help you move this forward?

BrunoQuaresma:
@jsjoeio I'm just aligning stuff with @bpmct before move forward, thanks for ask.

bpmct:
We will likely be able to render the root-level README (with shields, if we want). So we will likely not need a second README in docs. Still going over details :)

bpmct:
> Does it have to be duplicated, couldn't the README just be moved to the 'docs/' folder since according to [this](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-readmes):
> 
> > If you put your README file in your repository's root, docs, or hidden .github directory, GitHub will recognize and automatically surface your README to repository visitors.
> 
> Although that may decrease the intuitiveness/simplicity of the repository and thus be a bad idea.

@jsjoeio What are your thoughts on moving the readme as suggested above? This would reduce the complexity for the docs site

jsjoeio:
Thanks for the test repo. Seems fine to me too! I'll see if @code-asher or @oxy want to weigh in

BrunoQuaresma:
I opened up a new PR to move the README. I will close this one, when we get that merged to no break the code-server docs build.

jsjoeio:
I see #3663 has been merged. Does that mean we should close this @BrunoQuaresma?

BrunoQuaresma:
@jsjoeio yes, doing it right now.

PR #3578, opened 2021-06-08
chore(dev): migrate away from parcel

jsjoeio:
This PR migrates us away from Parcel by using `tsc` and `browserify` to accomplish the same thing.

_Originally, we needed to upgrade Parcel from v1 to v2 due to maintainers deprecating v1 and a bunch of security vulnerabilities that we couldn't patch in v1. We were going to use v2 but it has breaking changes. So we decided to replace it with `browserify` which is a bit more lightweight and does the same thing._

## Changes

- remove `parcel`
- use `tsc` to compile `.ts` files and `browserify` to bundle a few files used in the browser
- reference `css` files directly in HTML instead of including with `js` bundles
- remove minification of browser files (issues with es2015+ support using `browserify` plugin `tinyify`)

## Screenshot

Screenshot after building and running locally
Screen Shot 2021-06-10 at 2 23 28 PM

## Checklist
- [x] updated `CHANGELOG.md`

## Related

- https://github.com/cdr/code-server/pull/3614

PR #3586, opened 2021-06-09
fix(dev-deps): parcel vunlerabilities

jsjoeio:
This fixes a couple vulnerabilities introduced by `parcel`. 

## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,12 @@
     "postcss": "^8.2.1",
     "browserslist": "^4.16.5",
     "safe-buffer": "^5.1.1",
-    "vfile-message": "^2.0.2"
+    "vfile-message": "^2.0.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -82,7 +82,12 @@
     "postcss": "^8.2.1",
     "browserslist": "^4.16.5",
     "safe-buffer": "^5.1.1",
-    "vfile-message": "^2.0.2"
+    "vfile-message": "^2.0.2",
+    "parcel/@parcel/config-default/@parcel/optimizer-cssnano/cssnano/cssnano-preset-default/postcss-svgo/svgo/css-select/css-what": "5.0.1",
```
jawnsy, line 86: Are we forced to do this because we can't update parcel?

jawnsy (approved):
Changes seem reasonable to me!

package.json
jsjoeio: Well we're updating parcel to v2 in https://github.com/cdr/code-server/pull/3578 but it doesn't have the fixes either 😅 AFAICT so this is the only solution.

PR #3598, opened 2021-06-11
Allow code-server to identify raspberry pi 4 (armv7l)

### LINKED ISSUE #3612: Default to uname -m in arch() in ci/lib.sh
If no matching case statement for [`ci/lib.sh arch()`](https://github.com/cdr/code-server/blob/main/ci/lib.sh#L37), just return `uname -m`.

janiversen:
This is a one line fix, so adding a PR seems a lot (but will do if demanded).

This fix is only interesting for developers trying to use raspberry pi 4.

With this patch `yarn` passes. The general plan is to work through all issues in being able to build the code on a raspberry pi 4.

Next PR will concentrate on `yarn watch`.

Edit: 
Arch() will not either return the converted cpu type or raw “uname -m”
Fixes #3612
## Checklist

- [ ] updated `CHANGELOG.md`

### DIFF
--- a/ci/lib.sh
+++ b/ci/lib.sh
@@ -35,7 +35,8 @@ os() {
 }
 
 arch() {
-  case "$(uname -m)" in
+  cpu="$(uname -m)"
+  case "$cpu" in
   aarch64)
     echo arm64
     ;;
@@ -43,8 +44,7 @@ arch() {
     echo amd64
     ;;
   *)
-    echo "unknown architecture $(uname -a)"
-    exit 1
+    echo "$cpu"
     ;;
   esac
 }


ci/lib.sh
```diff
@@ -36,6 +36,9 @@ os() {
 
 arch() {
   case "$(uname -m)" in
+  armv7l)
+    echo arm64
+    ;;
```
jsjoeio, line 41: Wouldn't this break since armv7l is 32 bit? ([source](https://askubuntu.com/a/1042272))

jsjoeio (changes requested):
Thanks so much for your PR @janiversen! We discussed this internally and instead we were wondering if you want to [do this](https://github.com/cdr/code-server/issues/3612)?

ci/lib.sh
janiversen: Let me answer the 2 questions in one:
arm7l can and should run the upcoming 64bit raspbian (hopefully bullseye, run nicely but as a alfa version), but right now the official version is buster and that is only 32bit, so I will move it.

I do this, because I want to contribute to code-server. I am at the moment very active in homeassistant also an apache member and have been programming since 1976 (professionally from 1983). My setup is an ipad with a raspberry pi 4 (with a nvm-e attached).

I have been trying to build a version, but have run into several problems, which I am solving one by one. If you have a better way of getting a development environment working on my raspberry pi 4 (buster), I am more than happy to use that.

Being a network expert (I am NOT a UI gui) I have been tracing the communication between the UI and the backend, and I think I know why I see instabilities, including why reconnect can lead to an error that a file cannot be saved.

So in short, I just want to be helpful :-)

janiversen:
Changed. I am still fighting with “yarn watch” it seems to compile code-server on raspberry pi 4 is “interesting”. When I run “install.sh” it works (apart from interesting errors about the number of notify handles), so next step is to find the differences.

ci/lib.sh
jsjoeio: > arm7l can and should run the upcoming 64bit raspbian (hopefully bullseye, run nicely but as a alfa version), but right now the official version is buster and that is only 32bit, so I will move it.

Ah, gotcha! That makes sense. 

> If you have a better way of getting a development environment working on my raspberry pi 4

Oh man. That is a great question! I have a Raspberry Pi 4 but haven't tried developing code-server on it. I do also use an iPad occasionally for writing code, but I host code-server with Railway.

One idea that comes to mind is us creating a Railway template specifically for developing code-server. Then we wouldn't need to worry about issues with the Pi and you could develop from anywhere (including the iPad). Would that something you're interested in? 

> I do this, because I want to contribute to code-server
> So in short, I just want to be helpful :-)

Honestly, we can't tell you how much we appreciate that!

jsjoeio:
>  I am still fighting with “yarn watch” it seems to compile code-server on raspberry pi 4 is “interesting”. 

I know what you're thinking and I'm thinking it too — the `yarn watch` script that runs is...finicky. We're actually in the middle of migrating away from `parcel` which is used to bundle the browser files as part of the `yarn watch` process. But I would love to get this working on your environment (and all environments).

jsjoeio (approved):
Thanks so much for this fix! 🚀

This looks good to me. I'm going to have @oxy or @code-asher give the final approval though since they know this part of the codebase better than me.

ci/lib.sh
janiversen: I have an iMac, which I share with my wife, which is why I moved my development (apart from a big C++ project) to the Raspberry. My iPad is the center of all my activities, with a mouse and a external keyboard.

My idea from running code-server on a raspberry, is due to my involvement in HomeAssistant (a pure python project), but I can do fine with having a code-server in the cloud. At least that could get me up and running fast.

jsjoeio: Awesome! Okay, let me add it to my TODO list and see if I can put something together for you. It should be before the end of the week.

code-asher (approved):
Looks good! We could use `local cpu` to match the general style in that file but it's not a big deal.

janiversen:
I am not sure, what I can do about the failing CI ? (if that is a show-stopper for merging my first commit).

> I know what you're thinking and I'm thinking it too — the `yarn watch` script that runs is...finicky. We're actually in the middle of migrating away from `parcel` which is used to bundle the browser files as part of the `yarn watch` process. But I would love to get this working on your environment (and all environments).

I just got code-server up and running on my Mac (in the background) and my iPad as front-end. I will try to build code-server on the Mac, since I can easily cross compile.

Let me know if I can help with the migration or with the 1.57 branch (the new terminal windows are going to help my daily work a lot).

code-asher:
Failing CI is unrelated so I'll go ahead and merge this in. I think 1.57 is mostly finished but I'll leave some notes on that branch for what is left.

Thanks for the contribution!

ci/lib.sh
jsjoeio: @janiversen we took a look into it and unfortunately I don't think the Railway env will work 😢 @bpmct and I were looking into it and found some system requirements for VS Code that won't work on Railway's free plan. More details [here](https://github.com/bpmct/develop-code-server/issues/8#issuecomment-863411167).

> I will try to build code-server on the Mac, since I can easily cross compile

Maybe this will be the best solution

jsjoeio: Thank you for jumping in and contributing (even though your ran into roadblocks due to some of the limitations with our workflow and docs). 

> Thanks for merging my first (of hopefully many) commits.

Seriously, thank you for contributing! We have more work then bandwidth and every bit helps a ton! Hopefully looking forward to more commits!

janiversen: Speaking of bandwidth, I like to do the “dummy” tasks like doing intelligent replace of x to y+, because it is my way of getting into the code. So if any if you need that done, forward it to me and concentrate on the difficult issues.

Do we have a chat forum like discord ? it is a lot easier for discussions.

jsjoeio: > Speaking of bandwidth, I like to do the “dummy” tasks like doing intelligent replace of x to y+, because it is my way of getting into the cod

That's very kind of you to offer! I'm trying to think. I've added a bunch to the [Roadmap](https://github.com/cdr/code-server/projects/4). I think you should be able to see the Notes in there. Most of it is related to testing. Currently priority for me at least is to improve our code-coverage and get it to 100%. That might be a good place to start?

 > Do we have a chat forum like discord? 
 
We do have a Community Slack. https://community.coder.com/

Though if I'm being honest, I don't check it too much personally (not sure about the other maintainers) but it just becomes too much to deal with. We all check GitHub multiple times throughout the day though so a GitHub Discussion could work 🤔 If we find that too slow or doesn't work, happy to try something else with you!

PR #3614, opened 2021-06-15
chore(dev): remove parcel

jsjoeio:
After trying to fix many issues with upgrading to Parcel 2, we're thinking it may be easier to remove it completely.
## Checklist

- [x] address comments
- [x] clean up commits
- [ ] update `CHANGELOG.md`


ci/build/build-code-server.sh
```diff
@@ -3,7 +3,6 @@ set -euo pipefail
 
 # Builds code-server into out and the frontend into dist.
 
-# MINIFY controls whether parcel minifies dist.
 MINIFY=${MINIFY-true}
```
jsjoeio, line 6: I don't know if we need this?

Also not sure if this means we need to add [`tinyify`](https://www.npmjs.com/package/tinyify) (browserify plugin for optimizing production bundles)

ci/build/build-code-server.sh
```diff
@@ -32,14 +31,10 @@ main() {
...
-    --dist-dir dist \
-    $([[ $MINIFY ]] || echo --no-optimize) \
-    src/browser/register.ts \
-    src/browser/serviceWorker.ts \
-    src/browser/pages/login.ts \
-    src/browser/pages/vscode.ts
+  yarn browserify src/browser/register.ts -o out/browser/register.browserified.js
+  yarn browserify src/browser/serviceWorker.ts -o out/browser/serviceWorker.browserified.js
+  yarn browserify src/browser/pages/login.ts -o out/browser/pages/login.browserified.js
+  yarn browserify src/browser/pages/vscode.ts -o out/browser/pages/vscode.browserified.js
```
jsjoeio, line 37: Before, we just did `parcel`. Do you need to use `yarn` or can we call the CLI directly?

ci/dev/watch.ts
```diff
@@ -150,7 +142,7 @@ class Watcher {
         startingVscode = true
       } else if (startingVscode && line.includes("Finished compilation")) {
         if (startedVscode) {
-          bundle.then(restartServer)
+          restartServer()
```
jsjoeio, line 145: Do we need to call `bundleBrowserCode(browserFiles)` here or no?

ci/dev/watch.ts
```diff
@@ -173,30 +166,27 @@ class Watcher {
...
+  console.log(`[browser] bundling...`)
+  inputFiles.forEach(async (path: string) => {
+    const outputPath = path.replace(".js", ".browserified.js")
+    browserify()
+      .add(path)
+      .bundle()
+      .on("error", function (error: Error) {
+        console.error(error.toString())
+      })
+      .pipe(fs.createWriteStream(outputPath))
```
jsjoeio, line 187: Should we add a log `console.log(`[browser] bundled ${path} to `${outputPath}`)?

package.json
```diff
@@ -29,19 +29,15 @@
     "lint": "./ci/dev/lint.sh",
     "test": "echo 'Run yarn test:unit or yarn test:e2e' && exit 1",
     "ci": "./ci/dev/ci.sh",
-    "watch": "VSCODE_IPC_HOOK_CLI= NODE_OPTIONS=--max_old_space_size=32384 ts-node ./ci/dev/watch.ts",
+    "watch": "VSCODE_IPC_HOOK_CLI= NODE_OPTIONS='--max_old_space_size=32384 --trace-warnings' ts-node ./ci/dev/watch.ts",
```
jsjoeio, line 32: How do we feel about adding the `--trace-warnings`?

src/browser/pages/vscode.html
```diff
@@ -39,8 +39,43 @@
   <body aria-label=""></body>
 
   <!-- Startup (do not modify order of script tags!) -->
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/pages/vscode.js"></script>
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/register.js"></script>
+  <script>
```
jsjoeio, line 42: Wait this isn't supposed to be in here? I thought we could keep it in `vscode.ts`?

ci/dev/watch.ts
code-asher: Nope, we only need the one on line 157 since that's for code-server's code.

ci/build/build-code-server.sh
code-asher: I think we can just use `browserify` because `yarn` adds the bin directory to the path but that assumes no one calls the script directly so it seems safer this way.

ci/build/build-code-server.sh
code-asher: True, it might make sense to add that so our bundles are optimized.

ci/dev/watch.ts
code-asher: IMO the "bundling..." and "done bundling" logs are sufficient!

package.json
code-asher: Works for me!

src/browser/pages/vscode.html
code-asher: Yup, we can remove all this.

tsconfig.json
```diff
@@ -1,14 +1,14 @@
...
+    "target": "es2020",
     "module": "commonjs",
     "moduleResolution": "node",
     "strict": true,
     "noImplicitReturns": true,
     "noUnusedLocals": true,
     "forceConsistentCasingInFileNames": true,
     "outDir": "./out",
-    "declaration": true,
+    "declaration": false,
```
code-asher, line 11: :+1:

code-asher (commented):
-3308 lines, I like it.

ci/build/build-code-server.sh
```diff
@@ -32,14 +31,10 @@ main() {
...
-  parcel build \
-    --public-url "." \
-    --dist-dir dist \
-    $([[ $MINIFY ]] || echo --no-optimize) \
-    src/browser/register.ts \
-    src/browser/serviceWorker.ts \
-    src/browser/pages/login.ts \
-    src/browser/pages/vscode.ts
+  yarn browserify src/browser/register.ts -o out/browser/register.browserified.js
+  yarn browserify src/browser/serviceWorker.ts -o out/browser/serviceWorker.browserified.js
```
code-asher, line 35: We don't use any imports in the service worker so we could use it as-is!

ci/build/build-code-server.sh
jsjoeio: Ah, thanks for explaining that! Makes sense. Okay, interpreting that as "leave as is"!

ci/build/build-code-server.sh
jsjoeio: Oh fantastic! Then will remove that here and use the non-browserified `.js` file in the HTML file. Nice catch!

jsjoeio:
@code-asher ready for a second review 🎉

ci/build/build-code-server.sh
```diff
@@ -32,14 +29,9 @@ main() {
...
-    --public-url "." \
-    --dist-dir dist \
-    $([[ $MINIFY ]] || echo --no-optimize) \
-    src/browser/register.ts \
-    src/browser/serviceWorker.ts \
-    src/browser/pages/login.ts \
-    src/browser/pages/vscode.ts
+  yarn browserify -p tinyify src/browser/register.ts -o out/browser/register.browserified.js
+  yarn browserify -p tinyify src/browser/pages/login.ts -o out/browser/pages/login.browserified.js
+  yarn browserify src/browser/pages/vscode.ts -o out/browser/pages/vscode.browserified.js
```
code-asher, line 34: Should we minify `vscode.ts` as well?

ci/build/build-code-server.sh
```diff
@@ -3,9 +3,6 @@ set -euo pipefail
 
 # Builds code-server into out and the frontend into dist.
 
-# MINIFY controls whether parcel minifies dist.
-MINIFY=${MINIFY-true}
```
code-asher, line 7: Should we leave the ability to disable minification? Implementation could look something like:

```bash
local opts=()
[[ $MINIFY ]] && opts+=("-p" "tinyfy")

yarn browserify "${opts[@]}" src/browser/register.ts -o out/browser/register.browserified.js
```

ci/dev/watch.ts
```diff
@@ -86,28 +84,21 @@ class Watcher {
...
       plugin.stderr.on("data", (d) => process.stderr.write(d))
     }
 
+    const browserFiles = [
+      path.join(this.rootPath, "out/browser/register.js"),
+      path.join(this.rootPath, "out/browser/pages/login.js"),
+      path.join(this.rootPath, "out/browser/pages/vscode.js"),
+    ]
+
+    bundleBrowserCode(browserFiles)
```
code-asher, line 100: I think we could remove this since at this point `tsc` most likely won't have finished compiling yet.

ci/dev/watch.ts
```diff
@@ -173,30 +165,27 @@ class Watcher {
           console.log("[plugin]", original)
         }
         if (line.includes("Watching for file changes")) {
-          bundle.then(restartServer)
+          bundleBrowserCode(browserFiles)
```
code-asher, line 168: We can remove this since this is a `tsc` for plugins which won't recompile the browser files.

src/browser/pages/login.html
```diff
@@ -49,6 +49,6 @@ <h1 class="main">Welcome to code-server</h1>
       </div>
     </div>
   </body>
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/register.js"></script>
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/pages/login.js"></script>
+  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/out/browser/register.browserified.js"></script>
+  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/out/browser/login.browserified.js"></script>
```
code-asher, line 53:
```suggestion
  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/out/browser/pages/login.browserified.js"></script>
``

File: src/browser/pages/vscode.html [hunk 14]
```diff
@@ -39,8 +39,8 @@
   <body aria-label=""></body>
 
   <!-- Startup (do not modify order of script tags!) -->
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/pages/vscode.js"></script>
-  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/dist/register.js"></script>
+  <script data-cfasync="false" src="{{CS_STATIC_BASE}}/out/browser/register.browserified.js"></script>
```
code-asher, line 42: This is completely just a thought I had and not something we need to do in this PR but I wonder if we should do `import "../../register"` in `vscode.ts` and `login.ts` to avoid having to add a second script link (and to keep things to a single bundle especially since I think they both import the logger so we end up bundling it twice).

src/browser/pages/vscode.ts
```diff
@@ -52,4 +52,4 @@ try {
   document.body.style.background = JSON.parse(localStorage.getItem("colorThemeData")!).colorMap["editor.background"]
 } catch (error) {
   // Oh well.
-}
+}
```
code-asher, line 55: Missing newline? Definitely no big deal but doesn't match the other files.

package.json
```diff
@@ -85,12 +82,7 @@
     "postcss": "^8.2.1",
     "browserslist": "^4.16.5",
     "safe-buffer": "^5.1.1",
-    "vfile-message": "^2.0.2",
-    "parcel/@parcel/config-default/@parcel/optimizer-cssnano/cssnano/cssnano-preset-default/postcss-svgo/svgo/css-select/css-what": "5.0.1",
-    "parcel/@parcel/config-default/@parcel/optimizer-htmlnano/htmlnano/cssnano/cssnano-preset-default/postcss-svgo/svgo/css-select/css-what": "5.0.1",
-    "parcel/@parcel/config-default/@parcel/optimizer-htmlnano/htmlnano/svgo/css-select/css-what": "5.0.1",
-    "parcel/@parcel/config-default/@parcel/optimizer-cssnano/cssnano/cssnano-preset-default/postcss-normalize-url/normalize-url": "4.5.1",
-    "parcel/@parcel/config-default/@parcel/optimizer-htmlnano/htmlnano/cssnano/cssnano-preset-default/postcss-normalize-url/normalize-url": "4.5.1"
```
code-asher, line 93: Oo yeah feels good.

code-asher (changes requested):

ci/build/build-code-server.sh
jsjoeio: I knew I was missing something. Nice catch!

ci/build/build-code-server.sh
jsjoeio: Honestly, I didn't know how I might implement this so I removed it 😅 Thanks for writing up the implementation and making it easy to copy/paste!

src/browser/pages/vscode.html
jsjoeio: Good call!

code-asher (approved):
:fire:

PR #3629, opened 2021-06-17
docs(CONTRIBUTING): add section on testing

jsjoeio:
## Checklist

- [x] updated `CHANGELOG.md`

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -9,6 +9,10 @@
   - [Updating VS Code](#updating-vs-code)
     - [Notes about Changes](#notes-about-changes)
 - [Build](#build)
+- [Testing](#testing)
+  - [Unit Tests](#unit-tests)
+  - [Integration Tests](#integration-tests)
+  - [End-to-End Tests](#end-to-end-tests)
 - [Structure](#structure)
   - [Modifications to VS Code](#modifications-to-vs-code)
   - [Currently Known Issues](#currently-known-issues)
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -112,6 +116,37 @@ In our GitHub Actions CI, we use CentOS 7 for maximum compatibility.
...
+
+There are three kinds of tests in code-server:
+
+1. unit tests
+2. integration tests
+3. end-to-end tests
+
+### Unit Tests
+
+Our unit tests are written in TypeScript and run using [Jest](https://jestjs.io/).
```
jawnsy, line 129: Is there a short explanation of what Jest is that we could add here, maybe like "Jest testing framework"? Is Jest used for both unit and integration tests, or just intended for integration tests?

docs/CONTRIBUTING.md
```diff
@@ -112,6 +116,37 @@ In our GitHub Actions CI, we use CentOS 7 for maximum compatibility.
...
+
+These live under [test/unit](../test/unit).
+
+### Integration Tests
+
+These are a work-in-progress. We build code-server a run a script called [test-standalone-release.sh`](../ci/build/test-standalone-release.sh)
```
jawnsy, line 135:
```suggestion
These are a work-in-progress. We build code-server and run a script called [test-standalone-release.sh`](../ci/build/test-standalone-release.sh)
```

jawnsy (approved):
Wow!! :fire: :fire: :fire: 

I love this, great pointers to help new contributors navigate the codebase and understand how things work

janiversen:
I am strongly in favor of of the 3 test levels, but I miss a definition of the definition and objective of each level. E.g. integration testing could be seen as the interface to to VScode.

docs/CONTRIBUTING.md
jsjoeio: Really good point!

jsjoeio:
> but I miss a definition of the definition and objective of each level. E.g. integration testing could be seen as the interface to to VScode

Great feedback! I think I should clarify this a bit more on what we mean by each.

docs/CONTRIBUTING.md
```diff
@@ -112,6 +117,43 @@ In our GitHub Actions CI, we use CentOS 7 for maximum compatibility.
...
+
+There are three kinds of tests in code-server:
+
+1. unit tests
+2. integration tests
+3. end-to-end tests
+
+### Unit Tests
+
+Our unit tests are written in TypeScript and run using [Jest](https://jestjs.io/), which is a JavaScript testing framework.
```
jawnsy, line 130: small note: I'd suggest phrasing as `[Jest testing framework]` because it makes the link bigger, and hence more clickable when you're on mobile. It can be really tough to click small links with your thumb on a small display! Links probably need some minimum length for accessibility purposes I think

jsjoeio: really really good point! i'll make that change now!

janiversen (approved):
LGTM

jsjoeio:
Sweet! Thanks for the reviews @janiversen @jawnsy - going to merge

PR #3630, opened 2021-06-17
docs: Update CONTRIBUTING.md

janiversen:
Node needs be v14.x not greater. If installing the standard version ‘brew install node’, both ‘yarn’ and ‘code-server’ (release version) complains.

Newest version is v16.x so we are pretty far behind.

## Checklist

- [ ] updated `CHANGELOG.md`


jsjoeio (approved):
Ah, nice fix!

Yeah I think we're limited by the Node.js version supported for developing VSCode but hoping they catch up soon and then we can catch up.

PR #3640, opened 2021-06-18
refactor: add func and tests for getNlsConfiguration

jsjoeio:
This PR refactors a couple lines inside `src/browser/pages/vscode.ts` so that it's easier to maintain and test.

## Changes
- add function `getNlsConfiguration`
- add tests

## Checklist

- [x] updated `CHANGELOG.md`

Fixes: N/A

### DIFF
--- a/src/browser/pages/vscode.ts
+++ b/src/browser/pages/vscode.ts
@@ -5,9 +5,56 @@ const options = getOptions()
 // TODO: Add proper types.
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-let nlsConfig: any
+// NOTE@jsjoeio
+// This lives here ../../../lib/vscode/src/vs/base/common/platform.ts#L106
+export const nlsConfigElementId = "vscode-remote-nls-configuration"
+
+type NlsConfiguration = {
+  locale: string
+  availableLanguages: { [key: string]: string } | {}
+  _languagePackId?: string
+  _translationsConfigFile?: string
+  _cacheRoot?: string
+  _resolvedLanguagePackCoreLocation?: string
+  _corruptedFile?: string
+  _languagePackSupport?: boolean
+  loadBundle?: any
+}
+
+/**
+ * A helper function to get the NLS Configuration settings.
+ *
+ * This is used by VSCode for localizations (i.e. changing
+ * the display language).
+ *
+ * Make sure to wrap this in a try/catch block when you call it.
+ **/
+export function getNlsConfiguration(document: Document) {
+  const errorMsgPrefix = "[vscode]"
+  const nlsConfigElement = document?.getElementById(nlsConfigElementId)
+  const nlsConfig = nlsConfigElement?.getAttribute("data-settings")
+
+  if (!document) {
+    throw new Error(`${errorMsgPrefix} Could not parse NLS configuration. document is undefined.`)
+  }
+
+  if (!nlsConfigElement) {
+    throw new Error(
+      `${errorMsgPrefix} Could not parse NLS configuration. Could not find nlsConfigElement with id: ${nlsConfigElementId}`,
+    )
+  }
+
+  if (!nlsConfig) {
+    throw new Error(
+      `${errorMsgPrefix} Could not parse NLS configuration. Found nlsConfigElement but missing data-settings attribute.`,
+    )
+  }
+
+  return JSON.parse(nlsConfig) as NlsConfiguration
+}
+
 try {
-  nlsConfig = JSON.parse(document.getElementById("vscode-remote-nls-configuration")!.getAttribute("data-settings")!)
+  const nlsConfig = getNlsConfiguration(document)
   if (nlsConfig._resolvedLanguagePackCoreLocation) {
     const bundles = Object.create(null)
     nlsConfig.loadBundle = (bundle: any, _language: any, cb: any): void => {
@@ -26,26 +73,25 @@ try {
         .catch(cb)
     }
   }
+  ;(self.require as any) = {
+    // Without the full URL VS Code will try to load file://.
+    baseUrl: `${window.location.origin}${options.csStaticBase}/lib/vscode/out`,
+    recordStats: true,
+    paths: {
+      "vscode-textmate": `../node_modules/vscode-textmate/release/main`,
+      "vscode-oniguruma": `../node_modules/vscode-oniguruma/release/main`,
+      xterm: `../node_modules/xterm/lib/xterm.js`,
+      "xterm-addon-search": `../node_modules/xterm-addon-search/lib/xterm-addon-search.js`,
+      "xterm-addon-unicode11": `../node_modules/xterm-addon-unicode11/lib/xterm-addon-unicode11.js`,
+      "xterm-addon-webgl": `../node_modules/xterm-addon-webgl/lib/xterm-addon-webgl.js`,
+      "tas-client-umd": `../node_modules/tas-client-umd/lib/tas-client-umd.js`,
+      "iconv-lite-umd": `../node_modules/iconv-lite-umd/lib/iconv-lite-umd.js`,
+      jschardet: `../node_modules/jschardet/dist/jschardet.min.js`,
+    },
+    "vs/nls": nlsConfig,
+  }
 } catch (error) {
-  /* Probably fine. */
-}
-
-;(self.require as any) = {
-  // Without the full URL VS Code will try to load file://.
-  baseUrl: `${window.location.origin}${options.csStaticBase}/lib/vscode/out`,
-  recordStats: true,
-  paths: {
-    "vscode-textmate": `../node_modules/vscode-textmate/release/main`,
-    "vscode-oniguruma": `../node_modules/vscode-oniguruma/release/main`,
-    xterm: `../node_modules/xterm/lib/xterm.js`,
-    "xterm-addon-search": `../node_modules/xterm-addon-search/lib/xterm-addon-search.js`,
-    "xterm-addon-unicode11": `../node_modules/xterm-addon-unicode11/lib/xterm-addon-unicode11.js`,
-    "xterm-addon-webgl": `../node_modules/xterm-addon-webgl/lib/xterm-addon-webgl.js`,
-    "tas-client-umd": `../node_modules/tas-client-umd/lib/tas-client-umd.js`,
-    "iconv-lite-umd": `../node_modules/iconv-lite-umd/lib/iconv-lite-umd.js`,
-    jschardet: `../node_modules/jschardet/dist/jschardet.min.js`,
-  },
-  "vs/nls": nlsConfig,
+  console.error(error)
 }
 
 try {
--- a/test/unit/browser/vscode.test.ts
+++ b/test/unit/browser/vscode.test.ts
@@ -1,0 +1,61 @@
+/**
+ * @jest-environment jsdom
+ */
+import { JSDOM } from "jsdom"
+import { getNlsConfiguration, nlsConfigElementId } from "../../../src/browser/pages/vscode"
+
+describe("vscode", () => {
+  describe("getNlsConfiguration", () => {
+    beforeEach(() => {
+      const { window } = new JSDOM()
+      global.document = window.document
+    })
+
+    it("should throw an error if Document is undefined", () => {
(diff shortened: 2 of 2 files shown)


src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
...
+type NlsConfiguration = {
+  locale: string
+  availableLanguages: { [key: string]: string } | {}
+  _languagePackId?: string
+  _translationsConfigFile?: string
+  _cacheRoot?: string
+  _resolvedLanguagePackCoreLocation?: string
+  _corruptedFile?: string
+  _languagePackSupport?: boolean
+  loadBundle?: any
```
jsjoeio, line 19: Incremental type improvements 😄

One day, we will be able to remove that `/* eslint-disable @typescript-eslint/no-explicit-any */` but today is not that today.

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
...
+ * This is used by VSCode for localizations (i.e. changing
+ * the display language).
+ **/
+export function getNlsConfiguration(document: Document) {
+  const nlsConfigElement = document?.getElementById(nlsConfigElementId)
+  const nlsConfig = nlsConfigElement?.getAttribute("data-settings")
+
+  if (!document || !nlsConfigElement || !nlsConfig) {
+    return undefined
+  }
```
jsjoeio, line 34: The downside of this approach is:
- we don't log any errors
- we don't know which piece of this came back false/`undefined`

Would it be worth it to break this up and use a `console.error` for each? Not sure how often people experience errors with the NLS Configuration.

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
...
+
+  try {
+    return JSON.parse(nlsConfig) as NlsConfiguration
+  } catch (error) {
+    console.error("[vscode] Failed to read or parse the NLS configuration.", error)
```
jsjoeio, line 39: I had originally used our `logger` but I know there was a previous discussion about using `console` for logging on the browser so that's what I did.

Plus, it's harder to mock and test our `logger`. I couldn't make it work consistently.

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * @jest-environment jsdom
+ */
```
jsjoeio, line 3: A little trick I learned from StackOverflow! This means no `Location is not defined` errors and you can write your tests peacefully!

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { JSDOM } from "jsdom"
+import { getNlsConfiguration, nlsConfigElementId } from "../../../src/browser/pages/vscode"
+import { createLoggerMock } from "../../utils/helpers"
```
jsjoeio, line 6: remove

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+describe("vscode", () => {
+  describe("getNlsConfiguration", () => {
+    const loggerModule = createLoggerMock()
```
jsjoeio, line 10: remove

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    beforeEach(() => {
+      jest.clearAllMocks()
+      jest.mock("@coder/logger", () => loggerModule)
```
jsjoeio, line 13: remove

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      const { window } = new JSDOM()
+      global.window = window as unknown as Window & typeof globalThis
+      global.document = window.document
+      global.navigator = window.navigator
+      global.location = window.location
```
jsjoeio, line 18: we only use `document` - is all this necessary?

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    })
+
+    afterAll(() => {
+      jest.restoreAllMocks()
+    })
```
jsjoeio, line 23: remove

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      const dataSettings = {
+        first: "Jane",
+        last: "Doe",
+      }
+
+      mockElement.setAttribute("id", nlsConfigElementId)
+      mockElement.setAttribute("data-settings", JSON.stringify(dataSettings))
+      document.body.appendChild(mockElement)
+      // TODO add element to the Dom
+      // TODO test when DOM is there b
```
jsjoeio, line 82: remove

test/unit/browser/vscode.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      const actual = getNlsConfiguration(global.document)
+      expect(actual).toStrictEqual(dataSettings)
+      document.body.removeChild(mockElement)
```
jsjoeio, line 85: in consistent line breaks. clean up.

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
...
+  loadBundle?: any
+}
+
+/**
+ * A helper function to get the NLS Configuration settings.
+ *
+ * This is used by VSCode for localizations (i.e. changing
+ * the display language).
+ **/
+export function getNlsConfiguration(document: Document) {
```
jsjoeio, line 28: How do we feel about this function signature? `document` is available globally in the browser, so it might be silly to pass it in as a prop. Though it does make it a bit easier for testing. Not sure, might be a bikeshed comment.

code-asher: It does feel a bit odd to me since it's an API that is explicitly only for tests, but if the alternative (setting document global in the tests) makes it harder to get coverage then I would say it's worth it.

src/browser/pages/vscode.ts
code-asher: IMO we should throw individual errors for each one, then wherever we call this we can catch and do a single `console.error`.

src/browser/pages/vscode.ts
code-asher: If we go with throwing errors for missing elements we could just let this throw bubble up and log it elsewhere.

Yeah we have so little logging on the frontend it doesn't really seem worth bundling the logger just for one or two logging statements.

Out of curiosity, what problems did you run into testing `logger` that were resolved by switching to `console`?

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
...
+
+  try {
+    return JSON.parse(nlsConfig) as NlsConfiguration
+  } catch (error) {
+    console.error("[vscode] Failed to read or parse the NLS configuration.", error)
+    return undefined
+  }
+}
+
+let nlsConfig: NlsConfiguration | undefined
```
code-asher, line 44: If we have `getNlsConfiguration` throw instead of return undefined we could do something like:

```
try {
  const nlsConfig = getNlsConfiguration(document)
  if (nlsConfig._resolvedLanguagePackCoreLocation) {
    // etc
  }
  // etc
} catch (error) {
  console.error(error)
}

File: test/unit/browser/vscode.test.ts (hunk 4 above)
Comment (code-asher) [reply test/unit/browser/vscode.test.ts:3]: Neat!!

File: src/browser/pages/vscode.ts (hunk 2 above)
Comment (jsjoeio) [reply src/browser/pages/vscode.ts:34]: something like?
```javascript
try {
  if (!document) {
    throw Error("no document)
  }
  
  // etc..
} catch (e) {
  console.error("something broke", e)
}
```

genius! I'll refactor

src/browser/pages/vscode.ts
jsjoeio: > If we go with throwing errors for missing elements we could just let this throw bubble up and log it elsewhere.

Good idea! That's easy to test too.

> Out of curiosity, what problems did you run into testing logger that were resolved by switching to console?

When I tried mocking the `logger`, I couldn't get it to work. I think it's because it will only work if you mock the `logger` then `require` the file and then use the function from the `require` import.

When I used `console`, I could overwrite it to mock it and it worked

src/browser/pages/vscode.ts
jsjoeio: Good idea! I'll refactor

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,54 @@ const options = getOptions()
 // TODO: Add proper types.
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-let nlsConfig: any
+export const nlsConfigElementId = "vscode-remote-nls-configuration"
+
+type NlsConfiguration = {
```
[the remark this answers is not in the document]
jsjoeio: Good idea! I'll take a look

src/browser/pages/vscode.ts
jsjoeio: > but if the alternative (setting document global in the tests) makes it harder to get coverage then I would say it's worth it

Yeah, I think it might be so I'll leave as is for now

src/browser/pages/vscode.ts
jsjoeio: > wherever we call this we can catch and do a single

Oh, don't use `try/catch` inside. Just wherever we call it. Got it!

src/browser/pages/vscode.ts
jsjoeio: Worked brilliantly!

jsjoeio:
@code-asher ready for a a second review!

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,63 @@ const options = getOptions()
...
+  }
+
+  if (!nlsConfigElement) {
+    throw new Error(
+      `${errorMsgPrefix} Could not parse NLS configuration. Could not find nlsConfigElement with id: ${nlsConfigElementId}`,
+    )
+  }
+
+  if (!nlsConfig) {
+    return undefined
```
code-asher, line 48: Should this throw too? I think we can consider it an error.

jsjoeio: Oh! I thought the `nlsConfig` was optional but you're right, this should throw!

jsjoeio: Fixed!

src/browser/pages/vscode.ts
```diff
@@ -5,18 +5,65 @@ const options = getOptions()
...
+  }
+
+  return JSON.parse(nlsConfig) as NlsConfiguration
+}
+
 try {
-  nlsConfig = JSON.parse(document.getElementById("vscode-remote-nls-configuration")!.getAttribute("data-settings")!)
-  if (nlsConfig._resolvedLanguagePackCoreLocation) {
+  const nlsConfig = getNlsConfiguration(document)
+  if (nlsConfig?._resolvedLanguagePackCoreLocation) {
```
code-asher, line 58: I think it should no longer be possible for `nlsConfig` to be undefined?

jsjoeio: Oh right! because now, it can never be `undefined`. I didn't even think to check that again - nice catch!

src/browser/pages/vscode.ts
```diff
@@ -16,7 +63,7 @@ try {
         return cb(undefined, result)
       }
       // FIXME: Only works if path separators are /.
-      const path = nlsConfig._resolvedLanguagePackCoreLocation + "/" + bundle.replace(/\//g, "!") + ".nls.json"
+      const path = nlsConfig?._resolvedLanguagePackCoreLocation + "/" + bundle.replace(/\//g, "!") + ".nls.json"
```
code-asher, line 66: Another `?` we can remove. Otherwise looks great! Feel free to merge.

code-asher (approved):

src/browser/pages/vscode.ts
jsjoeio: Nice! I'll remove this an merge. Thanks for the stellar feedback man!

PR #3647, opened 2021-06-21
Update link desc to Coder.com in README

khorne3:
Updating the link desc to be clearer about where the URL goes/why it's of interest


jsjoeio (approved):
Woo! Thanks @khorne3 🎉

bpmct:
thanks Katie! just saw this

PR #3664, opened 2021-06-24
Parallel tests

code-asher:
I recommend reviewing by commit (and reading the commit messages).

Setting as a draft until I'm sure the tests pass in CI.

## Checklist

- [X] updated `CHANGELOG.md`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -334,8 +334,9 @@ jobs:
     needs: package-linux-amd64
     runs-on: ubuntu-latest
     env:
-      PASSWORD: e45432jklfdsab
-      CODE_SERVER_ADDRESS: http://localhost:8080
+      # Since we build code-server we might as well run tests from the release
+      # since VS Code will load faster due to the bundling.
+      CODE_SERVER_TEST_ENTRY: "./release-packages/code-server-linux-amd64"
     steps:
       - uses: actions/checkout@v2
 
@@ -362,9 +363,11 @@ jobs:
           name: release-packages
           path: ./release-packages
 
-      - name: Untar code-server file
+      - name: Untar code-server release
         run: |
-          cd release-packages && tar -xzf code-server*-linux-amd64.tar.gz
+          cd release-packages
+          tar -xzf code-server*-linux-amd64.tar.gz
+          mv code-server*-linux-amd64 code-server-linux-amd64
 
       - name: Install dependencies
         if: steps.cache-yarn.outputs.cache-hit != 'true'
@@ -380,9 +383,7 @@ jobs:
           yarn install --check-files
 
       - name: Run end-to-end tests
-        run: |
-          ./release-packages/code-server*-linux-amd64/bin/code-server --log trace &
-          yarn test:e2e
+        run: yarn test:e2e
 
       - name: Upload test artifacts
         if: always()
--- a/ci/dev/test-e2e.sh
+++ b/ci/dev/test-e2e.sh
@@ -3,10 +3,35 @@ set -euo pipefail
 
 main() {
   cd "$(dirname "$0")/../.."
+  source ./ci/lib.sh
+
+  local dir="$PWD"
+  if [[ ! ${CODE_SERVER_TEST_ENTRY-} ]]; then
+    echo "Set CODE_SERVER_TEST_ENTRY to test another build of code-server"
+  else
+    pushd "$CODE_SERVER_TEST_ENTRY"
+    dir="$PWD"
+    popd
+  fi
+
+  echo "Testing build in '$dir'"
+
+  # Simple sanity checks to see that we've built. There could still be things
+  # wrong (native modules version issues, incomplete build, etc).
+  if [[ ! -d $dir/out ]]; then
+    echo >&2 "No code-server build detected"
+    echo >&2 "You can build it with 'yarn build' or 'yarn watch'"
+    exit 1
+  fi
+
+  if [[ ! -d $dir/lib/vscode/out ]]; then
+    echo >&2 "No VS Code build detected"
+    echo >&2 "You can build it with 'yarn build:vscode' or 'yarn watch'"
+    exit 1
+  fi
+
   cd test
-  # We set these environment variables because they're used in the e2e tests
-  # they don't have to be these values, but these are the defaults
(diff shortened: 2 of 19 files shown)


oxy (approved):
Looks good to me!

.github/workflows/ci.yaml
```diff
@@ -334,8 +334,9 @@ jobs:
     needs: package-linux-amd64
     runs-on: ubuntu-latest
     env:
-      PASSWORD: e45432jklfdsab
-      CODE_SERVER_ADDRESS: http://localhost:8080
+      # Since we build code-server we might as well run tests from the release
+      # since VS Code will load faster due to the bundling.
+      CODE_SERVER_TEST_ENTRY: "./release-packages/code-server-linux-amd64"
```
jsjoeio, line 339: genius 🎉

ci/dev/test-e2e.sh
```diff
@@ -3,10 +3,35 @@ set -euo pipefail
...
+    echo >&2 "No VS Code build detected"
+    echo >&2 "You can build it with 'yarn build:vscode' or 'yarn watch'"
+    exit 1
+  fi
+
   cd test
-  # We set these environment variables because they're used in the e2e tests
-  # they don't have to be these values, but these are the defaults
-  PASSWORD=e45432jklfdsab CODE_SERVER_ADDRESS=http://localhost:8080 yarn playwright test "$@"
+  yarn playwright test "$@"
```
jsjoeio, line 34: > This also means you are no longer required to specify the password and
address yourself (or the extension directory once we add a test
extension). `yarn test:e2e` should just work as-is.

My favorite part of this PR ♥️

test/e2e/baseFixture.ts
```diff
@@ -1,12 +1,71 @@
...
-    await codeServer.navigate()
-    await use(codeServer)
+/**
+ * Wraps `test.describe` to create and manage an instance of code-server. If you
+ * don't use this you will need to create your own code-server instance and pass
+ * it to `test.use`.
+ *
+ * If `includeCredentials` is `true` page requests will be authenticated.
+ */
+export const describe = (name: string, includeCredentials: boolean, fn: (codeServer: CodeServer) => void) => {
```
jsjoeio, line 12: This is awesome!

test/e2e/baseFixture.ts
```diff
@@ -1,12 +1,71 @@
...
+  test.describe(name, () => {
+    // This will spawn on demand so nothing is necessary on before.
+    const codeServer = new CodeServer(name)
+
+    // Kill code-server after the suite has ended. This may happen even without
+    // doing it explicitly but it seems prudent to be sure.
+    test.afterAll(() => {
+      codeServer.close()
```
jsjoeio, line 20: Do we need to `await` this since `close()` is async?
```suggestion
    test.afterAll(async () => {
      await codeServer.close()
```

test/e2e/baseFixture.ts
```diff
@@ -1,12 +1,71 @@
...
+    })
+
+    const storageState = JSON.parse(process.env.STORAGE || "{}")
+
+    // Sanity check to ensure the cookie is set.
```
jsjoeio, line 25: 👏

test/e2e/baseFixture.ts
```diff
@@ -1,12 +1,71 @@
...
+ * ready.
+ */
+export const test = base.extend<TestFixtures>({
+  authenticated: false,
+  codeServer: undefined, // No default; should be provided through `test.use`.
+  codeServerPage: async ({ authenticated, codeServer, page }, use) => {
+    // It's possible code-server might prevent navigation because of unsaved
+    // changes (seems to happen based on timing even if no changes have been
+    // made too). In these cases just accept.
+    page.on("dialog", (d) => d.accept())
```
jsjoeio, line 62: Does this need to be async? Playwright does say `dialog.accept()` returns `<Promise<void>>` [source](https://playwright.dev/docs/api/class-dialog#dialog-accept)
```suggestion
    page.on("dialog", async (d) => await d.accept())
```

test/e2e/models/CodeServer.ts
```diff
@@ -1,21 +1,163 @@
...
+    const dir = await tmpdir(workspaceDir)
+    await fs.mkdir(path.join(dir, ".vscode"))
+    await fs.writeFile(
+      path.join(dir, ".vscode/settings.json"),
+      JSON.stringify({
+        "workbench.startupEditor": "none",
+      }),
+      "utf8",
+    )
+    return dir
```
jsjoeio, line 52: This is huge. Thank you for adding this!

test/e2e/models/CodeServer.ts
```diff
@@ -1,21 +1,163 @@
...
+ * Navigation and setup performed by this model will cause the code-server
+ * process to spawn if it hasn't yet.
+ */
+export class CodeServerPage {
+  private readonly editorSelector = "div.monaco-workbench"
+
+  constructor(private readonly codeServer: CodeServer, public readonly page: Page) {}
+
+  async address() {
+    return this.codeServer.address()
```
jsjoeio, line 152: Do we need to `await` this since `address()` is async?
```suggestion
    return await this.codeServer.address()
```

jsjoeio (changes requested):
Fantastic work man 👏

I think there are a few places we need to add `async`/`await` but everything else looks good!

test/e2e/baseFixture.ts
code-asher: Nice catch!

test/e2e/baseFixture.ts
code-asher: This form of arrow function (without curly braces) will return automatically so it's equal to:

```javascript
() => {
  return d.accept()
}
```

Which has the same effect as an `async` and `await` (in this case at least since the returned promise resolves to void either way).

test/e2e/models/CodeServer.ts
code-asher: Whoops I think the `async` was left over from when I was doing this a bit differently. Will remove!

test/e2e/baseFixture.ts
jsjoeio: right! but don't we still need to await the promise to resolve?

jsjoeio: discussed offline briefly — we don't need to wait for the promise to resolve since it returns `void`

jsjoeio (approved):
Nice work man! This is huge. 🎉

PR #3684, opened 2021-06-28
docs: Update README to point to code-server docs site

BrunoQuaresma:


docs/README.md
```diff
@@ -26,7 +26,7 @@ You can use whatever linux distribution floats your boat but in our [guide](./gu
 There are three ways you can get started:
 
 1. Using the [install script](https://github.com/cdr/code-server/blob/main/install.sh), which automates most of the process. The script uses the system package manager (if possible)
-2. Manually installing code-server; see [Installation](./install.md) for instructions applicable to most use cases
+2. Manually installing code-server; see [Installation](https://coder.com/docs/code-server/latest/install) for instructions applicable to most use cases
```
bpmct, line 29: @jsjoeio as a part of the [release process](https://github.com/cdr/code-server/blob/main/docs/MAINTAINING.md#release), do you still do a bulk find and replace to increment the version number in any docs? 

If so, I think we should replace `latest` with the version number in this case actually so that old docs will link to the correct branch.

bpmct (approved):
yay!

docs/README.md
jsjoeio: Great question! Yes, that happens [here](https://github.com/cdr/code-server/blob/main/ci/build/release-prep.sh#L81-L84).

> If so, I think we should replace latest with the version number in this case actually so that old docs will link to the correct branch

Makes sense to me!

jsjoeio (changes requested):
Two requests:
- rebase with `main` (this will fix CI)
- address @bpmct's comment about the installation link

docs/README.md
BrunoQuaresma: It is quite hard to understand what is happening on this line and when I try to run that locally, after comment the send a PR step, I get an error:
```
xargs: sd: No such file or directory
```

BrunoQuaresma: I don't know why sd is being used here... thoughts?

BrunoQuaresma: I changed latest to use 3.10.2 so the current script are able to replace it.

jsjoeio: Wait a second...I should clarify. You (@BrunoQuaresma) don't need to run that script. That's only used when someone needs to prep a release. I was only pointing that out to show the logic.

All you need to do is update the link Ben pointed out to use the version (3.10.2) (which sounds like you did and tested, thank you ♥️)

> I don't know why sd is being used here... thoughts?

Apologies - I could have sworn I added a check and explanation for it in this file but I guess I forgot. `sd` is a replacement for `sed` written in Rust that has a friendly API. See [docs](https://github.com/chmln/sd).

jsjoeio (approved):
This LGTM! Thanks for doing this Bruno! Let me know when you want me to merge (or you can I think?)

PR #3692, opened 2021-06-29
Fix broken link on docs site

elliott-king:
Links don't work on docs site unless they have md formatting. Fix is minor.


bpmct (approved):
Thanks, I think GitHub renders links without the brackets necessary, but our docs engine site does not.

jsjoeio:
Thanks for the fix @elliott-king 🎉

PR #3695, opened 2021-06-29
fix: escape error.message on login failure

### LINKED ISSUE #3382: Sanitize error messages before writing to login page
Issue identified by CodeQL in `src/node/util.ts`

See issue explanation and recommendation: https://github.com/cdr/code-server/security/code-scanning/44?query=ref%3Arefs%2Fheads%2Fmain

jsjoeio:
This PR fixes a security issue by sanitizing the `error.message` before sending it back to the client on a login failure.

## Changes
- add function `escapeHtml` + tests
- fix: escape `error.message` before sending to client on login failure

## Checklist
- [x] tested locally
- [x] added a test

Fixes #3382

### DIFF
--- a/src/node/routes/login.ts
+++ b/src/node/routes/login.ts
@@ -112,7 +112,7 @@ router.post("/", async (req, res) => {
 
     throw new Error("Incorrect password")
   } catch (error) {
-    const htmlToRender = await getRoot(req, error)
-    res.send(htmlToRender)
+    const renderedHtml = await getRoot(req, error)
+    res.send(renderedHtml)
   }
 })
--- a/src/node/routes/login.ts
+++ b/src/node/routes/login.ts
@@ -41,7 +41,7 @@ const getRoot = async (req: Request, error?: Error): Promise<string> => {
     req,
     content
       .replace(/{{PASSWORD_MSG}}/g, passwordMsg)
-      .replace(/{{ERROR}}/, error ? `<div class="error">${error.message}</div>` : ""),
+      .replace(/{{ERROR}}/, error ? `<div class="error">${escapeHtml(error.message)}</div>` : ""),
   )
 }
 
@@ -112,8 +112,7 @@ router.post("/", async (req, res) => {
 
     throw new Error("Incorrect password")
   } catch (error) {
-    const html = await getRoot(req, error)
-    const escapedHtml = escapeHtml(html)
-    res.send(escapedHtml)
+    const htmlToRender = await getRoot(req, error)
+    res.send(htmlToRender)
   }
 })
--- a/src/node/routes/login.ts
+++ b/src/node/routes/login.ts
@@ -4,7 +4,7 @@ import { RateLimiter as Limiter } from "limiter"
 import * as path from "path"
 import { rootPath } from "../constants"
 import { authenticated, getCookieDomain, redirect, replaceTemplates } from "../http"
-import { getPasswordMethod, handlePasswordValidation, humanPath, sanitizeString } from "../util"
+import { getPasswordMethod, handlePasswordValidation, humanPath, sanitizeString, escapeHtml } from "../util"
 
 export enum Cookie {
   Key = "key",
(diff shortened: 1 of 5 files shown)


src/node/util.ts
```diff
@@ -509,3 +509,17 @@ export const isFile = async (path: string): Promise<boolean> => {
     return false
   }
 }
+
+/**
+ * Escapes any HTML string special characters, like &, <, >, ", and '.
+ *
+ * Source: https://stackoverflow.com/a/6234804/3015595
+ **/
+export function escapeHTML(unsafe: string): string {
```
jsjoeio, line 518: nit: rename to `escapeHtml`

test/unit/routes/login.test.ts
```diff
@@ -34,4 +37,41 @@ describe("login", () => {
...
+
+    // Store whatever might be in here so we can restore it afterward.
+    // TODO: We should probably pass this as an argument somehow instead of
+    // manipulating the environment.
+    const previousEnvPassword = process.env.PASSWORD
+
+    beforeEach(async () => {
+      process.env.PASSWORD = "test"
+      _codeServer = await integration.setup(["--auth=password"], "")
+    })
```
jsjoeio, line 57: we dont' use the password so maybe remove?

src/node/util.ts
```diff
@@ -509,3 +509,17 @@ export const isFile = async (path: string): Promise<boolean> => {
...
+ *
+ * Source: https://stackoverflow.com/a/6234804/3015595
+ **/
+export function escapeHtml(unsafe: string): string {
+  return unsafe
+    .replace(/&/g, "&amp;")
+    .replace(/</g, "&lt;")
+    .replace(/>/g, "&gt;")
+    .replace(/"/g, "&quot;")
+    .replace(/'/g, "&#039;")
```
oxy, line 524: small nit:
I think we could use `&apos;` here instead of `&#039;`, the former is more immediately recognizable in text!

The reason `&#039` is sometimes recommended is because of old browsers that don't support `&apos`, but we basically target the latest versions of Chrome/FF/Safari and have no need to support IE8, so *should* be fine!

oxy (approved):
one small nit; though I'll leave it to you on if you want it to stay as-is or update the code!

src/node/util.ts
jsjoeio: TIL! Thanks for the suggestion. I'll change that then merge!

src/node/routes/login.ts
```diff
@@ -111,6 +112,7 @@ router.post("/", async (req, res) => {
 
     throw new Error("Incorrect password")
   } catch (error) {
-    res.send(await getRoot(req, error))
+    const htmlToRender = await getRoot(req, error)
```
kylecarbs, line 115: Very minor nit, but this is `renderedHtml`, right?

kylecarbs (approved):
:lock:

src/node/util.ts
```diff
@@ -509,3 +509,17 @@ export const isFile = async (path: string): Promise<boolean> => {
...
 }
+
+/**
+ * Escapes any HTML string special characters, like &, <, >, ", and '.
+ *
+ * Source: https://stackoverflow.com/a/6234804/3015595
+ **/
+export function escapeHtml(unsafe: string): string {
+  return unsafe
+    .replace(/&/g, "&amp;")
```
jawnsy, line 520: seems kinda unfortunate that there isn't a built-in function for doing this? I'm always wary of escaping things incorrectly, but this seems reasonable enough

jawnsy (approved):

src/node/routes/login.ts
jsjoeio: lol technically I think you're right. i'll change it

src/node/util.ts
jsjoeio: Well, it is [built-in](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) but:
> Although `escape()` is not strictly deprecated (as in "removed from the Web standards"), it is defined in Annex B of the ECMA-262 standard, whose introduction states:
> Programmers should not use or assume the existence of these features and behaviors when writing new ECMAScript code

PR #3697, opened 2021-06-30
docs(guide): add WebSockets to requirements

jsjoeio:
This PR updates the requirements in `guide.md` by adding a line about WebSockets.

We've had people in the past not enable WebSockets in their environment, leading to code-server not working.

Evidence
- https://github.com/cdr/code-server/discussions/3667#discussioncomment-934888
- https://github.com/cdr/code-server/discussions/3175#discussioncomment-640924

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -45,6 +45,8 @@ For a good experience, we recommend at least:
 
 You can use whatever linux distribution floats your boat but in this guide we assume Debian on Google Cloud.
 
+In order to work properly, your environment should have WebSockets enabled, which code-server uses to communicate between the browser and server
+
 ### Google Cloud
 
 For demonstration purposes, this guide assumes you're using a VM on GCP but you should be


docs/guide.md
```diff
@@ -45,6 +45,10 @@ For a good experience, we recommend at least:
 
 You can use whatever linux distribution floats your boat but in this guide we assume Debian on Google Cloud.
 
+In order to work properly, your enviornment should have the following enabled:
+
+- WebSockets (used to communicate between code-server client and server)
```
kylecarbs, line 50: Since there's one requirement, it reads a bit weird being a list. Something like:

```
WebSocket's are required to use code-server.
```

kylecarbs (changes requested):

docs/guide.md
```diff
@@ -45,6 +45,10 @@ For a good experience, we recommend at least:
 
 You can use whatever linux distribution floats your boat but in this guide we assume Debian on Google Cloud.
 
+In order to work properly, your enviornment should have the following enabled:
```
jawnsy, line 48:
```suggestion
In order to work properly, your environment should have the following enabled:
```

docs/guide.md
jawnsy: I guess the rationale for a list is that we're probably going to add more here later?

Perhaps:

```suggestion
* WebSockets, which code-server uses to communicate between the browser and server
```

jawnsy (approved):

docs/guide.md
jsjoeio: Yeah, I'll change to one line instead of a list though!

PR #3698, opened 2021-06-30
fix(isHashMatch): check that hash starts with $

jsjoeio:
This PR adds an additional check to `isHashMatch` to check that the hash starts with a "$" to prevent us from calling `argon2.verify` if it doesn't.

It also throws an error if something goes wrong with the `.verify()` function.

**Why do we need to add this?**

The other day I noticed this error: `'TypeError: pchstr must contain a $ as first char\n'`

I believe it happens because we're still in between a migration from the old hashing function `sha256` to the new one `argon2`. And locally, I switch between the latest code-server and the development one. This means it may be calling the function to verify the hash expecting it to be a hash made with `argon2` and instead it's made with `sha256`. 

Once we release the newest version of code-server with the `argon2` algorithm, we don't want it to throw unexpected errors, like this one. We may need to revisit some of the logic for verifying password hashes in case there are issues.

Either way, this adds an extra check to prevent this error from happening.

Fixes N/A

### DIFF
--- a/test/unit/node/util.test.ts
+++ b/test/unit/node/util.test.ts
@@ -189,6 +189,17 @@ describe("isHashMatch", () => {
     const actual = await util.isHashMatch(password, _hash)
     expect(actual).toBe(false)
   })
+  it("should return false and not throw an error if the hash doesn't start with a $", async () => {
+    const password = "hellowpasssword"
+    const _hash = "n2i$v=19$m=4096,t=3,p=1$EAoczTxVki21JDfIZpTUxg$rkXgyrW4RDGoDYrxBFD4H2DlSMEhP4h+Api1hXnGnFY"
+    expect(async () => await util.isHashMatch(password, _hash)).not.toThrow()
+    expect(await util.isHashMatch(password, _hash)).toBe(false)
+  })
+  it("should reject the promise and throw if error", async () => {
+    const password = "hellowpasssword"
+    const _hash = "$ar2i"
+    expect(async () => await util.isHashMatch(password, _hash)).rejects.toThrow()
+  })
 })
 
 describe("hashLegacy", () => {
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -166,14 +166,13 @@ export const hash = async (password: string): Promise<string> => {
  * Used to verify if the password matches the hash
  */
 export const isHashMatch = async (password: string, hash: string) => {
-  if (password === "" || hash === "") {
+  if (password === "" || hash === "" || !hash.startsWith("$")) {
     return false
   }
   try {
     return await argon2.verify(hash, password)
   } catch (error) {
-    logger.error(error)
-    return false
+    throw new Error(error)
   }
 }
 


kylecarbs (approved):
Seems like a reasonable check, but I'm not entire sure why we need to do this.

test/unit/node/util.test.ts
```diff
@@ -189,6 +189,17 @@ describe("isHashMatch", () => {
...
     expect(actual).toBe(false)
   })
+  it("should return false and not throw an error if the hash doesn't start with a $", async () => {
+    const password = "hellowpasssword"
+    const _hash = "n2i$v=19$m=4096,t=3,p=1$EAoczTxVki21JDfIZpTUxg$rkXgyrW4RDGoDYrxBFD4H2DlSMEhP4h+Api1hXnGnFY"
+    expect(async () => await util.isHashMatch(password, _hash)).not.toThrow()
+    expect(await util.isHashMatch(password, _hash)).toBe(false)
+  })
+  it("should reject the promise and throw if error", async () => {
+    const password = "hellowpasssword"
```
jawnsy, line 199: Just curious - what does it mean when the hash begins with a `$`? Is it used as a signifier to designate that the hash is Argon2? I had thought that the other test case was an Argon2 hash, too, so this is a little confusing for me

jawnsy (approved):
Seems reasonable. :smile:

jsjoeio:
> Seems like a reasonable check, but I'm not entire sure why we need to do this.

My fault. I updated the PR description with the _why_ behind this.

test/unit/node/util.test.ts
jsjoeio: > Just curious - what does it mean when the hash begins with a $? Is it used as a signifier to designate that the hash is Argon2?

AFAIK, I believe that's exactly why. 

>  had thought that the other test case was an Argon2 hash, too, so this is a little confusing for me

If you point out which test case you're referring to, happening to clear up any confusion!

jsjoeio:
Side note: @code-asher brought up a good point about UX here so I opened another issue: https://github.com/cdr/code-server/issues/3704

test/unit/node/util.test.ts
jawnsy: Sorry I meant that this:

```
    const _hash = "n2i$v=19$m=4096,t=3,p=1$EAoczTxVki21JDfIZpTUxg$rkXgyrW4RDGoDYrxBFD4H2DlSMEhP4h+Api1hXnGnFY"
```

Looks like a valid hash to me, but I guess it's not? What is this hash?

I guess it makes sense that it has to start with `$` since a valid hash looks like `    const _hash = "$argon2i$v=19$m=4096,t=3,p=1$EAoczTxVki21JDfIZpTUxg$rkXgyrW4RDGoDYrxBFD4H2DlSMEhP4h+Api1hXnGnFY"`?

Should we look for "`$argon2i$`" instead of just the `$`?

PR #3700, opened 2021-06-30
fix: add timeouts to CI jobs

jsjoeio:
This PR adds timeouts to CI jobs to prevent them from running longer than reasonable.

Case: Pre-build checks ran for [1h 51m today](https://github.com/cdr/code-server/actions/runs/987241200) because I didn't close an HTTP server in a unit test.

Fixes N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,8 +16,9 @@ on:
 
 jobs:
   prebuild:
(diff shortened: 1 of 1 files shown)


kylecarbs (approved):

jawnsy (approved):
Nice! Timeouts are always good.

PR #3707, opened 2021-07-01
fix: update install script for alpine

### LINKED ISSUE #3421: Fix install script to detect Alpine and suggest installing via npm/yarn
As a user running Alpine, the install script should exit and tell me to install using npm/yarn per [the docs](https://github.com/cdr/code-server/blob/main/docs/install.md#yarn-npm).

Related: 
- https://github.com/cdr/code-server/issues/3413
- https://github.com/cdr/code-server/discussions/2261
- https://github.com/cdr/code-server/discussions/2139

jsjoeio:
This PR fixes the install script to work on Alpine 64-bit. Our `arch` function doesn't recognize 32-bit Alpine, which fallsback to `npm`/`yarn` install (which is why it works on 32-bit Alpine), but tries to use the standalone install for 64-bit, which doesn't work.

This fixes that.

Fixes #3421

Also, thank you to @b-trav for filing an issue about the broken links in `install.sh`!

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -58,7 +58,7 @@ Usage:
   - If Homebrew is not installed it will install the latest standalone release
     into ~/.local
 
-- For FreeBSD, it will install the npm package with yarn or npm.
+- For FreeBSD or Alpine, it will install the npm package with yarn or npm.
 
 - If ran on an architecture with no releases, it will install the
   npm package with yarn or npm.
@@ -234,6 +234,17 @@ main() {
       exit 1
     fi
     echoh "No precompiled releases available for $OS."
+    install_npm
+    return
+  fi
+
+  if [ "$OS" = "linux" ] && [ "$(distro)" = "alpine" ]; then
+    if [ "$METHOD" = standalone ]; then
+      echoerr "No precompiled releases available for alpine."
+      echoerr 'Please rerun without the "--method standalone" flag to install from npm.'
+      exit 1
+    fi
+    echoh "No precompiled releases available for alpine."
     install_npm
     return
   fi
@@ -471,6 +482,11 @@ distro() {
 
       echo "$ID"
     )
+    return
+  fi
+
+  if [ -f /etc/alpine-release ]; then
+    echo "alpine"
     return
   fi
 }
--- a/install.sh
+++ b/install.sh
@@ -2,7 +2,7 @@
 set -eu
 
 # code-server's automatic install script.
-# See https://github.com/cdr/code-server/blob/main/docs/install.md
+# See https://github.com/cdr/code-server/blob/main/docs/install
 
 usage() {
   arg0="$0"
@@ -67,7 +67,7 @@ Usage:
 
 It will cache all downloaded assets into ~/.cache/code-server
 
-More installation docs are at https://github.com/cdr/code-server/blob/main/docs/install.md
+More installation docs are at https://coder.com/docs/code-server/v3.10.2/install
 EOF
 }
 
@@ -430,7 +430,7 @@ install_npm() {
   echoh
   echoerr "Please install npm or yarn to install code-server!"
   echoerr "You will need at least node v12 and a few C dependencies."
-  echoerr "See the docs https://github.com/cdr/code-server/blob/v3.10.2/docs/install.md#yarn-npm"
+  echoerr "See the docs https://coder.com/docs/code-server/v3.10.2/install#yarn-npm"
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


jsjoeio:
I don't have access to Alpine 64-bit. @code-asher can you test? (Or is there a way for me to get access to some VM running Alpine 64-bit?

jawnsy (approved):
LGTM!

install.sh
```diff
@@ -473,6 +484,11 @@ distro() {
     )
     return
   fi
+
+  if [ -f /etc/alpine-release ]; then
```
jawnsy, line 488: It's too big of a change for this, I just wanted to mention that there's a standard file that contains distro info, `os-release`, which also seems to work on Alpine:

```
15:17 coder@jawnsy-m ~/projects/m $ docker run --rm -it fedora cat /etc/os-release | grep "^ID="
ID=fedora
15:17 coder@jawnsy-m ~/projects/m $ docker run --rm -it ubuntu cat /etc/os-release | grep "^ID="
ID=ubuntu
15:17 coder@jawnsy-m ~/projects/m $ docker run --rm -it debian cat /etc/os-release | grep "^ID="
ID=debian
15:17 coder@jawnsy-m ~/projects/m $ docker run --rm -it alpine cat /etc/os-release | grep "^ID="
ID=alpine
```

Here's the info about it: https://www.freedesktop.org/software/systemd/man/os-release.html

There's also a doc link earlier in the file here, do we want to use the new docs site instead? https://github.com/cdr/code-server/blob/5a6af177256056e316d0056b291fb3a8159c5776/install.sh#L70

code-asher: Ooo we use that a few lines up actually. @jsjoeio let's expand the case statement to include `alpine` instead of checking `/etc/alpine-release`. Actually I'm not sure why we need the case at all. Maybe let's just return whatever ID is in that file.

code-asher: Wait I misread. We only have the case for ID_LIKE then we fall back to ID. So this should already be returning `alpine` which means no changes are necessary here.

jsjoeio: > There's also a doc link earlier in the file here, do we want to use the new docs site instead?

Good catch! I'll update those.

jsjoeio: > I just wanted to mention that there's a standard file that contains distro info, os-release, which also seems to work on Alpine

TIL that's a thing! Thanks for the tip @jawnsy  🙌

> So this should already be returning alpine which means no changes are necessary in distro (can remove the /etc/alpine-release block).

Ah, didn't realize that. Sweet! Less code to add. Removed.

install.sh
```diff
@@ -419,7 +430,7 @@ install_npm() {
   echoh
   echoerr "Please install npm or yarn to install code-server!"
   echoerr "You will need at least node v12 and a few C dependencies."
-  echoerr "See the docs https://github.com/cdr/code-server/blob/v3.10.2/docs/install.md#yarn-npm"
+  echoerr "See the docs https://coder.com/docs/code-server/v3.10.2/install#yarn-npm"
```
jawnsy, line 433: Are you planning to bump the version number each release, or should we link to `/latest/` here?

jawnsy (approved):

install.sh
jsjoeio: > If so, I think we should replace latest with the version number in this case actually so that old docs will link to the correct branch.

I think @bpmct and @BrunoQuaresma thought we should use the version ([source](https://github.com/cdr/code-server/pull/3684#discussion_r659870342)).

We do have a script called `release-prep.sh` which handles updating these links for us. See [here](https://github.com/cdr/code-server/blob/main/ci/build/release-prep.sh#L81-L84)

install.sh
```diff
@@ -2,7 +2,7 @@
 set -eu
 
 # code-server's automatic install script.
-# See https://github.com/cdr/code-server/blob/main/docs/install.md
+# See https://github.com/cdr/code-server/blob/main/docs/install
```
code-asher, line 5: Should this also go to coder.com? Without `.md` it seems to 404.

install.sh
BrunoQuaresma: Yes, we want to keep the version so the script `release-prep` handle the update for us.

install.sh
jsjoeio: Dang it. I did a serach for `.md` and must have accidentally done a search and replace. I'll fix.

jsjoeio: https://github.com/cdr/code-server/pull/3711

Issue #3585, opened 2021-06-09
Install script tries to use root if install destination does not yet exist even in standalone mode.


nugend:
https://github.com/cdr/code-server/blob/d8c3ba6a179d185aa42ceb6761dd8a2be339d68f/install.sh#L378

The '-w' test fails if  "$STANDALONE_INSTALL_PREFIX" doesn't exist. Since this variable defaults to $HOME/.local the user may not have created that hidden directory before attempting to install.

jsjoeio:
Ah, nice catch! Thanks for pointing this out and raising an issue.

PR #3712, opened 2021-07-02
Make installer work on any os/arch and add installer tests

code-asher:


.github/workflows/ci.yaml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: CI
```
jsjoeio, line 1: should we revert back the `Pre-Build` change? 😅

code-asher: Ahaha I'll go ahead and do that!

jsjoeio:
Will this fix https://github.com/cdr/code-server/issues/3585?

code-asher:
Nope but I'll add a fix.

.github/workflows/installer.yml
```diff
@@ -0,0 +1,57 @@
+name: Installer integration
```
jawnsy, line 1: I think we should avoid adding too many workflows, rather than adding jobs to an existing workflow, because the workflows show up completely separately in the UI. This is why the monorepo uses two very large workflow files, and `files_changed` to detect whether to skip jobs.

.github/workflows/scripts.yml
```diff
@@ -0,0 +1,27 @@
...
+      - "installer.sh"
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  test:
+    name: Run script unit tests
+    runs-on: ubuntu-latest
+    # This runs on Alpine to make sure we're testing with actual sh.
```
jawnsy, line 17: I'm confused, doesn't the script force things to run as bash due to the shebang? Also, Ubuntu should include Bourne shell (e.g. dash) too

ci/dev/test-scripts.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+main() {
+  cd "$(dirname "$0")/../.."
```
jawnsy, line 5: If you're trying to identify the root of the repository, this might help:

`PROJECT_ROOT="$(git rev-parse --show-toplevel)"`

Then you can `cd $PROJECT_ROOT`, though I'd suggest doing pushd/popd like this:

```
pushd $PROJECT_ROOT
  bats ./test/scripts
popd
```

You can also redirect to /dev/null if you want to suppress the default messages that pushd/popd will emit

.github/workflows/ci.yaml
```diff
@@ -1,4 +1,4 @@
-name: ci
+name: Build
```
jsjoeio, line 1: I'm all in favor of the change. Will it break anything though?

.github/workflows/ci.yaml
```diff
@@ -16,7 +16,7 @@ on:
 
 jobs:
   prebuild:
-    name: Pre-Build checks
+    name: Pre-build checks
```
jsjoeio, line 19: yesssss!

.github/workflows/installer.yml
jsjoeio: > because the workflows show up completely separately in the UI

What's the downside of that? 

> This is why the monorepo uses two very large workflow files

Isn't it easy to read/maintain if you have separate workflow files? Then they're isolated from each other and you're less likely to break something if you modify something in the workflow.

.github/workflows/installer.yml
```diff
@@ -0,0 +1,57 @@
...
+    container: "alpine:3.14"
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Install curl
+        run: apk add curl
+
+      - name: Add user
+        run: adduser coder --disabled-password
```
jsjoeio, line 39: Why do we have to add a user and why do we use `--disabled-password`?

.github/workflows/installer.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      # Standalone should work without root.
+      - name: Test standalone to a non-existent prefix
+        run: su coder -c "./install.sh --method standalone --prefix /tmp/does/not/yet/exist"
```
jsjoeio, line 43: TIL that `su` makes you a superuser? (according to explainshell)

test/scripts/install.bats
```diff
@@ -0,0 +1,163 @@
...
+
+SCRIPT_NAME="install.sh"
+SCRIPT="$BATS_TEST_DIRNAME/../../$SCRIPT_NAME"
+
+# Override version so it doesn't have to curl and to avoid caching in case the
+# user already has the latest version installed.
+export VERSION="9999.99.9"
+
+function should-use-deb() {
+  DISTRO=$1 ARCH=$2 OS=linux run "$SCRIPT" --dry-run
```
jsjoeio, line 11: NICE! I love this API you've come up with for testing across various environments - really clean 👏

jsjoeio (approved):
Nice work! Really happy you laid a foundation for writing script tests. Maybe this'll motivate me to write some for our other scripts.

.github/workflows/installer.yml
jawnsy: It's a matter of personal preference, and your call to make of course. I just find this:

image

to make it more difficult to tell which of the tests passed/failed, than this view:

image

GitHub Actions also supports inter-job dependencies but it's harder to do inter-workflow dependencies. It could be that the tradeoffs we were making are different from the tradeoffs you're making, of course, so you'll have to do what makes more sense for you :grin:

jsjoeio: Ah, I see what you mean now. Thanks for the screenshots! Yeah, I can see why that might be easier. I guess we'll leave it to @code-asher then to make the call!

.github/workflows/installer.yml
jawnsy: My guess is that this is running as root by default, so `su coder` means "run as coder user" and the `-c` defines the shell command to run. This explanation looks decent: https://unix.stackexchange.com/questions/257698/how-are-setuid-suid-sudo-and-su-all-related

jsjoeio: > The most common use of su is to become root

oh! that makes more sense now. thanks for the link and explanation!

.github/workflows/installer.yml
code-asher: My first impression was that I liked them split up weirdly enough. But it's definitely possible we may find that to be unwieldy so I think we could see how this goes and then if ends up being frustrating we can combine them.

.github/workflows/scripts.yml
code-asher: The installer shebang is `#!/bin/sh` but since that's often a symlink to bash/dash/etc I've used Alpine to make sure it's really using `sh` to catch any bashisms that might sneak in there.

ci/dev/test-scripts.sh
code-asher: True! I tend to favor `dirname "$0"` because I can run it without having to `cd` to the repository first. It feels a little more flexible overall to me (at the cost of having to update the `cd` if the file is moved of course).

.github/workflows/ci.yaml
code-asher: Nothing that I can think of but hopefully we don't find out the hard way hahahaha

.github/workflows/installer.yml
code-asher: It's to make sure the script works without being ran as root (to make sure we don't have permission issues). It's basically to make sure https://github.com/cdr/code-server/issues/3585 doesn't happen. I linked that in the commit message but it might make sense to add as a comment.

.github/workflows/scripts.yml
jawnsy: Ah! Good point, I was looking at `test-scripts.sh`, which uses bash explicitly: `#!/usr/bin/env bash`

This makes sense then. I think Ubuntu should be using dash for /bin/sh by default, though I'm not sure if GitHub Actions does some customization to the environment:

```
$ docker run --rm -it ubuntu:latest
root@081f1b5d886d:/# which sh
/usr/bin/sh
root@081f1b5d886d:/# ls -al /usr/bin/sh
lrwxrwxrwx 1 root root 4 Jul 18  2019 /usr/bin/sh -> dash
```

There's also a handy `checkbashisms` script in `devscripts` (install it with apt) that you may find useful. I use it in my dotfiles build: https://github.com/jawnsy/dotfiles/blob/0a27557f15aba599056171abc5ff936e7b530d13/.github/workflows/build.yml#L49

.github/workflows/ci.yaml
jsjoeio: lolol only one way to find out

.github/workflows/scripts.yml
code-asher: Oh that's dope! I'll definitely add that.

PR #3727, opened 2021-07-06
fix(ci): update codeql upload-sarif to v1.0.4

### LINKED ISSUE #3517: Revert code-ql-action/upload-sarif
Revert https://github.com/cdr/code-server/issues/3459

We have a short term workaround thanks to @adityasharad 

> Short term workaround: please try removing the TAR file /release-images/code-server-amd64-3.10.2.tar (and others like it) from the workspace before running the upload step.

UPDATE: a PR has been merged https://github.com/github/codeql-action/pull/550

So we should be able to re-enable without using the workaround.

jsjoeio:
This re-enables the `trivy-scan-image` job in `ci` which is used to scan our Docker image for vulnerabilities. We had to disable it due to this issue in codeql-action/upload-sarif: https://github.com/github/codeql-action/issues/528

Fixes #3517

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -433,9 +433,6 @@ jobs:
   trivy-scan-image:
     runs-on: ubuntu-20.04
     needs: docker-images
-    # NOTE@jsjoeio: disabling due to a memory issue upstream
-    # See: https://github.com/github/codeql-action/issues/528
-    if: 1 == 2
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
@@ -462,7 +459,7 @@ jobs:
         run: cat trivy-image-results.sarif && ls -l trivy-image-results.sarif
 
       - name: Upload Trivy scan results to GitHub Security tab
-        uses: github/codeql-action/upload-sarif@v1
+        uses: github/codeql-action/upload-sarif@v1.0.4
         with:
           sarif_file: "trivy-image-results.sarif"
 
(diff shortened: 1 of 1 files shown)


jsjoeio:
image

Weird, I swear it worked before but now I'm getting this error. Thinking out loud here...

1. We build docker images and then upload them as artifacts under `release-images`
2. In `trivy-scan-image`, we download those release images
3. Then we run the Trivy vulnerability scanner and pass in the `.tar` file as input 🤔 Is this the correct thing to do? Looking under [inputs](https://github.com/aquasecurity/trivy-action#inputs) in the docs, yes, `.tar` is expected. 
4. Looks like it wants a `manifest.json` in the Docker image, which we may or may not have...
5. And it also wants `index.json` I think? hmm

jsjoeio:
@oxy any ideas?

jawnsy:
@jsjoeio You could also push the image to GitHub's container registry and then pull/scan from there, but we'd have to prune old images or else the cost will add up

jsjoeio:
Hmm...I've never used GitHub's container registry. (I would be worried that someone forgets to prune or something breaks with that flow and we get charged a lot 😂)

I'm open to it though! Would you recommend that instead?

jsjoeio:
Possibly related: https://github.com/aquasecurity/trivy/issues/1080

jsjoeio:
> GitHub's container registry

I looked into this and [pushing is easy](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images). 

I found [this action](https://github.com/bots-house/ghcr-delete-image-action) for deleting by tag

We could probably combine the three:
1. Push to GitHub container registry
2. Use with trivy-scan-image
3. Delete from GitHub container registry

jsjoeio:
We also need to make sure the [repo has access](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility) to the container packages.

jsjoeio:
After thinking about this a bit more, I've decided to put this on hold. It would take some time to set everything up for the GitHub Container Registry. I'd rather wait to see if we get a response on https://github.com/aquasecurity/trivy/issues/1080 before taking on the work to make this happen.

PR #3743, opened 2021-07-07
feat: add more tests for browser/pages/vscode

jsjoeio:
This PR adds more tests for `browser/pages/vscode.ts`.

Fixes N/A

## Notes

The main purpose of this PR is to test this piece of code
```js
document.body.style.background = JSON.parse(localStorage.getItem("colorThemeData")!).colorMap["editor.background"]
```

This code grabs the `colorThemeData` from `localStorage` to get the editor background color based on the user's VSCode theme and then sets the background of the `body` to that color.

### DIFF
--- a/src/browser/pages/vscode.ts
+++ b/src/browser/pages/vscode.ts
@@ -95,8 +95,64 @@ try {
   console.error(error)
 }
 
+export function setBodyBackgroundToThemeBackgroundColor(document: Document, localStorage: Storage) {
+  const errorMsgPrefix = "[vscode]"
+
+  if (!document) {
+    throw new Error(`${errorMsgPrefix} Could not set body background to theme background color. Document is undefined.`)
+  }
+
+  if (!localStorage) {
+    throw new Error(
+      `${errorMsgPrefix} Could not set body background to theme background color. localStorage is undefined.`,
+    )
+  }
+
+  const colorThemeData = localStorage.getItem("colorThemeData")
+
+  if (!colorThemeData) {
+    throw new Error(
+      `${errorMsgPrefix} Could not set body background to theme background color. Could not find colorThemeData in localStorage.`,
+    )
+  }
+
+  let _colorThemeData
+  try {
+    // We wrap this JSON.parse logic in a try/catch
+    // because it can throw if the JSON is invalid.
+    // and instead of throwing a random error
+    // we can throw our own error, which will be more helpful
+    // to the end user.
+    _colorThemeData = JSON.parse(colorThemeData)
+  } catch {
+    throw new Error(
+      `${errorMsgPrefix} Could not set body background to theme background color. Could not parse colorThemeData from localStorage.`,
+    )
+  }
+
+  const hasColorMapProperty = Object.prototype.hasOwnProperty.call(_colorThemeData, "colorMap")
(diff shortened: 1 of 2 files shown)


src/browser/pages/vscode.ts
```diff
@@ -95,8 +95,64 @@ try {
   console.error(error)
 }
 
+export function setBodyBackgroundToThemeBackgroundColor(document: Document, localStorage: Storage) {
```
jsjoeio, line 98: At first, I didn't like the function signature — passing in `document` and `localStorage` — but then I justified it because:
- this function is only called in one place 
- it makes it _a lot_ easier to test when `document` or `localStorage` is missing

src/browser/pages/vscode.ts
```diff
@@ -95,8 +95,64 @@ try {
...
+
+  const editorBgColor = _colorThemeData.colorMap["editor.background"]
+
+  if (!editorBgColor) {
+    throw new Error(
+      `${errorMsgPrefix} Could not set body background to theme background color. colorThemeData.colorMap["editor.background"] is undefined.`,
+    )
+  }
+
+  document.body.style.background = editorBgColor
```
jsjoeio, line 148: Here we assume that `body` is part of the `document` but I think it would be really rare/unusual if this weren't the case. I think we can leave as is and if we run into the rare issue, I can modify down the road.

src/browser/pages/vscode.ts
```diff
@@ -95,8 +95,64 @@ try {
...
+  return null
+}
+
 try {
-  document.body.style.background = JSON.parse(localStorage.getItem("colorThemeData")!).colorMap["editor.background"]
+  setBodyBackgroundToThemeBackgroundColor(document, localStorage)
 } catch (error) {
-  // Oh well.
+  console.error("Something went wrong setting the body background to the theme background color.")
+  console.error(error)
```
jsjoeio, line 157: `setBodyBackgroundToThemeBackgroundColor` throws specific errors, which will be caught by our `catch` block and logged to the console in the browser.

src/browser/pages/vscode.ts
jawnsy: makes sense to me, this also allows you to test this by mocking document/localStorage, and avoids global state

test/unit/browser/vscode.test.ts
```diff
@@ -58,4 +62,117 @@ describe("vscode", () => {
       document.body.removeChild(mockElement)
     })
   })
+  describe("setBodyBackgroundToThemeBackgroundColor", () => {
+    beforeEach(() => {
+      // We need to set the url in the JSDOM constructor
+      // to prevent this error "SecurityError: localStorage is not available for opaque origins"
+      // See: https://github.com/jsdom/jsdom/issues/2304#issuecomment-622314949
+      const { window } = new JSDOM("", { url: "http://localhost" })
```
jawnsy, line 70: I guess this is because Node.js doesn't have `window`?

jawnsy (approved):
LGTM!

test/unit/browser/vscode.test.ts
jsjoeio: Exactly! `window` is only available in browser environments because it's a browser global

PR #3749, opened 2021-07-08
fix(dependabot): ignore major updates for node

jsjoeio:
This PR tells dependabot to ignore major updates for Node.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,7 +19,7 @@ updates:
       time: "11:00"
     ignore:
       - dependency-name: "@types/node"
-        versions: ["15.x", "14.x", "13.x"]
+        update-types: ["version-update:semver-major"]
       - dependency-name: "xdg-basedir"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -19,7 +19,7 @@ updates:
       time: "11:00"
     ignore:
       - dependency-name: "@types/node"
-        versions: ["15.x", "14.x", "13.x"]
+        versions: ["16.x", 15.x", "14.x", "13.x"]
```
jawnsy, line 22: wonder if it's better to ignore all major version updates? https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/

jawnsy (approved):

.github/dependabot.yml
jsjoeio: Definitely! TIL! I'll change now

PR #3760, opened 2021-07-09
feat: add registerRequireOnSelf on function

jsjoeio:
This adds a new function called `registerRequireOnSelf` to `vscode.ts`. We do this so that we can extract this logic into a function, which makes it easier to test. 

## Screenshot

https://user-images.githubusercontent.com/3806031/125342840-35fec800-e30a-11eb-99e0-7c9c11409e6a.mov

Fixes N/A

TODOs
- [ ] add tests for registerLoadBundleOnNlsConfig
- [x] add follow-up item for "This might be a lot of work though, not sure. We also might need to load loader.js first and then do require.config(config) instead." https://github.com/cdr/code-server/issues/3780
- [x] add follow-up item: https://github.com/cdr/code-server/issues/3781

### DIFF
--- a/src/browser/pages/vscode.ts
+++ b/src/browser/pages/vscode.ts
@@ -1,10 +1,8 @@
-import { getOptions } from "../../common/util"
+import { getOptions, Options } from "../../common/util"
 import "../register"
 
-const options = getOptions()
-
-// TODO: Add proper types.
-/* eslint-disable @typescript-eslint/no-explicit-any */
+// TODO@jsjoeio: Add proper types.
+type FixMeLater = any
 
 // NOTE@jsjoeio
 // This lives here ../../../lib/vscode/src/vs/base/common/platform.ts#L106
@@ -19,7 +17,20 @@ type NlsConfiguration = {
   _resolvedLanguagePackCoreLocation?: string
   _corruptedFile?: string
   _languagePackSupport?: boolean
-  loadBundle?: any
+  loadBundle?: FixMeLater
+}
+
+/**
+ * Helper function to create the path to the bundle
+ * for getNlsConfiguration.
+ */
+export function createBundlePath(_resolvedLanguagePackCoreLocation: string, bundle: string) {
+  // NOTE@jsjoeio - this comment was here before me
+  // Refers to operating systems that use a different path separator.
+  // Probably just Windows but we're not sure if "/" breaks on Windows
+  // so we'll leave it alone for now.
+  // FIXME: Only works if path separators are /.
+  return _resolvedLanguagePackCoreLocation + "/" + bundle.replace(/\//g, "!") + ".nls.json"
 }
 
 /**
@@ -30,14 +41,10 @@ type NlsConfiguration = {
  *
  * Make sure to wrap this in a try/catch block when you call it.
  **/
-export function getNlsConfiguration(document: Document) {
+export function getNlsConfiguration(_document: Document, base: string) {
   const errorMsgPrefix = "[vscode]"
-  const nlsConfigElement = document?.getElementById(nlsConfigElementId)
-  const nlsConfig = nlsConfigElement?.getAttribute("data-settings")
-
-  if (!document) {
-    throw new Error(`${errorMsgPrefix} Could not parse NLS configuration. document is undefined.`)
-  }
+  const nlsConfigElement = _document?.getElementById(nlsConfigElementId)
+  const dataSettings = nlsConfigElement?.getAttribute("data-settings")
 
   if (!nlsConfigElement) {
     throw new Error(
@@ -45,27 +52,34 @@ export function getNlsConfiguration(document: Document) {
     )
   }
 
-  if (!nlsConfig) {
+  if (!dataSettings) {
     throw new Error(
       `${errorMsgPrefix} Could not parse NLS configuration. Found nlsConfigElement but missing data-settings attribute.`,
     )
   }
 
-  return JSON.parse(nlsConfig) as NlsConfiguration
-}
+  const nlsConfig = JSON.parse(dataSettings) as NlsConfiguration
 
-try {
-  const nlsConfig = getNlsConfiguration(document)
   if (nlsConfig._resolvedLanguagePackCoreLocation) {
-    const bundles = Object.create(null)
-    nlsConfig.loadBundle = (bundle: any, _language: any, cb: any): void => {
+    // NOTE@jsjoeio
+    // Not sure why we use Object.create(null) instead of {}
+    // They are not the same
+    // See: https://stackoverflow.com/a/15518712/3015595
+    // We copied this from ../../../lib/vscode/src/bootstrap.js#L143
+    const bundles: {
+      [key: string]: string
+    } = Object.create(null)
+
+    type LoadBundleCallback = (_: undefined, result?: string) => void
+
+    nlsConfig.loadBundle = (bundle: string, _language: string, cb: LoadBundleCallback): void => {
       const result = bundles[bundle]
       if (result) {
         return cb(undefined, result)
       }
       // FIXME: Only works if path separators are /.
-      const path = nlsConfig._resolvedLanguagePackCoreLocation + "/" + bundle.replace(/\//g, "!") + ".nls.json"
-      fetch(`${options.base}/vscode/resource/?path=${encodeURIComponent(path)}`)
+      const path = createBundlePath(nlsConfig._resolvedLanguagePackCoreLocation || "", bundle)
+      fetch(`${base}/vscode/resource/?path=${encodeURIComponent(path)}`)
         .then((response) => response.json())
         .then((json) => {
           bundles[bundle] = json
@@ -74,17 +88,61 @@ try {
         .catch(cb)
     }
   }
-  ;(self.require as any) = {
+
+  return nlsConfig
+}
+
+type GetLoaderParams = {
+  nlsConfig: NlsConfiguration
+  options: Options
+  _window: Window
+}
+
+/**
+ * Link to types in the loader source repo
+ * https://github.com/microsoft/vscode-loader/blob/main/src/loader.d.ts#L280
+ */
+type Loader = {
+  baseUrl: string
+  recordStats: boolean
+  // TODO@jsjoeio: There don't appear to be any types for trustedTypes yet.
+  trustedTypesPolicy: FixMeLater
+  paths: {
+    [key: string]: string
+  }
+  "vs/nls": NlsConfiguration
+}
+
+/**
+ * A helper function which creates a script url if the value
+ * is valid.
+ *
+ * Extracted into a function to make it easier to test
+ */
+export function _createScriptURL(value: string, origin: string): string {
+  if (value.startsWith(origin)) {
+    return value
+  }
+  throw new Error(`Invalid script url: ${value}`)
+}
+
+/**
+ * A helper function to get the require loader
+ *
+ * This used by VSCode/code-server
+ * to load files.
+ *
+ * We extracted the logic into a function so that
+ * it's easier to test.
+ **/
+export function getConfigurationForLoader({ nlsConfig, options, _window }: GetLoaderParams) {
+  const loader: Loader = {
     // Without the full URL VS Code will try to load file://.
     baseUrl: `${window.location.origin}${options.csStaticBase}/lib/vscode/out`,
     recordStats: true,
-    // TODO: There don't appear to be any types for trustedTypes yet.
-    trustedTypesPolicy: (window as any).trustedTypes?.createPolicy("amdLoader", {
+    trustedTypesPolicy: (_window as FixMeLater).trustedTypes?.createPolicy("amdLoader", {
       createScriptURL(value: string): string {
-        if (value.startsWith(window.location.origin)) {
-          return value
-        }
-        throw new Error(`Invalid script url: ${value}`)
+        return _createScriptURL(value, window.location.origin)
       },
     }),
     paths: {
@@ -100,25 +158,16 @@ try {
     },
     "vs/nls": nlsConfig,
   }
-} catch (error) {
-  console.error(error)
-  /* Probably fine. */
+
+  return loader
 }
 
-export function setBodyBackgroundToThemeBackgroundColor(document: Document, localStorage: Storage) {
+/**
+ * Sets the body background color to match the theme.
+ */
+export function setBodyBackgroundToThemeBackgroundColor(_document: Document, _localStorage: Storage) {
   const errorMsgPrefix = "[vscode]"
-
-  if (!document) {
-    throw new Error(`${errorMsgPrefix} Could not set body background to theme background color. Document is undefined.`)
-  }
-
-  if (!localStorage) {
-    throw new Error(
-      `${errorMsgPrefix} Could not set body background to theme background color. localStorage is undefined.`,
-    )
-  }
-
-  const colorThemeData = localStorage.getItem("colorThemeData")
+  const colorThemeData = _localStorage.getItem("colorThemeData")
 
   if (!colorThemeData) {
     throw new Error(
@@ -155,14 +204,48 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
     )
   }
 
-  document.body.style.background = editorBgColor
+  _document.body.style.background = editorBgColor
 
   return null
 }
 
+/**
+ * A helper function to encapsulate all the
+ * logic used in this file.
+ *
+ * We purposely include all of this in a single function
+ * so that it's easier to test.
+ */
+export function main(_document: Document | undefined, _window: Window | undefined, _localStorage: Storage | undefined) {
+  if (!_document) {
+    throw new Error(`document is undefined.`)
+  }
+
+  if (!_window) {
+    throw new Error(`window is undefined.`)
+  }
+
+  if (!_localStorage) {
+    throw new Error(`localStorage is undefined.`)
+  }
+
+  const options = getOptions()
+  const nlsConfig = getNlsConfiguration(_document, options.base)
+
+  const loader = getConfigurationForLoader({
+    nlsConfig,
+    options,
+    _window,
+  })
+
+  ;(self.require as unknown as Loader) = loader
+
+  setBodyBackgroundToThemeBackgroundColor(_document, _localStorage)
+}
+
 try {
-  setBodyBackgroundToThemeBackgroundColor(document, localStorage)
+  main(document, window, localStorage)
 } catch (error) {
-  console.error("Something went wrong setting the body background to the theme background color.")
+  console.error("[vscode] failed to initialize VS Code")
   console.error(error)
 }
--- a/test/unit/browser/pages/vscode.test.ts
+++ b/test/unit/browser/pages/vscode.test.ts
@@ -5,48 +5,47 @@ import { JSDOM } from "jsdom"
 import {
   getNlsConfiguration,
   nlsConfigElementId,
+  getConfigurationForLoader,
   setBodyBackgroundToThemeBackgroundColor,
+  _createScriptURL,
+  main,
+  createBundlePath,
 } from "../../../../src/browser/pages/vscode"
 
 describe("vscode", () => {
   describe("getNlsConfiguration", () => {
+    let _document: Document
+
     beforeEach(() => {
-      const { window } = new JSDOM()
-      global.document = window.document
+      // We use underscores to not confuse with global values
+      const { window: _window } = new JSDOM()
+      _document = _window.document
     })
 
-    it("should throw an error if Document is undefined", () => {
-      const errorMsgPrefix = "[vscode]"
-      const errorMessage = `${errorMsgPrefix} Could not parse NLS configuration. document is undefined.`
-
-      expect(() => {
-        getNlsConfiguration(undefined as any as Document)
-      }).toThrowError(errorMessage)
-    })
     it("should throw an error if no nlsConfigElement", () => {
       const errorMsgPrefix = "[vscode]"
       const errorMessage = `${errorMsgPrefix} Could not parse NLS configuration. Could not find nlsConfigElement with id: ${nlsConfigElementId}`
 
       expect(() => {
-        getNlsConfiguration(document)
+        getNlsConfiguration(_document, "")
       }).toThrowError(errorMessage)
     })
     it("should throw an error if no nlsConfig", () => {
-      const mockElement = document.createElement("div")
+      const mockElement = _document.createElement("div")
       mockElement.setAttribute("id", nlsConfigElementId)
-      document.body.appendChild(mockElement)
+      _document.body.appendChild(mockElement)
 
       const errorMsgPrefix = "[vscode]"
       const errorMessage = `${errorMsgPrefix} Could not parse NLS configuration. Found nlsConfigElement but missing data-settings attribute.`
 
       expect(() => {
-        getNlsConfiguration(document)
+        getNlsConfiguration(_document, "")
       }).toThrowError(errorMessage)
 
-      document.body.removeChild(mockElement)
+      _document.body.removeChild(mockElement)
     })
     it("should return the correct configuration", () => {
-      const mockElement = document.createElement("div")
+      const mockElement = _document.createElement("div")
       const dataSettings = {
         first: "Jane",
         last: "Doe",
@@ -54,22 +53,52 @@ describe("vscode", () => {
 
       mockElement.setAttribute("id", nlsConfigElementId)
       mockElement.setAttribute("data-settings", JSON.stringify(dataSettings))
-      document.body.appendChild(mockElement)
-      const actual = getNlsConfiguration(global.document)
+      _document.body.appendChild(mockElement)
+      const actual = getNlsConfiguration(_document, "")
 
       expect(actual).toStrictEqual(dataSettings)
 
-      document.body.removeChild(mockElement)
+      _document.body.removeChild(mockElement)
+    })
+    it("should return have loadBundle property if _resolvedLangaugePackCoreLocation", () => {
+      const mockElement = _document.createElement("div")
+      const dataSettings = {
+        locale: "en",
+        availableLanguages: ["en", "de"],
+        _resolvedLanguagePackCoreLocation: "./",
+      }
+
+      mockElement.setAttribute("id", nlsConfigElementId)
+      mockElement.setAttribute("data-settings", JSON.stringify(dataSettings))
+      _document.body.appendChild(mockElement)
+      const nlsConfig = getNlsConfiguration(_document, "")
+
+      expect(nlsConfig._resolvedLanguagePackCoreLocation).not.toBe(undefined)
+      expect(nlsConfig.loadBundle).not.toBe(undefined)
+
+      _document.body.removeChild(mockElement)
+    })
+  })
+  describe("createBundlePath", () => {
+    it("should return the correct path", () => {
+      const _resolvedLangaugePackCoreLocation = "./languages"
+      const bundle = "/bundle.js"
+      const expected = "./languages/!bundle.js.nls.json"
+      const actual = createBundlePath(_resolvedLangaugePackCoreLocation, bundle)
+      expect(actual).toBe(expected)
     })
   })
   describe("setBodyBackgroundToThemeBackgroundColor", () => {
+    let _document: Document
+    let _localStorage: Storage
+
     beforeEach(() => {
       // We need to set the url in the JSDOM constructor
       // to prevent this error "SecurityError: localStorage is not available for opaque origins"
       // See: https://github.com/jsdom/jsdom/issues/2304#issuecomment-622314949
-      const { window } = new JSDOM("", { url: "http://localhost" })
-      global.document = window.document
-      global.localStorage = window.localStorage
+      const { window: _window } = new JSDOM("", { url: "http://localhost" })
+      _document = _window.document
+      _localStorage = _window.localStorage
     })
     it("should return null", () => {
       const test = {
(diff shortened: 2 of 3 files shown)


src/browser/pages/vscode.ts
```diff
@@ -54,6 +54,84 @@ export function getNlsConfiguration(document: Document) {
...
+
+  if (!options || !options.csStaticBase) {
+    throw new Error(
+      `${errorMsgPrefix} Could not register require on self. options or options.csStaticBase is undefined or missing.`,
+    )
+  }
+
+  if (!nlsConfig) {
+    throw new Error(`${errorMsgPrefix} Could not register require on self. nlsConfig is undefined.`)
+  }
```
code-asher, line 103: These should be handled by the type checker right?

src/browser/pages/vscode.ts
```diff
@@ -54,6 +54,84 @@ export function getNlsConfiguration(document: Document) {
...
+    },
+    "vs/nls": nlsConfig,
+  }
+
+  // TODO@jsjoeio
+  // I'm not sure how to properly type cast this
+  // This might be our best bet
+  // Source: https://stackoverflow.com/a/30740935
+  type FixMeLater = any
+  ;(self.require as FixMeLater) = requireOnSelf
```
code-asher, line 128: We could remove this and return the object to make the function free of side effects. Then we'd just have something like:

```
self.require = getRequire()
```

Although this function is basically just wrapping a simple object right? 

What do you think about doing `require("vscode")` and testing it that way? Or wrap the whole try/catch in a function called `configureLoader` or something like that?

jsjoeio: > We could remove this and return the object to make the function free of side effects. 

Even better, I like this!

> What do you think about doing require("vscode") and testing it that way?

Eh...I wouldn't say I'm in favor. I mean with this way, we're able to isolate this logic and test as a unit. Plus doing it this way, we have some better error handling so we know what went wrong if this fails. Writing a separate test which does that could be done later though! I would vote it's out of the scope of this. Thoughts?

> wrap the whole try/catch in a function

I thought about doing this — calling this function in a separate try/catch block, but we need `nlsConfig` which would be block-scoped to the other try/catch block, unless I move the declaration to a `let` and move it outside the try/catch block. 

> function called configureLoader

Are you suggesting renaming the current function to `configureLoader`? If so, I'm onboard!

code-asher: Yeah that's fair, I keep being hesitant about adding new API only for use in tests because it increases the area for future bugs (for example you can now call `registerRequireOnSelf` with the wrong arguments or in files where it doesn't make sense which you couldn't do before).

Maybe for all the browser files that run themselves we should adopt a pattern like this:
```javascript
function someFunc(document) {
}
function someOtherFunc(window) {
}

try {
  someFunc(document)
  someOtherFunc(window)
} catch (error) {
  console.error(error.message)
}
```

That way we minimize the amount of code that uses globals and needs to be tested via the `require` method. So in this case it would be something like:

```
try {
  const nlsConfig = getNLSConfiguration(document)
  const options = getOptions()
  configureLoader(self, window.location.origin, nlsConfig, options)
} catch (error) {
  console.error(error)
}
```

code-asher: BTW in case it helps I'm calling it things like `loaderConfig` because setting this object on `self.require` is what configures the loader and tells it how to load files that get imported.

src/browser/pages/vscode.ts
code-asher: For `nlsConfig` we already throw if unable to get it (`getNlsConfiguration`) and for `getOptions` we always return something even if getting the options fails so both those can be omitted.

It's not possible for `self` and `window` to be missing unless this is being loaded outside a browser context in which case nothing will work anyway haha

code-asher: Although those last two points are moot now with the latest changes! But I don't think we need to check that origin is undefined. If `window.location` exists `origin` will also exist so if somehow `window` is missing we will have already errored before getting here (`no such property location on window` or something like that).

code-asher: I would also argue that passing an undefined to a function that only accepts a string is an error on the caller's side so if we do really need to typecheck we could do it before calling the func. So for example:

```javascript
if (!window) {
  throw new Error()
}
registerRequireOnSelf(window)
```

Or, if we want to keep the typecheck inside the function we could change the types to `Window | undefined` to accurately reflect what the function accepts.

src/browser/pages/vscode.ts
jsjoeio: > Maybe for all the top-level browser files that run themselves we should adopt a pattern like this

I like the idea of this. That would make it cleaner, easier to maintain & test. I'll refactor to use this approach. 

> BTW in case it helps I'm calling it things like loaderConfig because setting this object on self.require is what configures the loader and tells it how to load files that get imported

Yes! Super helpful. Thank you for explaining that (I figured it was _like_ that but wasn't certain).

src/browser/pages/vscode.ts
jsjoeio: > For nlsConfig we already throw if unable to get it (getNlsConfiguration) and for getOptions we always return something even if getting the options fails so both those can be omitted.

Ah true! Good point. I'll remove those.

jsjoeio: > Oh are these typechecks for unit tests that don't load a browser context?

Yes exactly!

> If so that makes sense to me, maybe we can move them out in that case to the calling side

Ah I didn't think of that. Almost like an integration test (if i can stretch the definition that far lol) but I see what you mean. Almost like defining a context where "X" is missing, we should `catch` before we call the function. Smart! I'll refactor.

jsjoeio:
@code-asher refactored! I hope I understood your suggestions. Feel free to nitpick/correct again if I didn't!

jsjoeio:
> will decrease coverage by 0.67%.

Ah, that's super frustrating :( Before all the refactoring, code coverage had increased. And now that I removed a bunch, I guess it's gone down. Is what it is. 

@code-asher I'll mark as draft and add more tests, but do you mind reviewing still?

jsjoeio:
All in all, I think this will actually be a lot better than the first iteration.

src/browser/pages/vscode.ts
```diff
@@ -160,9 +156,72 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+ * A helper function to encapsulate all the
+ * logic used in this file.
+ *
+ * It does the following:
+ * - grabs the options using getOptions
+ * - gets the nlsConfig
+ * - registers loadBundle on nlsConfig
+ * - gets the loader
+ * - sets loader on self.require
+ * - sets the body background color to match the theme
```
code-asher, line 193: Minor nit, we might want to avoid comments that duplicate what the code is doing; they tend to become out of date quickly (someone updates the code but not the comment, etc). :sweat_smile:  

On another note, I think `main` is a pretty common pattern (at least in our repo it is) so feel free to avoid the justifications if you want! :smile:

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +177,151 @@ describe("vscode", () => {
...
+      }
+      const loader = getLoader({
+        origin: "localhost",
+        nlsConfig: nlsConfig,
+        options,
+      })
+
+      expect(loader).toStrictEqual({
+        baseUrl: "localhost//lib/vscode/out",
+        paths: {
```
code-asher, line 271: Testing the paths like this might not catch anything since we'd be copying any mistakes we already have.

The ideal might be to:
1. Ensure that each path resolves to a real Node module.
2. Check VS Code's workbench.html and make sure we have the same paths we do (maybe could do this later).

src/browser/pages/vscode.ts
```diff
@@ -54,39 +52,35 @@ export function getNlsConfiguration(document: Document) {
...
     recordStats: true,
-    // TODO: There don't appear to be any types for trustedTypes yet.
-    trustedTypesPolicy: (window as any).trustedTypes?.createPolicy("amdLoader", {
-      createScriptURL(value: string): string {
-        if (value.startsWith(window.location.origin)) {
-          return value
-        }
-        throw new Error(`Invalid script url: ${value}`)
-      },
-    }),
```
code-asher, line 89: `trustedTypesPolicy` was removed?

src/browser/pages/vscode.ts
```diff
@@ -160,9 +156,72 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+    origin: window.location.origin,
+  })
+
+  // TODO@jsjoeio
+  // I'm not sure how to properly type cast this
+  // This might be our best bet
+  // Source: https://stackoverflow.com/a/30740935
+  // This object on self.require is what configures the loader
+  // and tells it how to load files that get imported.
+  ;(self.require as FixMeLater) = loader
```
code-asher, line 218: Maybe we can copy down the `.d.ts` for the loader to get the right type.

src/browser/pages/vscode.ts
```diff
@@ -160,9 +156,72 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+
   setBodyBackgroundToThemeBackgroundColor(document, localStorage)
+}
+
+try {
+  main()
```
code-asher, line 224: Thoughts on moving globals to the top? So `main(document, window)`, etc? That way we don't have to mess with globals in the unit test.

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +177,151 @@ describe("vscode", () => {
       localStorage.removeItem("colorThemeData")
     })
   })
+  describe("getLoader", () => {
+    it("should throw an error if options.csStaticBase is undefined or an empty string in context", () => {
```
code-asher, line 181: My bad, I miscommunicated. This just tests that the test itself throws which I don't think we need to have.

Take for example `setBodyBackgroundToThemeBackgroundColor`. It claims to take a document but checks if it's undefined. But it can only be a document according to the types.

So there are two solutions to this mismatch:

1. Change the type of `document` on `setBodyBackgroundToThemeBackgroundColor` to `Document | undefined`. Now it truly reflects what the function expects to handle.
2. Leave the type. This means that code which calls the function needs to be sure it is calling it with something valid. Usually if the types say it's valid it is so nothing needs to be done. But if something does need to be checked then we'd have something on the calling side. In this example we'd have something like this inside `main`:

```
if (!document) throw new Error()
setBodyBackgroundToThemeBackgroundColor(document)
```

However in our case we can make `main` also take a `document` negating the need for any checks.

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +177,151 @@ describe("vscode", () => {
...
+          throw new Error(errorMessage)
+        }
+        getLoader({
+          origin: "localhost",
+          nlsConfig: nlsConfig,
+          options,
+        })
+      }).toThrowError(errorMessage)
+    })
+    it("should throw an error if nlsConfig is undefined", () => {
```
code-asher, line 224: Same story as the previous comment!

src/browser/pages/vscode.ts
code-asher: This might be a lot of work though, not sure. We also might need to load `loader.js` first and then do `require.config(config)` instead.

test/unit/browser/pages/vscode.test.ts
code-asher: Or another way of saying it is that manual typechecks are possibly best done at the boundaries of the code where the types first enter (idk if there's any formal proof or evidence of this though haha).

src/browser/pages/vscode.ts
```diff
@@ -54,39 +52,35 @@ export function getNlsConfiguration(document: Document) {
...
+/**
+ * A helper function to get the require loader
+ *
+ * This used by VSCode/code-server
+ * to load files.
+ *
+ * We extracted the logic into a function so that
+ * it's easier to test.
+ **/
+export function getLoader({ origin, nlsConfig, options }: GetLoaderParams) {
```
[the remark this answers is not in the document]
jsjoeio: Honestly, that's my favorite kind of nitpicky 😂 I'd rather use the best names so I will change it to `getConfigurationForLoader` (unless you have a better name!)

src/browser/pages/vscode.ts
```diff
@@ -54,39 +52,35 @@ export function getNlsConfiguration(document: Document) {
...
-        })
-        .catch(cb)
-    }
+type GetLoaderParams = {
+  origin: string
+  nlsConfig: NlsConfiguration
+  options: Options
+}
+
+type Loader = {
```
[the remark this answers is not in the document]
jsjoeio: Good idea! (and thanks for the link). I wish they did too. Maybe they're hidden somewhere with a different name. I'll look a little bit

src/browser/pages/vscode.ts
```diff
@@ -160,9 +156,72 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
   return null
 }
 
-try {
+export function registerLoadBundleOnNlsConfig(nlsConfig: NlsConfiguration, base: string) {
```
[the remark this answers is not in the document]
jsjoeio: Good point! Yeah, this was just the first way to do this that came to mind 😂 but i like your idea! will refactor

src/browser/pages/vscode.ts
```diff
@@ -160,9 +156,72 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+  const bundles: {
+    [key: string]: string
+  } = {}
```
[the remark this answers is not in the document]
jsjoeio: I thought it was strange we were doing `Object.create(null)` but I'll take a closer look.  Thanks for the heads-up

src/browser/pages/vscode.ts
jsjoeio: Good point! I guess I tend to overcomment things but I can see why this might not be a good idea here. I'll slim down!

src/browser/pages/vscode.ts
jsjoeio: Um...that's weird. I don't remember doing that? Maybe a bad rebase on my part? most likely. I'lll add in!

test/unit/browser/pages/vscode.test.ts
jsjoeio: Okay! I'll take a look adding here. Worst case, I'll create a follow-up issue and improving there

src/browser/pages/vscode.ts
jsjoeio: > Maybe we can copy down the .d.ts

I'll take a look.

> This might be a lot of work though, not sure. We also might need to load loader.js first and then do require.config(config) instead.

Hmm! I'll look and if it is, we can do as a followup item

src/browser/pages/vscode.ts
jsjoeio: Honestly, that would be my preference! Much easier to unit test. Good call!

test/unit/browser/pages/vscode.test.ts
jsjoeio: > However in our case we can make main also take a document negating the need for any checks.

Actually, why don't I do that instead and have the error live there? We have multiple uses for `Document` so it would make sense to do that

src/browser/pages/vscode.ts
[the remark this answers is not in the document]
code-asher: That works for me!

test/unit/browser/pages/vscode.test.ts
code-asher: Makes sense! Maybe eventually we could have something like this in each top-level browser file:

```
try {
  assertExists(document, window) // We make sure the globals we expect exist
  main(document, window) // Then pass in the globals so all our functions are pure(ish)
} catch (error) {
  console.log(error)
}
```

And we could create `assertExists` in one of our util files. Just a thought though, doesn't have to be this way.

jsjoeio: Great idea! I like that approach

src/browser/pages/vscode.ts
jsjoeio: > This might be a lot of work though, not sure. We also might need to load loader.js first and then do require.config(config) instead.

To be honest, I don't fully understand how this all works, so I'm going to punt and say this should be a follow-up item.

jsjoeio: Though after re-reading your comment, I think we might be talking about two separate things. I'm trying to get the right types for `self.require` (which may not be solvable beyond using `any` and it sounds like you're talking about getting the right types for `loader`. Is that right?

https://github.com/cdr/code-server/issues/3780

test/unit/browser/pages/vscode.test.ts
jsjoeio: Added this as a follow-up item: https://github.com/cdr/code-server/issues/3781

src/browser/pages/vscode.ts
code-asher: My bad, I meant to imply we can tackle this later and not in this PR but I did not make that clear!

I think they do something like this, broadly speaking:

```
const options = self.require
self.require = new requireFuncFactory(options) // This returns an IRequireFunc which is the loader
```

I thought they declared `require` as `IRequireFunc` but they do not so you can ignore my previous comment. I think we can just do something like:

```
;(self.require as unknown as IConfigurationOptions) = { options }
```

jsjoeio: image

I'm having trouble finding `IConfigurationOptions` but I did something similar

jsjoeio:
Eventually when this gets to a good state and is approved and merged, the `vscode` file will be bulletproof 😎

src/browser/pages/vscode.ts
```diff
@@ -19,7 +17,18 @@ type NlsConfiguration = {
...
+}
+
+/**
+ * Helper function to create the path to the bundle
+ * for getNlsConfiguration.
+ */
+export function createBundlePath(_resolvedLanguagePackCoreLocation: string, bundle: string) {
+  // NOTE@jsjoeio - this comment was here before me but not sure
+  // why it's here or how to fix.
+  // FIXME: Only works if path separators are /.
```
code-asher, line 30: Refers to operating systems that use a different path separator. Probably just Windows although I'm not 100% sure using `/` here breaks Windows or not.

src/browser/pages/vscode.ts
```diff
@@ -30,42 +39,45 @@ type NlsConfiguration = {
  *
  * Make sure to wrap this in a try/catch block when you call it.
```
code-asher, line 40: It's really a shame TypeScript has no support for throwing.

I gave this a thumbs up: https://github.com/microsoft/TypeScript/issues/13219

Thoughts on (I'm thinking we could do this later since we have a lot of `throw` around) using a sort of DocBlock syntax with `@throws`? Similar to the example in this issue: https://github.com/microsoft/TypeScript/issues/43528

We could also consider (at a later date, not in this PR) not using `throw` at all in our codebase and always returning something like Rust's `Result`.

src/browser/pages/vscode.ts
```diff
@@ -30,42 +39,45 @@ type NlsConfiguration = {
...
  **/
-export function getNlsConfiguration(document: Document) {
+export function getNlsConfiguration(_document: Document, base: string) {
```
code-asher, line 42: Is the underscore to distinguish it from the global? I wonder if we should leave it without the underscore to prevent accidentally using the global (since the types wouldn't catch the mistake).

src/browser/pages/vscode.ts
```diff
@@ -100,25 +157,16 @@ try {
...
-  /* Probably fine. */
+
+  return loader
 }
 
-export function setBodyBackgroundToThemeBackgroundColor(document: Document, localStorage: Storage) {
+/**
+ * Sets the body background color to match the theme.
+ */
+export function setBodyBackgroundToThemeBackgroundColor(_document: Document, _localStorage: Storage) {
```
code-asher, line 167: Same comment as above for the underscore.

src/browser/pages/vscode.ts
```diff
@@ -74,17 +86,62 @@ try {
...
+/**
+ * A helper function to get the require loader
+ *
+ * This used by VSCode/code-server
+ * to load files.
+ *
+ * We extracted the logic into a function so that
+ * it's easier to test.
+ **/
+export function getConfigurationForLoader({ origin, nlsConfig, options, _window }: GetLoaderParams) {
```
code-asher, line 137: Same comment as above for the underscore.

src/browser/pages/vscode.ts
```diff
@@ -155,14 +203,50 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
 }
 
+/**
+ * A helper function to encapsulate all the
+ * logic used in this file.
+ *
+ * We purposely include all of this in a single function
+ * so that it's easier to test.
+ */
+export function main(_document: Document | undefined, _window: Window | undefined, _localStorage: Storage | undefined) {
```
code-asher, line 218: Same comment as above for the underscore.

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +185,171 @@ describe("vscode", () => {
...
     })
   })
+  describe("getConfigurationForLoader", () => {
+    beforeEach(() => {
+      const { window } = new JSDOM()
+      global.document = window.document
+    })
+    afterEach(() => {
+      global.window = undefined as unknown as Window & typeof globalThis
+      global.document = undefined as unknown as Document & typeof globalThis
```
code-asher, line 195: We shouldn't need any more globals right? Since we pass everything in to the individual functions now. In fact avoiding globals is probably good for testing since it'll catch any mistakes where we might accidentally use a global.

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +185,171 @@ describe("vscode", () => {
...
+          ...options,
+        }
+      }
+
+      const mockFn = jest.fn(mockCreatePolicy)
+
+      // @ts-expect-error we are adding a custom property to window
+      global.window.trustedTypes = {
+        createPolicy: mockFn,
+      }
```
code-asher, line 262: Same comment about globals as above.

test/unit/browser/pages/vscode.test.ts
```diff
@@ -175,4 +185,171 @@ describe("vscode", () => {
...
+    })
+  })
+  describe("main", () => {
+    beforeEach(() => {
+      // We need to set the url in the JSDOM constructor
+      // to prevent this error "SecurityError: localStorage is not available for opaque origins"
+      // See: https://github.com/jsdom/jsdom/issues/2304#issuecomment-62231494
+      const { window } = new JSDOM("", { url: "http://localhost" })
+      global.document = window.document
+      global.localStorage = window.localStorage
```
code-asher, line 305: Same comment about globals.

src/browser/pages/vscode.ts
```diff
@@ -155,14 +203,50 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+  if (!_document) {
+    throw new Error(`${errorMsgPrefix} Could not run code-server vscode logic. document is undefined.`)
+  }
+
+  if (!_window) {
+    throw new Error(`${errorMsgPrefix} Could not run code-server vscode logic. window is undefined.`)
+  }
+
+  if (!_localStorage) {
+    throw new Error(`${errorMsgPrefix} Could not run code-server vscode logic. localStorage is undefined.`)
```
code-asher, line 230: IMO errors should be purely the error since that reduces boilerplate a bit and makes them more flexible. (`document is undefined` for example.)

Then in the catch where we log the error:
```javascript
catch (error) {
  console.error("[vscode] failed to initialize VS Code")
  console.error(error.message) // Or console.log(error) to print a stack trace
}
```

src/browser/pages/vscode.ts
```diff
@@ -155,14 +203,50 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+  })
+
+  ;(self.require as unknown as Loader) = loader
+
+  setBodyBackgroundToThemeBackgroundColor(_document, _localStorage)
+}
+
 try {
-  setBodyBackgroundToThemeBackgroundColor(document, localStorage)
+  main(document, window, localStorage)
```
code-asher, line 249: Beautiful new entry!

src/browser/pages/vscode.ts
```diff
@@ -155,14 +203,50 @@ export function setBodyBackgroundToThemeBackgroundColor(document: Document, loca
...
+  }
+
+  const options = getOptions()
+  const nlsConfig = getNlsConfiguration(_document, options.base)
+
+  const loader = getConfigurationForLoader({
+    nlsConfig,
+    options,
+    _window,
+    origin: _window.location.origin,
```
code-asher, line 240: I suppose since we ended up needing `window` we could skip `origin` since it's redundant.

code-asher (approved):

src/browser/pages/vscode.ts
code-asher: Oops nevermind ignore me on the second part, you already linked to that `.d.ts` file in a comment so I think you were just saying it's not in VS Code and not that you couldn't find `IConfigurationOptions` at all.

src/browser/pages/vscode.ts
jsjoeio: Ahhh! Got it. Okay, I can update the comment then with those details. Any suggestions on how I could write a test for this function to see how it behaves on Windows?

src/browser/pages/vscode.ts
jsjoeio: Ha funny, looks like I did too. I'll subscribe to it.

image

I love either of those ideas! I'll add as a followup todo:

jsjoeio: https://github.com/cdr/code-server/issues/3825

src/browser/pages/vscode.ts
code-asher: I'm not really sure how effective unit tests will be here tbh, I think we'll need to do in-browser tests on different platforms.

TeffenEllis: Can we get away with `path.join` as an OS agnostic path builder?

jsjoeio: ^that would be awesome! I will leave for now and eventually we can follow-up later!

src/browser/pages/vscode.ts
jsjoeio: Exactly! Well, I actually made this decision for _the opposite reason_. Here's my thinking:

When we're testing these functions, they should be pure (like we talked about before with `main`). But if the parameter and the global share the same name, it's hard to tell which one is being used. Therefore to prevent accidentally relying on the global, I opted to prefix with the `_`. 

You're right that the types wouldn't catch the mistake, but at least we can guarantee (for now) these functions are pure and relying on the values (i.e. `_document`) being passed in rather than the `document`. This tradeoff has limitations like you brought up, but I think it makes testing, reasoning and maintainability easier for future us.

src/browser/pages/vscode.ts
jsjoeio: > IMO errors should be purely the error since that reduces boilerplate a bit and makes them more flexible. (document is undefined for example.)

Great point! I'm on board with that. I'll refactor!

test/unit/browser/pages/vscode.test.ts
jsjoeio: Hmm...well we need `window` and `document` to come from somewhere. I guess the point you're making is, define them locally in each test rather than on the `global` object?

src/browser/pages/vscode.ts
jsjoeio: Oh yeah true, good catch

TeffenEllis (approved):

test/unit/browser/pages/vscode.test.ts
jsjoeio: I took your point as that and refactored to not use globals in test. Good call!

src/browser/pages/vscode.ts
code-asher: If we avoid setting globals in the tests we'd catch the errors.

It seems to me doing it this way *increases* the probability you'd accidentally use the global and has *less* guarantee it's pure. The reason being: you can't use the global in this function no matter how much you wanted to since it's shadowed but by adding an underscore we've now created the option to use the global where it didn't exist before.

jsjoeio: > If we avoid setting globals in the tests we'd catch the errors.

I guess that's a valid point! (and I did remove them so technically you're right, we would catch them).

> The reason being: you can't use the global in this function no matter how much you wanted to since it's shadowed

Oh! I guess I worried that variable scoping/shadowing is something people don't always understand (myself included). I thought having the `_` would help readability/maintainability by saying, "Look, we are NOT using the global values" but maybe your argument makes more sense. I'm easy to convince if you think I should remove them!

code-asher: Fair! To be honest I had to do some quick tests to make sure the shadowing worked the way I expected haha. But I think with the knowledge that the shadowing does work we should remove them to make it impossible for future developers to use the globals in these functions.

code-asher: Or if we could somehow change the types for these files to make using globals an error, that might be even better!

Maybe we could make it so globals can only be used in the entry files we specify (I'm not sure if this is possible though) and we move the non-global code to other files that way it's clear you aren't supposed to use globals (and the compiler won't let you even if you try).

jsjoeio: Good ideas! Documented here to follow-up later: https://github.com/cdr/code-server/issues/3856

PR #3769, opened 2021-07-12
refactor: clean up testing files

jsjoeio:
I accidentally merged a PR that didn't have a file formatted correctly and it made CI on `main` fail. This fixes that but also clean ups some testing files to be better organized.

### DIFF
--- a/test/unit/browser/pages/login.test.ts
+++ b/test/unit/browser/pages/login.test.ts
@@ -1,5 +1,5 @@
 import { JSDOM } from "jsdom"
-import { LocationLike } from "../../unit/util.test"
+import { LocationLike } from "../../util.test"
 
 describe("login", () => {
   describe("there is an element with id 'base'", () => {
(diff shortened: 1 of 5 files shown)


jawnsy:
Ah, just curious, shouldn't the same formatting checks run on PRs and main? How'd it get merged in the first place?

jawnsy (approved):

jsjoeio:
I hit enable auto-merge which says "Automatically merge when all requirements are met." - let me checking the repo settings

jsjoeio:
image

Weird, pre-build is not required, but should be? Enabling now: cc @oxy and @code-asher

jsjoeio:
actually, i bet i broke it when i changed it to "Pre-Build" instead of "Pre-build"

jsjoeio:
argh, auto update imports didn't work lol fixing

jawnsy:
Haha, I dislike auto merge because of these sorts of problems, but I see how it can be helpful :man_shrugging:

jsjoeio:
I feel ya, luckily this has only happened once or twice

jsjoeio:
Moment of truth

code-asher:
Hmm but we require end-to-end tests which requires pre-build? Is it sneaking in the auto-merge in the gaps before the next check starts??

jsjoeio:
I guess it might be? Maybe it's a bug? I went into the settings and added "Pre-Build" as a required check. Maybe it's sneaking in because the previous run had both of those passing? I'm not too sure

PR #3783, opened 2021-07-15
Remove leftover merge conflict

notpushkin:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -390,19 +390,7 @@ Codespaces web dashboard instead of connecting directly to it.
 On the other hand, code-server is free, open-source, and can be run on any
 machine with few limitations.
 
-<<<<<<< HEAD
-
-## Does code-server have any security login validation?
-
-code-server only supports a single password and limits logins to two per minute plus twelve per hour.
-
-## Enterprise
-
-=======
-
 ## Are there community projects involving code-server?
-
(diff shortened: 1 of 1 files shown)


code-asher (changes requested):
Thank you for finding this!

I think we want to keep the `Does code-server have any security login validation` section but we can remove `Enterprise` (it will also need to be removed in the table of contents at the top).

docs/FAQ.md
```diff
@@ -390,6 +389,10 @@ Codespaces web dashboard instead of connecting directly to it.
 On the other hand, code-server is free, open-source, and can be run on any
 machine with few limitations.
 
+## Does code-server have any security login validation?
+
+code-server only supports a single password and limits logins to two per minute plus twelve per hour.
```
jawnsy, line 394:
```suggestion
code-server supports setting a single password limiting logins to two per minute and twelve per hour. 
```

a small tweak. we might want to have @khorne3 take a look through these docs too, if she has the bandwidth?

jawnsy (approved):

docs/FAQ.md
khorne3: Maybe `code-server supports setting a single password and limits logins to two per minute and twelve per hour` otherwise it sounds a bit like the password is the thing limiting logins

PR #3812, opened 2021-07-19
refactor: only accept string in pathToFsPath

jsjoeio:
This PR reverts some work done in https://github.com/cdr/code-server/pull/3751 to make `pathToFsPath` better.

### DIFF
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -458,16 +458,10 @@ enum CharCode {
  * Taken from vs/base/common/uri.ts. It's not imported to avoid also importing
  * everything that file imports.
  */
-export function pathToFsPath(path: string | string[], keepDriveLetterCasing = false): string {
+export function pathToFsPath(path: string, keepDriveLetterCasing = false): string {
   const isWindows = process.platform === "win32"
-  const uri = { authority: undefined, path: getFirstString(path), scheme: "file" }
+  const uri = { authority: undefined, path: getFirstString(path) || "", scheme: "file" }
   let value: string
-
-  if (typeof uri.path !== "string") {
-    throw new Error(
-      `Could not compute fsPath from given uri. Expected path to be of type string, but was of type ${typeof uri.path}.`,
-    )
-  }
 
   if (uri.authority && uri.path.length > 1 && uri.scheme === "file") {
     // unc path: file://shares/c$/far/boo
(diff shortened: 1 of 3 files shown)


src/node/util.ts
```diff
@@ -458,17 +458,11 @@ enum CharCode {
  * Taken from vs/base/common/uri.ts. It's not imported to avoid also importing
  * everything that file imports.
  */
-export function pathToFsPath(path: string | string[], keepDriveLetterCasing = false): string {
+export function pathToFsPath(path: string, keepDriveLetterCasing = false): string {
   const isWindows = process.platform === "win32"
-  const uri = { authority: undefined, path: getFirstString(path), scheme: "file" }
+  const uri = { authority: undefined, path: getFirstString(path) || "", scheme: "file" }
```
jawnsy, line 463: just wondering - would this be more readable in expanded form, e.g.

```typescript
const uri = {
  authority: undefined,
  path: getFirstString(path) || "",
  scheme: "file",
}
```

I'm not sure if prettier prefers this form though

test/unit/node/util.test.ts
```diff
@@ -464,19 +467,8 @@ describe("pathToFsPath", () => {
...
-      util
-        // @ts-expect-error We need to check other types
-        .pathToFsPath({}),
-    ).toThrow(`Could not compute fsPath from given uri. Expected path to be of type string, but was of type undefined.`)
-  })
-  it("should not throw an error for a string array", () => {
-    // @ts-expect-error We need to check other types
-    expect(() => util.pathToFsPath(["/hello/foo", "/hello/bar"]).not.toThrow())
-  })
   it("should use the first string in a string array", () => {
```
jawnsy, line 470: there's no more string array, right? so the description is now outdated

jawnsy (approved):
LGTM! Codecov shamin' us a little for lack of test coverage here though :joy:

HOME1640 (approved):

src/node/util.ts
jsjoeio: Definitely is for me! I think Prettier autoformats it though :(

test/unit/node/util.test.ts
jsjoeio: Good catch! I'll update that

jsjoeio:
> LGTM! Codecov shamin' us a little for lack of test coverage here though 😂

ugh I know 😂 I think it's cause I touched some of the routes files, but I didn't write any tests for them so it's unhappy. I'm going to say it's okay for now since at least I'm increasing coverage.

PR #3831, opened 2021-07-21
chore: Update dependency requirements. Fix node version manager.

TeffenEllis:
This PR addresses some minor doc changes, and adds files that allow NVM and nodenv to automatically switch to our preferred version.


.nvmrc
```diff
@@ -0,0 +1 @@
+.node-version
```
TeffenEllis, line 1: Included as a symlink to avoid version drifting between NVM and nodenv

docs/CONTRIBUTING.md
```diff
@@ -29,6 +29,7 @@ Here is what is needed:
 
 - `node` v14.x
 - `git` v2.x or greater
+- [`git-lfs`](https://git-lfs.github.com)
```
TeffenEllis, line 32: I encountered a somewhat mysterious error when building the project on a clean slate. Turns out we need this too!

.node-version
```diff
@@ -0,0 +1 @@
+14
```
jawnsy, line 1: I dunno anything about nvm, though I'm curious why these tools don't inspect package.json's engine info for this?

Seems like a reasonable change to me though!

jawnsy (approved):

.node-version
jsjoeio: Agreed, you would think it would be smarter!

jsjoeio (approved):
Awesome work! Thanks for updating all this 🎉

.node-version
TeffenEllis: Ahh, [open source](https://github.com/nvm-sh/nvm/issues/651) 😆

jsjoeio:
Minor note: if your PR isn't related to an issue, add the Milestone on the right side so we associate this work with the next release.

Issue #3814, opened 2021-07-20
Docker image for 3.11 not published


alpesh1988:
Hello Guys,

Somehow docker latest image is still running on vscode previous version and not with 1.57.1. You can check by looking at build size for latest and 3.10.2 images in the link below, Both have same sizes.

See [here](https://hub.docker.com/r/codercom/code-server/tags?page=1&ordering=last_updated)

Request you to fix that issue,

Looking forward to receive your response,

Regards,
Alpesh

stilleshan:
The same problem, the image pulled shows that it was created 8 weeks ago

GitHub action workflows had error 4 days ago
https://github.com/cdr/code-server/actions/runs/1039022443

jsjoeio:
Yes, we're having some troubles updating the Docker image for 3.11. Please read this thread for details: https://github.com/cdr/code-server/discussions/3779#discussioncomment-1010599

alpesh1988:
Is it really taking too long or missed to update an issue ?

jsjoeio:
As you probably saw #3838 was merged today.

It "partially addresses this issue, though a new release needs to be made for the workflow to kick in."

Hopefully we can do that tomorrow on Monday. cc @oxy

jsjoeio:
![image](https://user-images.githubusercontent.com/3806031/127701626-94cec3da-345c-4641-b4f2-3e524de950c6.png)

@code-asher has updated the images. Thanks for your patience!

PR #3838, opened 2021-07-22
fix(ci): build+push image in release flow

oxy:
Previously, we would build docker images on each CI run, export them, then re-import them and push them on release. However, there appears to be a bug in buildx's export flow, and images can't be re-imported into Docker.

To fix this, I've moved to building and publishing with buildx in the release workflow in one step, and removed the image build step from the main CI flow. I also moved the docker publish process to its own workflow instead of `release.yml` so that it can be re-run independently of the npm+homebrew step.

Partially adresses #3814, though a new release needs to be made for the workflow to kick in.


ci/steps/build-docker-image.sh
```diff
@@ -1,12 +1,37 @@
 #!/usr/bin/env bash
```
code-asher, line 1: Should we rename this file? Maybe `build-and-push-docker-image.sh`?

jsjoeio (commented):
As someone who is less familiar with Docker/buildx, I don't think I have enough exp to approve this.

However from a general standpoint, 
- great PR description 
- this approach feels clean and maintainable
- i appreciate you taking extra precautions and separating into it's own workflow

ci/steps/build-docker-image.sh
```diff
@@ -1,12 +1,37 @@
...
+
+  # Download the release-packages artifact
+  download_artifact release-packages ./release-packages
+
+  # Login to Docker
+  if [[ ${CI-} ]]; then
+    echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
+  fi
+
+  docker buildx bake -f ci/release-image/docker-bake.hcl --push
```
code-asher, line 34: Legend! So much better.

code-asher (approved):

code-asher:
Maybe we should rename the publish workflow as well since it's not the full publish process anymore.

ci/steps/build-docker-image.sh
oxy: I did consider this but felt it was a bit of a mouthful - though am not opposed to it!

code-asher: It does seem a bit long haha but I don't want to mislead anyone who might try to run this!

code-asher: Could maybe shorten to `docker-buildx-push.sh` or something.

oxy: Sounds great!

jsjoeio:
@oxy just checking in on this. Anything we can do to help you move this forward?

oxy:
@jsjoeio @code-asher ready for a second look and merge!

jsjoeio (approved):
LGTM! Great work on keeping it clean and easy to follow 👏

code-asher (approved):
:tada:

PR #3839, opened 2021-07-22
chore: add docs manifest

BrunoQuaresma:
Adds the manifest.json, which is the configuration file to generate the navigation of code-server docs in https://coder.com/docs/code-server/latest.


jsjoeio (approved):
LGTM! Thanks for adding this in!

PR #3840, opened 2021-07-22
docs(install): add uninstall instructions

jsjoeio:
This PR adds an "Uninstall" section to the `install.md` docs. I decied to add this because I noticed a trend in GitHub Discusions where people have asked how to do this:
- https://github.com/cdr/code-server/discussions/3525
- https://github.com/cdr/code-server/discussions/3823
- https://github.com/cdr/code-server/discussions/1989

## Caveat

This does not cover all scenarios but it does cover the ones that have been asked about. I figured we can add more as it gets asked (issue/discussion drive development?).

## Todos

- [x] clean up commits
- [x] open issue documenting idea from @TeffenEllis https://github.com/cdr/code-server/issues/3857

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -15,6 +15,11 @@
 - [Raspberry Pi](#raspberry-pi)
 - [Termux](#termux)
 - [Cloud providers](#cloud-providers)
+- [Uninstall](#uninstall)
+  - [install.sh](#installsh-1)
+  - [Homebrew](#homebrew)
+  - [yarn, npm](#yarn-npm-1)
+  - [Debian, Ubuntu](#debian-ubuntu-1)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -246,3 +251,54 @@ information.
...
+
+To delete settings and data:
+
+```shell
+rm -rf ~/.local/share/code-server ~/.config/code-server
+```
+
+### install.sh
+
+If you installed with the install script, by default code-server will be in `~/usr/lib/code-server` and you can remove it with `rm -rf`. e.g.
```
jsjoeio, line 267: ❓ Is `~/usr/lib/code-server` the right path? 🤔

jsjoeio: According to `install.sh`
> The release is unarchived into ~/.local/lib/code-server-X.X.X
>      and the binary symlinked into ~/.local/bin/code-server

So no that is not right.

jsjoeio: Updated for the unarchived. Wondering if we also need to tell them to remove the binary in `/bin`

docs/install.md
```diff
@@ -246,3 +251,54 @@ information.
 We maintain [one-click apps and install scripts for cloud
 providers](https://github.com/cdr/deploy-code-server) such as DigitalOcean,
 Railway, Heroku, and Azure.
+
+## Uninstall
+
+Removing code-server depends on where you installed it.
```
TeffenEllis, line 257:
```suggestion
code-server can be completely uninstalled by removing the application directory, and your user configuration directory. 
```

TeffenEllis (approved):
Some minor nits but otherwise this is a great addition. It would be especially nice if we could follow up with a programmatic means to identify the installation path e.g.

```sh
code-server --installation-prefix
code-server --config-prefix
```

jsjoeio:
> ```shell
> code-server --installation-prefix
> code-server --config-prefix
> ```

I think I understand what you mean, but to be extra clear, you're suggesting adding these as flags/args to run with `code-server` to identify these things as an end user? Fantastic idea!

Issue #3841, opened 2021-07-22
Investigate use of proxy-agent


jsjoeio:
## Background

Recently, we [accepted a PR](https://github.com/cdr/code-server/pull/3784) from dependabot bumping `proxy-agent` from `4.0.1` to `5.0.0`. Due to passing CI and no concerns after reading the release notes, we merged it.

It sparked a discussion amongst maintainers where we asked ourselves: _why do we have proxy-agent?_

The goal of this issue is to investigate our use and see if it's still needed in the codebase.

### What is `proxy-agent`?

an [`npm`package](https://www.npmjs.com/package/proxy-agent) which:
> Maps proxy protocols to http.Agent implementations

#### What is a proxy protocol?

There are two parts to understanding this. First, let's discuss what we mean by _proxy_.

**proxy**: an interface to something else. 
e.g. let's say you have a basic React app that needs to talk to a weather API. It requires an API key, which you don't want to expose on the client because then anyone could see and use it. You might use a proxy which could be a serverless function, a server, etc. which has access to the API key and passes on your request to the weather API.

Second, let's discuss what we mean by _protocol_.
**protocol**: a set of rules and guidelines for communicating data. 
e.g. HTTP stands for "Hypertext Transfer Protocol" and is the stand for how data should be communicated on the web.

Now that we're on the same page for those two words, I bet you can guess what we mean by "proxy protocol".

**proxy protocol**: a set of rules and guidelines for communicating data between proxies

#### What is an `http.Agent`?

We can read about `http.Agent` which is a class in Node.js. `Agent` is defined as:

> An Agent is responsible for managing connection persistence and reuse for HTTP clients. It maintains a queue of pending requests for a given host and port, reusing a single socket connection for each until the queue is empty, at which time the socket is either destroyed or put into a pool where it is kept to be used again for requests to the same host and port.

Source: [Node.js docs](https://nodejs.org/api/http.html#http_class_http_agent)

That definition feels a bit hard to understand IMO. Looking at [this StackOverflow question](https://stackoverflow.com/questions/64117101/what-is-the-role-of-https-agent-in-node), the OP explains it as:

> an Agent is 'in charge' of managing a connection - and clearly, the fact that https.Agent exists on top of the 'plain' http.Agent exists would imply that it is 'in charge' of managing an HTTPS connection

which helps a little bit but I'm still a little unsure. 

Spoke to @jawnsy who gave a couple analogies.

[...]

PR #3846, opened 2021-07-23
feat: add tests for shouldEnableProxy

jsjoeio:
This PR adds tests for `shouldEnableProxy()` which is used in `proxy_agent.ts`.

Related:
- https://github.com/cdr/code-server/issues/3841

### DIFF
--- a/test/unit/node/proxy_agent.test.ts
+++ b/test/unit/node/proxy_agent.test.ts
@@ -1,59 +1,39 @@
 import { shouldEnableProxy } from "../../../src/node/proxy_agent"
-
-/**
- * Helper function to set an env variable.
- *
- * Returns a function to cleanup the env variable.
- */
-function setEnv(name: string, value: string) {
-  process.env[name] = value
-
-  return {
-    cleanup() {
-      delete process.env[name]
-    },
-  }
-}
+import { useEnv } from "../../utils/helpers"
 
 describe("shouldEnableProxy", () => {
-  // Source: https://stackoverflow.com/a/48042799
-  const OLD_ENV = process.env
+  const [setHTTPProxy, resetHTTPProxy] = useEnv("HTTP_PROXY")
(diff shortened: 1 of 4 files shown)


test/unit/node/proxy_agent.test.ts
```diff
@@ -0,0 +1,59 @@
+import { shouldEnableProxy } from "../../../src/node/proxy_agent"
+
+/**
+ * Helper function to set an env variable.
+ *
+ * Returns a function to cleanup the env variable.
+ */
+function setEnv(name: string, value: string) {
```
TeffenEllis, line 8: Nit: given this is a specific test helper, it would be nice if we could move the `cleanup`  behavior into the `beforeEach` so that it may only have to be declared once. This is a bit like a React hook, and allows us to target a specific key, rather than dereferencing the entire `process.env` object. 

```ts
// test_helpers.ts 
function createSetEnv(key: string) {
  const initialValue = process.env[key]
  const setValue = (nextValue: string | undefined) => process.env[key] = nextValue
  const resetValue = () => process.env[key] => initialValue

  return [setValue, resetValue]
}

// test.ts

const [setHTTPProxy, resetHTTPproxy] = createSetEnv(“HTTP_PROXY”)
const [setNoProxy, resetNoProxy] = createSetEnv(“NO_PROXY”)
{
  beforeEach() {
    jest.resetModule()
    resetHTTPproxy()
    resetNoProxy()
  }
```

TeffenEllis (changes requested):

test/unit/node/proxy_agent.test.ts
jsjoeio: Oh waaaay better than my solution. The original `cleanup()` I wrote was partially inspired by previous experience cleaning up `useEffect` functions. I really like your implementation. Thanks for writing that out! I'll make these changes.

TeffenEllis (approved):
Incredibly thorough ✨

Nicely done.

jsjoeio:
🙌 Team work!

PR #3858, opened 2021-07-26
feat: add more tests for proxy.ts

jsjoeio:
This PR adds more tests for `proxy.ts`, specifically to test that our proxy handles error events appropriately.

Fixes N/A

P.S. huge shoutout to @vapurrmaid for pairing with me on this 🎉

### DIFF
--- a/test/unit/helpers.test.ts
+++ b/test/unit/helpers.test.ts
@@ -1,5 +1,5 @@
 import { promises as fs } from "fs"
-import { tmpdir, useEnv } from "../../test/utils/helpers"
+import { getAvailablePort, tmpdir, useEnv } from "../../test/utils/helpers"
 
 /**
  * This file is for testing test helpers (not core code).
@@ -37,5 +37,18 @@ describe("useEnv", () => {
     expect(process.env[envKey]).toEqual("hello there")
     resetValue()
     expect(process.env[envKey]).toEqual("test environment variable")
+  })
+})
+
+describe("getAvailablePort", () => {
+  it("should return a valid port", async () => {
+    const port = await getAvailablePort()
+    expect(port).toBeGreaterThan(0)
+    expect(port).toBeLessThanOrEqual(65535)
+  })
+  it("should return different ports for different calls", async () => {
+    const portOne = await getAvailablePort()
+    const portTwo = await getAvailablePort()
+    expect(portOne).not.toEqual(portTwo)
   })
 })
--- a/test/unit/node/proxy.test.ts
+++ b/test/unit/node/proxy.test.ts
@@ -6,6 +6,7 @@ import { HttpCode } from "../../../src/common/http"
 import { proxy } from "../../../src/node/proxy"
 import * as httpserver from "../../utils/httpserver"
 import * as integration from "../../utils/integration"
+import { getAvailablePort } from "../../utils/helpers"
 
 describe("proxy", () => {
   const nhooyrDevServer = new httpserver.HttpServer()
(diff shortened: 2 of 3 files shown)


jsjoeio:
cc @code-asher would love to get your eyes on this and help me understand what I'm missing

test/unit/proxy.test.ts
```diff
@@ -102,4 +108,135 @@ describe("proxy", () => {
...
+
+  afterEach(async () => {
+    await testServer.close()
+    await proxyTarget.close()
+  })
+
+  it("should return a 500 when proxy target errors ", async () => {
+    // Close the proxy target so that proxy errors
+    await proxyTarget.close()
+    const errorResp = await nodeFetch.default(`${URL}/error`)
```
vapurrmaid, line 232: As noted during pair programming - Using the `http` and `node-fetch` libraries directly is _easier_ (LoC and maintainability). I have concerns about the custom class using in testing that abstracts these

test/unit/proxy.test.ts
```diff
@@ -102,4 +106,107 @@ describe("proxy", () => {
...
+// and specifically tests the proxy in isolation vs. using
+// the httpserver abstraction we've built.
+//
+// Leaving this as a separate test suite for now because
+// we may consider refactoring the httpserver abstraction
+// in the future.
+//
+// If you're writing a test specifically for code in
+// src/node/proxy.ts, you should probably add it to
+// this test suite.
```
jsjoeio, line 167: ✍️ @code-asher made a point that the next eng. coming in here might be confused as to where to put their test. I added this comment to address that.

test/unit/proxy.test.ts
```diff
@@ -102,4 +106,107 @@ describe("proxy", () => {
...
+describe("proxy (standalone)", () => {
+  const PORT = 9003
+  const PROXY_PORT = 8003
```
jsjoeio, line 170: ❓Chose these ports at random. Not sure if there's a better way to chose two random ports.

vapurrmaid: Could use something like https://www.npmjs.com/package/get-port but it's probably not necessary at this moment in time.

test/unit/proxy.test.ts
vapurrmaid: TIL that to obtain a random port, we just create and close an arbitrary server around a locking process, and return the port it was assigned:

https://github.com/sindresorhus/get-port/blob/260c667f457c4260574f915c15c470d2c8eaa12d/index.js#L23-L33

jsjoeio: Nice find! I refactored to use that instead.

test/unit/helpers.test.ts
```diff
@@ -39,3 +39,15 @@ describe("useEnv", () => {
...
+
+describe("getAvailablePort", () => {
+  it("should return a valid port", async () => {
+    const port = await getAvailablePort()
+    expect(port).toBeGreaterThan(0)
+  })
+  it("should return different ports for different calls", async () => {
+    const portOne = await getAvailablePort()
+    const portTwo = await getAvailablePort()
+    expect(portOne).not.toEqual(portTwo)
```
vapurrmaid, line 51: I'd add an `expect` that the ports are within the valid port range (0 to 65535)

jsjoeio: Good call! TIL there's a specific range!

vapurrmaid (approved):

oxy (approved):
Looks good to me! :sparkles:

PR #3869, opened 2021-07-29
docs(install): Fix link to helm readme

### LINKED ISSUE #3867: Getting 404 on Helm link
Hello, the link on installation using Helm seems to be broken and returning 404. I can't seem to find the original link referenced from the markdown either. Please look into it. Link to the doc is below.

https://github.com/cdr/code-server/blob/main/docs/install.md#helm

vapurrmaid:
Fixes #3867

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -234,7 +234,7 @@ alternative](https://hub.docker.com/r/linuxserver/code-server).
 
 ## Helm
 
-You can install code-server via [Helm](../ci/helm-chart/README.md).
+You can install code-server via [Helm](https://github.com/cdr/code-server/blob/main/ci/helm-chart/README.md).
 
 ## Raspberry Pi
 


docs/install.md
```diff
@@ -234,7 +234,7 @@ alternative](https://hub.docker.com/r/linuxserver/code-server).
 
 ## Helm
 
-You can install code-server via [Helm](../ci/helm-chart/README.md).
+You can install code-server via [Helm](https://github.com/cdr/code-server/blob/main/ci/helm-chart/README.md).
```
jawnsy, line 237: Hmm, that's strange, why does this need to be an absolute link? The relative link does seem to work ok ([see preview](https://coder.com/docs/code-server/v3.11.0/install#helm)) but it links to `https://coder.com/docs/code-server/v3.11.0/ci/helm-chart/README` (missing the .md extension)

I don't know if something else is going on here :thinking:

vapurrmaid: Hey @jawnsy - From what I can tell, content located in `ci` is not rendered on coder.com.

Adding the `.md` extension doesn't fix this, try this link:

https://coder.com/docs/code-server/v3.11.0/ci/helm-chart/README.md

image

I think it needs to be an absolute link unless we also intend on hosting `ci/helm-chart/README.md` on coder.com/docs from what I see

jsjoeio: Yeah I think @BrunoQuaresma knows more about this but at least we can fix this issue for now, get people to the right docs and then improve later if needed!

jsjoeio (approved):
Thanks for fixing this @vapurrmaid and thanks @jawnsy for asking good questions and helping review! 🙌

docs/install.md
BrunoQuaresma: Yes, if we want to add it to the docs page, we have to move it to live inside of the docs folder.

jsjoeio: Sweet! made a followup issue: https://github.com/cdr/code-server/issues/3877

PR #3881, opened 2021-07-30
Query artifacts by branch name

code-asher:
This way workflows will still work even if there have been enough runs
to push the version branch run off the first page.


jsjoeio (approved):
I love one liners 🔥

Good call!

PR #3892, opened 2021-08-03
feat(ci): armv7 cross builds

### LINKED ISSUE #1337: Add arm32v7 builder
Create a new file called `arm.Dockerfile` that allows you to run it on arm32v7 CPUs such as the Raspberry Pi's CPU.

oxy:
Hopefully should cross-compile and work for armv7!

Fixes #1337.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -234,23 +234,32 @@ jobs:
   # It is not feasible to cross-compile with CentOS.
 
   # Cross-compile notes: To compile native dependencies for arm64,
-  # we install the aarch64 cross toolchain and then set it as the default
+  # we install the aarch64/armv7l cross toolchain and then set it as the default
   # compiler/linker/etc. with the AR/CC/CXX/LINK environment variables.
   # qemu-user-static on ubuntu-16.04 currently doesn't run Node correctly,
-  # so we just build with "native"/x86_64 node, then download arm64 node
-  # and then put it in our release. We can't smoke test the arm64 build this way,
+  # so we just build with "native"/x86_64 node, then download arm64/armv7l node
+  # and then put it in our release. We can't smoke test the cross build this way,
   # but this means we don't need to maintain a self-hosted runner!
-  package-linux-arm64:
-    name: Linux ARM64 cross-compile build
+  package-linux-cross:
+    name: Linux cross-compile builds
     needs: build
     runs-on: ubuntu-16.04
     timeout-minutes: 15
+    strategy:
+      matrix:
+        include:
+          - prefix: aarch64-linux-gnu
+            arch: arm64
+          - prefix: arm-linux-gnueabihf
+            arch: armv7l
+
     env:
-      AR: aarch64-linux-gnu-ar
-      CC: aarch64-linux-gnu-gcc
-      CXX: aarch64-linux-gnu-g++
-      LINK: aarch64-linux-gnu-g++
-      NPM_CONFIG_ARCH: arm64
+      AR: ${{ format('{0}-ar', matrix.prefix) }}
+      CC: ${{ format('{0}-gcc', matrix.prefix) }}
+      CXX: ${{ format('{0}-g++', matrix.prefix) }}
+      LINK: ${{ format('{0}-g++', matrix.prefix) }}
+      NPM_CONFIG_ARCH: ${{ matrix.arch }}
+      NODE_VERSION: v14.17.4
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


jsjoeio (approved):
Overall, I don't see anything concerning here. Nice work 👏

.github/workflows/ci.yaml
```diff
@@ -234,23 +234,32 @@ jobs:
...
-      CC: aarch64-linux-gnu-gcc
-      CXX: aarch64-linux-gnu-g++
-      LINK: aarch64-linux-gnu-g++
-      NPM_CONFIG_ARCH: arm64
+      AR: ${{ format('{0}-ar', matrix.prefix) }}
+      CC: ${{ format('{0}-gcc', matrix.prefix) }}
+      CXX: ${{ format('{0}-g++', matrix.prefix) }}
+      LINK: ${{ format('{0}-g++', matrix.prefix) }}
+      NPM_CONFIG_ARCH: ${{ matrix.arch }}
+      NODE_VERSION: v14.17.4
```
jawnsy, line 262: Is it possible to install the "latest" node? The rest of the build scripts seem to bundle the latest Node 14.x, which would help us avoid security problems.

I don't have strong feelings about pinning specific versions or using the "latest" minor/patch release, but I do think we should be consistent with the versions we're using between the native vs cross-compiled builds

oxy: Unfortunately, there's no stable URL I can point to for latest node, which is why I resolved to this :(
Even in https://nodejs.org/dist/latest-v14.x/ - you need to know the latest version to have the right URL.

jawnsy: Apparently, they publish a JSON file, so we can write a script to extract the latest version using jq or whatever: https://nodejs.org/dist/index.json

it seems... unfortunate, that there doesn't seem to be an easier way to do this

anyway, shouldn't let this hold up this PR, this LGTM

.github/workflows/ci.yaml
```diff
@@ -234,23 +234,32 @@ jobs:
...
+          - prefix: arm-linux-gnueabihf
+            arch: armv7l
+
     env:
-      AR: aarch64-linux-gnu-ar
-      CC: aarch64-linux-gnu-gcc
-      CXX: aarch64-linux-gnu-g++
-      LINK: aarch64-linux-gnu-g++
-      NPM_CONFIG_ARCH: arm64
+      AR: ${{ format('{0}-ar', matrix.prefix) }}
```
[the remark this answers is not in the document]
jsjoeio: ^that looks a little easier to read IMO

ci/build/arch-override.json
```diff
@@ -0,0 +1,8 @@
+{
+  "rpm": {
+    "armv7l": "armhfp"
+  },
+  "deb": {
+    "armv7l": "armhf"
```
jsjoeio, line 6: I worry about adding an extra file just because it seems like more to maintain. Also the json means we can't really put comments to explain why we have this. Any thoughts on writing a function to pass in `rpm` or `deb` and then returning values instead?

ci/build/build-packages.sh
```diff
@@ -43,14 +43,27 @@ release_gcp() {
...
 }
 
+get_nfpm_arch() {
+  if jq -re ".${PKG_FORMAT}.${ARCH}" ./ci/build/arch-override.json > /dev/null; then
+    jq -re ".${PKG_FORMAT}.${ARCH}" ./ci/build/arch-override.json
+  else
+    echo "$ARCH"
+  fi
+}
+
```
jsjoeio, line 53: Maybe we could add a comment above explaining why we have to override the arch? Or why we have this function?

jsjoeio (changes requested):
Wondering if we can explain why we want to to go with the `arch-override.json` approach instead of providing a function and returns values based on input. 

I'd rather us do that then add more files like this json file but open to hearing why the current approach is a better approach

ci/build/arch-override.json
oxy: My reasoning to put this in a separate file is that its easier to add overrides for other architectures should we need it in the future (say powerpc/s390x) and that it also separates code (building packages) from data (what overrides to apply when)

ci/build/build-packages.sh
oxy: Sure! I can do that - just give me a minute.

ci/build/arch-override.json
jsjoeio: Ah, okay that makes sense then :) Let's leave as is!

ci/build/build-packages.sh
jsjoeio: Only request then is to add a comment or two explaining why we have to override the arch and  then we should approve and merge!

.github/workflows/ci.yaml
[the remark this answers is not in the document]
oxy: Oh! Didn't know I could do it this way :o
TIL

ci/build/build-packages.sh
oxy: Done!

jsjoeio (approved):
ship it!

PR #3895, opened 2021-08-03
fix(lib/vscode): patch authority in asWebviewUri

jsjoeio:
This PR patches `asWebviewUri` in `lib/vscode/src/vs/workbench/api/common/shared/webview.ts` to remove  the port number from the `authority` field used to create the URI.

For now, we found that removing the port fixes the issue (resources load as expected in webviews), but long-term, we want to understand how this is meant to be used (having the port in the URI authority field).

Related:
- https://github.com/cdr/code-server/discussions/3883
- https://excalidraw.com/#json=6022466832957440,nM6CH8Ga9F2AiLSzjJ55-A

## Video 

Here is a video of an extension using a webview which loads both CSS and JS working.

https://user-images.githubusercontent.com/3806031/128255007-c342bf0f-f444-48be-84cd-fa8059995d33.mov

Fixes N/A

TODOS
- [x] fix audit-ci vulnerability in separate PR https://github.com/cdr/code-server/pull/3899
- [x] clone vscode
- [x] get tests working
- [x] submit new test for `asWebviewUri` https://github.com/microsoft/vscode/pull/130148

### DIFF
--- a/lib/vscode/src/vs/workbench/api/common/shared/webview.ts
+++ b/lib/vscode/src/vs/workbench/api/common/shared/webview.ts
@@ -55,9 +55,12 @@ export function asWebviewUri(
 		});
 	}
 
+	// NOTE@coder: Add the port separately because if the port is in the domain the
+	// URL will be invalid and the browser will not request it.
+	const authorityUrl = new URL(`${resource.scheme}://${resource.authority}`);
 	return URI.from({
 		scheme: Schemas.https,
-		authority: `${resource.scheme}+${resource.authority}.${webviewRootResourceAuthority}`,
+		authority: `${resource.scheme}+${authorityUrl.hostname}.${webviewRootResourceAuthority}${authorityUrl.port ? (':' + authorityUrl.port) : ''}`,
 		path: resource.path,
 		fragment: resource.fragment,
 		query: resource.query,
--- a/lib/vscode/src/vs/workbench/contrib/webview/browser/pre/service-worker.js
+++ b/lib/vscode/src/vs/workbench/contrib/webview/browser/pre/service-worker.js
@@ -256,7 +256,7 @@ async function processResourceRequest(event, requestUrl) {
 
 	const firstHostSegment = requestUrl.hostname.slice(0, requestUrl.hostname.length - (resourceBaseAuthority.length + 1));
 	const scheme = firstHostSegment.split('+', 1)[0];
-	const authority = firstHostSegment.slice(scheme.length + 1); // may be empty
+	const authority = firstHostSegment.slice(scheme.length + 1) + requestUrl.port ? (':' + requestUrl.port) : ''; // may be empty
 
 	for (const parentClient of parentClients) {
 		parentClient.postMessage({


lib/vscode/src/vs/workbench/api/common/shared/webview.ts
```diff
@@ -55,9 +55,15 @@ export function asWebviewUri(
...
+	// NOTE@coder (@jsjoeio)
+	// We're not exactly sure how the resource.authority is used in the URI
+	// but we noticed resources not loading correctly when the authority
+	// contained a port like localhost:8080.
+	// As a temporary workaround, we split the authority at the first colon
+	// and use the first part e.g. localhost
 	return URI.from({
 		scheme: Schemas.https,
-		authority: `${resource.scheme}+${resource.authority}.${webviewRootResourceAuthority}`,
+		authority: `${resource.scheme}+${resource.authority.split(':')[0]}.${webviewRootResourceAuthority}`,
```
jawnsy, line 66: Be careful with this - authority could be an IPv6 address, which will contain colons and possibly also a port number, for example: `https://[2607:f8b0:4005:80f::200e]` (this is Google's IPv6 address)

It might be preferable to use something like `new URL("https://localhost:8443/test/abc").hostname` to do this instead?

jsjoeio: Oh, didn't know or consider that. Okay, maybe we should do something like this instead?

```diff
+${new URL(`http://${resource.authority}/test/abc``).hostname}
-${resource.authority.split(':')[0]}
```

Would that be okay?

vapurrmaid: Note - the above doesn't account for any exceptions thrown from the URL constructor.

https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

jsjoeio: Good call @vapurrmaid - let's have @code-asher chime in with his thoughts. I think he's in favor of the least intrusive patch.

vapurrmaid: All good, it's probably benign - just something to weigh

jawnsy: I'm not too sure what guarantees we get regarding `remoteInfo`, but I think we can figure this out together.

* On line 46-48, if `resource` has a scheme of `http` or `https` then we return it as-is
* On line 50-56, if `remoteInfo` is set and the original resource is a `file` (e.g. `file:///home/jsjoeio/coolfile.json`), then we use the scheme `vscode-remote` with the remote authority and file path (so, without running the code, I'd guess it looks something like: `vscode-remote://joes-server.com/home/jsjoeio/coolfile.json`)
* If we get to line 64, then we know that resource is neither http/https (otherwise it would've exited early). The original resource may have been a non-HTTP/HTTPS URI, or it may be a remote resource. So then we construct a new URI with `https` scheme and authority set to something like: `https://file+joes-server.com.vscode-resource.vscode-webview.net`)

If someone passes a remote resource with a port number, then we currently end up with something like: `https://file+joes-server.com:5555.vscode-resource.vscode-webview.net` because the original scheme was `file`, the remote authority is `joes-server.com:5555` and everything else is appended.

I'm not sure whether the correct behavior is to drop the original port number (in which case, we want to change `remoteInfo.authority` from `joes-server.com:5555` to `joes-server.com`) or if we want to preserve the port number, but move it to the appropriate part of the new URL. For example, that would look like: `https://file+joes-server.com.vscode-resource.vscode-webview.net:5555`

The VSCode URI is designed to work with generic URIs (which can either be a URN or URL) but we know that we're working with a URL in this case, since the schema is `https`. For the distinction between URI, URL, and URN, see the ASCII diagram here: https://github.com/microsoft/vscode/blob/77f04697139d5c86dd284eb53d712b966c6d7dec/src/vs/base/common/uri.ts#L82-L97

jsjoeio: Thanks for the excellent notes @jawnsy 🙌  I wonder if moving it to the appropriate part of the new URL would work 🤔 I can test locally.

jsjoeio: Heeeey! Look at that - it worked @jawnsy 

```typescript
	const [host, port] = resource.authority.split(':', 2)
	const authority = `${resource.scheme}+${host}.${webviewRootResourceAuthority}${port ? `:${port}` : ''}`;
	return URI.from({
		scheme: Schemas.https,
		authority,
		path: resource.path,
		fragment: resource.fragment,
		query: resource.query,
	});
```

So is the behavior we want? I wonder if I should submit a fix upstream to handle this scenario?

code-asher: Ahhhh IPv6 is an interesting edge case.

To think out loud and summarize a bit, what happens here is that VS Code has a service worker that intercepts all requests in the web view (the iframe). It then extracts the "real" URI from this fake one and sends it to the parent (via `postMessage`) where the actual request is made after which the response is sent back (again through `postMessage`) for the service worker to resolve with.

So it doesn't matter *too* much what this URL is as long as:

1. The browser thinks it is valid and so actually makes the request inside the iframe for the service worker to then catch.
2. The URL has the info in it required to reconstruct the real URL and complete the real request successfully.

On the first point, the browser does nothing at all if there is a colon anywhere in the hostname (encoding it doesn't work either). For example, these all silently do nothing (no error at all from the browser indicating these are invalid):

```
<script src="https://test:5000.com/test.js"></script>
<script src="https://test%3A5000.com/test.js"></script>
<script src="https://test:.com/test.js"></script>
<script src="https://test.com:d/test.js"></script>
```

Maybe there is an argument to be made here that VS Code's `URI` class should have thrown an error rather than spit out an invalid URL but I'm not sure.

In any case it seems that the solution must involve removing the port from the domain. `new URL(url).hostname` feels like an excellent solution to this. I think we will be safe here since we use the host header as the remote authority so it should be valid.

That leads to the second point. Is that port necessary for VS Code to reconstruct the real URL? If it is, we should add it back in the way @jawnsy is proposing. This may mean we also have to tweak the reconstruction side of things to correctly extract the port assuming VS Code hasn't accounted for this case.

But since it apparently works without the port (and without the authority entirely), do we even need to worry about it? Actually I'm confused why the authority would be required at all. The scheme and path should be all VS Code needs since it already knows where the backend is.

Anyway, the reconstructions seems to happen here:

https://github.com/cdr/code-server/blob/570cb6983281abbad617a8bbc9e1aa6a3d0048b5/lib/vscode/src/vs/workbench/contrib/webview/browser/pre/service-worker.js#L257-L271

So sure enough if we strip the port then this `authority` here will be port-less but we could patch this to add it back in. This is probably the "correct" solution. I wonder why it works without the authority though? Note the `may be empty` in the above code.

Well if we chase this down it eventually gets passed through to our file provider through the file provider channel. This channel (like all the channels) are set up during the initial connection and remain open so the authority has no use here. The file provider simply uses `fsPath` off the URI and ignores all the other properties of the URI.

In fact if you log all the requests to the file provider you'll see that many of them don't have the authority at all. So maybe we could just remove it but...

That doesn't mean there is absolutely no case where the authority won't be used. For example if a plugin implements their own scheme handler and they try to read the authority for some reason, maybe that could be an issue.

So I think the most correct solution is to stick the port in the proper place then add it back to the authority during the reconstruction linked above.

code-asher: Something like this maybe?

```javascript
const authority = firstHostSegment.slice(scheme.length + 1) + requestUrl.port ? (":" + requestUrl.port) : ""
```

jsjoeio: @code-asher all that makes sense to me. Should we still submit a patch upstream?

lib/vscode/src/vs/workbench/contrib/webview/browser/pre/service-worker.js
```diff
@@ -256,7 +256,7 @@ async function processResourceRequest(event, requestUrl) {
 
 	const firstHostSegment = requestUrl.hostname.slice(0, requestUrl.hostname.length - (resourceBaseAuthority.length + 1));
 	const scheme = firstHostSegment.split('+', 1)[0];
-	const authority = firstHostSegment.slice(scheme.length + 1); // may be empty
+	const authority = firstHostSegment.slice(scheme.length + 1) + port ? (":" + port) : ""
```
code-asher, line 259: Looks like CodeTogether cut out before it got the rest of the changes :stuck_out_tongue: 

Anyway, together with the other change this would look like:

```suggestion
	const authority = firstHostSegment.slice(scheme.length + 1) + requestUrl.port ? (":" + requestUrl.port) : ""; // may be empty
```

code-asher (changes requested):

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
code-asher: It would be nice if we could get it fixed upstream. I suppose worst thing that happens is that it gets closed.

jsjoeio: Cool cool! Let me fix this up here then re-request a review from everyone and then I'll look into following up with a PR upstream to fix there too

jsjoeio:
Something strange is happening so converting back to a draft

jsjoeio:
All fixed - ready for review!

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
```diff
@@ -55,9 +55,12 @@ export function asWebviewUri(
 		});
 	}
 
+	// NOTE@coder: Add the port separately because if the port is in the domain the
+	// URL will be invalid and the browser will not request it.
+	const authorityUrl = new URL(`http://${resource.authority}`);
```
jawnsy, line 60: Are we guaranteed that the incoming authority is a valid URL? otherwise as Grey mentioned before, we'd have to try/catch a possible error right?

Also it does't matter since we're not using the scheme part, but you might want to consider using https here just because it looks a little nicer

jawnsy (approved):

jawnsy:
LGTM. Are we still planning to offer this patch upstream?

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
TeffenEllis: AFAIK `authority` may represent any fully qualified host e.g. `localhost:8080`, `example.com/any/path`. The  scheme [may not necessarily be http ](https://github.com/cdr/vscode/blob/main/src/vs/base/common/network.ts)

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
```diff
@@ -55,9 +55,12 @@ export function asWebviewUri(
...
 	return URI.from({
 		scheme: Schemas.https,
-		authority: `${resource.scheme}+${resource.authority}.${webviewRootResourceAuthority}`,
+		authority: `${resource.scheme}+${authorityUrl.hostname}.${webviewRootResourceAuthority}${authorityUrl.port ? (':' + authorityUrl.port) : ''}`,
```
TeffenEllis, line 63: @jsjoeio IMO the bug stems from modifying URLs as strings, rather than leaning on the browser’s innate URL constructor Would something like this help with URL?
```ts
// e.g. new URL(‘abc://localhost:3000/foo/bar?baz=qux')
// The scheme isn’t too big of a deal here but it helps to lean on what’s known from above.
const authorityUrl = new URL(`${resource.scheme}://${resource.authority}`);

authorityUrl.hostname = `${authorityUrl.hostname}+${resource.authority}.${webviewRootResourceAuthority}`;
// Remove the protocol
const authority = authorityUrl.toString().slice(`${authorityUrl.protocol}//`.length);

return URI.from({
  scheme: Schemas.https,
  authority,
  path: resource.path,
  fragment: resource.fragment,
  query: resource.query,
});
```

harrylepotter-win:
@jsjoeio  thankyou so much for this.

jsjoeio:
> LGTM. Are we still planning to offer this patch upstream?

Yessir! Added to the PR description just now: https://github.com/microsoft/vscode/pull/130148

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
jsjoeio: Converting to draft stage and then i'll re-request a review

vapurrmaid: Should we `try/catch` the URL construction?

jsjoeio: Sure! Good call!

jsjoeio: Okay...looking at your solution, I don't think I fully understand how it helps. I tried it, but I think either the hostname or the port gets mangled (though I might be doing something else wrong). I'm going to leave as is for now. And if you want to follow-up with more details or have me submit another PR to improve this, happy to do so!

lib/vscode/src/vs/workbench/api/common/shared/webview.ts
jsjoeio: Good call! I'll use the `resource.scheme` here instead of `http` or `https`

code-asher: We set the remote authority here: 

https://github.com/cdr/code-server/blob/cde28a0a0e29381fa5a631da092f7065bde35ccf/src/node/routes/vscode.ts#L34

So potentially it could be blank although I'm not sure how often the host header is missing.

code-asher: Personally I think letting the error throw is fine although maybe we should catch and re-throw it with a more helpful message ("host header is invalid" or something like that).

code-asher: Although VS Code does not make the authority mandatory so we should probably gracefully handle a blank authority.

PR #3910, opened 2021-08-06
chore(ci): replace playwright gh action

vapurrmaid:
This action is no longer needed, see README for the action here: https://github.com/microsoft/playwright-github-action#%EF%B8%8F-you-dont-need-this-github-action-%EF%B8%8F

The suggested way to install OS deps for playwright is simply:

```
npx playwright install-deps
```

Reference: https://playwright.dev/docs/ci/#github-actions

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -352,8 +352,8 @@ jobs:
         with:
           node-version: "14"
 
-      - name: Install playwright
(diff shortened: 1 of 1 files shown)


jsjoeio (approved):
Beautiful! And it looks like all the e2e are still working! Thanks Grey 🙌

PR #4020, opened 2021-08-23
chore: add example for parameter/value to Helm Values

### LINKED ISSUE #4009: Unknown option --no-auth
## Values.yaml

```
# Optional additional arguments
extraArgs:
#  - --allow-http
  - --no-auth
   ```

## Steps to Reproduce

1. Uncomment the Extra Argument for --no-auth in the helm-chart/values.yaml file
2. Install the Helm chart

## Expected

Chart should install and code-server should not require a password

## Actual

Chart installs but the code-server pod is in error

## Logs

$ k logs code-server-694b6f4b98-j6srk -f
[2021-08-19T17:51:09.606Z] error Unknown option --no-auth

jawnsy:
Fixes #4009

### DIFF
--- a/ci/helm-chart/values.yaml
+++ b/ci/helm-chart/values.yaml
@@ -57,8 +57,18 @@ ingress:
 
 # Optional additional arguments
 extraArgs: []
-#  - --allow-http
-#  - --no-auth
+  # These are the arguments normally passed to code-server; run
+  # code-server --help for a list of available options.
+  #
+  # Each argument and parameter must have its own entry; if you use
+  # --param value on the command line, then enter it here as:
+  #
+  # - --param
+  # - value
+  #
+  # If you receive an error like "Unknown option --param value", it may be
+  # because both the parameter and value are specified as a single argument,
+  # rather than two separate arguments (e.g. "- --param value" on a line).
 
 # Optional additional environment variables
 extraVars: []


kylecarbs (approved):
Wonderful doc addition!

code-asher (approved):

jsjoeio:
I'm not sure why Docs preview failed 🤔 Re-running the other failed jobs too. Should be fine though!

jawnsy:
@jsjoeio It's probably because I'm doing this from a fork, if I can't get it into a good state then I'll push it as a branch instead - I wanted to use the same workflow that our external contributors would use, as I think that is an important use case for us to support

jsjoeio:
> @jsjoeio It's probably because I'm doing this from a fork, if I can't get it into a good state then I'll push it as a branch instead - I wanted to use the same workflow that our external contributors would use, as I think that is an important use case for us to support

Love that you're making sure we dogfood that workflow! I'm seeing this:

image

cc @BrunoQuaresma is there any reason this workflow doesn't work for forks? Is it because `m` is private? I'm not fully sure how this should work 🤔

BrunoQuaresma:
@bpmct I think it is because of that since forks don't have access to our internal secrets.

jawnsy:
Ah, I think we have to configure this for our secrets: https://docs.github.com/en/actions/reference/environments

Not worth doing right now though, I didn't change docs so I think this is safe to merge :crossed_fingers:

PR #4034, opened 2021-08-24
docs: add note about external domain in requirements

### LINKED ISSUE #3921: Add docs around Google Cloud https
"In order to enable https, you have to add an external domain."

We should give the docs for setting up code-server on Google Cloud a run through.

https://github.com/cdr/code-server/discussions/3907

jsjoeio:
Fixes #3921


docs/requirements.md
```diff
@@ -53,3 +53,4 @@ Notes:
 
 - To lower costs, you can shut down your server when you're not using it.
 - We recommend using the `gcloud cli` to avoid using the GCP Dashboard if possible.
+- To enable HTTPS, you have to add an external domain.
```
jawnsy, line 56: Is this strictly true? In air-gapped networks, you can use self-signed certificates or certificates signed by an internal certificate authority, and if you control DNS then everything is OK.

I think this statement is fine but would suggest we soften it to "For serving code-server over HTTPS, we recommend using an external domain name along with a service such as Let's Encrypt"

jawnsy (approved):

docs/requirements.md
jsjoeio: Ahhh okay, I didn't know. This came after a GitHub Discussion. Let's go with your wording instead.

jsjoeio, line 56:
```suggestion
- For serving code-server over HTTPS, we recommend using an external domain name along with a service such as Let's Encrypt
```

TeffenEllis (approved):
This looks great. I’m curious how we can test such domains in the future, maybe updating the health check endpoint to ping itself to see if the server is accessible from the outside.

PR #4036, opened 2021-08-24
refactor: update image in readme

### LINKED ISSUE #3843: Update image in README
> I noticed the hero image on the code-server docs is an older release of code-server - maybe keep that current to alleviate any confusion.

Mark brought up a good point saying the image in the `README` (also shows up in code-server docs on website) shows a very old version of code-server.

It's not worth automating IMO but we could update the image and remove any version references so it does not look outdated or cause confusion.

jsjoeio:
Updates the main image used in the `README`, specifically removing any mention of the code-server or VS Code version so that it doesn't get outdated

Fixes #3843


jawnsy:
This makes sense to me. Do we have an SVG version of this image?

jawnsy (approved):

jsjoeio:
> Do we have an SVG version of this image?

No, but I can export from Figma I think and make one! Should I do that instead?

jawnsy:
@jsjoeio SVGs are always nicer because they scale better and are smaller, plus we can do fancy animations like the [Expo CLI project](https://github.com/expo/expo-cli) that I think you retweeted about once!

TeffenEllis (approved):

Issue #1105, opened 2019-10-24
Intellij Idea Keybindings clipboardPasteAction not found


NickBolles:
- `code-server` version: 
```
code-server --version
v10.15.1
```
- OS Version: docker
- Extension: [Intellij Idea Keybindings](https://github.com/kasecato/vscode-intellij-idea-keybindings)

## Description

When pasting in the editor I get a `"editor.action.clipboardPasteAction' not found` toast notification and pasting fails. But copying works successfully. 

I can also copy and paste just fine in the terminal.

Issue on the extension repo: https://github.com/kasecato/vscode-intellij-idea-keybindings/issues/163 

## Steps to Reproduce

1. 
1.

NickBolles:
For anyone else finding this, I just overrode this keybinding with the following in my keybindings.json file
```
  // Remove incompatible paste action
  {
    "key": "ctrl+v",
    "command": "-editor.action.clipboardPasteAction",
    "when": "editorTextFocus && !editorReadonly"
  },
```

danger89:
Uh.. yup something is wrong. I'm using v2 of code-server.

![image](https://user-images.githubusercontent.com/628926/68033142-6b42be00-fcbf-11e9-94b9-131950c742ec.png)

code-asher:
You might be running into this: https://github.com/cdr/code-server/issues/1106

If so, the best solution is to use HTTPS. There's a workaround for pasting but I'm not sure there's one for copying (it might work to just unbind it).

danger89:
I'm using HTTPS. But I'm using Nginx reverse proxy in between.

code-asher:
> **EDIT:** Lets put it differently, I do not need the HTTPS from code-server. I can use plain communication between Nginx and the localhost:8080, and my separate Nginx reverse proxy server is taking care of the SSL stuff (which is running on port 443, publicly available).

That sounds good; I use the same setup. As long as the browser itself is seeing an HTTPS all is well. But I realized I missed some information about Firefox so I elaborated here: https://github.com/cdr/code-server/issues/1106#issuecomment-548911682

bgold0:
Should note that pasting works fine when nothing is selected. It's only when things are selected that you get this error.

NickBolles:
> Should note that pasting works fine when nothing is selected. It's only when things are selected that you get this error.

My initial issue was happening whether or not I had anything selected. That issue is resolved (by removing the command in the intellij extension). Maybe we should create a new issue for any other root causes.

[...]

Issue #33, opened 2019-03-06
Collaboration (multi user) support


xPaw:
It's not documented whether multiple users can use one self hosted installation at once or not. Is this a supported use case?

sr229:
If I remember, this is a planned feature. But if this was a 1:1 code-oss port. VS Live Share could work on this.

kylecarbs:
Currently, we do not support multiple users collaboratively editing on the same `code-server` instance. We've thought of this and would love to hear what your thoughts are. 

In our world, it'd be great to provide collaborative access, but never allow the collaborator to have control over the host machine in any way (for security reasons).

sr229:
I think we should do this the way how VS Live share does it - RBAC by limiting views to read-only and restricting terminal access.

xPaw:
I currently use https://github.com/c9/core for collaboration, where multiple users can log into the same server and work together.

code-server interests me as it uses VS Code so it heavily supported, unlike Cloud9. For my use case, permissions aren't really a problem.

nhooyr:
@kylecarbs does vscode live share work?

kylecarbs:
I believe it _would_ if the extension was open-sourced. Unfortunately, we can't provide support for it.

See: https://github.com/Microsoft/vscode/issues/38733#issuecomment-388213016

zk-g:
This is a must for a cloud ide, Without this feature I don't even see an advantage of using a cloud editor as there are few ways to share the workspace through the client: and in plus require less resources than the browser.
So definitely +1

zk-g:
Maybe also with an interactive button in the status bar showing connected user and options to manage permissions.
This could be a really game changer

PhilsLab:
I'd have a look at how MS Word Online handles it:
![Word Online](https://user-images.githubusercontent.com/15452526/53984472-aaf9db80-4119-11e9-9616-7e88776a309c.png)

I like the idea of having a different cursor for every person actively editing,
this could be implemented via WebSockets / WebRTC.

Also, when providing a login functionality, there should be a modern backend for this, not just local users.
A solution that would fit very many people would be to integrate SAML and/or OpenID authentication.

As I mentioned in #85, you could use Keycloak (which can handle both SAML and OpenID) as an IdP:
Keycloak: [keycloak/keycloak](https://github.com/keycloak/keycloak)
Keycloak NodeJS integration: [keycloak/keycloak-nodejs-connect](https://github.com/keycloak/keycloak-nodejs-connect)

[...]

PR #4048, opened 2021-08-26
chore: add stale bot

jsjoeio:
@jawnsy brought this up the other day. Then GitHub mentioned we should it so I thought why not 🤷‍♂️

I will take responsibility if it goes horribly wrong 😂

The way this works is we have a `stale.yml` under `.github` which is the configuration and then @jawnsy enabled [Stale bot](https://github.com/probot/probot) it at the `cdr` org level. 

Fixes N/A

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,12 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 180
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 5
+# Label to apply when stale.
+staleLabel: stale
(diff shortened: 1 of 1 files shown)


Lemmingh:
I guess you forgot to set `exemptLabels` and `exemptMilestones`. An On Deck bug is closed:
#1105

jsjoeio:
@Lemmingh thanks for pointing that out! 

I guess we should add that. If no one responds though, it's still probably stale and makes sense to close, but I'll open an issue to do this.

bpmct:
Does that mean #33 will be considered stale at some point?

jsjoeio:
@bpmct yes if there is no activity for 180 days!

PR #4060, opened 2021-08-28
revert: "docs(guide): add ssh into code-server on vs code"

jawnsy:
This reverts commit efbfe060cdba9a01ee3d7a48b0f6c862ca8e91f8 as
some of the contents broke our doc site generator.

PR #4042, opened 2021-08-25
docs(guide): add ssh into code-server on vs code

jsjoeio:
This PR adds steps to `guide.md` for ssh'ing into code-server using Cloudeflare's Argo Tunnels. 

Inspired by https://github.com/cdr/code-server/discussions/3961

Thanks to @AyaanZaveri @im-coder-lg for figuring this out and sharing!

Fixes #4017

PR #4071, opened 2021-08-30
fix: revert revert and add ssh back into guide docs

jsjoeio:
#4042 broke our docs website, #4060 reverted the change.

> There are 2 problems in there. There are 4 back ticks at the end of one block. Changing the line “```output” to  “```console” should fix it

Thanks to @mterhar, I think we found the issue. 

We're working with @BrunoQuaresma to prevent issues like this in the future.

Related:
- https://github.com/cdr/code-server/pull/4042
- https://github.com/cdr/code-server/pull/4060
- https://github.com/cdr/code-server/discussions/4057

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -15,6 +15,9 @@
   - [Stripping `/proxy/<port>` from the request path](#stripping-proxyport-from-the-request-path)
   - [Proxying to create a React app](#proxying-to-create-a-react-app)
   - [Proxying to a Vue app](#proxying-to-a-vue-app)
+- [SSH into code-server on VS Code](#ssh-into-code-server-on-vs-code)
+  - [Option 1: cloudflared tunnel](#option-1-cloudflared-tunnel)
+  - [Option 2: ngrok tunnel](#option-2-ngrok-tunnel)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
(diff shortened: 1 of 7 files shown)


jsjoeio:
Checked the preview. Looks good! @BrunoQuaresma any idea why the docs are making this shield round?

When I look at it in the browser, it's square like the others. See [here](https://img.shields.io/badge/Terminal-2E2E2E?style=for-the-badge&logo=Windows+Terminal&logoColor=ffffff)

image

BrunoQuaresma:
@jsjoeio I will check it. Probably a random CSS class. Also, when adding badges, it is recommended to make it a link, even if it is only an image. When it is an image, you can make it a link pointing to the image itself. It is how GH does to make their badges inline. You can inspect how our README.md looks like on GH inspecting the badges.

BrunoQuaresma:
@jsjoeio also, looks like we are not closing a code block here: 
Screen Shot 2021-08-31 at 10 37 57
It looks like the `bash` block is inside of another block.

jsjoeio:
> When it is an image, you can make it a link pointing to the image itself. It is how GH does to make their badges inline. You can inspect how our README.md looks like on GH inspecting the badges.

Ah okay, I'll match that.

jsjoeio:
> also, looks like we are not closing a code block here:

Nice catch! I'll fix that too

jsjoeio:
> > When it is an image, you can make it a link pointing to the image itself. It is how GH does to make their badges inline. You can inspect how our README.md looks like on GH inspecting the badges.
> 
> Ah okay, I'll match that.

I just checked and these are already links 🤔 Not sure what else we can do then. 

image

jsjoeio:
Ready for another review! @BrunoQuaresma

BrunoQuaresma:
@jsjoeio I sent a PR just fixing the link on the terminal badge.
Screen Shot 2021-09-01 at 10 14 03

BrunoQuaresma (approved):
LGTM

jsjoeio:
I'll merge after I get this merged in: https://github.com/cdr/code-server/pull/4085

im-coder-lg:
@jsjoeio I need to tell, the ngrok URL changes every time, as well as the port. One time you run ngrok, you get 0.tcp.ngrok.io and port 19028, but the next time you run it, you will get 2.ngrok.io and 10320, and it goes on. So, add `in this case,` to the ngrok guide where the port is mentioned and add a note that the port and the URL always changes so you have to use the URL and the port given on the terminal.

jsjoeio:
Ahh good call @im-coder-lg - thanks for mentioning that!

im-coder-lg:
:+1:

PR #4085, opened 2021-09-01
fix(security): tar vulnerability

jsjoeio:
This PR is based off of https://github.com/cdr/code-server/pull/4071 and fixes a `tar` vulnerability.

### DIFF
--- a/lib/vscode/package.json
+++ b/lib/vscode/package.json
@@ -218,6 +218,7 @@
     "elliptic": "^6.5.3",
     "nwmatcher": "^1.4.4",
     "chrome-remote-interface": "^0.30.0",
-    "glob-parent": "^5.1.2"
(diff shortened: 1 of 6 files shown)


TeffenEllis:
@jsjoeio I believe this may be outdated as we no longer need to tar for extensions

jsjoeio:
> @jsjoeio I believe this may be outdated as we no longer need tar for extensions in an upcoming fork clean up PR

Sweet! I wonder when that will land though? If we have PRs coming in before that, should we fix this sooner? 

(though looks like something is still failing)

jsjoeio:
Maybe I need to re-run `yarn`? I'll check this

https://github.com/cdr/code-server/pull/4085/checks?check_run_id=3489226386

jawnsy (approved):

Issue #4094, opened 2021-09-02
Add test coverage for src/node/app.ts


jsjoeio:
UPDATED

![image](https://user-images.githubusercontent.com/3806031/133159434-03126d2f-8366-48bc-91df-20d1d45ee963.png)

We're currently missing test coverage for the following lines:
- 36-40
- 45-52
- ~~72~~ coverage added in #4099
- ~~77~~ coverage added in #4099

PR #4099, opened 2021-09-03
feat: add tests for src/node/app.ts

jsjoeio:
This adds a couple tests for ensureAddress.

(Note: related to #4094. I still need to add more tests to complete the issue but will do so in separate PRs.)


jawnsy (approved):
tests are always nice :+1:

PR #4105, opened 2021-09-04
Resolve remote authority to initialize tunnel API

kylecarbs:
This enables the `vscode.workspace.registerRemoteAuthorityResolver` API to be executed.

It's specifically scoped to the "coder-link" scheme at the moment to reduce external dependency on forking VS Code functionality.

The remote host doesn't resolve to an extension host like the API expects, but instead we only utilize the tunnel functionality.


jsjoeio (approved):
Tests are passing and I don't see how this would break anything. 

Nice comments too 🚀

Issue #1397, opened 2020-03-05
Publish windows release


nhooyr:
See https://github.com/cdr/code-server/issues/1306

danny-laa:
@nhooyr Any updates? Does 3.0 release support Windows?

nhooyr:
There are no self contained windows releases yet.

We haven't tested but the npm package might work on windows. `yarn global add @coder/code-server`

maxloh:
> There are no self contained windows releases yet.
> 
> We haven't tested but the npm package might work on windows. `yarn global add @coder/code-server`

I was able to install the npm package in windows, but there are a few bugs,

1. src/browser/media/manifest.json not found

2. All vscode-remote-resource path returns 404 not found, but this bug can easily be solved by removing leading slash in file path agument, eg.  
`http://127.0.0.1:8080/vscode-remote-resource?path=/c:/Users/Max/AppData/Local/Yarn/Data/global/node_modules/@coder/code-server/lib/vscode/extensions/markdown-basics/language-configuration.json` ➡ 404 not found  
`http://127.0.0.1:8080/vscode-remote-resource?path=c:/Users/Max/AppData/Local/Yarn/Data/global/node_modules/@coder/code-server/lib/vscode/extensions/markdown-basics/language-configuration.json` ➡ Correct path

3. Syntax highlighting, extensions and themes not working (because of 2)

nhooyr:
@maxloh Awesome thanks for looking into it!

nhooyr:
The package is now just `code-server` on npm. We were able to get ownership from the previous owner.

> src/browser/media/manifest.json not found

Fixed.

> All vscode-remote-resource path returns 404 not found, but this bug can easily be solved by removing leading slash in the path agument, eg.

@code-asher 

> Automatic update not working because https://github.com/cdr/code-server/releases/download/3.2.0/code-server-3.2.0-win32-x86_64.tar.gz does not exist

We'll be removing automatic updates soon. See #1532 .

> Furthermore, running the post install scripts requires Visual Studio and "VC++ 2015.3 v140 toolset for desktop (x86,x64)" to be installed, but since running code-server does not require electron to be built, I think it may be unnecessary.

Will mention this in docs, it is required as vscode uses native modules that have to be built.

nhooyr:
Asher has fixed the paths, will publish a new release soon.

nhooyr:
We're going to need to test/rewrite any shell scripts that will have to run on Windows.

[...]

Issue #4027, opened 2021-08-24
Investigate changing docs preview workflow 


jsjoeio:
We recently added a docs preview to code-server (thank you @BrunoQuaresma ) but now have an issue. If someone forks code-server, submits a PR, the docs preview fails. My hunch says we can't fix this since `m` is private and there are some tokens/secrets we rely on.

https://github.com/cdr/code-server/blob/main/.github/workflows/docs-preview.yaml
https://github.com/cdr/code-server/pull/4020/checks?check_run_id=3413784650

We should investigate fixing this.

jsjoeio:
@mterhar made a good suggestion

> We could have it check if it's a fork and if so, deploy to github pages for that repo
> `if: github.repository == 'cdr/code-server'` then run our preview app.
> `if: github.repository != 'cdr/code-server'` then run a github pages deploy

jsjoeio:
@jawnsy said to look here: https://docs.github.com/en/actions/reference/environments

https://github.com/cdr/code-server/pull/4020#issuecomment-905019508

jsjoeio:
Another option would be to skip this workflow if it's from a fork, but then we might not catch bugs 🤔

jsjoeio:
@bpmct found a good solution to this:

> However, it looks like we can allow forks to use the secrets from the upstream repo and even add approvals. https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci

vapurrmaid:
While this comment doesn't help `code-server` repository in isolation, I'd like to throw-in that maybe we should seek a way to preview docs changes that don't require cloning a monorepo - there's probably a minimal environment we can move to a public repo like `cdr/docs-engine` or something along those lines

im-coder-lg:
Hmm... I know this is not my place, but can't we get the Vercel app(noticed the Coder site is deployed on Vercel)? I mean, after installing Vercel over here, and enabling that, if anyone makes a PR, even if or if it's not a docs change, it will give a preview, just to be on the safe side. You don't need extra files, all you would need is the Vercel app. Since the repo is directly connected there, it's easier than tokens and workflows, right?

jawnsy:
@im-coder-lg Great question, and thanks for your contributions! We definitely want to make it easier for people outside our company to contribute to code-server, hence this issue.

[...]

PR #4116, opened 2021-09-07
fix: run postinstall script explicitly with bash

maxloh:
Currently, Windows default script-shell (cmd) fails to run the postinstall script.

This commit fixes the problem by running postinstall.sh explicitly with the default bash executable of the OS.

Related: #1397

### DIFF
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -48,7 +48,7 @@ bundle_code_server() {
   {
     "commit": "$(git rev-parse HEAD)",
     "scripts": {
-      "postinstall": "./postinstall.sh"
+      "postinstall": "bash ./postinstall.sh"
     }
   }
 EOF


jsjoeio:
First, thank you @maxloh for adding this! 🎉

I think this is a great first step toward improving the support on Windows. 

I am not on expert in this realm, but I assume this shouldn't cause any issues on other platforms (macOS/Linux)?

@code-asher I'll let you give the final approval since you have more expertise here.

jsjoeio:
side note: don't worry about the docs preview failing, we need to look into that working on forks. See https://github.com/cdr/code-server/issues/4027

code-asher (approved):
Interesting, I guess `cmd` must not take hashbangs into account. This change does make my senses tingle but I think it should be OK.

Long-term maybe we should rewrite this script to use Node.

jsjoeio:
> Long-term maybe we should rewrite this script to use Node.

You had me at Node 😍 But agreed, then we could write tests for it more easily

maxloh:
@code-asher cmd does not recognize forward slashes. It tries to run `.` instead

```
C:\>./postinstall.sh
'.' is not recognized as an internal or external command,
operable program or batch file.
```

If we replace forward slashes with backslashes, Windows will open the .sh file with the associated application, which maybe a random text editor but not bash.

```
C:\>.\postinstall.sh
```

To run the .sh file with the default bash command executable, we will need to run it explicitly with bash

```
C:\>bash ./postinstall.sh
```

code-asher:
Ohh that makes sense! Thank you for the clarification.

PR #4120, opened 2021-09-08
Push latest Docker tag along with version tag

code-asher:
https://github.com/cdr/code-server/discussions/3920#discussioncomment-1281283
https://docs.docker.com/engine/reference/commandline/buildx_bake/

### DIFF
--- a/ci/release-image/docker-bake.hcl
+++ b/ci/release-image/docker-bake.hcl
@@ -12,6 +12,9 @@ group "default" {
 
 target "code-server" {
     dockerfile = "ci/release-image/Dockerfile"
-    tags = ["docker.io/codercom/code-server:${VERSION}"]
+    tags = [
+        "docker.io/codercom/code-server:latest",
+        notequal("latest",VERSION) ? "docker.io/codercom/code-server:${VERSION}" : "",
+    ]
     platforms = ["linux/amd64", "linux/arm64"]
 }


ci/release-image/docker-bake.hcl
```diff
@@ -12,6 +12,9 @@ group "default" {
 
 target "code-server" {
     dockerfile = "ci/release-image/Dockerfile"
-    tags = ["docker.io/codercom/code-server:${VERSION}"]
+    tags = [
+        "docker.io/codercom/code-server:latest",
+        notequal("latest",VERSION) ? "docker.io/codercom/code-server:${VERSION}" : "",
```
jawnsy, line 17: Hmm, this doesn't seem like it'd work correctly in the case where we're trying to re-build an old version (not sure if we ever do that though) since latest would move "backwards"

Seems we'd need some way to check the "current" latest version and compare before we update the tag, to make that work correctly

I guess this seems like a decent enough solution :man_shrugging:

jawnsy (approved):
Seems reasonable to me

ci/release-image/docker-bake.hcl
code-asher: We might be covered by the check here, more as a side effect though:

https://github.com/code-asher/code-server/blob/b4df05f8b9ed3cced01d8a95d7a26f6e2928be58/ci/steps/docker-buildx-push.sh#L21

Maybe we could add a check to see if it is greater than the latest version available and only then add latest.

code-asher: fwiw this does bring us back to what we had before we switched to `buildx` (always push to latest)

jawnsy: Fair enough, don't let me stand in your way! :see_no_evil:

code-asher: Made an issue: https://github.com/cdr/code-server/issues/4121 Thanks for the suggestion!

PR #4010, opened 2021-08-19
Clean up and refactor VSCode integration

TeffenEllis:
This PR addresses portions of #3835, specifically changes associated with making our modifications to upstream more aligned with their coding architecture and style. To keep this PR reviewable, a second PR based off the remaining changes in our [fork branch](https://github.com/cdr/vscode/pull/2) will follow. 

- Fixed linter/prettier mixups during development.
- Cleaned up web config injection, removed separate NLS injection.
- Cleaned up comments for use with intellisense
- Updated types to more often use VScode's built in types.
- Cleaned up URI helpers.
- Organized connection classes into separate files.
- Documented protocol usage, adding trace logging support.
- Fleshed out partial support for remote debugging.
- Updated tests.

### How to review this PR

Despite the file count, much of the changes here are linter corrections, removing duplicate interfaces, and moving existing interfaces into better documented areas. I recommend reviewing with [whitespace changes disabled](https://github.com/cdr/code-server/pull/4010/files?diff=unified&w=1).

PR #4135, opened 2021-09-10
Add vscode via vendor package.

### LINKED ISSUE #3917: Upgrade to VS Code 1.60
Upgrade to VS Code 1.59

TeffenEllis:
While #4010 resolves some incompatibilities with 1.60.0, it's still in review and needs additional testing. This PR updates `lib/vscode` to use a yarn package pointing to our [VS Code fork](https://github.com/cdr/vscode/tree/1.60.0-code-server-patch). Once this is merged, any remaining fork clean up branches will be ported over to the vendored .

Fixes #3917

### Notable changes from upstream

- The log service now defers initialization.
- Several terminal channels were added.
- Several type definitions were adjusted. 

# Our changes
- Remove asar linking.
- Remove module lint check.
- Use yarn for vscode vendoring.

### How to review this PR

This is a somewhat unusual PR since we don’t yet have a diff to start with, however, I’ve made a branch on our fork with the previous changes form `lib/vscode`. This allows you compare all the changes from upstream, as well as the adjustments to our code to fix them in 1.60.0

https://github.com/cdr/vscode/compare/1.57.1-initial...cdr:code-server

### Follow up

#4136 can be merged after this. They are only separated to make reviewing the changes easier to follow.


vendor/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "vendor",
+  "version": "1.0.0",
+  "license": "MIT",
+  "private": true,
+  "scripts": {
+    "postinstall": "./postinstall.sh"
+  },
+  "devDependencies": {
+    "code-oss-dev": "cdr/vscode#96a09a7846538c3bbedb6a2aeca729537bb8202b"
```
kylecarbs, line 10: Can this be in the root package.json?

jsjoeio:
> Remove asar linking.

Happy we're doing this. Those npm changes upstream were not fun to debug so glad we can avoid that.

ci/dev/fmt.sh
```diff
@@ -19,7 +19,7 @@ main() {
     "*.sh"
   )
   prettier --write --loglevel=warn $(
-    git ls-files "${prettierExts[@]}" | grep -v "lib/vscode" | grep -v 'helm-chart'
+    git ls-files "${prettierExts[@]}" | grep -v "lib/vscode" | grep -v "vendor/modules/code-oss-dev" | grep -v 'helm-chart'
```
jsjoeio: Do we still need `grep -v "lib/vscode"`?

ci/dev/lint.sh
```diff
@@ -4,17 +4,14 @@ set -euo pipefail
 main() {
   cd "$(dirname "$0")/../.."
 
-  eslint --max-warnings=0 --fix $(git ls-files "*.ts" "*.tsx" "*.js" | grep -v "lib/vscode")
-  stylelint $(git ls-files "*.css" | grep -v "lib/vscode")
+  eslint --max-warnings=0 --fix $(git ls-files "*.ts" "*.tsx" "*.js" | grep -v "vendor/modules/code-oss-dev" | grep -v "lib/vscode")
+  stylelint $(git ls-files "*.css" | grep -v "vendor/modules/code-oss-dev" | grep -v "lib/vscode")
   tsc --noEmit --skipLibCheck
-  shellcheck -e SC2046,SC2164,SC2154,SC1091,SC1090,SC2002 $(git ls-files "*.sh" | grep -v "lib/vscode")
+  shellcheck -e SC2046,SC2164,SC2154,SC1091,SC1090,SC2002 $(git ls-files "*.sh" | grep -v "vendor/modules/code-oss-dev" | grep -v "lib/vscode")
```
jsjoeio, line 10: Do we still need `grep -v "lib/vscode"`?

ci/dev/update-vscode.sh
```diff
@@ -1,133 +0,0 @@
-#!/usr/bin/env bash
```
jsjoeio, line 1: So happy we no longer need this 🥳

jsjoeio (commented):
This is awesome 🎉 

Seems like it's mostly location changes (I'm not seeing a ton of changes related to VS Code itself). 

The biggest question for me is what's the  explanation behind `vendor/modules`? Is this a yarn naming convention or our own? And it doesn't have to happen here, but it would be nice for us to document _how_ VS Code exists in code-server. i.e. how `yarn` plays into the story and how this all works (for instance is this still considered a git submodule).

(I wish we would have done this when we switched to a git subtree so that the reasoning behind the change + how it works would have been documented).

This could come in a followup PR as a docs addition though.

vendor/package.json
TeffenEllis: Unfortunately not for a few reasons. We don’t want the `postinstall` scripts to run because they’re made for already built and released package which differs in shape than the development version. This will also be installed without the dev dependencies which we for our own build process. Lastly, we work out of `vendor/modules` because their Webpack configuration can’t discern between `code-server/node_modules/` and `code-server/node_modules/vscode/node_modules/`, so it assumes that its own dependencies are not accessible. Anything short of a privileged chroot makes this really difficult to overcome. 

-  has webpack

ci/dev/fmt.sh
TeffenEllis: Until #4136 is release 😄

vendor/package.json
TeffenEllis: @jsjoeio ^

We can follow up with further documentation after #4136

.github/workflows/ci.yaml
```diff
@@ -124,26 +124,26 @@ jobs:
...
-      # that changed a file inside `lib/vscode`. If a commit changes any file in there,
+      # that changed a file inside `vendor/modules/code-oss-dev`. If a commit changes any file in there,
       # the hash returned will change, and we rebuild vscode. If the hash did not change,
       # (for example, a change to `src/` or `docs/`), we reuse the same build as last time.
       # This saves a lot of time in CI, as compiling VSCode can take anywhere from 5-10 minutes.
-      - name: Get latest lib/vscode rev
+      - name: Get latest vendor/modules/code-oss-dev rev
         id: vscode-rev
-        run: echo "::set-output name=rev::$(git log -1 --format='%H' ./lib/vscode)"
+        run: echo "::set-output name=rev::$(cat vendor/modules/code-oss-dev/package.json | jq -r .version)"
```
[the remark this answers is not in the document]
TeffenEllis: I believe so?
https://github.com/cdr/code-server/blob/vscode-1.60.0-yarn/.github/workflows/ci.yaml#L190

jsjoeio:
Seems to be running just fine locally!

I do see this error but guessing extension related (not your changes):

image

Weird...a reload of code-server fixes it 🤷‍♂️

Search for files and the terminal aren't working :(  I tried`cd lib/vscode && yarn` again but that didn't fix it.

https://user-images.githubusercontent.com/3806031/132900766-f0387543-838b-4bc6-ac59-941b59d0b422.mov

image

What am I missing? Do I need to manually delete node-modules?

jsjoeio:
> The terminal process failed to launch: A native exception occurred during launch (The module '/Users/jp/Dev/code-server/vendor/modules/code-oss-dev/node_modules/node-pty/build/Release/pty.node' was compiled against a different Node.js version using NODE_MODULE_VERSION 89. This version of Node.js requires NODE_MODULE_VERSION 83. Please try re-compiling or re-installing the module (for instance, using `npm rebuild` or `npm install`).).

Yeah I tried deleting `node_modules` in `lib/vscode` and running `yarn install` but that didn't help.

> Check the Node.js version that's used by Electron

I see the docs say this but I'm not sure how to check that

vendor/package.json
kylecarbs: Ahh neato. Makes sense!

jsjoeio:
Oh no...I think I broke something. So I ran `cd vendor && yarn install --modules-folder modules --ignore-scripts` and now when I go back to the root and run `yarn watch` I get this error:

```shell
yarn run v1.22.10
$ VSCODE_IPC_HOOK_CLI= NODE_OPTIONS='--max_old_space_size=32384 --trace-warnings' ts-node ./ci/dev/watch.ts
[tsc] [11:59:26 AM] Starting compilation in watch mode...
[vscode] $ npm-run-all -lp watch-client watch-extensions
/bin/sh: npm-run-all: command not found
error Command failed with exit code 127.
[vscode] info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
vs code watcher terminated unexpectedly
killing vs code watcher
killing tsc
killing watch
error Command failed with exit code 127.
```

I'm wondering if I should **not** use `--ignore-scripts` when developing locally (i.e. running `yarn watch`).

TeffenEllis:
@jsjoeio,

I recommend doing `yarn clean` to completely erase your modules, then `yarn` from the project directory. Vendor modules will automatically be installed and `yarn watch` can be run. However, if you encounter that terminal error, your node version differs from a built package and you’ll have to do the following:

```sh
 cd vendor/modules/code-oss-dev
 nom rebuild
 ```

From there, `yarn watch` should behave correctly.

**Comment by jsjoeio** [comment]
Oh man... `yarn clean` didn't work.  Going to try the second approach

image

File: ci/dev/fmt.sh [hunk 6]
```diff
@@ -32,13 +32,14 @@ main() {
...
   doctoc --title '# Termux' docs/termux.md > /dev/null
 
-  if [[ ${CI-} && $(git ls-files --other --modified --exclude-standard) ]]; then
-    echo "Files need generation or are formatted incorrectly:"
-    git -c color.ui=always status | grep --color=no '\[31m'
-    echo "Please run the following locally:"
-    echo "  yarn fmt"
-    exit 1
-  fi
+  ## TODO: replace with a method that generates fewer false positives.
```
code-asher, line 35: Was it a yarn.lock triggering this? Maybe we are missing a `--frozen-lockfile` somewhere. :thinking: 

I think we should leave this check in because if a yarn.lock is changing that is a bug we want to catch (I assume it means CI may be using different versions than local).

The real bug here is probably the error message. This state can be caused by a lack of formatting, volatile lock files, and possibly other things (although nothing else comes to mind at the moment).

.gitignore
```diff
@@ -8,7 +8,8 @@ release-packages/
 release-gcp/
 release-images/
 node_modules
-/lib/vscode/node_modules.asar
+vendor/modules
+/node_modules.asar
```
code-asher, line 12: I think this does not exist.

.github/workflows/ci.yaml
```diff
@@ -108,42 +108,42 @@ jobs:
...
       - name: Install dependencies
-        if: steps.cache-yarn.outputs.cache-hit != 'true'
+        # if: steps.cache-yarn.outputs.cache-hit != 'true'
         run: yarn --frozen-lockfile
 
       - name: Build code-server
         run: yarn build
 
-      # Parse the hash of the latest commit inside lib/vscode
+      # Parse the hash of the latest commit inside vendor/modules/code-oss-dev
```
code-asher, line 127: Thoughts on keeping it in `lib`? (So `lib/modules`). Reason being we download other dependencies there (link agent and node).

jsjoeio:
I'm using Node v14.17.0 if that helps. I wish I knew how to check which Node version corresponded to version 83 vs 89

jsjoeio:
I'm going to delete the code-server repo, reclone and try again

jsjoeio:
Still no luck. I'm going to blame my Node version. What version are you using @TeffenEllis ?

image

ci/lib.sh
```diff
@@ -104,21 +104,3 @@ export OS
...
 RELEASE_PATH="${RELEASE_PATH-release}"
-
-# VS Code bundles some modules into an asar which is an archive format that
-# works like tar. It then seems to get unpacked into node_modules.asar.
-#
-# I don't know why they do this but all the dependencies they bundle already
-# exist in node_modules so just symlink it. We have to do this since not only VS
-# Code itself but also extensions will look specifically in this directory for
-# files (like the ripgrep binary or the oniguruma wasm).
-symlink_asar() {
```
code-asher, line 115: We likely need to keep these symlinks. When I built a release there was no `vendor/modules/code-oss-dev/node_modules.asar` directory which some extensions rely on.

To reproduce, install the Bracket Pair Colorizer 2 extension. You will see an error in the browser console and the brackets will not be colored.

I think VS Code does something where they duplicate some modules into this directory so maybe eventually use their build process instead although it seems like a waste of space. :man_astronaut:

ci/dev/postinstall.sh
```diff
@@ -3,17 +3,19 @@ set -euo pipefail
...
   cd ..
 
-  cd lib/vscode
-  yarn ${CI+--frozen-lockfile}
+  cd vendor
+  echo 'Installing vendor dependencies...'
 
-  symlink_asar
+  # Ignore scripts to prevent partial install which omits development dependencies.
+  yarn install --modules-folder modules --ignore-scripts
```
code-asher, line 17: Ah, this is probably where we are missing that `--frozen-lockfile`.

I am not clear on why we need to ignore scripts; how do they cause a partial install?

.github/workflows/ci.yaml
```diff
@@ -108,42 +108,42 @@ jobs:
...
 
-      - name: Fetch dependencies from cache
-        id: cache-yarn
-        uses: actions/cache@v2
-        with:
-          path: "**/node_modules"
-          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            yarn-build-
+      # - name: Fetch dependencies from cache
```
code-asher, line 111: Hmm strange this seems like it should work. :thinking: Maybe there was a change to the postinstall but not to the yarn.lock so the updated install was not ran? If so maybe we can fix this by also adding the postinstall scripts to the key hash.

It would be nice if we could leave all deps in the root instead of a separate vender directory but I imagine that could wreak havoc on certain parts of the build process.

code-asher (changes requested):
I have not looked at the fork at all but the changes here look great! Everything seems to run as expected.

Nice work!

code-asher:
With regards to the Node module version errors, that might be because the `code-oss-dev` module we are pulling in has `.yarnrc` files? They tell Node to target Electron 13.1.8 instead of whatever version of Node you are running.

jsjoeio:
```text
       v14.15.0   (LTS: Fermium)
       v14.15.1   (LTS: Fermium)
       v14.15.2   (LTS: Fermium) (83?)
       v14.15.3   (LTS: Fermium)
       v14.15.4   (LTS: Fermium)
       v14.15.5   (LTS: Fermium)
       v14.16.0   (LTS: Fermium)
       v14.16.1   (LTS: Fermium)
->     v14.17.0   (LTS: Fermium) (89)
       v14.17.1   (LTS: Fermium)
       v14.17.2   (LTS: Fermium)
       v14.17.3   (LTS: Fermium)
       v14.17.4   (LTS: Fermium)
       v14.17.5   (LTS: Fermium)
       v14.17.6   (Latest LTS: Fermium)
```

jsjoeio:
## Video

✅ Besides Asher's requested changes, everything looks good on my side

https://user-images.githubusercontent.com/3806031/132926155-b4d1a513-fc6d-473c-9ad7-0bf7a8799bc0.mov

jsjoeio:
> codecov/patch Failing after 1s — 33.33% of diff hit (target 64.12%)

I thought I disabled this in a previous PR but apparently I didn't 😅 Here is PR to fix that.

Feel free to rebase after that gets merged in.

https://github.com/cdr/code-server/pull/4159

code-asher (approved):
Fantastic!!

ci/dev/postinstall.sh
```diff
@@ -3,17 +3,28 @@ set -euo pipefail
...
+
+  # * We install in 'modules' instead of 'node_modules' because VS Code's extensions
+  # use a webpack config which cannot differentiate between its own node_modules
+  # and itself being in a directory with the same name.
+  #
+  # * We ignore scripts because NPM/Yarn's default behavior is to assume that
+  # devDependencies are not needed, and that even git repo based packages are
+  # assumed to be compiled. Because the default behavior for VS Code's `postinstall`
+  # assumes we're also compiled, this needs to be ignored.
+  yarn install --modules-folder modules --ignore-scripts --frozen-lockfile
```
jsjoeio, line 24: 🎉👏 I love comments like this!

jsjoeio (approved):
Woohoo!!!

PR #4136, opened 2021-09-10
Remove subtree

TeffenEllis:
**This PR should be merged after #4135, which migrates the app to a vendor package.**


jsjoeio:
> This PR should be merged after #4135, which migrates the app to a vendor package.

What if we merge this into `vscode-1.60.0-yarn` (#4135's branch) instead of into `main`? That way, we don't accidentally merge this and those two land in `main` together since they depend on each other. Just a thought

TeffenEllis:
> > This PR should be merged after #4135, which migrates the app to a vendor package.
> 
> What if we merge this into `vscode-1.60.0-yarn` (#4135's branch) instead of into `main`? That way, we don't accidentally merge this and those two land in `main` together since they depend on each other. Just a thought

I believe if we merge in that direction then the diff `vscode-1.60.0-yarn` will be unreadable as it’ll contain its files as well as the many deletions present here.

jsjoeio:
> > > This PR should be merged after #4135, which migrates the app to a vendor package.
> > 
> > 
> > What if we merge this into `vscode-1.60.0-yarn` (#4135's branch) instead of into `main`? That way, we don't accidentally merge this and those two land in `main` together since they depend on each other. Just a thought
> 
> I believe if we merge in that direction then the diff `vscode-1.60.0-yarn` will be unreadable as it’ll contain its files as well as the many deletions present here.

Ahh...didn't consider that. Okay, sounds like you have the best plan in mind already! Ignore me :)

PR #4159, opened 2021-09-13
fix(codecov): use status > patch correctly

jsjoeio:
I thought I fixed this in a previous PR but I guess not. I can't seem to find it. Turns out, our codecov was technically invalid because we weren't using `patch: off` under `status`.

To prevent this in the future, validate the codecov:

```shell
cd .github && curl --data-binary @codecov.yml https://codecov.io/validate
```

Fixes N/A

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -6,7 +6,8 @@ coverage:
   precision: 2
   round: down
   range: "40...70"
-  patch: off
+  status:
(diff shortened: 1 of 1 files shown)


bryphe-coder (approved):
Looks reasonable to me! 👍

jawnsy (approved):

PR #4160, opened 2021-09-13
feat: add tests for src/node/app.ts

### LINKED ISSUE #4094: Add test coverage for src/node/app.ts

jsjoeio:
This PR adds more tests for `src/node/app.ts` to hit 100% coverage for this file.

image

Fixes #4094

## Notes

<details>
  <summary>See more</summary>

What is this file? 
`src/node/app.ts`

What does it do? 
It exports two functions: `createApp` and `ensureAddress`

What are these functions used for?
`createApp` creates the main Express app used inside code-server which lets communication happen between the code-server client (in the browser) and the server running code-server on either local or remote machine.

`ensureAddress` ensures that the server has an address and returns it, including the protocol.

What code is untested?

36-40
```ts
if (!resolved) {
  reject(err)
} else {
  // Promise resolved earlier so this is an unrelated error.
  util.logError(logger, "http server error", err)
}
```

45-52
```ts
try {
  await fs.unlink(args.socket)
} catch (error) {
  if (error.code !== "ENOENT") {
    logger.error(error.message)
  }
}
server.listen(args.socket, resolve)
```

### Testing the first block
Let's take a look at 36-40 and understand when this code runs.

This is part of an event listener defined on `server` (which is the `httpolyglot` or `http` server) runs our callback function for the `error` event.

It says, "when the 'error' event is emitted, run this callback."

"This call back is an if/else statement which checks if `resolved` is not true, then reject the outer Promise with the `err` object that came with the 'error' event. Otherwise, the outer Promise resolved earlier so this is an unrelated error. Therefore, use our `util` and call `logError` with our logger and the message 'http server error' along with the `err` object"

The first thing we need to do is set up a test for this. Then we can throw some `console.log`s in there to ensure things are working. After that, we'll need to figure out how to get the server to emit the `error` event.

Woo! I have coverage for 36, 38-40 so far. I need to figure out how to test that `reject` line.

Let's take a look at those this code block as a whole and add comments

```ts
  let resolved = false
  await new Promise<void>(async (resolve2, reject) => {
    const resolve = () => {
      resolved = true
      resolve2()
    }
    server.on("error", (err) => {
      if (!resolved) {
        reject(err) // this line is ONLY called if resolved is false
      } else {
        // Promise resolved earlier so this is an unrelated error.
        util.logError(logger, "http server error", err)
      }
    })

    if (args.socket) {
      try {
        await fs.unlink(args.socket)
      } catch (error) {
        if (error.code !== "ENOENT") {
          logger.error(error.message)
        }
      }
      // resolve is passed here to listen which is used as the listeningListener 
      server.listen(args.socket, resolve)
    } else {
      // resolve is passed here to listen which is used as the listeningListener
      server.listen(args.port, args.host.replace(/^\[|\]$/g, ""), resolve)
    }
  })
```

From what I can tell, it seems like it's absolutely impossible for that `reject` line to be called unless we close and then listen again on a new port? we can try it.

No no, nevermind...`resolved` never gets reset so even if I close the server and listen again, that value is still `true`. I don't think that line can be hit.

STOPPED HERE 

My thinking is this:

If it can't be hit, I'm thinking of refactoring to remove it.
If it can be hit, I think the only way to get testing coverage is to extract some of that logic into a function so I can test resolved being true and false

Woohoo! We did it.

</details>

### DIFF
--- a/src/node/util.ts
+++ b/src/node/util.ts
@@ -524,3 +524,12 @@ export function escapeHtml(unsafe: string): string {
     .replace(/"/g, "&quot;")
     .replace(/'/g, "&apos;")
 }
+
+/**
+ * A helper function which returns a boolean indicating whether
+ * the given error is a NodeJS.ErrnoException by checking if
+ * it has a .code property.
+ */
+export function isNodeJSErrnoException(error: unknown): error is NodeJS.ErrnoException {
+  return error instanceof Error && (error as NodeJS.ErrnoException).code !== undefined
+}
--- a/test/unit/node/__snapshots__/app.test.ts.snap
+++ b/test/unit/node/__snapshots__/app.test.ts.snap
@@ -1,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`handleServerError should log an error if resolved is true 1`] = `"Cannot read property 'handle' of undefined"`;
--- a/test/unit/node/app.test.ts
+++ b/test/unit/node/app.test.ts
@@ -1,6 +1,168 @@
+import { logger } from "@coder/logger"
+import { promises, rmdirSync } from "fs"
 import * as http from "http"
-import { ensureAddress } from "../../../src/node/app"
-import { getAvailablePort } from "../../utils/helpers"
+import * as https from "https"
+import * as path from "path"
+import { createApp, ensureAddress, handleArgsSocketCatchError, handleServerError } from "../../../src/node/app"
+import { OptionalString, setDefaults } from "../../../src/node/cli"
+import { generateCertificate } from "../../../src/node/util"
+import { getAvailablePort, tmpdir } from "../../utils/helpers"
+
+describe("createApp", () => {
+  let spy: jest.SpyInstance
+  let unlinkSpy: jest.SpyInstance
+  let port: number
+  let tmpDirPath: string
+  let tmpFilePath: string
+
+  beforeAll(async () => {
+    tmpDirPath = await tmpdir("unlink-socket")
+    tmpFilePath = path.join(tmpDirPath, "unlink-socket-file")
+  })
+
+  beforeEach(async () => {
+    spy = jest.spyOn(logger, "error")
+    // NOTE:@jsjoeio
+    // Be mindful when spying.
+    // You can't spy on fs functions if you do import * as fs
+    // You have to import individually, like we do here with promises
+    // then you can spy on those modules methods, like unlink.
+    // See: https://github.com/aelbore/esbuild-jest/issues/26#issuecomment-893763840
+    unlinkSpy = jest.spyOn(promises, "unlink")
+    port = await getAvailablePort()
+  })
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterAll(() => {
+    jest.restoreAllMocks()
+    // Ensure directory was removed
+    rmdirSync(tmpDirPath, { recursive: true })
+  })
+
+  it("should return an Express app, a WebSockets Express app and an http server", async () => {
+    const defaultArgs = await setDefaults({
+      port,
+      _: [],
+    })
+    const [app, wsApp, server] = await createApp(defaultArgs)
+
+    // This doesn't check much, but it's a good sanity check
+    // to ensure we actually get back values from createApp
+    expect(app).not.toBeNull()
+    expect(wsApp).not.toBeNull()
+    expect(server).toBeInstanceOf(http.Server)
+
+    // Cleanup
+    server.close()
+  })
+
+  it("should handle error events on the server", async () => {
+    const defaultArgs = await setDefaults({
+      port,
+      _: [],
+    })
+
+    // This looks funky, but that's because createApp
+    // returns an array like [app, wsApp, server]
+    // We only need server which is at index 2
+    // we do it this way so ESLint is happy that we're
+    // have no declared variables not being used
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    const testError = new Error("Test error")
+    // We can easily test how the server handles errors
+    // By emitting an error event
+    // Ref: https://stackoverflow.com/a/33872506/3015595
+    server.emit("error", testError)
+    expect(spy).toHaveBeenCalledTimes(1)
+    expect(spy).toHaveBeenCalledWith(`http server error: ${testError.message} ${testError.stack}`)
+
+    // Cleanup
+    server.close()
+  })
+
+  it("should reject errors that happen before the server can listen", async () => {
+    // We listen on an invalid port
+    // causing the app to reject the Promise called at startup
+    const port = 2
+    const defaultArgs = await setDefaults({
+      port,
+      _: [],
+    })
+
+    async function masterBall() {
+      const app = await createApp(defaultArgs)
+      const server = app[2]
+
+      const testError = new Error("Test error")
+
+      server.emit("error", testError)
+
+      // Cleanup
+      server.close()
+    }
+
+    expect(() => masterBall()).rejects.toThrow(`listen EACCES: permission denied 127.0.0.1:${port}`)
+  })
+
+  it("should unlink a socket before listening on the socket", async () => {
+    await promises.writeFile(tmpFilePath, "")
+    const defaultArgs = await setDefaults({
+      _: [],
+      socket: tmpFilePath,
+    })
+
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    expect(unlinkSpy).toHaveBeenCalledTimes(1)
+    server.close()
+  })
+  it("should catch errors thrown when unlinking a socket", async () => {
+    const tmpDir2 = await tmpdir("unlink-socket-error")
+    const tmpFile = path.join(tmpDir2, "unlink-socket-file")
+    // await promises.writeFile(tmpFile, "")
+    const socketPath = tmpFile
+    const defaultArgs = await setDefaults({
+      _: [],
+      socket: socketPath,
+    })
+
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    expect(spy).toHaveBeenCalledTimes(1)
+    expect(spy).toHaveBeenCalledWith(`ENOENT: no such file or directory, unlink '${socketPath}'`)
+
+    server.close()
+    // Ensure directory was removed
+    rmdirSync(tmpDir2, { recursive: true })
+  })
+
+  it("should create an https server if args.cert exists", async () => {
+    const testCertificate = await generateCertificate("localhost")
+    const cert = new OptionalString(testCertificate.cert)
+    const defaultArgs = await setDefaults({
+      port,
+      cert,
+      _: [],
+      ["cert-key"]: testCertificate.certKey,
+    })
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    // This doesn't check much, but it's a good sanity check
+    // to ensure we actually get an https.Server
+    expect(server).toBeInstanceOf(https.Server)
+
+    // Cleanup
+    server.close()
+  })
+})
 
 describe("ensureAddress", () => {
   let mockServer: http.Server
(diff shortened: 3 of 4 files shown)


src/node/util.ts
```diff
@@ -524,3 +524,10 @@ export function escapeHtml(unsafe: string): string {
     .replace(/"/g, "&quot;")
     .replace(/'/g, "&apos;")
 }
+
+/**
+ *
+ */
```
jsjoeio, line 530: Add comments explaining what this function is and why we have it

test/unit/node/__snapshots__/app.test.ts.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`handleServerError should log an error if resolved is true 1`] = `"Cannot read property 'handle' of undefined"`;
```
jsjoeio, line 3: It's been a while since I've used snapshots in Jest. Is this right?

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+import { OptionalString, setDefaults } from "../../../src/node/cli"
+import { getAvailablePort, tmpdir } from "../../utils/helpers"
+
+describe("createApp", () => {
+  let spy: jest.SpyInstance
+  let unlinkSpy: jest.SpyInstance
+
+  beforeEach(() => {
+    // https://github.com/aelbore/esbuild-jest/issues/26#issuecomment-893763840
+    // explain why we do it this way
```
jsjoeio, line 16: Add comment/docs here.

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterAll(() => {
+    jest.restoreAllMocks()
+  })
+
+  // TODO@jsjoeio refactor to use beforeEach and afterEach
```
jsjoeio, line 29: remove

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+    const defaultArgs = await setDefaults({
+      port,
+      _: [],
+    })
+
+    // This looks funky, but that's because createApp
+    // returns an array like [app, wsApp, server]
+    // We only need server which is at index 2
+    // we do it this way so ESLint is happy that we're
+    // have no declared variables not being used
```
jsjoeio, line 59: We do this multiple times so maybe remove?

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    const testError = new Error("Test error")
+    // Emitting error events on servers
+    // https://stackoverflow.com/a/33872506/3015595
```
jsjoeio, line 65: Elaborate on this comment

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+    server.emit("error", testError)
+    expect(spy).toHaveBeenCalledTimes(1)
+    expect(spy).toHaveBeenCalledWith(`http server error: ${testError.message} ${testError.stack}`)
+
+    // Cleanup
+    server.close()
+  })
+
+  it("should reject errors", async () => {
```
jsjoeio, line 74: Elaborate exactly what this is testing

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+    async function masterBall() {
+      const app = await createApp(defaultArgs)
+      const server = app[2]
+
+      const testError = new Error("Test error")
+      // Emitting error events on servers
+      // https://stackoverflow.com/a/33872506/3015595
+      server.emit("error", testError)
+      // expect(spy).toHaveBeenCalledTimes(1)
+      // expect(spy).toHaveBeenCalledWith(`http server error: ${testError.message} ${testError.stack}`)
```
jsjoeio, line 95: Remove

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+      // Cleanup
+      server.close()
+    }
+
+    expect(() => masterBall()).rejects.toThrow(`listen EACCES: permission denied 127.0.0.1:${port}`)
+  })
+
+  it("should unlink a socket before listening on the socket", async () => {
+    const tmpDir = await tmpdir("unlink-socket")
+    const tmpFile = path.join(tmpDir, "unlink-socket-file")
```
jsjoeio, line 106: Consider moving to beforeEach

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+      socket: socketPath,
+    })
+
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    expect(unlinkSpy).toHaveBeenCalledTimes(1)
+    // Ensure directory was removed
+    rmdirSync(socketPath, { recursive: true })
+    server.close()
```
jsjoeio, line 120: Consider moving to beforeEach

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+    expect(spy).toHaveBeenCalledTimes(1)
+    expect(spy).toHaveBeenCalledWith(`ENOENT: no such file or directory, unlink '${socketPath}'`)
+    // Ensure directory was removed
+    rmdirSync(socketPath, { recursive: true })
+    server.close()
+  })
+
+  it("should create an https server if args.cert exists", async () => {
+    // No idea why we have this weird optional string thing
+    const port = await getAvailablePort()
```
jsjoeio, line 144: Move to beforeEach

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,165 @@
...
+      cert,
+      _: [],
+      ["cert-key"]: "./test/utils/test.key",
+    })
+    const app = await createApp(defaultArgs)
+    const server = app[2]
+
+    // This doesn't check much, but it's a good sanity check
+    // to ensure we actually get back values from createApp
+    expect(server).toBeInstanceOf(https.Server)
```
jsjoeio, line 157: Clean up

test/unit/node/app.test.ts
```diff
@@ -28,3 +187,121 @@ describe("ensureAddress", () => {
...
+// no one else can touch it besides root
+
+// make the socket on the file path and i'll get that..
+
+// create a directory and pass that in as the socket
+// with one file and use the directory as the socket path
+
+// The other thing I can do is mock fs.unlink
+// and make it throw an error
+// Stopped
```
jsjoeio, line 241: Remove

test/unit/node/app.test.ts
```diff
@@ -28,3 +187,121 @@ describe("ensureAddress", () => {
...
+
+  afterEach(() => {
+    jest.clearAllMocks()
+  })
+
+  afterAll(() => {
+    jest.restoreAllMocks()
+  })
+
+  it("should log an error if its not an isNodeJSErrnoException", () => {
```
jsjoeio, line 258: add "when" like Grey does

test/utils/test.crt
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
+MIIFjjCCA3YCCQCVV/TwiFnSozANBgkqhkiG9w0BAQsFADCBiDELMAkGA1UEBhMC
```
jsjoeio, line 2: Is there a better way to do this? Maybe instead I can add this to a string and write this to a file in the test?

That way we don't bloat code-server with new files

test/utils/test.key
```diff
@@ -0,0 +1,52 @@
+-----BEGIN PRIVATE KEY-----
```
jsjoeio, line 1: same as above

jsjoeio:
@BrunoQuaresma any idea why this is failing?

```shell
Run yarn ts-node ./product/coder.com/site/scripts/checkDocs.ts
yarn run v1.22.11
$ /home/runner/work/code-server/code-server/node_modules/.bin/ts-node ./product/coder.com/site/scripts/checkDocs.ts
info - Checking Coder docs...
info - Checking code-server docs...
error - https://codercom-2tynkpczs-codercom.vercel.app/docs/code-server/v3.11.1/collaboration returned an error
error - Response code 404 (Not Found)
Error: Process completed with exit code 1.
```

[This page](https://coder.com/docs/code-server/v3.11.1/collaboration) does exist 🤔

BrunoQuaresma:
> @BrunoQuaresma any idea why this is failing?
> 
> ```shell
> Run yarn ts-node ./product/coder.com/site/scripts/checkDocs.ts
> yarn run v1.22.11
> $ /home/runner/work/code-server/code-server/node_modules/.bin/ts-node ./product/coder.com/site/scripts/checkDocs.ts
> info - Checking Coder docs...
> info - Checking code-server docs...
> error - https://codercom-2tynkpczs-codercom.vercel.app/docs/code-server/v3.11.1/collaboration returned an error
> error - Response code 404 (Not Found)
> error Command failed with exit code 1.
> info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
> Error: Process completed with exit code 1.
> ```
> 
> [This page](https://coder.com/docs/code-server/v3.11.1/collaboration) does exist 🤔

Looks like any of the previous commits fixed this.

jsjoeio:
> Looks like a commit already fixed this.

Or you showed up and now it's fixed 😎 Thanks @BrunoQuaresma !

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,262 @@
...
 import * as http from "http"
-import { ensureAddress } from "../../../src/node/app"
-import { getAvailablePort } from "../../utils/helpers"
+import * as https from "https"
+import * as path from "path"
+import { createApp, ensureAddress, handleArgsSocketCatchError, handleServerError } from "../../../src/node/app"
+import { OptionalString, setDefaults } from "../../../src/node/cli"
+import { getAvailablePort, tmpdir } from "../../utils/helpers"
+
+// wip: todo@jsjoeio fix
```
jsjoeio, line 10: nit: remove

test/unit/node/app.test.ts
```diff
@@ -1,6 +1,262 @@
...
+H+x8Qx9a+eyNq/miV9czK++miaLyGQRkSDE35iFWUKpa1C1BcKl53uKfYCo6u83a
+nAWr/CATnyFgVZzMEJloSvYfnuA2+klJtZ9cRV0lp4CxefR+AKnkcNtmCr71LX12
+b7XTXHw68gEQp/iVZc0cP23gaDs26KIduLez2moF+iTjqn4J77epV6vGjuj27WL3
+hK8Z7KUFqUU6DSVJ2Z/qGBw/ZP78l//o2Jbu4KFfg5UviUzrLXWfrMJw9uqgfD4c
+OiahmBsqB/9JHGFGOhQy1/F8k/7Y3k1o1hdYwNCmB7iwPJ9eklhYHCrQc/yIarZz
+Z4rwv90aq+cFtoVvrTPA0rjSu4jwo2tI/nYfdZOfcCXhcw==
+-----END CERTIFICATE-----
+`
+
+const FAKE_KEY = `-----BEGIN PRIVATE KEY-----
```
jawnsy, line 45: FYI these sorts of keys might cause problems with security scan tools. Might want to generate a small (e.g. 1024-bit) key every time the tests start

jawnsy (approved):

test/unit/node/app.test.ts
jsjoeio: Ahh! Nice catch. 

> Might want to generate a small (e.g. 1024-bit) key every time the tests start

I wonder how easy that is 🤔 I think when I made these locally, it ran me through an interactive prompt. I'll look into this.

jawnsy: Try this instead of using OpenSSL directly to generate keys: https://www.npmjs.com/package/pem -- with very weak key settings, generating keys is very fast. We do something similar with our monorepo tests, for the same reason (to avoid having hardcoded keys flagged as a potential security issue)

jawnsy: I also came across this: https://nodejs.org/api/tls.html#tls_tls_createsecurepair_context_isserver_requestcert_rejectunauthorized_options but I don't know Node.js well enough to know if this is helpful. You can also just merge this as-is and see if scanners pick it up or not, I don't think it's a big deal since this is just test code. It might be that Node.js doesn't provide any straightforward way to create x509 certificates for TLS

jsjoeio: Well, I'll at least take a look at both of those and see. Dude, thank you so much!

jawnsy: Oh, looks like Node.js 15 introduced a new API; see: https://nodejs.org/api/crypto.html#crypto_new_x509certificate_buffer and https://www.geeksforgeeks.org/node-js-new-x509certificate-function/

jsjoeio: Turns out we had `pem` already and even a `generateCertificate` function 😅 Ended up using that. (Would have never thought to look without your suggestion so thank you ♥️)

https://github.com/cdr/code-server/blob/main/src/node/util.ts#L103-L146

PR #4175, opened 2021-09-15
fix: comment out broken e2e test

jsjoeio:
The main purpose of this PR is to get the assets from CI so we can update code-server on Docker.

I've run e2e tests twice and they keep failing. Maybe we removed the Log out button when we updated VS Code? cc @TeffenEllis 

I'm going to disable those tests for now.

```shell
  ✓ terminal.test.ts:26:3 › [WebKit] Integrated Terminal should echo a string to a file (18s)


  1) logout.test.ts:4:3 › [Firefox] logout should be able logout ===================================

    page.click: Timeout 30000ms exceeded.
    =========================== logs ===========================
    waiting for selector "[aria-label="Application Menu"]"
    ============================================================
    Note: use DEBUG=pw:api environment variable to capture Playwright logs.

       5 |     // Recommended by Playwright for async navigation
       6 |     // https://github.com/microsoft/playwright/issues/1987#issuecomment-620182151
    >  7 |     await Promise.all([codeServerPage.page.waitForNavigation(), codeServerPage.navigateMenus(["Log Out"])])
```

### DIFF
--- a/test/e2e/logout.test.ts
+++ b/test/e2e/logout.test.ts
@@ -1,11 +1,12 @@
-import { describe, test, expect } from "./baseFixture"
+// NOTE@jsjoeio commenting out until we can figure out what's wrong
+// import { describe, test, expect } from "./baseFixture"
 
-describe("logout", true, () => {
-  test("should be able logout", async ({ codeServerPage }) => {
-    // Recommended by Playwright for async navigation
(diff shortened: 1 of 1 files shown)


jawnsy (approved):

TeffenEllis:
@jsjoeio I believe upstream made some changes to how they create menus, and it’s possible that our test isn’t accurately targeting the menu entry. Could we move the test to vscode’s suite?

PR #4194, opened 2021-09-16
feat: add test for get_nfpm_arch()

jsjoeio:
This adds tests for build-packages.sh, specifically 

This PR introduces a new file called `build-lib.sh` which extracts the `get_nfpm_arch()` function out of `build-packages.sh` so that it can be unit tested. I did this because if you `source build-packages.sh` it runs the `main` function. 

With this new approach, we can move individual functions related to the build process into `build-lib.sh` and unit test them. 

The inspiration for this came from some recent rockiness with the 3.12.0 release and our interest in stabilizing the build and release pipeline.

image

Fixes N/A

_Note: I would love if we could fully test all our scripts used in the build and release process. I also tossed around the idea of rewriting these scripts in JS/TS but that's a discussion for the future 😉_

### DIFF
--- a/ci/build/build-lib.sh
+++ b/ci/build/build-lib.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+# This is a library which contains functions used inside ci/build
+#
+# We separated it into it's own file so that we could easily unit test
+# these functions and helpers.
+
+# On some CPU architectures (notably node/uname "armv7l", default on Raspberry Pis),
+# different package managers have different labels for the same CPU (deb=armhf, rpm=armhfp).
+# This function returns the overriden arch on platforms
+# with alternate labels, or the same arch otherwise.
+get_nfpm_arch() {
+  local PKG_FORMAT="${1:-}"
+  local ARCH="${2:-}"
+
+  case "$ARCH" in
+    armv7l)
+      if [ "$PKG_FORMAT" = "deb" ]; then
+        echo armhf
+      elif [ "$PKG_FORMAT" = "rpm" ]; then
+        echo armhfp
+      fi
+      ;;
+    *)
+      echo "$ARCH"
+      ;;
+  esac
+}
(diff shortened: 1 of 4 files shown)


jsjoeio:
I don't know why this is passing locally but failing in CI 🤔 

image

jsjoeio:
We think it's because the runner doesn't have `jq` installed. Going to refactor this function to a case statement instead

ci/build/build-lib.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
```
jawnsy, line 2: suggest using `set -euo pipefail` here, this also means `$ARCH` needs to be `${ARCH:-}` etc

ci/build/build-lib.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# We separated it into it's own file so that we could easily unit test
+# these functions and helpers.
+
+# On some CPU architectures (notably node/uname "armv7l", default on Raspberry Pis),
+# different package managers have different labels for the same CPU (deb=armhf, rpm=armhfp).
+# This function returns the overriden arch on platforms
+# with alternate labels, or the same arch otherwise.
+get_nfpm_arch() {
+  case "$ARCH" in
```
jawnsy, line 13: Alternatively you can pass the parameters as arguments, e.g. `get_nfpm_arch armv7l deb` and then use $1 and $2 instead of $ARCH and $PKG_FORMAT. This way you can call the function without modifying a global env var state. This is what this does: https://github.com/cdr/m/blob/f0de9bff47859bfe8965474b62a2cd9e900863d9/lib/bash/trace.sh#L23-L26

jawnsy (approved):

ci/build/build-lib.sh
jsjoeio: AHHH! I had that and it kept failing and I didn't know why. Thanks for the missing piece of the puzzle 🙌

ci/build/build-lib.sh
jsjoeio: OOO good idea. I'm all about not relying on global env vars. I'll refactor! Thanks for the suggestion!

PR #4211, opened 2021-09-20
feat: add tests for shouldRunVsCodeCli and bindAddrFromArgs

### LINKED ISSUE #4210: Add tests for shouldRunVsCodeCli and bindAddrFromArgs
These functions live in `src/node/cli.ts`:
- `shouldRunVsCodeCli`
- `bindAddrFromArgs`

jsjoeio:
This PR adds more tests for `src/node/cli.ts`. Specifically, these functions: `shouldRunVsCodeCli` and `bindAddrFromArgs`

Fixes #4210

TODOS
- [x] `shouldRunVsCodeCli`
- [x] `bindAddrFromArgs`
- [x] fix broken tests? (it seems one test is broken locally for me, probably a local issue)

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -626,7 +626,20 @@ function bindAddrFromAllSources(...argsConfig: Args[]): Addr {
 }
 
 export const shouldRunVsCodeCli = (args: Args): boolean => {
-  return !!args["list-extensions"] || !!args["install-extension"] || !!args["uninstall-extension"]
+  // Create new interface with only these keys
+  // Pick<Args, "list-extensions" | "install-extension" | "uninstall-extension">
+  // Get the keys of new interface
+  // keyof ...
+  // Turn that into an array
+  // Array<...>
+  type ExtensionArgs = Array<keyof Pick<Args, "list-extensions" | "install-extension" | "uninstall-extension">>
+  const extensionRelatedArgs: ExtensionArgs = ["list-extensions", "install-extension", "uninstall-extension"]
+
+  const argKeys = Object.keys(args)
+
+  // If any of the extensionRelatedArgs are included in args
+  // then we don't want to run the vscode cli
+  return extensionRelatedArgs.some((arg) => argKeys.includes(arg))
 }
 
 /**
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -594,7 +594,11 @@ interface Addr {
   port: number
 }
 
-function bindAddrFromArgs(addr: Addr, args: Args): Addr {
+/**
+ * This function creates the bind address
+ * using the CLI args.
+ */
+export function bindAddrFromArgs(addr: Addr, args: Args): Addr {
   addr = { ...addr }
   if (args["bind-addr"]) {
     addr = parseBindAddr(args["bind-addr"])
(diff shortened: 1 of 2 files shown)


jsjoeio:
I'm a bit dumbfounded. The tests pass in CI: https://github.com/cdr/code-server/pull/4211/checks?check_run_id=3657430012

But 1 test is not passing locally 🤔 It's the `should catch errors thrown when unlinking a socket` test.

```ts
 it("should catch errors thrown when unlinking a socket", async () => {
    const tmpDir2 = await tmpdir("unlink-socket-error")
    const tmpFile = path.join(tmpDir2, "unlink-socket-file")
    // await promises.writeFile(tmpFile, "")
    const socketPath = tmpFile
    const defaultArgs = await setDefaults({
      _: [],
      socket: socketPath,
    })

    const app = await createApp(defaultArgs)
    const server = app[2]

    expect(spy).toHaveBeenCalledTimes(1)
    expect(spy).toHaveBeenCalledWith(`ENOENT: no such file or directory, unlink '${socketPath}'`)

    server.close()
    // Ensure directory was removed
    rmdirSync(tmpDir2, { recursive: true })
  })
```

Basically, what's happening is:
- I pass in a path to a file that doesn't exist for `socket`
- It tries to `unlink` the file
- Then it should call the `spy` (which is our logger), to log the error

BUT it does all this but then gets to `server.listen` and instead throws and error saying "address already in use:

See below: 
```ts
      try {
        await fs.unlink(args.socket)
      } catch (error: any) {
        handleArgsSocketCatchError(error)
      }
      server.listen(args.socket, resolve
```

image

image

jsjoeio:
help @code-asher - any ideas?

src/node/cli.ts
```diff
@@ -594,7 +594,11 @@ interface Addr {
   port: number
 }
 
-function bindAddrFromArgs(addr: Addr, args: Args): Addr {
+/**
+ * This function creates the bind address
+ * using the CLI args.
+ */
+export function bindAddrFromArgs(addr: Addr, args: Args): Addr {
```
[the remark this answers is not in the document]
bryphe-coder: Cool!

src/node/cli.ts
```diff
@@ -626,7 +630,20 @@ function bindAddrFromAllSources(...argsConfig: Args[]): Addr {
...
 
 export const shouldRunVsCodeCli = (args: Args): boolean => {
-  return !!args["list-extensions"] || !!args["install-extension"] || !!args["uninstall-extension"]
+  // Create new interface with only these keys
+  // Pick<Args, "list-extensions" | "install-extension" | "uninstall-extension">
+  // Get the keys of new interface
+  // keyof ...
+  // Turn that into an array
+  // Array<...>
+  type ExtensionArgs = Array<keyof Pick<Args, "list-extensions" | "install-extension" | "uninstall-extension">>
```
bryphe-coder, line 639: I like the overall strategy to make this type-safe and get the benefit from the compiler! The only thing I find a bit awkward is that we have to reference the `"list-extensions" | "install-extensions" | "uninstall-extensions"` twice here.

I wonder if we could get most of the benefits, without having to define these multiple times, by doing:

```
type ArgumentKeys = Array<keyof Args>;
const extensionRelatedArgs: ArgumentKeys = ["list-extensions", "install-extension", "uninstall-extension"];
```

That way - the compiler will still flag an error if these get out-of-sync, but you only have to define them once here in `shouldRunVsCodeCli`

test/unit/node/cli.test.ts
```diff
@@ -463,3 +472,173 @@ describe("splitOnFirstEquals", () => {
...
+    const expected = {
+      host: "localhost",
+      port: 3000,
+    }
+
+    expect(actual).toStrictEqual(expected)
+  })
+
+  it("should use the args.port over process.env.PORT if both set", () => {
+    const [setValue, resetValue] = useEnv("PORT")
```
bryphe-coder, line 622: Nice set of tests 👍

bryphe-coder (approved):
Looks great to me - nice work @jsjoeio ! I like your approach to leveraging the TS type system as much as possible.

 Just had minor feedback about streamlining having to define the `"list-extensions" | "install-extension" | "uninstall-extension"`. Up to you if you want to incorporate that or not 👍

test/unit/node/cli.test.ts
jsjoeio: Thanks!

src/node/cli.ts
jsjoeio: Oh this is a huge improvement. Waaaaay better. Didn't think about that. Thanks for providing this suggestion! I'll incorporate this suggestion before merging! 🙌

vapurrmaid (commented):
✔️ Nice work!

Issue #4182, opened 2021-09-16
iPad pro editor lose focus after swipping back from safari browser


fuchanglin:
## OS/Web Information

- Web Browser: Safari PWA mode
- Local OS: iPad OS 15 with magic keyboard
- Remote OS: Centos 8
- Remote Architecture: amd64
- `code-server --version`: 3.12.0

## Steps to Reproduce

1. Edit code in Safari PWA
2. Swipe out to safari browser, view some article, select some word
3. Swipe back to editor
4. Sometime, it will lose focus, after re-focus, can not type Enter, tab, up/down/left/right arrow, a-z works fine.

## Expected

When swipe back, the editor should still work

## Actual

Very annoying, in order to bring Enter back, it seems I need to 1. restart Safari browser and the PWA app 
Or 2. detach iPad from magic keyboard, type Enter using the on-screen keyboard

## Notes

This issue can be reproduced in VS Code: Yes/No

jsjoeio:
Sorry to hear this is happening! I will have to see if I can reproduce on my iPad. Thanks for your patience!

spouradrix:
@fuchanglin By the way, you can regain focus of the code-server window by clicking on the search box. Sometimes this fix does not work but it mostly does.

bpmct:
Was able to reproduce this in split view

Tonyhaenn:
I’ve noticed that if I tap into the terminal, and then into the editor window, I’m able to start editing again.
IPad Pro / ipad os 15
Code-server 3.12

fuchanglin:
> I’ve noticed that if I tap into the terminal, and then into the editor window, I’m able to start editing again. IPad Pro / ipad os 15 Code-server 3.12

Sometime it works, sometime we will lost all the focus and cannot type Enter.

ivanpenchev:
Something similar happens to me on iPad PRO with or without PWA and with an external keyboard. While I'm editing something, if I switch to another app on the iPad, when I get back to the code-server view, the caret is still visible in the editor but in fact it behaves like the editor has lost focus thus the keyboard is not working. Taps or clicks throughout the code, changes the position of the caret but the keyboard remains non-functional.

Touching anything else on the interface and back into the editor makes it regain focus and the keyboard is back to working.

thanhj:
Any update on this topic, I am still facing this issue today 2024.06.
I am running vscodeclient as a safari home app, every time i switch back from safari, I can not type anything on the vscode and have to touch/click the screen several times.

[...]

Issue #3824, opened 2021-07-20
3.11 - Terminal not properly visible on iPad


asheswook:
## OS/Web Information

- Web Browser: Safari 15
- Local OS: iPadOS 15 dev2
- Remote OS: Ubuntu 20.04.2 LTS
- Remote Architecture: Linux
- `code-server --version`: 3.11.0 4e8cd09ef0412dfc7b148b7639a692e20e4fd6dd

## Explanation
Terminal does not display normally. It only shows the color. Sometimes, when I refresh the web, it appears well. The setting values for VSCode are all default. Even if the screen doesn't appear, its process is going well. It's just that I can't see it. Also, Debug console works well.

## Screenshot
![C81484A8-6913-436A-BC60-43EBB8E6A573](https://user-images.githubusercontent.com/25760310/126360231-83cd9897-1e34-4e9e-a301-0551441bd347.jpeg)
![308860AC-AC93-4160-96E3-8A23F3370283](https://user-images.githubusercontent.com/25760310/126360248-ba1658f2-efa2-4e53-94eb-abbde0d192e1.jpeg)
![C4468FC1-75E3-4DD7-AF7B-1D92A41E9EF9](https://user-images.githubusercontent.com/25760310/126360249-82732efa-d14b-4d76-b91a-161ac0971f0f.jpeg)

jsjoeio:
Hmm I have yet to run this on my iPad so I will have to check.

> Terminal does not display normally

Hard to tell from the photo exactly. How should it look?

asheswook:
The letters printed on the terminal should be visible like this:
![20D6E72B-974D-442F-947F-34FF4E09A6FC](https://user-images.githubusercontent.com/25760310/126424687-79a0faf9-2529-4c0d-8908-6a165079cc3f.jpeg)

asheswook:
https://user-images.githubusercontent.com/25760310/126425196-5e0cbe84-3d71-4169-a42b-f76b43e7929c.mov

asheswook:
What I want is, unlike the second console, like the first console in the video, all consoles print content well.

dhguswns23:
Same here. My workaround is to change default shell profile (in my case, bash to sh) then revert that change. But it is tiresome.

jsjoeio:
Thanks for the screenshots! (the video won't play :( )

Good to know. We will look into this soon!

AjRic07:
[...]

Issue #1455, opened 2020-03-25
Can’t scroll on iPad with a trackpad.


awahlig:
On an iPad running iPadOS 13.4 (which added full trackpad/mouse support), two-finger-scrolling on a trackpad doesn’t work (nothing happens).

Tested in different panels (editor, terminal, settings).
Tested using Magic Trackpad 2 (Apple’s iPad Pro keyboard with trackpad is not yet available).
Tried both the :latest (About window shows 1.41.1) and :3.0.1 docker images.

dvelopin:
Magic Mouse also doesn't "scroll" it seems to simulate "up" and "down" from the keyboard instead. There may be behavior settings we need to adjust?

mjsamberg:
Also, the scroll wheel on a mouse scrolls the entire window instead of just the text editor area.

TheWindyMan:
On a Logitech MX Vertical, binding Scroll Up and Scroll Down in Assistive Touch settings to the forward and back buttons allows for scrolling in the terminal/code windows separately. I haven’t found a way to bind those actions to the gesture of scrolling on the Magic Trackpad 2

kylecarbs:
We're making this a priority. Thanks for the upvotes everyone, shows us it's in high demand 👀.

neilalexander:
@mjsamberg The app container scrolling should be fixed in the code-server 3.0 releases — certainly it appears to be better for me now in code-server 3.0.2. 

I can confirm though that neither the Magic Mouse 2 nor the Magic Trackpad 2 scrolling works in 3.0.2 though. Scrolling using the touchscreen does work, although it feels slow due to the fact that there is no scroll momentum.

mjsamberg:
@neilalexander App container scrolling still occurs with my logitech mouse and the scroll wheel.

cmoog:
_Another update_
Please reference this [WebKit issue](https://bugs.webkit.org/show_bug.cgi?id=210071) for progress towards a patch.

diegosucaria:
Hello, has somebody found any other workaround for this? (even if it is buggy) I mean, other than using the scrollbars.

code-server and an Ipad pro with magic keyboard (which includes trackpad) is awesome!

mjsamberg:
I did notice, FWIW, that scrolling does work as expected in Google Docs website on iPad in multiple directions and even when there are multiple scrollable elements on the screen.

perelin:
[...]

PR #4228, opened 2021-09-23
update iPad docs to reflect info in Sep 2021

bpmct:
This updates our iPad.md docs to reflect some of the latest changes I observed when using code-server on an iPad.

- Mention the scrolling issue (#1455) has been fixed
- Add the terminal text issue (#3824) under known issues
- Add lose focus issue (#4182) under known issues

I also moved the self-signed certificate instructions towards the bottom of the docs since it is not something we recommend.


jsjoeio (approved):
Awesome! Thanks for doing this Ben 🚀

PR #4236, opened 2021-09-23
fix(brew-bump.sh): add checks and handle errors 

### LINKED ISSUE #3962: Fix brew-bump.sh script
It looks like the `brew-bump.sh` is not working as expected 🤔

```logs
./ci/steps/brew-bump.sh: line 19: upstream: command not found
fatal: cannot run /Users/runner/.git-askpass.sh: No such file or directory
fatal: could not read Password for 'https://cdr-oss@github.com': Device not configured
```

https://github.com/cdr/code-server/runs/3305506315?check_suite_focus=true#step:5:5676

We'll need to fix this before the next release.

Resources:
https://github.com/mislav/bump-homebrew-formula-action

jsjoeio:
This PR adds a lot of bash helper functions (+ tests) used in `brew-bump.sh` in order to ensure this script doesn't fail when publishing a release.

image

## TODOs

- [x] add BATS test for `steps-lib.sh`
- [x] add check + test for `$VERSION`
- [x] add check for `cdr/ci/homebrew`
- [x] add check for remote upstream
- [x] add check for `$HOMEBREW_GITHUB_API_TOKEN` 
- [x] add check for `git-ask-pass.sh`
- [x] add check for cleaning up homebrew-core
- [x] fix Trivy vulnerability (https://github.com/cdr/code-server/pull/4251)
- [x] test the `GIT_ASKPASS` in isolated environment

Fixes #3962

## Test Plan

To test out the other parts of the script (specifically `GIT_ASKPASS`), here is what I did:
1. Create new Workspace on `https://master.cdr.dev/workspaces`
2. "log out" of git with:
```shell
git config --global --unset user.name
git config --global --unset user.email
git config --global --unset credential.helper
```
3.  Verify log out by trying to clone via HTTPS a private repo:
```shell
git clone https://github.com/jsjoeio/dip.chat-v2.git
```
4. Create script similar to `brew-bump.sh` and add `steps-lib.sh` to Workspace.
5. Make script executable `chmod +x test-brew.sh`
6. Run script: `VERSION="1.2.3" HOMEBREW_GITHUB_API_TOKEN="<redcated>" ./test-brew.sh` (make sure the token had `repo` and `workflow` in scope).

## Notes

Ran into issues with scripts that used `set -u` so that's why I created a new file called `steps-lib.sh`.

See: https://github.com/sstephenson/bats/issues/171

### DIFF
--- a/ci/steps/brew-bump.sh
+++ b/ci/steps/brew-bump.sh
@@ -5,6 +5,21 @@ main() {
   cd "$(dirname "$0")/../.."
   # Only sourcing this so we get access to $VERSION
   source ./ci/lib.sh
+  source ./ci/steps/steps-lib.sh
+
+  echo "Checking environment variables"
+
+  # We need VERSION to bump the brew formula
+  if [[ $(is_env_var_set "VERSION") -eq 1 ]]; then
+    echo "VERSION is not set"
+    exit 1
+  fi
+
+  # We need HOMEBREW_GITHUB_API_TOKEN to push up commits
+  if [[ $(is_env_var_set "HOMEBREW_GITHUB_API_TOKEN") -eq 1 ]]; then
+    echo "HOMEBREW_GITHUB_API_TOKEN is not set"
+    exit 1
+  fi
 
   # NOTE: we need to make sure cdrci/homebrew-core
   # is up-to-date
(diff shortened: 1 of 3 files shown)


ci/steps/brew-bump.sh
```diff
@@ -13,27 +28,65 @@ main() {
...
 
+  # Make sure the git clone step is successful
+  if [[ $(directory_exists "homebrew-core") -eq 1 ]]; then
+    echo "git clone failed. Cannot find homebrew-core directory."
+    ls -la
+    exit 1
+  fi
+
   echo "Changing into homebrew-core directory"
   cd homebrew-core && pwd
```
jawnsy, line 39: Unrelated to your change but it looks like using pushd/popd would be better here, since they print the path stack when they change things. On line 39 we're changing into homebrew-core and then on line 51 we're going back up a level

ci/steps/brew-bump.sh
```diff
@@ -5,6 +5,21 @@ main() {
   cd "$(dirname "$0")/../.."
   # Only sourcing this so we get access to $VERSION
   source ./ci/lib.sh
+  source ./ci/steps/steps-lib.sh
+
+  echo "Checking environment variables"
+
+  # We need VERSION to bump the brew formula
+  if [[ $(is_env_var_set "VERSION") -eq 1 ]]; then
```
jawnsy, line 13: If your function returns 0, it's considered true (zero exit code means success) and if it returns 1 then it's considered false. That means you can also use a pattern like this, which looks a little nicer:

```bash
#!/bin/bash

set -euo pipefail

is_true() {
  return 0
}

is_false() {
  return 1
}

if is_true; then
  echo "is_true: true"
else
  echo "is_true: false"
fi

if is_false; then
  echo "is_false: true"
else
  echo "is_false: false"
fi
```

This program prints:

```shell-session
$ bash test.sh
is_true: true
is_false: false
```

ci/steps/brew-bump.sh
```diff
@@ -13,27 +28,65 @@ main() {
...
+    echo "Could not find upstream in list of remotes."
+    git remote -v
+    exit 1
+  fi
+
+  # TODO@jsjoeio - can I somehow check that this succeeded?
   echo "Fetching upstream Homebrew/hombrew-core commits"
   git fetch upstream
 
+  # TODO@jsjoeio - can I somehow check that this succeeded?
```
jawnsy, line 56: you can likely check the exit status for `git merge` but this is likely a bit tricky to test, because the only time merges fail are when there's a merge conflict

ci/steps/brew-bump.sh
```diff
@@ -13,27 +28,65 @@ main() {
...
   echo "Merging in latest Homebrew/homebrew-core changes"
   git merge upstream/master
 
   echo "Pushing changes to cdrci/homebrew-core fork on GitHub"
+
+  # GIT_ASKPASS lets us use the password when pushing without revealing it in the process list
```
jawnsy, line 62: nice :+1:

test/scripts/steps-lib.bats
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bats
+
+SCRIPT_NAME="steps-lib.sh"
+SCRIPT="$BATS_TEST_DIRNAME/../../ci/steps/$SCRIPT_NAME"
+
+source "$SCRIPT"
+
+@test "is_env_var_set should return 1 if env var is not set" {
+  run is_env_var_set "ASDF_TEST_SET"
+  [ "$output" = 1 ]
```
jawnsy, line 10: If you change the functions to return stuff instead of echo, then you'd have to change this to check the `$status` instead of `$output`

jawnsy (approved):
left some really minor suggestions for your consideration, this is good-to-go as it is

ci/steps/brew-bump.sh
jsjoeio: Oh I like this way better. I'll refactor to do this. Thank you!

ci/steps/brew-bump.sh
jsjoeio: Ahh got it! Okay, maybe I'll just remove this line and ignore for now. I don't anticipate any merge conflicts since we're only ever updating our fork

test/scripts/steps-lib.bats
jsjoeio: Good point! Okay if/after I refactor, I'll update these!

ci/steps/brew-bump.sh
jsjoeio: I haven't used pushd/popd but Ill take your word and see if that's an easy refactor!

jawnsy: I think the proposed change would be, on line 39, change `cd homebrew-core` to `pushd homebrew-core` - this remembers the previous location and enters a directory. Then on line 51, call `popd` which changes to the directory you were in before you called pushd - it's a stack.

jsjoeio: Woah! That's so cool. Thanks for the guidance there. Very very easy to make the change. TIL!

PR #4265, opened 2021-09-28
feat(cli): add test for defaultConfigFile

jsjoeio:
This PR adds a test for `defaultConfigFile` in `src/node/cli.ts`.

image

Fixes N/A

TODOS 
- [x] export function
- [x] add JSDoc block to function
- [x] write basic test


src/node/cli.ts
```diff
@@ -502,10 +502,21 @@ export async function setDefaults(cliArgs: Args, configArgs?: ConfigArgs): Promi
...
+ * - bind-addr: 127.0.0.1:8080
+ * - auth: password
+ * - password: <password>
+ * - cert: false
+ */
+export function defaultConfigFile(password: string): string {
   return `bind-addr: 127.0.0.1:8080
 auth: password
-password: ${await generatePassword()}
+password: ${password}
```
jsjoeio, line 519: 🛠️ I believe functions should be _pure_ (unless they can't for some reason) so I refactored this to make `password` a param and pass in `await generatePassword()` when we call this function.

bryphe-coder: Nice!

test/unit/node/cli.test.ts
```diff
@@ -642,3 +643,15 @@ describe("bindAddrFromArgs", () => {
     resetValue()
   })
 })
+
+describe("defaultConfigFile", () => {
+  it("should return the dfeault config file as a string", async () => {
```
bryphe-coder, line 648:
```suggestion
  it("should return the default config file as a string", async () => {
```

bryphe-coder (approved):
LGTM!

test/unit/node/cli.test.ts
jsjoeio: lol nice catch, thank you!

vapurrmaid (commented):
Looks great ✔️

PR #4274, opened 2021-09-29
docs: update section around testing

### LINKED ISSUE #4177: Add testing guidelines
I would like to add a section to `MAINTAINING` which adds Testing Guidelines. This should explain:
- how testing works in code-server
- where to put your tests
- how to figure out what needs more tests
- how tests work in CI

jsjoeio:
This PR adds more docs around our testing strategy including:
- new section in `MAINTAINING.md`
- new section for scripts in `CONTRIBUTING.md`

Fixes #4177

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,20 +1,23 @@
 <!-- START doctoc generated TOC please keep comment here to allow auto update -->
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
 # Contributing
 
-- [Requirements](#requirements)
-- [Creating pull requests](#creating-pull-requests)
-  - [Commits and commit history](#commits-and-commit-history)
-- [Development workflow](#development-workflow)
-  - [Updates to VS Code](#updates-to-vs-code)
-  - [Build](#build)
+- [Contributing](#contributing)
+  - [Requirements](#requirements)
+  - [Creating pull requests](#creating-pull-requests)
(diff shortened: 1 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -129,13 +130,14 @@ yarn package
...
 
-There are three kinds of tests in code-server:
+There are four kinds of tests in code-server:
 
 1. Unit tests
-2. Integration tests
-3. End-to-end tests
+2. Script tests
+3. Integration tests
+4. End-to-end tests
```
vapurrmaid, line 140: **non-blocking(nice-to-have)** it's typically less error-prone to just use `1.` for all of them and let markdown convert them into 1 2 3 4 etc.

vapurrmaid (approved):
♦️ Thanks for adding specific, concise instructions in MAINTAINING.md - it gives new and recurring contributors a very easy decision framework to work from, which makes the community more inclusive!

docs/CONTRIBUTING.md
jsjoeio: I wish I could remember the point, but @code-asher mentioned a point _against_ doing this so I haven't, but yeah, it probably is less error-prone!

code-asher: Ahaha yeah because I read these in my editor and all I see is 1 1 1 11 1 1 1 1 1 1111 :stuck_out_tongue: 

Really this should not be done manually, I would use a VS Code plugin or something to number them automatically.

PR #4284, opened 2021-09-30
feat(cli): add test for readSocketPath

jsjoeio:
This PR adds tests for `readSocketPath`.

## TODOs

- [x] extract `readSocketPath` and export
- [x] fix TS error
- [x] write test for `readSocketPath`

Fixes N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,14 +31,18 @@ jobs:
       - name: Install helm
         uses: azure/setup-helm@v1.1
 
-      - name: Fetch dependencies from cache
-        id: cache-yarn
-        uses: actions/cache@v2
-        with:
-          path: "**/node_modules"
-          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            yarn-build-
+      # NOTE@jsjoeio
+      # disabling this until we can audit the build process
+      # and the usefulness of this step
+      # See: https://github.com/cdr/code-server/issues/4287
+      # - name: Fetch dependencies from cache
+      #   id: cache-yarn
+      #   uses: actions/cache@v2
+      #   with:
+      #     path: "**/node_modules"
+      #     key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+      #     restore-keys: |
+      #       yarn-build-
 
       - name: Install dependencies
         # if: steps.cache-yarn.outputs.cache-hit != 'true'
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -662,7 +662,7 @@ function bindAddrFromAllSources(...argsConfig: Args[]): Addr {
 /**
  * Reads the socketPath based on path passed in.
  *
- * The one usually pased in is the DEFAULT_SOCKET_PATH.
+ * The one usually passed in is the DEFAULT_SOCKET_PATH.
  *
  * If it can't read the path, it throws an error and returns undefined.
  */
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -3,7 +3,9 @@ import { promises as fs } from "fs"
 import yaml from "js-yaml"
 import * as os from "os"
 import * as path from "path"
-import { canConnect, generateCertificate, generatePassword, humanPath, paths } from "./util"
+import { canConnect, generateCertificate, generatePassword, humanPath, paths, isNodeJSErrnoException } from "./util"
+
+const DEFAULT_SOCKET_PATH = path.join(os.tmpdir(), "vscode-ipc")
 
 export enum Feature {
   // No current experimental features!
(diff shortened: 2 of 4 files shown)


src/node/cli.ts
```diff
@@ -667,6 +669,26 @@ export const shouldRunVsCodeCli = (args: Args): boolean => {
   return extensionRelatedArgs.some((arg) => argKeys.includes(arg))
 }
 
+/**
+ * Reads the socketPath which defaults to a temporary directory
+ * with another directory called vscode-ipc.
+ *
+ * If it can't read the path, it throws an error and returns undefined.
+ */
+export async function readSocketPath(path: string): Promise<string | undefined> {
```
jsjoeio, line 678: 💅 Mainly stylistic changes:
- make pure by passing in path
- export so we can test
- add check for `isNodeJSErronoException` so TS knows it has `error.code` and so we only throw if not a Node.js exception.

src/node/util.ts
```diff
@@ -482,7 +482,7 @@ export function escapeHtml(unsafe: string): string {
  * it has a .code property.
  */
 export function isNodeJSErrnoException(error: unknown): error is NodeJS.ErrnoException {
-  return error instanceof Error && (error as NodeJS.ErrnoException).code !== undefined
+  return error !== undefined && (error as NodeJS.ErrnoException).code !== undefined
```
jsjoeio, line 485: @code-asher and I learned `ErrnoException` extends `Error` but it is not an instance of `Error` so we had to fix this check

src/node/cli.ts
jsjoeio: This is because if the file isn't found, we don't care which is why we ignore `"ENOENT"` because we create a socket later ourselves.

bryphe-coder: Thanks for the description!

src/node/cli.ts
```diff
@@ -667,6 +669,27 @@ export const shouldRunVsCodeCli = (args: Args): boolean => {
   return extensionRelatedArgs.some((arg) => argKeys.includes(arg))
 }
 
+/**
+ * Reads the socketPath based on path passed in.
+ *
+ * The one usually pased in is the DEFAULT_SOCKET_PATH.
```
bryphe-coder, line 675:
```suggestion
 * The one usually passed in is the DEFAULT_SOCKET_PATH.
```

src/node/cli.ts
bryphe-coder: I like how you refactored the code to be more testable 👍

.github/workflows/ci.yaml
```diff
@@ -33,17 +33,21 @@ jobs:
...
-          path: "**/node_modules"
-          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            yarn-build-
+      # NOTE@jsjoeio
+      # disabling this until we can audit the build process
+      # and the usefulness of this step
+      # See: https://github.com/cdr/code-server/issues/4287
+      # - name: Fetch dependencies from cache
+      #   id: cache-yarn
```
bryphe-coder, line 41: Build caches can be tough! 😬

bryphe-coder (approved):

.github/workflows/ci.yaml
```diff
@@ -33,17 +33,21 @@ jobs:
...
+      #   uses: actions/cache@v2
+      #   with:
+      #     path: "**/node_modules"
+      #     key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+      #     restore-keys: |
+      #       yarn-build-
```
vapurrmaid, line 47: I back disabling this - just be aware of any timeouts related to installing (refetching) on each run - we saw this in `m` for a large package JSON and needed to adjust the allowable timeout

src/node/util.ts
vapurrmaid: Thoughts on this approach: https://stackoverflow.com/a/51533259 ?

vapurrmaid (commented):
✔️ looks good to me - just had a question about using a type predicate on ErrnoException

jsjoeio:
Waiting on https://github.com/cdr/code-server/pull/4290

src/node/util.ts
jsjoeio: > Thoughts on this approach: https://stackoverflow.com/a/51533259 ?

Maybe I'm misunderstanding the suggested solution, but I believe that's exactly why this change is necessary. I don't think that works 😢

> I learned ErrnoException extends Error but it is not an instance of Error

The typeguard works so it comes back as `NodeJS.ErrnoException` which is correct but using `error instanceof Error` when passing in `NodeJS.ErrnoException` returns `false` because an error of type `NodeJS.ErrnoException` is not an instance of `Error`. 

But maybe I'm missing something?

code-asher: Hmmm...there might be something more at play.  Check out what I get on my machine:

```
$ node -e 'try { fs.readFile("/") } catch (error) { console.log(error instanceof Error) }'
true
```

What do you get?

jsjoeio: Oh wow....Nice find! Hmm...how should we handle this then?

code-asher: I am not sure...options are to switch off Jest, look to see if there is a fix in progress (maybe this one? https://github.com/facebook/jest/pull/8220 but it is rather old), or patch it ourselves.

jsjoeio: Maybe I'll make an issue and do this in a followup PR if that's okay with y'all.

code-asher: Sure thing.

jsjoeio: Tracking here: https://github.com/cdr/code-server/issues/4303

Issue #4264, opened 2021-09-28
Provide nightly builds for testing


jawnsy:
# Background

We currently provide a variety of release builds that users can install with install.sh, as well as a package published to npm so that users can install from source code. These are updated when we create a release, which means that users may have to wait until the next release to see whether a bug fix addressed their issue.

Reducing the cycle time for feedback means that we can iterate more rapidly on code-server, and one way to achieve that is to publish nightly builds, so that users can see whether we have resolved their issues.

# Proposal

* Publish a `code-server-next` or `code-server-latest` package that users can install via npm
* Publish nightly builds of our usual build artifacts, like a Debian/Ubuntu package, etc (we may want to limit these to the last 30 or 90 days to reduce storage costs)

jsjoeio:
Yes! We need this yesterday. cc @bpmct @code-asher since I've chatted with y'all about this before.

jsjoeio:
## Homebrew

One of the pieces of this puzzle is supporting nightlies on Homebrew. While `homebrew/core` doesn't support beta/development versions, we can maintain our own tap (grain-lang does this [here](https://github.com/grain-lang/homebrew-tap)).

### UX

As a user, not a lot would change.
- to install code-server's stable release, `brew install code-server`
- to install the nightly release, `brew tap cdr/code-server-nightly` (example)

### Notes

> the forumla can get complex if you need to build from source

> Formula are the "build from source" option, and Cask are the "prepackaged"  option

> Yeah, you can see that in our `url` config, we derive the github tag and then download our tool

## Resources

- https://docs.brew.sh/Versions#acceptable-versioned-formulae
- https://github.com/grain-lang/homebrew-tap
- https://github.com/Homebrew/discussions/discussions/2217

jsjoeio:
Not stale

PR #4288, opened 2021-10-01
refactor(ci): only run release workflows on released event

jsjoeio:
This PR updates the trigger event for both the "Publish on npm and brew" and the "Publish on Docker" release workflows so that they only run on `released` (instead of `published`).

## Why is this change necessary?

In order for us to publish pre-releases _without_ triggering these workflows, this change needs to happen.

We learned this through the [GitHub Actions docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release):

> Note: The prereleased type will not trigger for pre-releases published from draft releases, but the published type will trigger. If you want a workflow to run when stable and pre-releases publish, subscribe to published instead of released and prereleased.

 

Fixes N/A
Related https://github.com/cdr/code-server/issues/4264

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
   release:
-    types: [published]
+    types: [released]
 
 jobs:
   docker-images:
--- a/.github/workflows/npm-brew.yaml
+++ b/.github/workflows/npm-brew.yaml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
 
   release:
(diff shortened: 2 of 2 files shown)


jsjoeio:
waiting on https://github.com/cdr/code-server/pull/4290

jawnsy (approved):
very interesting

jsjoeio:
The build is failing and I'm not sure why. See [logs](https://github.com/cdr/code-server/runs/3772710703?check_suite_focus=true#step:11:9). I'm going to investigate.

image

PR #4293, opened 2021-10-02
Fix: use pipe to spawn http server process to enable logs pipe to files

tianjianchn:
Previously we use `inherit` in stdio param to spawn the http server process(or the child process), which will swallow the logs in child process and not be written to the `code-server-stdout.log`. Using `pipe` will fix this problem.

Fixes #


jsjoeio:
@code-asher do you mind reviewing this? I think last time we changed this code, it was when we found that bug thanks to the help from the `ts-node` maintainer.

code-asher (approved):
Ahhhh I was wondering why we stopped getting output in the log file.  Thank you for this fix!!

PR #4301, opened 2021-10-03
Fix marketplace in CLI

TeffenEllis:
This PR resolves an issue where our [marketplace modifications](https://github.com/cdr/vscode/commit/65b2462c212d415fbf521489307e58e5b691818b) are not applied to the CLI.

- Also fixes error logging from child process.

### DIFF
--- a/src/node/main.ts
+++ b/src/node/main.ts
@@ -22,8 +22,8 @@ export const runVsCodeCli = async (args: DefaultedArgs): Promise<void> => {
 
   try {
     await cliProcessMain(args)
-  } catch (error) {
-    logger.error("Got error from VS Code", field("error", error))
+  } catch (error: any) {
+    logger.error("Got error from VS Code", error)
   }
 
   process.exit(0)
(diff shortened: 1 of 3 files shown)


src/node/main.ts
```diff
@@ -22,8 +22,8 @@ export const runVsCodeCli = async (args: DefaultedArgs): Promise<void> => {
 
   try {
     await cliProcessMain(args)
-  } catch (error) {
-    logger.error("Got error from VS Code", field("error", error))
+  } catch (error: any) {
```
jawnsy, line 25: just curious, are there root error types in Node or TypeScript that we can use instead of `any`? seems unfortunate not to have any type information

jsjoeio: I think it's traditionally been `any` unfortunately. But in TypeScript 4.0, they allow you [to type it as `unknown`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-0/#unknown-on-catch) which they say is safer. But yeah, it's too bad they don't tell us _which_ error this is.

jsjoeio (approved):

PR #4304, opened 2021-10-04
Remove unused Lock config

jawnsy:
We previously used the Lock app to lock old issues. Since we use Stale to close out old issues, we no longer need this and can remove the config.


jsjoeio (approved):
Thanks @jawnsy! 🙌

PR #4306, opened 2021-10-04
fix: revert CI change + fix broken tests

jsjoeio:
This PR does a few things:
- fixes broken unit test
- reverts a CI change around testing made in https://github.com/cdr/code-server/pull/4010
- check if `vscode` exists before closing

Fixes N/A

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,10 @@ jobs:
         run: yarn lint
         if: success()
 
+      - name: Run code-server unit tests
+        run: yarn test:unit
+        if: success()
+
       - name: Upload coverage report to Codecov
         run: yarn coverage
         if: success()
@@ -403,9 +407,6 @@ jobs:
           cd test/
           rm -r node_modules/playwright
           yarn install --check-files
-
-      - name: Run end-to-end tests
-        run: yarn test:unit
 
       - name: Run end-to-end tests
         run: yarn test:e2e
--- a/src/node/routes/index.ts
+++ b/src/node/routes/index.ts
@@ -162,7 +162,7 @@ export const register = async (
   }
 
   server.on("close", () => {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -54,6 +54,10 @@ jobs:
         run: yarn lint
         if: success()
 
+      - name: Run code-server unit tests
+        run: yarn test:unit
+        if: success()
+
```
jsjoeio, line 60: I'm guessing this was removed by accident but it should be here and we need it before uploading coverage to Codecov (it outputs coverage information).

jawnsy (approved):

.github/workflows/ci.yaml
```diff
@@ -404,9 +408,6 @@ jobs:
           rm -r node_modules/playwright
           yarn install --check-files
 
-      - name: Run end-to-end tests
-        run: yarn test:unit
-
```
jsjoeio, line 409: No need to run unit tests twice

src/node/routes/index.ts
```diff
@@ -162,7 +162,7 @@ export const register = async (
   }
 
   server.on("close", () => {
-    vscode.vscodeServer.close()
+    vscode?.vscodeServer.close()
```
jsjoeio, line 165: This prevents us trying to call `vscodeServer.close()` on `undefined`

test/unit/node/routes/static.test.ts
```diff
@@ -7,7 +7,7 @@ import * as integration from "../../../utils/integration"
 
 const NOT_FOUND = {
   code: 404,
-  message: "not found",
+  message: "Not Found",
```
jsjoeio, line 10: 🛠️ Change needed to make test pass (casing shouldn't matter). I was going to use `.toMatch(/string/)` but this was quicker

PR #4312, opened 2021-10-05
nginx config minor bug fixed

SnorlaxYum:
without `^~`, resources in folders like `/static` cannot be accessed.
So I do this PR. It's a minor one in the doc. So I think there's no need to create an issue for it.

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -198,7 +198,15 @@ At this point, you should be able to access code-server via
        listen [::]:80;
        server_name mydomain.com;
 
-       location ^~ / {
+       location / {
+         proxy_pass http://localhost:8080/;
+         proxy_set_header Host $host;
+         proxy_set_header Upgrade $http_upgrade;
+         proxy_set_header Connection upgrade;
+         proxy_set_header Accept-Encoding gzip;
+       }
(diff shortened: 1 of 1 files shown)


jsjoeio:
@bpmct or @code-asher could you review this? I haven't used nginx so not sure I can help much here.

SnorlaxYum:
without `^~`, I'll encounter 404 responces when the code server page tries to access resources with the uri like `/static/b37ff28a0a582aee84a8f961755d0cb40a4081db/usr/lib/code-server/vendor/modules/code-oss-dev/out/vs/workbench/workbench.web.api.nls.js` (and it is a usual case in the default setting).
after adding `^~`, it began working.
Another option is add a same block with `/` changed to `/static`, which also works.

jsjoeio:
@SnorlaxYum were you able to try @code-asher's suggestion? Hoping we can either close or move forward!

jsjoeio:
@SnorlaxYum we haven't heard from you in a couple days so I'm going to assume you fixed your issue. I'm going to close this PR. If you want to discuss getting this in, feel free to open an issue and we'll work with you!

PR #4318, opened 2021-10-07
Remove deprecated plugin architecture.

TeffenEllis:
This PR removes our internal plugin architecture, given that Coder Enterprise now supports the initial intent of this feature via custom apps. Additionally, some changes in VS Code upstream don’t fully align with our legacy design. For example, VS Code is moving toward a multi-server approach to isolated web views, however it’s unclear how we could align with this implementation with a plugin-ified vscode route.


jawnsy:
100% concur with @vapurrmaid, I think we will need to maintain this for at least a few more releases of code-server (and Coder) and go through our Coder product deprecation lifecycle. Are there concerns with keeping this support as-is for another, say, 6 months? What's the maintenance cost associated with keeping it?

TeffenEllis:
@jawnsy @vapurrmaid,

Those are all great points. At the moment, VS Code is a bit more forgiving of a single server approach, but I’m guessing this’ll be phased out in an upstream version in the not too distant future. What this means for code-server users is unclear, with them possibly having to open two ports on their host machine, or us compensating with an API loopback endpoint.

Plugins assume a single server with an Express interface, and support for endpoint specific web socket routing. Both of these seem to clash with upstream’s intent, but our work-arounds have been [somewhat successful](https://github.com/cdr/code-server/blob/main/src/node/routes/vscode.ts). If it’s possible, an estimate from the Coder Enterprise team on when plugins can be fully removed is appropriated and will help us deliver version upgrades faster.

jsjoeio:
What do we need to close or move this forward?

PR #2659, opened 2021-02-02
Replace gallery with Open-VSX

oxy:
After discussing internally, we've decided to just switch our default extension gallery to Open-VSX, and then document how to switch back to our old gallery. This PR just changes the default, and documentation changes are still required before we can make a release!

The hack I attempted previously broke pagination, and there is no obvious, non-invasive fix in sight due to how VSCode was designed.

Note that switching back should essentially be setting  `GALLERY_URL="https://extensions.coder.com/api" ITEM_URL=` (intentionally blank).

Closes https://github.com/cdr/code-server/issues/1473.

TODOS
 - [ ] test with Vim

PR #4319, opened 2021-10-07
Open VSX switch, Part II

### LINKED ISSUE #1473: Integrate with open-vsx.org
Hi,

Open VSX was just released and it is a open source registry for VS Code extensions, just like Visual Studio Marketplace.

It may be a good idea to use and support Open VSX instead of creating yet another one, since the extension developers will have more places to publish.

It's just an idea.

https://open-vsx.org/
https://github.com/eclipse/openvsx

GirlBossRush:
This PR continues the work set in place by Oxy in #2659, migrating our default configuration extension marketplace to VSX. We’ve received reports of extensions refusing to install, or exhibiting odd behaviors. I believe this is one part due to outdated extensions in our marketplace, and another regarding Microsoft’s [upstream changes](https://github.com/microsoft/vscode/commit/959b5ea26908071a5bdada0ff4d5389ece856a44) to determine what makes an extension “web compatible."

See https://github.com/cdr/vscode/commit/5f00e799937119b719eb32a692e9af0af20d9ce4 for additional implementation details.

Closes #1473


jsjoeio:
I think we need to fix this failing unit test. I was going to use `.toMatch` but got lazy because that takes a regex I think and I forgot to add it in.
```javascript
  ● /_static › disabled authentication › should return a 404 when a file is not provided

    expect(received).toContain(expected) // indexOf

    Expected substring: "not found"
    Received string:    "Not Found"

      45 | 
      46 |       const content = await resp.json()
    > 47 |       expect(content.error).toContain(NOT_FOUND.message)
         |                             ^
      48 |     })
```

jsjoeio:
If we fix the conflicts, is this ready to be merged?

TeffenEllis:
@jsjoeio this can likely be rebased when #4414 is merged, which will include VSX due to version incompatibilities in our previous marketplace

Issue #1104, opened 2019-10-24
Source Control context menu missing some items


JtMotoX:
- `code-server` version: 2.1638-vsc1.39.2
- OS Version: macOS Mojave 10.14.6

## Description

Source Control context menu missing some items that I use frequently in VSC.
- Stage Selected Ranges
- Unstage Selected Ranges
- Revert Selected Ranges

## Steps to Reproduce

1. Git clone a repo
1. Make some changes to a file
1. Open the file diff working tree in Source Control
1. Right click one of the modified lines in the right pane
1. Under "Change All Occurrences" there should be 3 additional options listed in the description

![3](https://user-images.githubusercontent.com/7191259/67610154-be5bc480-f745-11e9-92d8-7bed36d4ffe3.jpg)

sr229:
Can you tell us what UA/Browser you're using? It will help us alot.

JtMotoX:
> Can you tell us what UA/Browser you're using? It will help us alot.

I am using Chrome version 77.  I have also updated the original post with a screenshot.

sr229:
Hey OP, I'll be closing this since Microsoft has stated they would want this be seen on the main vscode project. Make sure to prefix [web] and use the same exact details here (with the exception of code-server version). Thanks for helping us out spot bugs!

JtMotoX:
> I went ahead and patched this until VS Code makes a proper fix (since I'm not sure this is the ideal way to fix it). Will have a new release by the end of the day.

Thank you! 🥇   I opened issue with MS (link below) but they closed it right away without looking at it so I really appreciate your patch.

https://github.com/microsoft/vscode/issues/83454

PR #4414, opened 2021-10-29
Use upstream server WIP

TeffenEllis:
This PR updates code-server to lean more on upstream’s server behaviors.

See **code-server-v2** branch on our fork additional changes: https://github.com/cdr/vscode/compare/main...cdr:code-server-v2?expand=1 

Breaking changes so far:
- extra-extensions-dir flag removed
- extra-builtin-extensions-dir flag removed
- Static endpoint cannot reach outside code-server anymore (/vcode-remote-resource can though)
- Experimental compression feature flag removed (is probably the default now) (not really breaking as this is not exposed)
- install-source flag removed
- locale flag removed
  - can't [change display language](https://code.visualstudio.com/docs/getstarted/locales) 
- no "Logout" in menu 
- Help > About doesn't show code-server version
- Installed Extensions doesn't show installed extensions
- Vim extension does not work

TODO:
- [x] Patch product.json with our values, Open VSX
- [x] Locale flag needed?
- [x] Install source flag needed?
- [x] Add back nfpm fix
- [ ] Open the last opened workspace/directory when there is no folder/workspace query param (unless the ignore-last-opened flag is set)
- [ ] Add back support for non-encoded syntax for folder and workspace query params
- [ ] Theme images on the welcome page are broken
- [ ] Add back `type` query param to the web socket connections (used in product)
- [ ] Spawn VS Code on demand to avoid overhead?
- [ ] Proposed API (either implement or always enable it and remove the flag)
- [ ] CONTRIBUTING.md is probably wildly out of date now
- [ ] Replace our token handling with VS Code's native's token
- [ ] Reverse proxy: service worker with base path fails due to incorrect scope
- [ ] Reverse proxy: update endpoint is incorrect (incorrect base path and protocol)

To test:
- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Password authentication
- [x] Search, edit, other basic functionality
- [x] Caching
- [x] Opening to a directory or workspace on the command line
- [x] Disconnections
- [x] Heartbeat
- [x] Display language (localization)
- [x] Opening a file in an existing instance
- [x] Log out
- [x] PWA
- [x] Some extensions (Java, Python, Jupyter, Vim, themes, icon themes)
  - [x] icon themes
  - [x] Python
  - [x] Vim
  - [x] Jupyter
  - [x] Java
  - [x] themes
- [x] Reverse proxy (NGINX, Caddy) (HTTP, HTTPS, base path, no base path, trailing slash, no trailing)
- [x] Debugging
- [x] Terminal, terminal persistence
- [x] Hot reload (`pkill -SIGUSR1 -f code-server`)
- [x] Running code-server in code-server
- [x] Log level gets set correctly in client
- [x] Update notification
- [x] /vscode endpoint (not working without trailing slash, otherwise it seems /anystring/ shows VS Code)
- [x] Plugins
- [x] GitHub auth
- [x] Diff menu https://github.com/cdr/code-server/issues/1104
- [x] CODE_SERVER_* variable sanitization in terminal


ci/build/build-vscode.sh
```diff
@@ -13,7 +13,7 @@ main() {
 
   # extensions-ci compiles extensions and includes their media.
   # compile-web compiles web extensions. TODO: Unsure if used.
-  yarn gulp extensions-ci compile-web "vscode-server${MINIFY:+-min}"
+  yarn gulp extensions-ci compile-web "vscode-reh-web${MINIFY:+-min}"
```
jsjoeio, line 16: ✍🏼 reh = remote extension host (note for future us)

src/node/app.ts
```diff
@@ -44,7 +44,7 @@ const listen = (server: http.Server, { host, port, socket }: ListenOptions) => {
       server.listen(socket, onListen)
     } else {
       // [] is the correct format when using :: but Node errors with them.
-      server.listen(port, host.replace(/^\[|\]$/g, ""), onListen)
+      server.listen(parseInt(port, 10), host.replace(/^\[|\]$/g, ""), onListen)
```
jsjoeio, line 47: ❓ Guessing this is to please TypeScript and sure port is indeed a `number`?

src/node/cli.ts
```diff
@@ -29,27 +29,7 @@ export enum LogLevel {
...
-    | "log"
-    | "verbose"
-    | "install-source"
-    | "list-extensions"
-    | "install-extension"
-    | "uninstall-extension"
-    | "locate-extension"
-    // | "telemetry"
-  > {
+export interface Args extends CodeServerLib.ServerParsedArgs {
```
jsjoeio, line 32: 👏 I love deletions like this!

src/node/cli.ts
```diff
@@ -65,7 +45,6 @@ export interface Args
   json?: boolean
   log?: LogLevel
   open?: boolean
-  port?: number
```
jsjoeio, line 68: ❌ is this a breaking change?

src/node/cli.ts
```diff
@@ -167,7 +147,7 @@ const options: Options<Required<Args>> = {
 
   // These two have been deprecated by bindAddr.
   host: { type: "string", description: "" },
-  port: { type: "number", description: "" },
+  port: { type: "string", description: "" },
```
jsjoeio, line 150: ❌ is this a breaking change?

src/node/main.ts
```diff
@@ -12,13 +12,15 @@ import { startLink } from "./link"
...
-    "vs/code/node/cli",
-    "shouldSpawnCliProcess",
+export const shouldSpawnCliProcess = async (args: CodeServerLib.ServerParsedArgs): Promise<boolean> => {
+  return (
+    !args["start-server"] &&
+    (!!args["list-extensions"] ||
+      !!args["install-extension"] ||
+      !!args["install-builtin-extension"] ||
+      !!args["uninstall-extension"] ||
+      !!args["locate-extension"])
```
jsjoeio, line 22: nit: I don't _love_ this pattern with `!!` because I think it's harder to reason about. I also think it might be worth extracting this into a function and testing? (I may have done that in a recently merged PR).

jsjoeio (commented):
Most of this looks good to me 👍

Biggest question: what's the motivation for changing port from `number` -> `string`? And is it considered a breaking change?

src/node/app.ts
TeffenEllis: Indeed, upstream anticipates a stringly number, I’m guessing because their server arguments interface isn’t as well defined as their desktop variant.

src/node/cli.ts
TeffenEllis: This was removed since any server function we’d pass the args into won’t accept a `string | number` union, so instead the only additions we add are code-server specific

src/node/main.ts
TeffenEllis: Funny enough, this came at straight from upstream:

https://github.com/cdr/vscode/blob/673165ca934c05d63fcab74bac7c819c32528c64/src/vs/server/main.js#L31

src/node/cli.ts
jsjoeio: Ahh okay, got it! Thanks for clarifying

jsjoeio (changes requested):
No concerns from me! 👍 Requesting changes since there are conflicts (and I see some todos in the description still). Once those are sorted, lmk and I'll test locally for regressions!

ci/build/build-vscode.sh
code-asher: Looks like this does not exist, maybe we want `vscode-web` and `vscode-web-min`?

src/node/cli.ts
```diff
@@ -176,31 +156,18 @@ const options: Options<Required<Args>> = {
...
   },
-  "enable-proposed-api": {
-    type: "string[]",
-    description:
-      "Enable proposed API features for extensions. Can receive one or more extension IDs to enable individually.",
-  },
   "uninstall-extension": { type: "string[]", description: "Uninstall a VS Code extension by id." },
   "show-versions": { type: "boolean", description: "Show VS Code extension versions." },
   "proxy-domain": { type: "string[]", description: "Domain used for proxying ports." },
-  "ignore-last-opened": {
```
code-asher, line 199: Removing this would cause a regression in behavior right?

src/node/cli.ts
```diff
@@ -176,31 +156,18 @@ const options: Options<Required<Args>> = {
   "user-data-dir": { type: "string", path: true, description: "Path to the user data directory." },
   "extensions-dir": { type: "string", path: true, description: "Path to the extensions directory." },
   "builtin-extensions-dir": { type: "string", path: true },
-  "extra-extensions-dir": { type: "string[]", path: true },
-  "extra-builtin-extensions-dir": { type: "string[]", path: true },
```
code-asher, line 180: Glad to be getting rid of these, they were buggy.

src/node/cli.ts
```diff
@@ -176,31 +156,18 @@ const options: Options<Required<Args>> = {
...
   force: { type: "boolean", description: "Avoid prompts when installing VS Code extensions." },
-  "install-source": { type: "string" },
   "locate-extension": { type: "string[]" },
   "install-extension": {
     type: "string[]",
     description:
       "Install or update a VS Code extension by id or vsix. The identifier of an extension is `${publisher}.${name}`.\n" +
       "To install a specific version provide `@${version}`. For example: 'vscode.csharp@1.2.3'.",
   },
-  "enable-proposed-api": {
```
code-asher, line 191: Should we keep this flag or do we enable it by default now?

src/node/cli.ts
```diff
@@ -176,31 +156,18 @@ const options: Options<Required<Args>> = {
   "user-data-dir": { type: "string", path: true, description: "Path to the user data directory." },
   "extensions-dir": { type: "string", path: true, description: "Path to the extensions directory." },
   "builtin-extensions-dir": { type: "string", path: true },
-  "extra-extensions-dir": { type: "string[]", path: true },
-  "extra-builtin-extensions-dir": { type: "string[]", path: true },
   "list-extensions": { type: "boolean", description: "List installed VS Code extensions." },
   force: { type: "boolean", description: "Avoid prompts when installing VS Code extensions." },
-  "install-source": { type: "string" },
```
code-asher, line 183: Do we know if anyone uses this flag and removing would cause a regression?

src/node/cli.ts
```diff
@@ -212,7 +179,6 @@ const options: Options<Required<Args>> = {
     description: "Force to open a file or folder in an already opened window.",
   },
 
-  locale: { type: "string" },
```
code-asher, line 215: Do we know if anyone uses this flag and removing would cause a regression?  I think this is something to do with display language support but am not sure it is useful.

src/node/routes/vscode.ts
```diff
@@ -37,19 +37,15 @@ export const createVSServerRouter = async (args: DefaultedArgs): Promise<VSServe
 
   // Signal processes that we got launched as CLI
   process.env["VSCODE_CLI"] = "1"
+  // Seems to be necessary to load modules properly.
+  process.env["VSCODE_DEV"] = "1"
```
code-asher, line 41: Is this dev as in development?  I wonder if this will cause issues in the production build.

code-asher (commented):
I have not been able to run this with watch or build yet but it looks very promising!!

src/node/cli.ts
jsjoeio: I don't know if I've ever seen it mentioned, but good question. Hard to know!

src/node/cli.ts
jsjoeio: I am also not sure if it's used. I guess we'll have to make this a breaking change and if there is a regression, we'll fix after?

jsjoeio: (I wish there were an easier way to know if this is actually used)

TeffenEllis: AFAIK locale isn’t supported in the server arguments, and they seem to be migrating to a client driven approach, [here](https://github.com/Microsoft/vscode/blob/8b8e1d1dfaf6fa6e42c4ca2743c40a012c4ac455/src/vs/server/extensionHostConnection.ts#L26) and [here](https://github.com/microsoft/vscode/blob/8b8e1d1dfaf6fa6e42c4ca2743c40a012c4ac455/src/vs/server/remoteAgentEnvironmentImpl.ts#L124)

src/node/routes/vscode.ts
TeffenEllis: I suspect this is yet to be fixed upstream since their server run command [assumes a development env](https://github.com/microsoft/vscode/blob/8b8e1d1dfaf6fa6e42c4ca2743c40a012c4ac455/resources/server/web.sh#L21), and their [prod command](https://github.com/microsoft/vscode/blob/main/resources/server/bin/server.sh) seems to error without this arg.

ci/build/build-vscode.sh
TeffenEllis: I believe this exists as a generated command. 

https://github.com/microsoft/vscode/blob/8b8e1d1dfaf6fa6e42c4ca2743c40a012c4ac455/build/gulpfile.reh.js#L365

src/node/main.ts
jsjoeio: Bummer! Okay oh well - something I need to nit about upstream 😂

src/node/cli.ts
TeffenEllis: AFAIK this argument isn’t referenced anywhere in code-server or upstream.

ci/build/build-vscode.sh
code-asher: Locally I get the same as CI here https://github.com/cdr/code-server/runs/4051125133?check_suite_focus=true#step:8:10

It looks like we are missing the platform part.  If we add it we would not be able to share the build step between platforms though, maybe we can get by with the general build.

src/node/cli.ts
code-asher: Oh snap that means we lost it so definitely a regression.

jsjoeio:
Just fixed the conflicts. But there are ton of TypeScript/testing errors to fix now that `port` is gone. Marking this as draft and going to start working on them.

ci/build/build-vscode.sh
TeffenEllis: Ahh this makes sense since upstream uses a separate build pipeline for each platform, and these gulp commands are only defined by the machine running the build. Maybe we could interpolate the platform name into a build step?

jsjoeio:
There is a lot to be fixed with this and we need to prioritize getting a release out ASAP so we're going to wait until @TeffenEllis is back to help with this.

jsjoeio:
Tested in Chrome. All looks good.

image

jsjoeio:
Do we need to update the `product.json`? (I think that's where the product is called Code - OSS Dev right?)

image

jsjoeio:
Tested in Safari.

Some browser errors but main functionality is working

image

jsjoeio:
Tested in Firefox.

Main functionality is working. Seeing the built-in extensions error though (might have seen this in Safari too)

image

jsjoeio:
PWA works 🎉

https://user-images.githubusercontent.com/3806031/140989959-0dfdf4b4-c5e0-48cd-a9ef-896c02207718.mov

jsjoeio:
Themes work!

image

Same with Jupyter !

image

jsjoeio:
## Joe/Asher Pair-programming Plan

- [ ] Fix e2e tests
- [ ] Finish testing checklist
- [ ] Make issues for TODO items (and any failing tests)
- [ ] Merge PR

jsjoeio:
Changing icons works!

image

jsjoeio:
Sign in with GitHub is asking for personal access token (Asher says we need to patch that back in)

image

jsjoeio:
Source Control context menu looks good 👍

image

jsjoeio:
code-server variables shouldn't be available in the terminal. Need to fix again

image

jsjoeio:
`/vscode` only works with trailing slash

image

jsjoeio:
Installing extension via CLI just hangs

image

jsjoeio:
Running `node . <file>` in the Integrated Terminal in code-server opens it in code-server (as expected)

image

jsjoeio:
update notification still works! 🎉

image

jsjoeio:
Need to look into e2e tests. I think we're getting a focus error with the Integrated Terminal test because something on the Getting Started page is taking focus which means we can't click the Command Palette in the menu.

image

code-asher (approved):
We will be knocking out the remaining tasks in separate PRs.

jsjoeio:
I will open new issues and create a milestone for the separate tasks!

Issue #4415, opened 2021-10-29
Yarn installation fails incompatible engine 'node' 


Justxd22:
## OS/Web Information

- Web Browser: ~
- Local OS: Android q aarch64 (arm64) 
- Remote OS: ~
- Remote Architecture: ~ 
- `code-server --version`: latest 

## Steps to Reproduce

1. Latest termux version 
2. apt update && apt upgrade && apt install build-essential python git nodejs-lts yarn

3.yarn global add code-server

## Expected

Install? 

## Actual

yarn complains about nodejs version expected v14 (the old lts) 
The problem is the nodejs-lts is now v16 how can i get v14? Should I go with the standalone install? 

## Screenshot

![Screenshot_20211029_232247](https://user-images.githubusercontent.com/66136622/139498267-78150fcf-90c7-41a9-a43e-ed9547d6a663.jpg)

Justxd22:
edit :

Using the ignore engines flag : 


yarn global add code-server --ignore-engines
yarn global v1.22.15
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "code-server > @coder/logger@1.1.16" has unmet peer dependency "@google-cloud/logging@^4.5.2".
[4/4] Building fresh packages...
[1/3] ⠠ argon2
[-/3] ⠐ waiting...
error /data/data/com.termux/files/home/.config/yarn/global/node_modules/argon2: Command failed.
Exit code: 1
Command: node-pre-gyp install --fallback-to-build
Arguments:
Directory: /data/data/com.termux/files/home/.config/yarn/global/node_modules/argon2
Output:
node-pre-gyp info it worked if it ends with ok
node-pre-gyp info using node-pre-gyp@1.0.6
node-pre-gyp info using node@17.0.1 | android | arm64
node-pre-gyp info check checked for "/data/data/com.termux/files/home/.config/yarn/global/node_modules/argon2/lib/binding/napi-v3/argon2.node" (not found)
node-pre-gyp http GET https://github.com/ranisalt/node-argon2/releases/download/v0.28.2/argon2-v0.28.2-napi-v3-android-arm64-unknown.tar.gz
node-pre-gyp ERR! install response status 404 Not Found on https://github.com/ranisalt/node-argon2/releases/download/v0.28.2/argon2-v0.28.2-napi-v3-android-arm64-unknown.tar.gz
[...]

PR #4472, opened 2021-11-09
Fix : recreate the termux guide to adapt the recent changes

arHSM:
Termux nodejs-lts changed from v14 to v16 and there are many issues people are facing such as with argon2. Hence I recommend changing it to this install process which is comparably better and has one less issue :^)

I've also added some extra things such as installing GO and Python, idk about the TOC tree but this is pretty much it.

Fixes
*Sort of fixes*
#4415
#4470 
#4394


im-coder-lg:
@arHSM this is good! @jsjoeio you can sanction the checks. @arHSM be sure to:
1. Clone locally, run `yarn fmt` rebase(use the GitHub UI and merge the latest commits), and push the change up here
2. not get scared that the docs are failing, since you edited on a fork, you don't have the tokens to publish changes to Coder's Vercel deployment, but it's not required. The build, e2e tests and pre-build checks are required though.

im-coder-lg:
Okay, now, do these steps:
1. Clone this repo and open a new tab with your code-server fork, and **remember, be on your PR branch**, ie: arHSM:patch-2
2. Open your terminal of choice, make sure Node JS and Yarn is installed, the CLI works and on your terminal, run `yarn install && yarn fmt`.
3. While it is running, switch to your branch inside your fork, and check for a `Fetch and merge` option. Fetch and merge the latest commits, and push the `yarn fmt` result here(there _will_ be a minor change which is important)

Then, wait for the checks to complete(if it doesn't, wait for a reply from @jsjoeio, he will help on this.)

im-coder-lg:
Oh no, only a maintainer can change the docs! Since a maintainer has access to the repo!

Let me explain elaborately. Imagine, just imagine that I am a maintainer of code-server. I'd have access to:
1. the repo itself
2. Coder's Vercel deployment(Coder.com)
3. Secret tokens

So, it means that if I make a docs edit, I'd make it from inside the repo since we need the tokens(secret) so that the tokens are applied. Whenever I try merging another branch inside code-server to the `main` branch, I'd get a docs deployment reply, I'd follow steps on making the PR, like running `yarn fmt` locally, etc.

So, if the docs are approved+the docs check finally works, great! If it doesn't, either the PR can get merged, or a maintainer can make the PR and we can review, which is better.

im-coder-lg:
Now, merge the latest commits from the base branch to your PR branch. This would help on merge.

arHSM:
image
looks good

jsjoeio (changes requested):
Hi @arHSM!

First, thank you so much for your PR ♥️ We really appreciate it.

I think there is some hesitation with this doc change mainly because it feels like it introduces a lot of extra steps compared to our previous guide. I'm wondering if we could find a happy middle ground?

The Installing Python and Go additions are fantastic! I'm wondering if we can simplify this guide somehow. In my head, I'm thinking something like:

Get Termux
Install dependencies (i.e. Node/npm, nvm)
Install code-server
Run code-server
Do you think it's possible to keep it as simple as that? I'd rather us not suggest switching users or doing much beyond those four steps but maybe it's not possible due to our current limitations.

Let us know!

arHSM:
> I'd rather us not suggest switching users or doing much beyond those four steps but maybe it's not possible due to our current limitations.

yeah i think we can remove those steps, but can we atleast add a note like "consider creating a new user instead of using root, read more [here](link to guide)"?

jsjoeio:
> yeah i think we can remove those steps, but can we atleast add a note like "consider creating a new user instead of using root, read more [here](link to guide)"?

Sure, if you think there's a specific use case that solves, we could definitely add that as a guide! The guide could even by "how to add a new user" or "why you should consider adding a new user"

arHSM:
I have now removed those unnecessary steps and added a quote black at the end of the process stating the following
> Consider using a new user instead of root, read [here](https://www.howtogeek.com/124950/htg-explains-why-you-shouldnt-log-into-your-linux-system-as-root/) why using root is not recommended.\
> Learn how to add a user [here](https://gist.github.com/arHSM/62242c343efc2827861ddc38e485d7df).

and i also noticed that Install & Upgrade was \#\# and not \#\#\# so  i made that change too

docs/termux.md
```diff
@@ -7,78 +7,66 @@
...
-
 ```bash
 curl -fsSL https://code-server.dev/install.sh | sh
 ```
 
-12. Every time you run `./start-debian.sh` (or `proot-distro login debian` if you used proot-distro), you can switch to your user `su - username` and run code-server.
+8. You can now start code server by simply running `code-server`.
+
+> Consider using a new user instead of root, read [here](https://www.howtogeek.com/124950/htg-explains-why-you-shouldnt-log-into-your-linux-system-as-root/) why using root is not recommended.\
+> Learn how to add a user [here](https://gist.github.com/arHSM/62242c343efc2827861ddc38e485d7df).
```
jsjoeio, line 60: I'm not against using a gist here but if you accidentally delete it in the future, it breaks our docs 😛

What if we move this into it's own section under Extra? And then you can use the text directly instead of linking the gist.

jsjoeio (changes requested):

jsjoeio:
> and i also noticed that Install & Upgrade was ## and not ### so i made that change too

you rock! I requested one last change and then I think we'll be good to go! ♥️

docs/termux.md
arHSM: Ok!

jsjoeio (approved):
Awesome! I think we're good to go once CI passes (ignore Docs Preview)

arHSM:
Wait don't merge yet, I have found a typo in step 5

jsjoeio:
@arHSM you give me the thumbs up when this is ready to merge!

PR #4492, opened 2021-11-10
Update changelog

code-asher:
I added the missing versions and some changelog entries for the latest
version.  I also added some extra details to the entries that should
help users understand what changed and what they need to do about
it.  The overall format is based on keepachangelog.com.

In that same spirit I removed entries that do not affect
users (documentation and development changes).

I removed the names because it seems unlikely users will be interested,
that information can be found in the PR, and code is really the work of
multiple developers (even if only one is writing code and the other is
reviewing) which is not something that is reflected in this document.

Please feel free to debate the changes!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,175 +1,123 @@
 # Changelog
 
-<!--
+All notable changes to this project will be documented in this file.
 
-This should be updated on every PR.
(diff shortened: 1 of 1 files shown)


jsjoeio (approved):
I'm all for this! I agree, they can find the user in the PR (and PRs have multiple authors as you mentioned). Seems like a step forward for us! 🎉

Issue #4479, opened 2021-11-10
Replace our token handling with VS Code's native token


jsjoeio:
Introduced in #4414

jsjoeio:
@code-asher do you think we need to do this before we cut a release? Wondering if we should move to On Deck and do later.

code-asher:
No, I think this is low priority.

PR #4499, opened 2021-11-11
Spawn vscode on demand

### LINKED ISSUE #4481: Spawn VS Code on demand to avoid overhead
Introduced in #4414

@code-asher may have more details.

TeffenEllis:
Fixes #4481

### DIFF
--- a/src/node/routes/index.ts
+++ b/src/node/routes/index.ts
@@ -10,7 +10,7 @@ import { HttpCode, HttpError } from "../../common/http"
 import { plural } from "../../common/util"
 import { App } from "../app"
 import { AuthType, DefaultedArgs } from "../cli"
-import { commit, isDevMode, rootPath } from "../constants"
+import { commit, rootPath } from "../constants"
 import { Heart } from "../heart"
 import { ensureAuthenticated, redirect } from "../http"
 import { PluginAPI } from "../plugin"
@@ -23,7 +23,7 @@ import * as login from "./login"
 import * as logout from "./logout"
 import * as pathProxy from "./pathProxy"
 import * as update from "./update"
-import { createVSServerRouter, VSServerResult } from "./vscode"
+import { CodeServerRouteWrapper } from "./vscode"
 
 /**
  * Register all routes and middleware.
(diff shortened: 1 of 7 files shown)


code-asher (approved):
Thank you for doing this!  I know it sucks to maintain legacy behavior.  Once plugins are removed we can revert this behavior as well if we want. :smile:

src/node/routes/index.ts
```diff
@@ -138,20 +138,12 @@ export const register = async (app: App, args: DefaultedArgs): Promise<Disposabl
...
-      logger.warn("VS Server router may still be compiling.")
-    } else {
-      throw error
-    }
+  const vsServerRouteHandler = new CodeServerRouteWrapper()
+
+  // Note that the root route is replaced in Coder Enterprise by the plugin API.
+  for (const routePrefix of ["/", "/vscode"]) {
+    app.router.use(routePrefix, vsServerRouteHandler.router)
+    app.wsRouter.use(routePrefix, vsServerRouteHandler.wsRouter)
```
jsjoeio, line 146: so much cleaner 👏

src/node/routes/vscode.ts
```diff
@@ -1,65 +1,105 @@
...
     }
+
     next()
-  })
+  }
 
-  router.all("*", ensureAuthenticated, (req, res, next) => {
-    req.on("error", (error: any) => {
+  private $proxyRequest: express.Handler = async (req, res, next) => {
+    // We allow certain errors to propagate so that other routers may handle requests
```
jsjoeio, line 37: 👏

❓ why start the name with `$`? is that some sort of special pattern?

src/node/routes/vscode.ts
```diff
@@ -1,65 +1,105 @@
...
+     * @file ../../../vendor/modules/code-oss-dev/src/vs/server/main.js
+     */
+    const createVSServer = await loadAMDModule<CodeServerLib.CreateServer>(
+      "vs/server/remoteExtensionHostAgent",
+      "createServer",
+    )
+
+    try {
+      this._codeServerMain = await createVSServer(null, {
+        connectionToken: "0000",
```
jsjoeio, line 80: ❓ What's the `connectionToken` used for again?

jsjoeio (approved):
Nice work!

code-asher:
One observation (this is not from this PR since I remember seeing this before) but I get a 404 when VS Code is still loading rather than the "VS Code may still be compiling" error.

src/node/routes/vscode.ts
TeffenEllis: It’s something of a convention I’ve seen help with readability, similarly to prefixing private methods with `_`. Prefixing a method like `$proxyRequest` helps indicate a unique purpose compared to a more lengthy `proxyRequestHandler`

src/node/routes/vscode.ts
TeffenEllis: Connection token is used by VS Code as a basic form of client-side security. At the moment we’ve disabled its usage, but I think it’s worth looking at again as upstream matures the feature.

code-asher: Related: https://github.com/cdr/code-server/issues/4479

jsjoeio (approved):

PR #4511, opened 2021-11-15
feat(testing): refactor humanPath and add tests

jsjoeio:
This PR refactors `humanPath` to be pure (i.e. no side effects) and it adds two tests for the function.

Fixes N/A

#TestingMondays


jawnsy (approved):
Nice!

PR #4520, opened 2021-11-16
Fix watcher not restarting code-server on VS Code compilation

code-asher:
The "Starting watch-client" string no longer appears in the latest build
output.  We could look for "Finished compilation with" to avoid
restarting when other tasks restart (since they also include the name
i.e. "Finished compilation extensions with 0 errors") but I figure we
might as well restart code-server when any compilation task completes in
case other tasks include changes that need to be reloaded.


jsjoeio (approved):
Nice cleanup!

PR #4542, opened 2021-11-22
feat: add test for onLine throw error

jsjoeio:
This PR adds an additional test for the `onLine` util function to cover the scenario where the child.Process passed to the function is missing the `stdout` method. This adds code coverage for [this line](https://github.com/cdr/code-server/blob/main/src/node/util.ts#L36).

## Screenshot

image

## References
- ["If the child was spawned with stdio set to anything other than 'pipe', then subprocess.stdout will be null."](https://stackoverflow.com/a/46024006/3015595)
- [Child Process - subprocess stdout - Node.js docs](https://nodejs.org/api/child_process.html#child_process_subprocess_stdout)
- [Child Process - options `stdio` - Node.js docs](https://nodejs.org/api/child_process.html#child_process_options_stdio)

Fixes N/A
#TestingMondays

Blocked by https://github.com/cdr/code-server/pull/4543

### DIFF
--- a/test/unit/node/util.test.ts
+++ b/test/unit/node/util.test.ts
@@ -447,6 +447,28 @@ describe("onLine", () => {
 
     expect(await received).toEqual(expected)
   })
+
+  describe("used with a process missing stdout ", () => {
+    it("should throw an error", async () => {
+      // Initialize a process that does not have stdout.
+      // "If the child was spawned with stdio set to anything
+      // other than 'pipe', then subprocess.stdout will be null."
+      // Source: https://stackoverflow.com/a/46024006/3015595
+      // Other source: https://nodejs.org/api/child_process.html#child_process_subprocess_stdout
+      // NOTE@jsjoeio - I'm not sure if this actually happens though
+      // which is why I have to set proc.stdout = null
+      // a couple lines below.
+      const proc = cp.spawn("node", [], {
+        stdio: "ignore",
+      })
+      const mockCallback = jest.fn()
+
+      expect(() => util.onLine(proc, mockCallback)).toThrowError(/stdout/)
+
+      // Cleanup
+      proc?.kill()
+    })
+  })
 })
 
 describe("escapeHtml", () => {


test/unit/node/util.test.ts
```diff
@@ -447,6 +447,33 @@ describe("onLine", () => {
...
+      // a couple lines below.
+      const proc = cp.spawn("node", [], {
+        stdio: [null],
+      })
+      const mockCallback = jest.fn()
+
+      // @ts-expect-error TypeScript knows proc.stdout should return a Readable
+      // We overwrite it so that we can get testing coverage for the scenario
+      // in onLine that would throw an error.
+      proc.stdout = null
```
jsjoeio, line 469: According to the docs, it doesn't seem like I should have to do this but no matter what values I pass in the array to `stdio` it seems like `proc.stdout` is always set? I would love a second opinion here cc @vapurrmaid @bryphe-coder

bryphe-coder: I think you might need different settings for `cp.spawn` - checking

bryphe-coder: If I use the `stdio` flag `'ignore'`, it seems to work as you expect (in order to exercise the code under test):

```
const proc = cp.spawn("node", [], {
   stdio: 'ignore'
})
```

bryphe-coder: At least, tested this out in the NodeJS console: 
image

jsjoeio: Here is a quote from the [v14 docs](https://nodejs.org/docs/latest-v14.x/api/child_process.html#child_process_subprocess_stdout):
> The `subprocess.stdout` property can be `null` if the child process could not be successfully spawned.

Seems like that's what this person was testing for. I'm not sure how to simulate this though (also wondering if it's worth the effort/time since my current implementation simulates it in _one way_ (albeit hacky).

bryphe-coder: Some useful docs around the various settings for `options.stdio` here: https://nodejs.org/api/child_process.html#optionsstdio

jsjoeio: Oh wow! Nice work 🙌 Okay I'll try that locally and see if it works for me as well.

bryphe-coder (approved):
Curious if make the change to use `stdio: [null]` -> `stdio: 'ignore'` works for you, in order to remove the manual setting of `proc.stdout`.

Otherwise, LGTM 👍

test/unit/node/util.test.ts
```diff
@@ -447,6 +447,33 @@ describe("onLine", () => {
...
+      // Initialize a process that does not have stdout.
+      // "If the child was spawned with stdio set to anything
+      // other than 'pipe', then subprocess.stdout will be null."
+      // Source: https://stackoverflow.com/a/46024006/3015595
+      // Other source: https://nodejs.org/api/child_process.html#child_process_subprocess_stdout
+      // NOTE@jsjoeio - I'm not sure if this actually happens though
+      // which is why I have to set proc.stdout = null
+      // a couple lines below.
+      const proc = cp.spawn("node", [], {
+        stdio: [null],
```
vapurrmaid, line 462: Does

```
stdio: [0, "empty"]
```
work?

jsjoeio: image

Looks like TS doesn't like this but `stdio: 'ignore'` worked!

test/unit/node/util.test.ts
vapurrmaid: this one is better than my suggestion - glad it worked!

test/unit/node/util.test.ts
vapurrmaid: Nice!

vapurrmaid (approved):

PR #4548, opened 2021-11-24
Proxy path fixes

### LINKED ISSUE #4503: Logout does not work behind reverse proxy
_Updated by @jsjoeio_

## What's the problem I'm solving?

With the [latest changes](https://github.com/cdr/code-server/pull/4414) to use the upstream server in code-server, we're now running into issues with our reverse proxy not using the correct URL for the Log out functionality.. The specific code we're talking about lives [here](https://github.com/cdr/vscode/blob/main/src/vs/server/webClientServer.ts#L317).

When using a reverse proxy (i.e. [Caddy](https://caddyserver.com/)/[NGINX](https://nginx.org/en/)), the /logout endpoint is incorrect because it has the wrong base path and protocol.

## What are some ways to go about it?

The most ideal way to solve this is to use _relative endpoints_. This way, code-server doesn't care about the base path or protocol. It only knows and uses relative paths.

## How will things change/what is the need for maintenance?

I am not sure if there is historical context I may be missing which would explain why we didn't use a relative path from the start (or rather why upstream didn't). Patching this in `vscode` means it's something we'll need to keep an eye on as we continue staying in line with upstream.

I can't imagine this introducing any breaking changes but it could potentially break in specific environments if they don't support using relative paths for some reason 🤔 (though none specifically come to mind).

## How to reproduce

Asher outlined steps below to reproduce this issue locally. To do so, follow these steps:
1. install [Caddy](https://caddyserver.com/) (`brew install caddy`)
2. start code-server locally with `yarn watch`
3. serve it at as base path like `/code` with Caddy
```Caddyfile
http://localhost:8082/code/* {
	uri strip_prefix /code
	reverse_proxy 127.0.0.1:8080
}
```
4. navigate to [http://localhost:8082/code/](http://localhost:8082/code/) (trailing slash is important!) 
5. Open the Menu and click Logout

### Expected

You are redirected to `http://mydomain.com/logout`

### Actual 

You should be redirected to `http://mydomain.com/code/logout`

Should be similar to the fix for [this issue](https://github.com/cdr/code-server/issues/4476). See PR [here](https://github.com/cdr/vscode/pull/17).

TeffenEllis:
This PR addresses several issues surrounding using reverse proxies with code-server. 

Fixed #4503

See https://github.com/cdr/vscode/commit/5e0c6f3b95ed032e62c49101ae502a46c62ef202

- Updated vscode args to match latest upstream.
- Fixed issues surrounding trailing slashes affecting base paths.
- Updated cookie names to better match upstream's usage, debuggability.

### DIFF
--- a/src/common/http.ts
+++ b/src/common/http.ts
@@ -13,7 +13,7 @@ export enum HttpCode {
  * used in the HTTP response.
  */
 export class HttpError extends Error {
-  public constructor(message: string, public readonly status: HttpCode, public readonly details?: object) {
+  public constructor(message: string, public readonly statusCode: HttpCode, public readonly details?: object) {
     super(message)
     this.name = this.constructor.name
   }
--- a/src/common/http.ts
+++ b/src/common/http.ts
@@ -18,3 +18,7 @@ export class HttpError extends Error {
     this.name = this.constructor.name
   }
 }
+
+export enum CookieKeys {
+  Session = "code-server-session",
+}
--- a/src/node/http.ts
+++ b/src/node/http.ts
@@ -6,7 +6,7 @@ import * as net from "net"
 import path from "path"
 import qs from "qs"
 import { Disposable } from "../common/emitter"
-import { HttpCode, HttpError } from "../common/http"
+import { CookieKeys, HttpCode, HttpError } from "../common/http"
 import { normalize } from "../common/util"
 import { AuthType, DefaultedArgs } from "./cli"
 import { version as codeServerVersion } from "./constants"
@@ -62,10 +62,6 @@ export const replaceTemplates = <T extends object>(
     .replace("{{OPTIONS}}", () => escapeJSON(serverOptions))
 }
 
-export enum Cookie {
-  SessionKey = "code-server-session",
-}
-
 /**
  * Throw an error if not authorized. Call `next` if provided.
  */
@@ -97,7 +93,7 @@ export const authenticated = async (req: express.Request): Promise<boolean> => {
       const passwordMethod = getPasswordMethod(hashedPasswordFromArgs)
       const isCookieValidArgs: IsCookieValidArgs = {
         passwordMethod,
-        cookieKey: sanitizeString(req.cookies[Cookie.SessionKey]),
+        cookieKey: sanitizeString(req.cookies[CookieKeys.Session]),
         passwordFromArgs: req.args.password || "",
         hashedPasswordFromArgs: req.args["hashed-password"],
       }
--- a/src/node/http.ts
+++ b/src/node/http.ts
@@ -62,6 +62,10 @@ export const replaceTemplates = <T extends object>(
     .replace("{{OPTIONS}}", () => escapeJSON(serverOptions))
 }
 
+export enum Cookie {
+  SessionKey = "code-server-session",
+}
+
 /**
  * Throw an error if not authorized. Call `next` if provided.
  */
@@ -93,7 +97,7 @@ export const authenticated = async (req: express.Request): Promise<boolean> => {
       const passwordMethod = getPasswordMethod(hashedPasswordFromArgs)
       const isCookieValidArgs: IsCookieValidArgs = {
         passwordMethod,
-        cookieKey: sanitizeString(req.cookies.key),
+        cookieKey: sanitizeString(req.cookies[Cookie.SessionKey]),
         passwordFromArgs: req.args.password || "",
         hashedPasswordFromArgs: req.args["hashed-password"],
       }
--- a/src/node/routes/login.ts
+++ b/src/node/routes/login.ts
@@ -4,12 +4,8 @@ import { RateLimiter as Limiter } from "limiter"
 import * as os from "os"
 import * as path from "path"
 import { rootPath } from "../constants"
-import { authenticated, getCookieDomain, redirect, replaceTemplates } from "../http"
+import { authenticated, Cookie, getCookieDomain, redirect, replaceTemplates } from "../http"
 import { getPasswordMethod, handlePasswordValidation, humanPath, sanitizeString, escapeHtml } from "../util"
-
-export enum Cookie {
-  Key = "key",
-}
 
 // RateLimiter wraps around the limiter library for logins.
 // It allows 2 logins every minute plus 12 logins every hour.
@@ -62,7 +58,7 @@ router.get("/", async (req, res) => {
   res.send(await getRoot(req))
 })
 
-router.post("/", async (req, res) => {
+router.post<{}, string, { password: string; base?: string }, { to?: string }>("/", async (req, res) => {
   const password = sanitizeString(req.body.password)
   const hashedPasswordFromArgs = req.args["hashed-password"]
 
@@ -87,13 +83,13 @@ router.post("/", async (req, res) => {
     if (isPasswordValid) {
       // The hash does not add any actual security but we do it for
       // obfuscation purposes (and as a side effect it handles escaping).
-      res.cookie(Cookie.Key, hashedPassword, {
+      res.cookie(Cookie.SessionKey, hashedPassword, {
         domain: getCookieDomain(req.headers.host || "", req.args["proxy-domain"]),
         // Browsers do not appear to allow cookies to be set relatively so we
         // need to get the root path from the browser since the proxy rewrites
         // it out of the path.  Otherwise code-server instances hosted on
         // separate sub-paths will clobber each other.
-        path: req.body.base ? path.posix.join(req.body.base, "..") : "/",
+        path: req.body.base ? path.posix.join(req.body.base, "..", "/") : "/",
         sameSite: "lax",
       })
 
--- a/src/node/routes/login.ts
+++ b/src/node/routes/login.ts
@@ -3,8 +3,9 @@ import { promises as fs } from "fs"
 import { RateLimiter as Limiter } from "limiter"
 import * as os from "os"
 import * as path from "path"
+import { CookieKeys } from "../../common/http"
 import { rootPath } from "../constants"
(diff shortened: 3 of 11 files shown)


src/common/http.ts
```diff
@@ -13,7 +13,7 @@ export enum HttpCode {
  * used in the HTTP response.
  */
 export class HttpError extends Error {
-  public constructor(message: string, public readonly status: HttpCode, public readonly details?: object) {
+  public constructor(message: string, public readonly statusCode: HttpCode, public readonly details?: object) {
```
TeffenEllis, line 16: Matches our same convention elsewhere
https://github.com/cdr/code-server/blob/main/src/node/routes/errors.ts#L31

src/node/http.ts
```diff
@@ -62,6 +62,10 @@ export const replaceTemplates = <T extends object>(
     .replace("{{OPTIONS}}", () => escapeJSON(serverOptions))
 }
 
+export enum Cookie {
+  SessionKey = "code-server-session",
```
TeffenEllis, line 66: Renamed from ‘key’ in the browser to assist in debugging, both from us and users.

src/node/routes/login.ts
```diff
@@ -62,7 +58,7 @@ router.get("/", async (req, res) => {
   res.send(await getRoot(req))
 })
 
-router.post("/", async (req, res) => {
+router.post<{}, string, { password: string; base?: string }, { to?: string }>("/", async (req, res) => {
```
TeffenEllis, line 61: Added types

src/node/routes/login.ts
```diff
@@ -87,13 +83,13 @@ router.post("/", async (req, res) => {
...
       // obfuscation purposes (and as a side effect it handles escaping).
-      res.cookie(Cookie.Key, hashedPassword, {
+      res.cookie(Cookie.SessionKey, hashedPassword, {
         domain: getCookieDomain(req.headers.host || "", req.args["proxy-domain"]),
         // Browsers do not appear to allow cookies to be set relatively so we
         // need to get the root path from the browser since the proxy rewrites
         // it out of the path.  Otherwise code-server instances hosted on
         // separate sub-paths will clobber each other.
-        path: req.body.base ? path.posix.join(req.body.base, "..") : "/",
+        path: req.body.base ? path.posix.join(req.body.base, "..", "/") : "/",
```
TeffenEllis, line 92: Previously this would assign the cookie to a resource path instead of a route path (e.g. `/code` instead of `/code/`). From what I’ve researched, there’s a [number of opinions](https://stackoverflow.com/questions/5948659/when-should-i-use-a-trailing-slash-in-my-url) in [both camps](https://cdivilly.wordpress.com/2014/03/11/why-trailing-slashes-on-uris-are-important/). I believe that since we’re now leaning so heavily on relative routes in VS Code, there’s little we can do to avoid this aside from convenient redirects and providing guidance for users.

- code-server may be accessed either at the root or at /vscode with no slash
- If you’re using a proxy that includes a path, it must end in a slash
- If you’d like to omit the slash with your proxy, this must be configured on your end

src/node/routes/logout.ts
```diff
@@ -1,17 +1,19 @@
...
+router.get<{}, undefined, undefined, { base?: string; to?: string }>("/", async (req, res) => {
+  const path = sanitizeString(req.query.base) || "/"
+  const to = sanitizeString(req.query.to) || "/"
+
   // Must use the *identical* properties used to set the cookie.
-  res.clearCookie(Cookie.Key, {
+  res.clearCookie(Cookie.SessionKey, {
     domain: getCookieDomain(req.headers.host || "", req.args["proxy-domain"]),
-    path: req.query.base || "/",
+    path: decodeURIComponent(path),
```
TeffenEllis, line 14: This is encoded on the VS Code side:

https://github.com/cdr/vscode/blob/5e0c6f3b95ed032e62c49101ae502a46c62ef202/src/vs/workbench/browser/client.ts#L184-L197

src/node/routes/vscode.ts
```diff
@@ -24,7 +24,7 @@ export class CodeServerRouteWrapper {
     const isAuthenticated = await authenticated(req)
 
     if (!isAuthenticated) {
-      return redirect(req, res, "login", {
+      return redirect(req, res, "login/", {
```
TeffenEllis, line 27: Note that the user won’t have to provide a trailing slash as we setup the redirect on their behalf.

src/node/routes/login.ts
jsjoeio: > If you’d like to omit the slash with your proxy, this must be configured on your end

I feel like this is a fair tradeoff but I didn't use proxies with code-server before so can't fully speak to that UX. @code-asher what are your thoughts?

jsjoeio (commented):
This looks good to me 👍 but I'll defer to @code-asher to give the final approval

src/node/routes/logout.ts
code-asher: I think this path has to be the same as the other cookie in order to overwrite it?  So I think we will need to apply the same logic here we did in the login, namely the whole `path.posix.join()` bit.

I suppose we will want `..` if browsing at `/vscode/` and `.` otherwise.

src/node/routes/login.ts
TeffenEllis: The issue with omitting the trailing slash from the cookie is that logging out from any arbitrary path requires that we send that very same path in the logout request. Trailing slashes denote the depth and access of a cookie, and while the spec is pretty forgiving on cookies to omit it, we can’t identify what part of the URL is where authenticated “starts” without knowing if the user first authenticated from `/ ` or `/code`. Both share the same path base `/` while adding a trailing slash defines cookies from `/code/`.

code-asher: Ah so a cookie set to the path `/code` actually scopes the cookie to `/`?  That does seem like a good reason to add the trailing slash.

code-asher: Hmm the docs here suggest that is not the case: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#define_where_cookies_are_sent

To quote the relevant portion:

For example, if you set Path=/docs, these request paths match:

    /docs
    /docs/
    /docs/Web/
    /docs/Web/HTTP

But these request paths don't:

    /
    /docsets
    /fr/docs

TeffenEllis: That’s correct, the spec lets the trailing slash be omitted however we still can’t know which depth they’re attempting to login or out of without a trailing slash since the URL spec is more strict.

code-asher: We need the depth either way right?

For example assuming the base path is `/apps/coder/code-server/` when we log out from `/apps/coder/code-server/vscode/` we now get redirected to `/apps/coder/code-server/logout?base=/apps/coder/code-server/vscode/` and from there we need to set the cookie path using `path.posix.join(base, "..")` or `path.posix.join(base, "..", "/")` (only difference is the trailing slash).

If we log out from `/apps/coder/code-server/` then we need to do the same thing in both cases except without a `..`.  If there was some future route like `/apps/coder/code-server/my/new/route/` we would need `../../..` in both cases.

Maybe there is some way we can get rid of needing to know what the current depth is?  Hmmmm... :thinking: 

In previous versions of code-server the way we made this work was to pass the relative path to the root to the frontend (so in the three examples above it would be `..`, `.`, and `../../..` respectively).  The client then resolved this against its current URL (in all three cases it would resolve to `/apps/coder/code-server`) and then we would redirect to `logout?base=/apps/coder/code-server` and we could set the path directly to that query variable since it was already resolved to the root.

code-asher: Maybe where we pass things like `logoutEndpoint` we can also include the current route so we can then derive the correct base.

src/node/routes/logout.ts
jsjoeio: I'm guessing we'll need to test that to verify. If you're correct, I think one of us will have to implement that then @code-asher. Do you want to or do you want me to?

src/node/routes/login.ts
jsjoeio: To summarize, @code-asher you're saying this change would break current behavior?

I think you and I will need to test and verify this ourselves. If it does lead to a breaking change, we should see what our options are like you said.

src/node/routes/logout.ts
code-asher: Yeah I think if you go to `http://localhost:8080/vscode/` and click "log out" then it will (un)set the cookie at `/vscode/` which will not log you out since the cookie would have been initially set at `/`.

If you get to it before I do feel free but otherwise I will take it on!

src/node/routes/login.ts
code-asher: Sort of.  This is mostly academic right now since you cannot host code-server at a base path without a trailing slash anyway because of the way browsers resolve relative paths (i.e. `/code` does not work because we would need the asset paths to be `./code/path/to/asset` which means we need to get `code` from somewhere which means we need to ask the user for the base path or we need to inject the asset paths on the client).  Current code-server does not work without a trailing slash (on the base path, it does work on our own routes like `/vscode` of course---I think I misspoke about this in Slack the other day so hopefully this clarifies).

There is one area we could potentially be breaking which is if existing users have a cookie at `/code` and we change the logic to use `/code/` then they can no longer log out.  But since we are also changing the cookie name that point is also moot.

I do still think we should avoid an arbitrary restriction though.  If one day we decide to support a slash-less base path it would be nice if the cookie logic already works.  So it is really just more about leaving support as broad as possible as long as there are no downsides in doing so.

code-asher: That is just this login portion though.  My comments about the logout stuff are still relevant though and I think need to be implemented.

code-asher (approved):
I am going to merge this in since I believe the current behavior has the same issues.  We can test/fix in another PR!

Issue #2734, opened 2021-02-16
Unable to connect with --link


SamHep0803:
- Web Browser: Safari
- Local OS: iPadOS 14.3
- Remote OS: Ubuntu 20.04
- Remote Architecture: x86_64
- `code-server --version`: 3.9.0 fc6d123da59a4e5a675ac8e080f66e032ba01a1b

When running `code-server --link`, the process starts successfully however when you go to the link given in the server output, the error `Error connecting to your code-server instance` shows up when you go to the site.

Initially, no error gets thrown in the terminal however after around 1 hour, 
`[2021-02-16T09:59:22.641Z] info 2021/02/16 09:59:22 [ERR] yamux: Failed to read header: failed to get reader: failed to read frame header: EOF`
`[2021-02-16T09:59:22.645Z] info 2021-02-16 09:59:22.644 [WARN] <proxy.go:48> code-server proxy exited {"error": "accept tcp [::]:36221: use of closed network connection"}
`
is logged.

Thanks, Sam.

SamHep0803:
Downgrading the version 3.8.0 did not help resolve the issue, I have a feeling this may be an issue on cdr's end. I have tried getting this to work in both GCP and DigitalOcean to no avail.

oxy:
Tried to repro this but could not - do you have any additional steps that could help in reproducing it?

SamHep0803:
Honestly not really, it's a very strange issue. 

I installed `code-server` with the install script, as is the recommended way in the docs and that was basically it. I have to mention that it was working 2 days ago but now it's not.

SamHep0803:
I have tried deleting the `~/.config/coder-cloud` folder and restarting the server and that hasn't worked either.

SamHep0803:
Update: after turning on my VPN, I seem to be able to connect to the server now, however I don't see why I would need to do that.

code-asher:
Did it give you an error ID? We can use that to investigate further.

parslov:
I have the same problem

- brand new AWS instance
- AWS free tier 
- Ubuntu 20.04
- Installed with code-server with this (from https://github.com/cdr/code-server): curl -fsSL https://code-server.dev/install.sh | sh
- Opened HTTP, HTTPS, Custom TCP (pot 8080), and SSH in inbound rules
- Started code-server with: code-server --link
- ...and then opened Safari on my mac and entered the https-address shown 
- ...and then get: 'Error connecting to your code-server instance'

Any ideas?

[...]

PR #4562, opened 2021-11-29
chore: add deprecation notice for --link

bpmct:
`code-server --link` has not been updated for a while, and some users have reported connection issues with it (see #2734)

While we will continue to host the --link servers, this PR adds a warning that we are not officially maintaining this version of --link. In the future, we will likely release an improved version, using a new architecture and codebase. However, there is no timeline for this.


jsjoeio (approved):
Thanks for taking this on @bpmct! 🙌

Issue #3051, opened 2021-04-06
Regression: New RegExp format causes blank page


L3P3:
## OS/Web Information

- Web Browser: Mypal 29.1.1 (Pale Moon/Firefox)
- Local OS: Windows XP SP3
- Remote OS: Debian 10
- Remote Architecture: x86-64
- `code-server --version`: 3.9.2 109d2ce3247869eaeab67aa7e5423503ec9eb859

## Steps to Reproduce

1. Open Mypal
2. Open code-server

## Expected

App opens.

## Actual

Blank page.

## Screenshot

![mpcsb2](https://user-images.githubusercontent.com/4629449/113763605-616d8780-9709-11eb-98b8-b28dc17e0a4b.PNG)

## Notes

Is most probably caused by following lines including quite recent RegExp syntax:

https://github.com/cdr/code-server/blob/f21433a130ab07db569a1ae43a04416e7ec306ba/lib/vscode/src/vs/editor/contrib/linesOperations/linesOperations.ts#L1060

Please either revert these lines (v3.9.0 it worked well) or make them more backwards-compatible (maybe feature check and fallback). Maybe, we can also consider this a bug in closure compiler as it should generate backwards-compatible code out of the box.

jsjoeio:
Hi @L3P3! Thanks for opening the issue. 

> This issue can be reproduced in VS Code: Yes/No

It looks like part of the bug report is missing this line. Can this be reproduced in VS Code? 

Based on what you linked, if that's truly the issue, then it looks like it was introduced/changed here: https://github.com/microsoft/vscode/commit/5c1ff8158b7731a8478cfe0e09a23e59502efb9f#diff-dde3b56145713053f086e5ce1fcea39c4cb171d4555b6ab91256c7cf2b39e9af

We suggest raising this upstream.

> Maybe, we can also consider this a bug in closure compiler as it should generate backwards-compatible code out of the box.

If that's the case, it may be worth raising this there.

L3P3:
Thanks for pointing that out. Did not know that vscode aims for backwards-compatiblity and I thought it is just used in modern electron.
Workaround for now is to simply delete SnakeCaseAction, I will now look into the compiler side of things.

L3P3:
Just for the record, if anyone else has this problem, run this command as root:
```
[...]

Issue #2826, opened 2021-03-06
Clicking issues in text area


L3P3:
- Web Browser: Mypal 29.0.1 (Pale Moon/Firefox)
- Local OS: Windows XP SP3
- Remote OS: Debian 10
- Remote Architecture: x86-64
- `code-server --version`: 3.9.0 fc6d123da59a4e5a675ac8e080f66e032ba01a1b

1. When right-clicking in the text area (and only there), the context menu does not appear. Following error is thrown:
```
  ERR this.shadowRootHostElement.attachShadow is not a function: setContainer@out/vs/workbench/workbench.web.api.js:448:617
setContainer@out/vs/workbench/workbench.web.api.js:1724:79
showContextView@out/vs/workbench/workbench.web.api.js:1724:180
showContextMenu@out/vs/workbench/workbench.web.api.js:2171:372
showContextMenu@out/vs/workbench/workbench.web.api.js:2524:362
_doShowContextMenu@out/vs/workbench/workbench.web.api.js:2036:481
showContextMenu@out/vs/workbench/workbench.web.api.js:2035:614
_onContextMenu@out/vs/workbench/workbench.web.api.js:2035:347
e/<@out/vs/workbench/workbench.web.api.js:2034:213
fire@out/vs/workbench/workbench.web.api.js:99:694
_createView/i.onContextMenu@out/vs/workbench/workbench.web.api.js:2357:766
emitContextMenu@out/vs/workbench/workbench.web.api.js:2215:957
emitContextMenu@out/vs/workbench/workbench.web.api.js:2007:54
_onContextMenu@out/vs/workbench/workbench.web.api.js:2202:979
p/<@out/vs/workbench/workbench.web.api.js:2201:318
onContextMenu/<@out/vs/workbench/workbench.web.api.js:479:433
```

2. When clicking on empty lines, nothing happens (cursor does not jump to that line), no error is thrown. Clicking on lines containing text behaves correctly. Is quite annoying, have to always click on nearest populated line and navigate down/up via keyboard.

jsjoeio:
Hmm... can you provide reproduction steps? I don't have a Windows XP machine around and I haven't used Mypal but willing to try and test on my machine locally.

code-asher:
I don't think this is something we'd patch here but rather would be
better to raise upstream (but best to see if it replicates in Codespaces
first to prove it's an issue for upstream).

Issue #2825, opened 2021-03-06
Regression: Syntax highlighting depends on WebAssembly


L3P3:
- Web Browser: Mypal 29.0.1 (Pale Moon/Firefox)
- Local OS: Windows XP SP3
- Remote OS: Debian 10
- Remote Architecture: x86-64
- `code-server --version`: 3.9.0 fc6d123da59a4e5a675ac8e080f66e032ba01a1b

With the previous code-server version, I was very productive with my favorite IDE on my favorite device which is still my daily driver for most things.
Now, syntax highlighting does not work anymore when wasm is not available.
In my opinion, we should make it possible to run code-server without wasm but with pure JS. There is a wasm-to-js polyfill available but I think it is better to just keep the JS version of things. I don't know how that wasm is generated by but usually, all wasm compilers also output asm.js which should work flawless too. Fixing this issue might be as easy as enabling asm.js fallback in compilation settings, which causes no overhead for wasm-enabled browsers but would also enable the others.
Hightlighting worked perfectly before, even with files with 10k of lines and I see no good reason for using wasm without fallbacks, it feels like cheating on the last steps to me, like using adobe flash or java applet just for one part of the app while 99% is done in JS. Then why not using java in the first place...
And I am honestly ready to pay for you to keep supporting bare JS environments.

I just found out that Mypal does in fact _experimentally_ support wasm, using about:config. Syntax highlighting works again. Phew! But it still scares me that it is required... Consider this issue as a theoretical debate or flexing with Windows XP.

![code-server-wasm](https://user-images.githubusercontent.com/4629449/110198337-00c10580-7e52-11eb-97e0-9789994cb0e3.PNG)

Full console history:
```
register.js:14:404: [Service Worker] registered 
workbench.web.api.js:4801:76: The web worker extension host is started without an iframe sandbox!
main.js:1:6795: no native wasm support detected

  ERR WebAssembly is not defined: r</<@node_modules/vscode-oniguruma/release/main.js:1:6836
t.loadWASM/<@node_modules/vscode-oniguruma/release/main.js:1:5293
t.loadWASM@node_modules/vscode-oniguruma/release/main.js:1:5275
_doGetVSCodeOniguruma@out/vs/workbench/workbench.web.api.js:5724:882
```

jsjoeio:
WOW! I have not seen the Windows XP UI in a long time.

Not sure what could have changed in `3.9.0` that would cause this but will keep that in mind.

> Syntax highlighting works again

[...]

PR #4579, opened 2021-12-05
Support browsers from before 2020

L3P3:
As reported in #2825, #2826 and #3051, almost everything works in older browsers. This setting here prevented me from updating and I think it is an obvious enhancement to not artificially restrict to browsers from 2020+. There should not be any noticable downsides of this change since es6 and es2020 are pretty similar with only minor differences.


jawnsy:
I guess [upstream vscode uses es2020 as the target](https://github.com/microsoft/vscode/blob/13c0ed811e7e6bf2a961b2ab8b6d086fa7822fb4/src/tsconfig.json#L8), so one risk is that divergence means that the result may not be as well-tested. I don't know enough to have an accurate assessment of the risk though.

jsjoeio:
> I guess upstream vscode uses es2020 as the target, so one risk is that divergence means that the result may not be as well-tested. I don't know enough to have an accurate assessment of the risk though.

Great point! I'm guessing we decided to use `es2020` after they upgraded. Looks like they dropped support for legacy edge and IE when they made [that upgrade](https://github.com/microsoft/vscode/issues/121144).

code-asher:
We can change our target but code-server will still not work since VS
Code has their own target.  I am not sure we even have any of our own
client-side JavaScript anymore in 4.0.0.

jsjoeio:
> We can change our target but code-server will still not work since VS
Code has their own target.  

Ah right! This would have to be in VS Code itself (or in [our fork](https://github.com/cdr/vscode)). But if we change in our fork, that would diverge from upstream and could cause other issues.

> I am not sure we even have any of our own
client-side JavaScript anymore in 4.0.0.

Wow, you're right. I didn't even realize that.

L3P3:
As stated, between es6 and es2020, there are no interoperability risks. In this case, they only differ in syntactic sugar. Not even polyfills required to ensure compatiblity between two modules.
Isn't this tsconfig also respected in submodules? I think it should be at least tested if that works. Or are you sure that there will still be es2020 code after this proposed change?

L3P3:
> I am not sure we even have any of our own client-side JavaScript anymore in 4.0.0.

What about the forked repository? Isn't that deemed "our own"?

code-asher:
VS Code is compiled separately using their build process so our tsconfig
has no effect on it.

But if we patch it there instead and everything works that seems like a
win to me.

We could also leave this change here as well in case we ever add
client-side JavaScript again.

Our tests are reporting the following with this change:

```
  test/unit/node/cli.test.ts:91:11 - error TS2550: Property 'flat' does not exist on type '(string | string[])[]'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2019' or later
```

So maybe we also need to add `lib`.

L3P3:
> So maybe we also need to add `lib`.

Yes, sure. Those are polyfills so that beside syntax change, also new methods work in older browsers. :wink:

jsjoeio:
> But if we patch it there instead and everything works that seems like a
win to me.

Good point! I agree. @code-asher should we ask @L3P3 to submit a PR to our fork [here](https://github.com/cdr/vscode/blob/main/src/tsconfig.json#L8)?

> We could also leave this change here as well in case we ever add
client-side JavaScript again.

Agreed. As long as the build passes, I think we're okay to merge.

jsjoeio (approved):
Even though we don't have any client-side JS in code-server at the moment, we're okay merging this in, in case we do want to support older browser in the future.

Thanks @L3P3 for adding this 👍

jsjoeio:
> Docs preview

P.S. you can ignore this. It's a known issue that I need to fix.

L3P3:
Sounds great. Done.

L3P3:
BTW if anyone wonders why they targeted es2020 in the first place: They use electron which is pretty much latest chrome. Web development and electron development has some differences. :wink:

jsjoeio:
```javascript
FAIL test/unit/node/routes/login.test.ts
  ● Test suite failed to run

    test/unit/node/routes/login.test.ts:81:9 - error TS2322: Type 'URLSearchParams' is not assignable to type 'BodyInit | undefined'.
      Type 'URLSearchParams' is missing the following properties from type 'URLSearchParams': entries, keys, values, [Symbol.iterator]

    81         body: params,
               ~~~~
```

Hmm... looks like one of the tests are failing. You might need to look into fixing `login.test.ts`

jsjoeio:
I still see the same error in the test [here](https://github.com/cdr/code-server/runs/4462927946?check_suite_focus=true#step:9:548). We'll push this for the next release.

L3P3:
Sorry, I thought that dom includes dom.iterable but it appears to not be the case. Just another try and hopefully no error. :see_no_evil:

PR #4589, opened 2021-12-07
Remove excessive reloading before VS Code is ready

code-asher:
The watch script was reloading the web server after every extension
compilation which is not necessary plus VS Code will not even be ready
at that point anyway.

Instead restart when the main compilation is finished.  The string to
match with includes a "with" because otherwise it would match "Finished
compilation extensions" which is not the main compilation task where we
actually need to restart the web server.

I also replaced this.log with console.log because the former does not
include a newline and it appears we want newlines with all
these (otherwise the next log starts on the same line which looks odd).

I removed the cache clean as well because the cache is meant to stay
there to speed up builds.

### DIFF
--- a/ci/dev/watch.ts
+++ b/ci/dev/watch.ts
@@ -2,7 +2,7 @@ import { spawn, fork, ChildProcess } from "child_process"
 import del from "del"
 import { promises as fs } from "fs"
 import * as path from "path"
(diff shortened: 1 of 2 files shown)


jsjoeio (approved):
Nice adjustments! I think this will be a solid improvement

Issue #4322, opened 2021-10-07
code-server spawned within code-server inherits plugin paths of parent code-server when it should not


jsjoeio:
@bryphe-coder and I found a bug in code-server running in Coder and realized it takes you to the plugins/applications page instead of showing you the editor. 

We have logic [here](https://github.com/cdr/vscode/blob/5cddbcc93d533db1282691b1f07b994b434ee0b6/src/vs/base/common/processes.ts#L114) to remove these but since Coder uses `CS_PLUGIN`, it doesn't get removed in the user's environment. 

## Todos 
- [ ] rename `CS_PLUGIN` to `CODE_SERVER_PLUGIN`
- [ ] support both for backwards compatbility

PR #4590, opened 2021-12-07
Fix spawning code-server within code-server

code-asher:
Also fixes https://github.com/cdr/code-server/issues/4322


jsjoeio (approved):
Woo!

PR #4594, opened 2021-12-08
Fix relative paths

code-asher:
Fixes https://github.com/cdr/code-server/issues/4503

### DIFF
--- a/ci/dev/watch.ts
+++ b/ci/dev/watch.ts
@@ -1,5 +1,4 @@
 import { spawn, fork, ChildProcess } from "child_process"
-import del from "del"
 import { promises as fs } from "fs"
 import * as path from "path"
 import { CompilationStats, onLine, OnLineCallback } from "../../src/node/util"
@@ -56,8 +55,6 @@ class Watcher {
     for (const event of ["SIGINT", "SIGTERM"]) {
       process.on(event, () => this.dispose(0))
     }
-
-    this.cleanFiles()
 
     for (const [processName, devProcess] of Object.entries(this.compilers)) {
       if (!devProcess) continue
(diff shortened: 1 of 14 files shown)


ci/dev/watch.ts
```diff
@@ -121,15 +118,6 @@ class Watcher {
...
 
-  /**
-   * Cleans files from previous builds.
-   */
-  private cleanFiles(): Promise<string[]> {
-    console.log("[Watcher]", "Cleaning files from previous builds...")
-
-    return del(["out/**/*"])
-  }
-
```
jsjoeio, line 132: I love when we get to delete code! 👏

src/node/http.ts
```diff
@@ -33,11 +34,11 @@ declare global {
 }
 
 export const createClientConfiguration = (req: express.Request): ClientConfiguration => {
-  const base = relativeRoot(req)
+  const base = relativeRoot(req.originalUrl)
 
   return {
-    base,
-    csStaticBase: normalize(path.posix.join(base, "_static/")),
+    base: base,
```
jsjoeio, line 40: nit: since these are the same name, we should be able to keep it `base` instead of `base: base`

src/node/http.ts
```diff
@@ -241,3 +255,36 @@ export function disposer(server: http.Server): Disposable["dispose"] {
...
+  // rewritten it.  That means we need to get the path from the frontend.
+
+  // The reason we need to set the path (as opposed to defaulting to /) is to
+  // avoid code-server instances on different sub-paths clobbering each other or
+  // from accessing each other's tokens (and to prevent other services from
+  // accessing code-server's tokens).
+
+  // When logging in or out the request must include the href (the full current
+  // URL of that page) and the relative path to the root as given to it by the
+  // backend.  Using these two we can determine the true absolute root.
```
jsjoeio, line 276: Thank you for adding this inline documentation 👏

jsjoeio (approved):
> This means the client has to do less work.

Favorite part of this whole PR 👏

PR #4599, opened 2021-12-10
Add separate function for VS Code arguments

code-asher:
The problem before was that the pop() caused the open in existing
instance functionality to break because the arguments no longer
contained the workspace or directory (files are left untouched).

We could simply remove the pop() but since `workspace` and `folder` are
not CLI arguments I think it makes sense to handle them in a separate
function which can be called at the point where they are needed.  This
also lets us de-duplicate some logic since we create these arguments in
two spots and lets us skip this logic when we do not need it.

The pop() is still avoided because manipulating a passed-in object
in-place seems like a risky move.  If we really need to do this we
should copy the positional argument array instead.

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -405,7 +405,10 @@ export const parse = (
     throw new Error("--cert-key is missing")
   }
 
-  logger.debug(() => ["parsed command line", field("args", { ...args, password: undefined })])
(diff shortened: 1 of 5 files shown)


jsjoeio (approved):
refactoring, cleanup and tests - what a lovely friday gift! nice work 😎

PR #4605, opened 2021-12-11
adding option --edge to install latest edge / preview version

Leask:
Adding option --edge to install latest edge / preview version.

```bash
$ ./install.sh --edge true
```

```bash
$ ./install.sh --edge=true
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,7 +32,7 @@ Usage:
   --version X.X.X
       Install a specific version instead of the latest.
 
-  --edge [true | false (default) ]
+  --edge
       Install the latest edge version instead of the latest stable version.
 
   --method [detect | standalone]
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -32,6 +32,9 @@ Usage:
   --version X.X.X
       Install a specific version instead of the latest.
 
+  --edge [true | false (default) ]
```
code-asher, line 35: Could we leave off the value and set this to true when the flag is set?  This is how we have been handling booleans, for example see `--dry-run`.

code-asher (dismissed):
This is a fantastic idea!

Leask:
Adding this commit: `updated arg --edge parsing to match the style of --dryrun`
https://github.com/cdr/code-server/pull/4605/commits/746f55a2b8046aea5eef0d2e13bec3a134b7d603
Thanks for the review :)

jsjoeio (approved):
Woohoo! Thanks so much

jsjoeio:
(You can ignore Docs preview - I need to fix it)

Leask:
Thanks!

Leask:
🆒

PR #4611, opened 2021-12-12
support git-lfs default

huangyoukun:
support git-lfs default

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -10,11 +10,13 @@ RUN apt-get update \
     man \
     nano \
     git \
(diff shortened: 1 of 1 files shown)


code-asher (approved):
I think adding this is reasonable, @jsjoeio thoughts?

jsjoeio:
> I think adding this is reasonable, @jsjoeio thoughts?

Agreed! I don't see any downsides.

jsjoeio (approved):

jsjoeio:
I'll merge when the rest is passing (ignore Docs preview)

PR #4633, opened 2021-12-16
Implement last opened functionality

code-asher:
Fixes https://github.com/cdr/code-server/issues/4619

This also restores the settings query file write which we were missing.


src/node/routes/vscode.ts
```diff
@@ -31,6 +32,30 @@ export class CodeServerRouteWrapper {
...
+    // Ew means the workspace was closed so clear the last folder/workspace.
+    const { query } = await settings.read()
+    if (req.query.ew) {
+      delete query.folder
+      delete query.workspace
+    }
+
+    // Redirect to the last folder/workspace if nothing else is opened.
+    if (
+      !req.query.folder &&
```
jsjoeio, line 44: note to self: would be good to write a unit test for this logic!

jsjoeio (approved):

src/node/routes/vscode.ts
code-asher: ahaha was just thinking the same thing yesterday, I am going to write one today.

jsjoeio: Twins!

code-asher:
Testing this took a bit of doing since I wanted to use the integration tests as it uses both the `req` object and redirections.

It was hard to see test failures among all the logger output so I added a logger mock and used that everywhere to get rid of the noise.

Then I noticed some accumulating test dirs so I took care of those.

I had issues with the settings from the tests stepping all over each other and conflicting with my own running instance of code-server so I had to add some separation to the instances which meant making the settings class use the path provided on the command line.

I also added some tests to make sure relative paths are being generated correctly.

src/node/routes/index.ts
```diff
@@ -47,6 +49,9 @@ export const register = async (app: App, args: DefaultedArgs): Promise<Disposabl
   app.router.use(cookieParser())
   app.wsRouter.use(cookieParser())
 
+  const settings = new SettingsProvider<CoderSettings>(path.join(args["user-data-dir"], "coder.json"))
+  const updater = new UpdateProvider("https://api.github.com/repos/cdr/code-server/releases/latest", settings)
```
jsjoeio, line 53: This looks fancy 👀 So it takes in a URL to a release and a SettingProvider and returns an updater?
```suggestion
  const updater = new UpdateProvider("https://api.github.com/repos/coder/code-server/releases/latest", settings)
```

src/node/update.ts
```diff
@@ -27,12 +27,11 @@ export class UpdateProvider {
      * The URL for getting the latest version of code-server. Should return JSON
      * that fulfills `LatestResponse`.
      */
-    private readonly latestUrl = "https://api.github.com/repos/cdr/code-server/releases/latest",
```
jsjoeio, line 30: 👏 love the cleanup here!

test/unit/node/routes/vscode.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    await resp.text()
+
+    resp = await codeServer.fetch("/")
+    url = new URL(resp.url)
+    expect(url.pathname).toBe("/")
+    expect(decodeURIComponent(url.search)).toBe("")
+    await resp.text()
+  })
+
+  it("should not redirect when last opened is ignored", async () => {
```
jsjoeio, line 139: This tests are such a nice addition 👏

jsjoeio (approved):
LGTM! Nice cleanup 👏

Approving but make sure you look at the URL provided to the `Updater`. I think it should use the new GitHub org

src/node/routes/index.ts
code-asher: Ooo nice catch.  Yup that is right.  The settings provider is used to store the last update info.

jsjoeio: Ah! I see. That makes sense

code-asher: Technically the updater only checks for updates and does not perform updates so the term `updater` might be inaccurate but I was not sure what else to call it. `updateChecker` maybe but it feels a bit long.

jsjoeio (approved):

PR #4651, opened 2021-12-18
docs(ipad): fix link to ios-ipad related issues

elsbrock:
Fixes a link to the iPad/iOS related issues that previously went into Nirvana (looks like the label was renamed).


jsjoeio (approved):
Thank you for fixing this!

O0peyDo0pey (approved):

PR #4677, opened 2022-01-04
fix(ci): add environment to docs-preview

### LINKED ISSUE #4027: Investigate changing docs preview workflow

jsjoeio:
This PR updates the `docs-preview.yaml` workflow to use an environment which should have the correct secrets for deploying a docs preview.

Fixes #4027

### DIFF
--- a/.github/workflows/docs-preview.yaml
+++ b/.github/workflows/docs-preview.yaml
@@ -21,6 +21,7 @@ jobs:
   preview:
     name: Docs preview
     runs-on: ubuntu-20.04
+    environment: CI
     steps:
       - name: Cancel Previous Runs
         uses: styfle/cancel-workflow-action@0.9.1


jsjoeio:
I'm not sure what is going wrong [here:](https://github.com/coder/code-server/runs/4704975825?check_suite_focus=true#step:3:43) 
image

I'll need to further investigate

jsjoeio:
@jawnsy and I discussed this. We're not sure why it's failing. We think merging it might fix so we're going to try that.

jawnsy (approved):

PR #4717, opened 2022-01-10
feat: add tests for node/settings.ts

jsjoeio:
This PR adds additional tests for `node/settings.ts` to ensure we log warnings when expected.

## Context
The `SettingsProvider` provides the application with access to the settings for code-server. This is then added to all routes under `req.settings` so they can easily access it. We're adding tests to ensure that our logger does indeed log warnings when something goes wrong.

### Before
image

### After
image

Fixes N/A
#TestingMondays

### DIFF
--- a/test/unit/node/settings.test.ts
+++ b/test/unit/node/settings.test.ts
@@ -1,0 +1,62 @@
+import { logger } from "@coder/logger"
+import { promises as fs } from "fs"
+import path from "path"
+import { SettingsProvider, CoderSettings } from "../../../src/node/settings"
+import { clean, mockLogger, tmpdir } from "../../utils/helpers"
+
+describe("settings", () => {
+  const testName = "settingsTests"
+  let testDir = ""
+
+  beforeAll(async () => {
+    mockLogger()
+    await clean(testName)
+    testDir = await tmpdir(testName)
+  })
+  describe("with invalid JSON in settings file", () => {
+    let mockSettingsFile = "coder.json"
+    let pathToMockSettingsFile = ""
+
+    beforeEach(async () => {
+      pathToMockSettingsFile = path.join(testDir, mockSettingsFile)
+      // Missing a quote, which makes it invalid intentionally
+      await fs.writeFile(pathToMockSettingsFile, '{"fakeKey":true,"helloWorld:"test"}')
+    })
+    afterEach(async () => {
+      jest.clearAllMocks()
+    })
+    it("should log a warning", async () => {
+      const settings = new SettingsProvider<CoderSettings>(pathToMockSettingsFile)
+      await settings.read()
+      // This happens when we can't parse a JSON (usually error in file)
+      expect(logger.warn).toHaveBeenCalledWith("Unexpected token t in JSON at position 29")
+    })
+  })
+  describe("with invalid settings file path", () => {
+    let mockSettingsFile = "nonExistent.json"
+    let pathToMockSettingsFile = ""
+
+    beforeEach(async () => {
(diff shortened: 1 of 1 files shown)


test/unit/node/settings.test.ts
```diff
@@ -0,0 +1,65 @@
...
+describe("settings", () => {
+  describe("with invalid JSON in settings file", () => {
+    const testName = "settingsInvalidJSON"
+    let testDir = ""
+    let mockSettingsFile = "coder.json"
+    let pathToMockSettingsFile = ""
+
+    beforeEach(async () => {
+      mockLogger()
+      await clean(testName)
```
code-asher, line 15: This is completely up to personal preference but using a single temporary directory per suite is a little easier since you would not need to duplicate this code for each test (this is how the other tests are doing it I believe).

If we want to do it this way we could move it to a `beforeEach` in the top-level describe (can also combine the logger mock if desired).

test/unit/node/settings.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      testDir = await tmpdir(testName)
+      pathToMockSettingsFile = `${testDir}/${mockSettingsFile}`
+      // Missing a quote, which makes it invalid intentionally
+      await fs.writeFile(pathToMockSettingsFile, '{"fakeKey":true,"helloWorld:"test"}')
+    })
+    afterEach(async () => {
+      await fs.rm(testDir, { recursive: true, force: true })
```
code-asher, line 22: No need, this will get cleaned up on the next test run.

Leaving the directory is nice because you can inspect it if something failed in the test.

test/unit/node/settings.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      testDir = await tmpdir(testName)
+      // Add hello so it's a directory that doesn't exist
+      // NOTE: if we don't have that, it fails the test
+      // That's because it will write a file if it doesn't exist
+      // but it throws if there's a directory in the path that
+      // doesn't exist.
+      pathToMockSettingsFile = `${testDir}/hello/${mockSettingsFile}`
+    })
+    afterEach(async () => {
+      await fs.rm(testDir, { recursive: true, force: true })
```
code-asher, line 50: Same as above.

code-asher (approved):
Looks great!  I only have architectural comments.

test/unit/node/settings.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    beforeEach(async () => {
+      mockLogger()
+      await clean(testName)
+      testDir = await tmpdir(testName)
+      // Add hello so it's a directory that doesn't exist
+      // NOTE: if we don't have that, it fails the test
+      // That's because it will write a file if it doesn't exist
+      // but it throws if there's a directory in the path that
+      // doesn't exist.
+      pathToMockSettingsFile = `${testDir}/hello/${mockSettingsFile}`
```
code-asher, line 47: Recommend using `path.join` for cross-platform compat.

jsjoeio: Ahh good call!

test/unit/node/settings.test.ts
jsjoeio: Oh! With `clean()`? Got it!

test/unit/node/settings.test.ts
jsjoeio: I'm all for less code so I'll do this. Thanks for the recommendation! I was worried expecting mock calls in each might cause conflicts, but I think as long as I clear the mocks `afterEach` it should be fine.

PR #4769, opened 2022-01-19
feat: publish beta and development tags on npm

### LINKED ISSUE #4744: Publish beta release on npm
## Problem

There is no easy way to test the latest changes on actual devices. i.e. I'm trying to test [these changes](https://github.com/coder/code-server/pull/4733) on Termux. 

## Solution

Modify our [publish-npm script](https://github.com/coder/code-server/blob/main/ci/steps/publish-npm.sh) to run on every PR into `main` and publish to npm. I think we would need two things:
- publish PRs into `main` and tag with PR number and commit
- publish merges into `main` as `beta`

This would require some refactoring but it should be a low amount of effort with high reward.

It would allow maintainers and community members to install these versions.

### Resources
- [`yarn publish --tag`](https://classic.yarnpkg.com/lang/en/docs/cli/publish/#toc-yarn-publish-tag)
- [publish beta to npm](https://maxisam.github.io/2017/03/29/publish-beta-to-npm/)

jsjoeio:
## Changes

This PR makes two major changes to how we release the npm package.

### PRs are published with PR number and commit

When a PR is opened into `main`, we publish the npm package under a tag with the PR number. For example, you could install code-server based on #4768 and the commit 1a44d9b8839642b048fe5816946208a9e910f132 by running:

```shell
yarn global add code-server@4768
```

If you needed to install a previous commit on the same PR, you can get more specific and run this:

```shell
yarn global add code-server@4768-1a44d9b8839642b048fe5816946208a9e910f132
```

This allows developers to test PR changes in a _production setting_ by installing code-server locally with a specific tag.

### Merges into `main` are published on npm under the `beta` tag

Anytime a PR is merged into the `main` branch, it triggers a GitHub Actions workflow to publish those changes as an npm package tagged with `beta`. 

This allows anyone using code-server to use the _bleeding edge_ version by running `yarn global add code-server@beta`. The benefit of this is allowing both users and maintainers to test new features before a release.

### Screenshots

image

### Testing

You can test the development version from this PR [here](https://www.npmjs.com/package/code-server/v/4.0.1-4769-ad7b23cfe6ffd72914e34781ef7721b129a23040). To install locally run:

```shell
yarn add code-server@4769
```

### Is there a limit to how often we can publish to npm?

According to [S/O](https://stackoverflow.com/a/34674049/3015595), no:

>  there is no restriction on how often a package can be published.

Fixes #4744

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -177,6 +177,15 @@ jobs:
           name: npm-package
           path: ./package.tar.gz
 
+      - name: Publish npm package with PR number and commit SHA
+        run: yarn publish:npm
+        env:
+          ENVIRONMENT: "development"
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_TAG: ${{ github.event.number }}
+          PR_NUMBER_AND_COMMIT_SHA: ${{ github.event.number }}-${{ github.event.pull_request.head.sha }}
+
   # TODO: cache building yarn --production
   # possibly 2m30s of savings(?)
   # this requires refactoring our release scripts
--- a/.github/workflows/npm-dev.yaml
+++ b/.github/workflows/npm-dev.yaml
@@ -1,0 +1,27 @@
+name: Publish on npm and tag with "beta"
+
+on:
+  # Shows the manual trigger in GitHub UI
+  # helpful as a back-up in case the GitHub Actions Workflow fails
+  workflow_dispatch:
+
+  push:
+    branches:
+      - main
+
+jobs:
+  # NOTE: this job requires curl, jq and yarn
+  # All of them are included in ubuntu-latest.
+  npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run ./ci/steps/publish-npm.sh
+        run: yarn publish:npm
+        env:
+          ENVIRONMENT: "staging"
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_TAG: "beta"
+          PR_NUMBER_AND_COMMIT_SHA: ${{ github.event.number }}-${{ github.event.pull_request.head.sha }}
--- a/ci/steps/publish-npm.sh
+++ b/ci/steps/publish-npm.sh
@@ -74,7 +74,9 @@ main() {
   # We only need to run npm version for "development" and "staging".
   # This is because our release:prep script automatically bumps the version
   # in the package.json and we commit it as part of the release PR.
-  if [[ ! ENVIRONMENT == "production" ]]; then
+  if [[ ENVIRONMENT == "production" ]]; then
+    NPM_VERSION="$VERSION"
+  else
     echo "Not a production environment"
     echo "Found environment: $ENVIRONMENT"
     echo "Manually bumping npm version..."
--- a/ci/steps/publish-npm.sh
+++ b/ci/steps/publish-npm.sh
@@ -74,7 +74,7 @@ main() {
   # We only need to run npm version for "development" and "staging".
   # This is because our release:prep script automatically bumps the version
   # in the package.json and we commit it as part of the release PR.
-  if [[ ENVIRONMENT == "production" ]]; then
+  if [[ "$ENVIRONMENT" == "production" ]]; then
     NPM_VERSION="$VERSION"
   else
     echo "Not a production environment"
--- a/ci/steps/publish-npm.sh
+++ b/ci/steps/publish-npm.sh
@@ -4,15 +4,60 @@ set -euo pipefail
 main() {
   cd "$(dirname "$0")/../.."
   source ./ci/lib.sh
+  source ./ci/steps/steps-lib.sh
 
-  # npm view won't exit with non-zero so we have to check the output.
-  local hasVersion
-  hasVersion=$(npm view "code-server@$VERSION" version)
-  if [[ $hasVersion == "$VERSION" ]]; then
-    echo "$VERSION is already published"
-    return
+  ## Authentication tokens
+  # Needed to publish on NPM
+  if ! is_env_var_set "NPM_TOKEN"; then
+    echo "NPM_TOKEN is not set. Cannot publish to npm without credentials."
+    exit 1
   fi
 
+  # NOTE@jsjoeio - only needed if we use the download_artifact
+  # because we talk to the GitHub API.
+  # Needed to use GitHub API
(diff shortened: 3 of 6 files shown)


ci/steps/publish-npm-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+  # https://github.com/actions/upload-artifact/issues/38
+  tar -xzf release-npm-package/package.tar.gz
+
+  # Ignore symlink when publishing npm package
+  # See: https://github.com/cdr/code-server/pull/3935
+  echo "node_modules.asar" > release/.npmignore
+  # TODO@jsjoeio
+  # There are two things that need to happen
+  # in order to publish on npm, we need to change the version
+  # and possibly change the tag
```
jsjoeio, line 54: nit: remove

ci/steps/publish-npm-dev.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
jsjoeio, line 1: ❓Should we refactor [`publish-npm.sh` script](https://github.com/coder/code-server/blob/main/ci/steps/publish-npm.sh) instead of adding this script? They're quite similar.

.github/workflows/ci.yaml
```diff
@@ -177,6 +177,15 @@ jobs:
           name: npm-package
           path: ./package.tar.gz
 
+      - name: Publish npm package with dev tag
+        run: yarn publish:npm:dev
```
jsjoeio, line 181: ❓The job this runs in already has the npm artifact. Should we modify the script to use that or should we keep as is? The benefit of refactoring would be to speed things up. The benefit of keeping the current approach is this step doesn't rely on any side effects and is _pure_.

jsjoeio (commented):
This works! See published version here: https://www.npmjs.com/package/code-server/v/4.0.1-4769-ad7b23cfe6ffd72914e34781ef7721b129a23040

I would like a second opinion on the implementation. There are ways we can refactor this but I am unsure if it's necessary.

ci/steps/publish-npm-dev.sh
code-asher: That does seem worthwhile to me.  Could tackle that later though.

.github/workflows/npm-dev.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  # NOTE: this job requires curl, jq and yarn
+  # All of them are included in ubuntu-latest.
+  npm:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Run ./ci/steps/publish-npm-dev.sh
+        run: ./ci/steps/publish-npm-dev.sh
```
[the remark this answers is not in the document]
jsjoeio: Yes! Good point

ci/steps/publish-npm-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+  cd "$(dirname "$0")/../.."
+  source ./ci/lib.sh
+  source ./ci/steps/steps-lib.sh
+
+  # We need VERSION to update the package.json
+  if ! is_env_var_set "TAG"; then
+    echo "TAG is not set. Cannot publish to npm without setting a tag."
+    exit 1
+  fi
+
```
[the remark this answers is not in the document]
jsjoeio: Oops! Will fix

ci/steps/publish-npm-dev.sh
```diff
@@ -0,0 +1,63 @@
...
+  # We need TAG to know what to publish under on npm
+  if ! is_env_var_set "TAG"; then
```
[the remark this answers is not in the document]
jsjoeio: I agree, the naming is confusing. Let's do that instead

.github/workflows/ci.yaml
code-asher: My first instinct is to use the artifact on disk.  For example if I ran this locally for some reason I think it would be unexpected if it pulled from CI instead?  

Network access I would say is especially impure so I might argue downloading is less pure than using the artifact that was just built in the job.

Lastly the download script has always felt to me like a janky workaround we have for passing artifacts around jobs; when/if we automate the release I think it is possible it ends up going away.

code-asher: But if we did have to separate it into a job we could just do that easily with the current setup, so I see what you mean by being pure.

code-asher: Hm I think you are right to have it this way, moving it into another job seems like a definite possibility to me.  In fact maybe we should do that?  It might make more sense than to be part of the build job.

jsjoeio: > Lastly the download script has always felt to me like a janky workaround we have for passing artifacts around jobs; when/if we automate the release I think it is possible it ends up going away.

I agree. I had to dig in to remember how it works. I'd say let's use the artifact on disk like you suggested.

ci/steps/publish-npm-dev.sh
jsjoeio: Eh I'll do it in this one! Shouldn't be too difficult

.github/workflows/ci.yaml
code-asher: Especially if we combine this with the existing NPM publish script because that one has to use the download as it is in a separate job.

I do still think we should replace the download script at some point, there has got to be a more official way.

code-asher (approved):

ci/steps/publish-npm.sh
```diff
@@ -24,7 +69,43 @@ main() {
   # Ignore symlink when publishing npm package
   # See: https://github.com/cdr/code-server/pull/3935
   echo "node_modules.asar" > release/.npmignore
-  yarn publish --non-interactive release
+
+  # NOTES:@jsjoeio
+  # We only need to run npm version for "development" and "staging".
+  # This is because our release:prep script automatically bumps the version
+  # in the package.json and we commit it as part of the release PR.
+  if [[ ! ENVIRONMENT == "production" ]]; then
```
code-asher, line 77:
```suggestion
  if [[ ENVIRONMENT == "production" ]]; then
    NPM_VERSION="$VERSION"
  else
```

ci/steps/publish-npm.sh
```diff
@@ -24,7 +69,43 @@ main() {
...
+    popd
+  fi
+
+  # We need to make sure we haven't already published the version.
+  # This is because npm view won't exit with non-zero so we have
+  # to check the output.
+  local hasVersion
+  hasVersion=$(npm view "code-server@$VERSION" version)
+  if [[ $hasVersion == "$VERSION" ]]; then
+    echo "$VERSION is already published"
```
code-asher, line 104:
```suggestion
  hasVersion=$(npm view "code-server@$NPM_VERSION" version)
  if [[ $hasVersion == "$NPM_VERSION" ]]; then
    echo "$NPM_VERSION is already published
```

PR #4776, opened 2022-01-20
fix: package bin/helper directory in the right spot

code-asher:
VS Code moves this as part of their packaging task.  We will probably
want to look into using theirs instead of replicating it.

Fixes https://github.com/coder/code-server/issues/4721


jsjoeio (approved):
That is unusual they change the path in production 🤔 I guess we'll have to investigate later. Thanks for fixing this!

code-asher:
Happens here btw:

https://github.com/coder/vscode/blob/5f069ef211aa08b23864032532b1161309ce03c7/src/vs/server/extensionHostConnection.ts#L36

And here is the build step:

https://github.com/coder/vscode/blob/5f069ef211aa08b23864032532b1161309ce03c7/build/gulpfile.reh.js#L329-L333

PR #4781, opened 2022-01-21
fix: update npm-dev.yaml

jsjoeio:
I didn't think it through in my previous PR but since this workflow only runs on a merge into `main`, there is no PR number. So instead, we use the word "beta" and then we also have to grab a different SHA since there is no PR SHA.

Prevents this from happening again: https://www.npmjs.com/package/code-server/v/4.0.1--
https://github.com/coder/code-server/runs/4903294989?check_suite_focus=true#step:3:16

Fixes N/A

### DIFF
--- a/.github/workflows/npm-dev.yaml
+++ b/.github/workflows/npm-dev.yaml
@@ -24,4 +24,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           NPM_TAG: "beta"
-          PR_NUMBER_AND_COMMIT_SHA: ${{ github.event.number }}-${{ github.event.pull_request.head.sha }}
+          # Since this only runs on a merge into main, we can't use github.event.number
+          # so we instead use the word "beta" and the PR merge commit SHA
+          PR_NUMBER_AND_COMMIT_SHA: beta-${{ github.sha }}


.github/workflows/npm-dev.yaml
```diff
@@ -24,4 +24,6 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
           NPM_TAG: "beta"
-          PR_NUMBER_AND_COMMIT_SHA: ${{ github.event.number }}-${{ github.event.pull_request.head.sha }}
+          # Since this only runs on a merge into main, we can't use github.event.number
+          # so we instead use the word "beta" and the PR merge commit SHA
+          PR_NUMBER_AND_COMMIT_SHA: beta-${{ github.sha }}
```
jsjoeio, line 29: I don't _love_ using "beta" where you would expect a PR_NUMBER but this was the least invasive change I could think of. I didn't want to have to add another environment variable and make this more complicated than it needs to be.

code-asher (approved):

.github/workflows/npm-dev.yaml
jsjoeio: But then again, we could always change this to `NPM_VERSION` but I don't know if that will make it more complicated. Open to ideas. Just wanted to open a fix quickly.

im-coder-lg:
Is this a fix to https://github.com/coder/code-server/runs/4903326846?check_suite_focus=true#step:14:13?

jsjoeio:
@im-coder-lg no, different fix. That should also be fixed though!

.github/workflows/npm-dev.yaml
code-asher: My bad, I reviewed too quickly and missed this I think :stuck_out_tongue: 

In case this is still relevant, I think it is a valid concern and `NPM_VERSION` would describe what this variable is for perfectly.

PR #4792, opened 2022-01-25
fix: make browser helper executable

code-asher:
Surprisingly it does not come already executable like the dev-bin
scripts.

Much to my shame I did not actually test a release with the helper
change.  This time I did and can confirm it works.

Truly fixes https://github.com/coder/code-server/issues/4721

However while it no longer errors it does not actually open a browser
window...so I will look into that separately.


jsjoeio (approved):
No worries! I also didn't test. But at least once I fix https://github.com/coder/code-server/issues/4790 it will be relatively straightforward to test things like this!

PR #4814, opened 2022-01-31
feat: add test for hash with empty string

jsjoeio:
This PR adds an additional test for the `hash` function.

Fixes N/A

#TestingMondays


test/unit/node/util.test.ts
```diff
@@ -99,11 +101,31 @@ describe("getEnvPaths", () => {
...
   })
+  it("should log an error and return an empty string on error", async () => {
+    // Although we don't expect to pass the wrong type to our `hash`
+    // function, we do want to make sure our function handles errors correctly.
+    // Therefore, we mock a scenario where the password is `undefined`
+    // which gets us into the catch block and checks that we handle errors by
+    // logging it and returning an empty string (as opposed to throwing an Error)
+    // and crashing our program for a silly password error.
+    // @ts-expect-error See description above
+    const hashed = await util.hash(undefined)
```
code-asher, line 125: Why would we not want to let this throw an error?

If we expect undefined values should we have a `typeof password == "undefined"` check?

jsjoeio: I think we had it that way before but then it crashed the program or something. I wish I wrote it down to give better context but it happened when I was migrating us over to the new argon2 package. 

> If we expect undefined values should we have a typeof password === "undefined" check?

We _don't_ expect `undefined` values so I don't think we should have that check. This was more of a mock scenario to get coverage for an uncovered area. Let me know if I should modify my comment to make that more clear.

jsjoeio: I didn't write enough context but here's where it happened: https://github.com/coder/code-server/pull/4733#discussion_r783498395

jsjoeio: The purpose of this test is to cover this block. Can you think of a better way to test it than what I'm proposing? 

image

code-asher: I think fundamentally the itch I am trying to scratch is that if we mistrust our types then we should fix them, otherwise the idea of type safety goes down the drain and TypeScript with it!

We might not need this catch.  My sense would be that we let this bubble up to the try/catch in the password handler.

code-asher: If we do need to test what happens when argon2 fails I think a mock of `argon2.hash` that throws would work pretty well.

jsjoeio: I see what you're saying. Well I always try to wrap `await` calls in `try/catch` which leads to this type of code.

To clarify, we _only_ need to test when argon2 fails if we have this catch block. But if we remove that, then we don't need to test. 

Okay...now saying this out loud...we know that `password` will _always_ be `string`. If not, we have a bug. Since we know that, and we know `.hash` even accepts empty strings (doesn't throw) then I think it's safe to remove the `try/catch` block and instead add a test for an empty string.

Thoughts?

code-asher: I agree!  That sounds great.

src/node/util.ts
```diff
@@ -157,12 +157,7 @@ export const generatePassword = async (length = 24): Promise<string> => {
  * Used to hash the password.
  */
 export const hash = async (password: string): Promise<string> => {
-  try {
-    return await argon2.hash(password)
-  } catch (error: any) {
-    logger.error(error)
-    return ""
-  }
```
jsjoeio, line 165: ✍🏼 as discussed with @code-asher, this will only throw if you throw in something other than `string` (like `number`). Since TS will prevent that, we don't need the `try/catch` block.

code-asher (approved):
:tada:

PR #4844, opened 2022-02-09
feat: Add support for ingressClassName in Helm Chart

wrmilling:
Fixes #NA - No issue created due to minor nature of change. 

This adds an optional parameter to the ingress object for `ingressClassName` which is available in K8s 1.18+. More information on `ingressClassName` can be found [here](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/).

Please let me know if you need any more information or other changes. I did a minor version bump for the chart, but happy to remove that if needed.

### DIFF
--- a/ci/helm-chart/values.yaml
+++ b/ci/helm-chart/values.yaml
@@ -35,13 +35,12 @@ service:
 ingress:
   enabled: false
   #annotations:
-  #  kubernetes.io/ingress.class: nginx
   #  kubernetes.io/tls-acme: "true"
   #hosts:
   #  - host: code-server.example.loc
   #    paths:
   #      - /
-  #ingressClassName: nginx
+  ingressClassName:
   #tls:
   #  - secretName: code-server
   #    hosts:
--- a/ci/helm-chart/values.yaml
+++ b/ci/helm-chart/values.yaml
@@ -41,7 +41,7 @@ ingress:
   #  - host: code-server.example.loc
   #    paths:
   #      - /
-
+  #ingressClassName: nginx
   #tls:
   #  - secretName: code-server
   #    hosts:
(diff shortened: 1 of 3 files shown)


jsjoeio (commented):
Thanks so much for adding this! 

Looks good to me but let me get a second opinion from the experts. 

cc @Matthew-Beckett @jawnsy

Matthew-Beckett (approved):
Looks good, other than the comment above about the legacy annotation no changes needed.

Thanks for helping!

❤️

ci/helm-chart/values.yaml
```diff
@@ -41,7 +41,7 @@ ingress:
   #  - host: code-server.example.loc
   #    paths:
   #      - /
-
+  #ingressClassName: nginx
```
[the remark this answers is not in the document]
wrmilling: I followed suit with the annotation example originally, but I am fine either way so I went ahead and set it as an empty value as suggested. Legacy annotation also removed.

ci/helm-chart/values.yaml
```diff
@@ -35,13 +35,12 @@ service:
...
   enabled: false
   #annotations:
-  #  kubernetes.io/ingress.class: nginx
   #  kubernetes.io/tls-acme: "true"
   #hosts:
   #  - host: code-server.example.loc
   #    paths:
   #      - /
-
+  ingressClassName:
```
jawnsy, line 43: add an empty string just to make it clearer that the field is intentionally empty:

```suggestion
  ingressClassName: ""
```

wrmilling:
Thanks @jsjoeio, @jawnsy, and @Matthew-Beckett. Feedback integrated into the latest commit.

ci/helm-chart/values.yaml
wrmilling: Whoops, thanks!

PR #4847, opened 2022-02-10
chore: update Docker image publish workflow

jawnsy:
* Use official action to log in to Docker Hub
* Run using pinned ubuntu-20.04 environment


jsjoeio (approved):
Thank you @jawnsy!

PR #4850, opened 2022-02-11
chore: allow overwriting Docker images

jawnsy:
Allow Docker images to be overwritten, which is required to update
dependencies in the base image when the upstream vendor resolves
security issues.


jsjoeio (approved):
Makes sense to me! @code-asher any concerns?

code-asher (approved):
:rocket:

PR #4866, opened 2022-02-14
feat(testing): add test for `parse` when error in args + config

### LINKED ISSUE #4858: [Testing]: add more coverage for src/node/cli.ts
For #TestingMondays, I'd like to increase our coverage for `src/node/cli.ts`. 293 is missing coverage.

We already have tests in place for `parse`.

To test this, we should create a test which uses a config file and adds an option that doesn't exist. We expect the function to throw an error with a specific message.

jsjoeio:
This adds a test for `parse` to hit a line where an error is thrown with a bad arg and a config file path are passed in.

Fixes #4858


test/unit/node/update.test.ts
```diff
@@ -218,8 +218,8 @@ describe("update", () => {
   it("should reject if response has status code 500", async () => {
     if (isAddressInfo(_address)) {
       const mockURL = `http://${_address.address}:${_address.port}/reject-status-code`
-      let provider = new UpdateProvider(mockURL, settings())
-      let update = await provider.getUpdate(true)
+      const provider = new UpdateProvider(mockURL, settings())
```
jsjoeio, line 221: Very strange...I didn't make this change but when I ran `yarn lint` or `yarn fmt` locally, this changed happened?

code-asher: Huh I think we are passing `--fix` to auto-fix but I did not know they would change `let` to `const`.

jsjoeio: I didn't either! Maybe a new thing? Or maybe I did something janky...no idea

code-asher: Could be something new!  Either way, const looks right since we are not changing these values.

code-asher (approved):

PR #4881, opened 2022-02-17
fix: state collision

code-asher:
VS Code names their databases after the workspace but if two workspaces share the same path (for example if they are on different virtual machines) but they share the same domain (since IndexedDB is scoped to domain) then those two states will have a naming collision.

Blocked on https://github.com/coder/vscode/pull/46.

### DIFF
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -1,3 +1,5 @@
+import * as path from "path"
+import { promises as fs } from "fs"
 import { describe, test, expect } from "./baseFixture"
 
 describe("CodeServer", true, [], () => {
@@ -23,5 +25,12 @@ describe("CodeServer", true, [], () => {
   test("should show the Integrated Terminal", async ({ codeServerPage }) => {
     await codeServerPage.focusTerminal()
     expect(await codeServerPage.page.isVisible("#terminal")).toBe(true)
+  })
+
+  test("should open a file", async ({ codeServerPage }) => {
+    const dir = await codeServerPage.dir()
+    const file = path.join(dir, "foo")
+    await fs.writeFile(file, "bar")
+    await codeServerPage.openFile(file)
   })
 })
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -28,14 +28,14 @@ describe("CodeServer", true, [], () => {
   })
 
   test("should open a file", async ({ codeServerPage }) => {
-    const dir = await codeServerPage.dir()
+    const dir = await codeServerPage.workspaceDir
     const file = path.join(dir, "foo")
     await fs.writeFile(file, "bar")
     await codeServerPage.openFile(file)
   })
 
   test("should not share state with other paths", async ({ codeServerPage }) => {
-    const dir = await codeServerPage.dir()
+    const dir = await codeServerPage.workspaceDir
     const file = path.join(dir, "foo")
     await fs.writeFile(file, "bar")
 
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -33,4 +33,25 @@ describe("CodeServer", true, [], () => {
     await fs.writeFile(file, "bar")
     await codeServerPage.openFile(file)
   })
+
+  test("should not share state with other paths", async ({ codeServerPage }) => {
+    const dir = await codeServerPage.dir()
+    const file = path.join(dir, "foo")
+    await fs.writeFile(file, "bar")
+
+    await codeServerPage.openFile(file)
+
+    // If we reload now VS Code will be unable to save the state changes so wait
+    // until those have been written to the database.  It flushes every five
+    // seconds so we need to wait at least that long.
+    await codeServerPage.page.waitForTimeout(5500)
+
+    // The tab should re-open on refresh.
+    await codeServerPage.page.reload()
+    await codeServerPage.waitForTab(file)
+
+    // The tab should not re-open on a different path.
+    await codeServerPage.setup(true, "/vscode")
+    expect(await codeServerPage.tabIsVisible(file)).toBe(false)
+  })
 })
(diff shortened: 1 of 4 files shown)


test/e2e/codeServer.test.ts
```diff
@@ -24,4 +26,40 @@ describe("CodeServer", true, [], () => {
...
+    const dir = await codeServerPage.dir()
+    const file = path.join(dir, "foo")
+    await fs.writeFile(file, "bar")
+
+    await codeServerPage.openFile(file)
+
+    // If we reload now VS Code will be unable to save the state changes so wait
+    // until those have been written to the database.  It flushes every five
+    // seconds so we need to wait at least that long.
+    await codeServerPage.page.waitForTimeout(5500)
```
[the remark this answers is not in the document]
code-asher: Except to look through the IndexedDB on a loop waiting until the state is written or something.  VS Code does use a `BroadcastChannel` which worked except it turns out they only broadcast global state changes and not workspace state changes.

jsjoeio:
Docs preview should be fixed: just waiting on a review for this 😉 https://github.com/coder/code-server/pull/4875

test/e2e/codeServer.test.ts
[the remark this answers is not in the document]
jsjoeio: Sometimes, this is the only thing that works! Very curious how you arrived at 5500 vs 5000 😂

> which I got working except it turns out they only broadcast global state changes and not workspace state changes.

Janky indeed. Well hey at least you tried a "proper" solution. Seems like we have the best solution atm so all good! Great comment too.

test/e2e/models/CodeServer.ts
```diff
@@ -13,14 +13,21 @@ interface CodeServerProcess {
   address: string
 }
 
-class CancelToken {
+class Context {
```
jsjoeio, line 16: ❓ What was the reason for changing this?

test/e2e/models/CodeServer.ts
```diff
@@ -47,11 +55,18 @@ export class CodeServer {
     return address
   }
 
+  async dir(): Promise<string> {
```
jsjoeio, line 58: nit: we may want to call this `workspaceDir` since I think that accurately describes what this does. I thought it created a new directory at first glance.

test/e2e/models/CodeServer.ts
```diff
@@ -183,12 +198,16 @@ export class CodeServerPage {
...
   /**
-   * Navigate to code-server.
+   * Navigate to a code-server endpoint.  By default go to the root.
    */
-  async navigate() {
-    const address = await this.codeServer.address()
-    await this.page.goto(address, { waitUntil: "networkidle" })
+  async navigate(path = "/") {
+    const to = new URL(path, await this.codeServer.address())
+    await this.page.goto(to.toString(), { waitUntil: "networkidle" })
```
jsjoeio, line 210: *insert joke "ha! you said 'tutu'"

test/e2e/models/CodeServer.ts
```diff
@@ -273,6 +292,29 @@ export class CodeServerPage {
     await this.page.waitForSelector("textarea.xterm-helper-textarea")
   }
 
+  /**
```
jsjoeio, line 295: Wonderful additions to this class 👏

test/e2e/models/CodeServer.ts
```diff
@@ -287,66 +329,126 @@ export class CodeServerPage {
...
-    }
+    context.finish()
+    logger.debug(`navigation took ${attempts} ${plural(attempts, "attempt")}`)
+  }
 
-    this.codeServer.logger.debug(`menu navigation retries: ${retryCount}`)
+  /**
+   * Navigate through a currently opened picker, retrying on failure.
+   */
+  async navigatePicker(items: string[]): Promise<void> {
```
jsjoeio, line 432: picker -> is this the file picker from the command palette box?

test/e2e/models/CodeServer.ts
```diff
@@ -287,66 +329,126 @@ export class CodeServerPage {
...
   }
 
   /**
    * Navigates to code-server then reloads until the editor is ready.
    *
    * It is recommended to run setup before using this model in any tests.
    */
-  async setup(authenticated: boolean) {
-    await this.navigate()
+  async setup(authenticated: boolean, endpoint = "/") {
```
jsjoeio, line 450: nit: we call it `endpoint` here and `path` in navigate

jsjoeio (approved):

test/e2e/codeServer.test.ts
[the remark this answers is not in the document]
code-asher: lol the extra 500ms is to be sure the interval has passed and the write has finished but it is completely arbitrary

test/e2e/models/CodeServer.ts
code-asher: I am using it to signal both canceling and being done so I felt `CancelToken` no longer fit.

test/e2e/models/CodeServer.ts
code-asher: loooooooooool tutu string hahahaha

test/e2e/models/CodeServer.ts
code-asher: Agreed!  I might make it not a function as well.

test/e2e/models/CodeServer.ts
code-asher: Yup, from the command palette or via File > Open File from the menu.  Although maybe I should call it `navigateQuickInput` since it should work for any "quick input" widgets as VS Code calls them.

test/e2e/models/CodeServer.ts
code-asher: Egads!  I will standardize on `endpoint`.

jsjoeio: TIL: https://www.urbandictionary.com/define.php?term=Egads%21

PR #4898, opened 2022-02-22
feat(testing): add new test for cli.ts

### LINKED ISSUE #4885: [Testing]: add more coverage for parse function in cli.ts
L359 in `cli.ts` is not covered (the `continue`):

https://github.com/coder/code-server/blob/main/src/node/cli.ts#L358-L360

I'd like to add a test, similar to https://github.com/coder/code-server/pull/4866

## How to test

Pass an arg that is an `OptionalString` and set it to `false` such as `--cert`

jsjoeio:
This PR adds a test to ensure we have coverage for this block:

https://github.com/coder/code-server/blob/main/src/node/cli.ts#L358-L360

Fixes #4885

### DIFF
--- a/test/unit/node/cli.test.ts
+++ b/test/unit/node/cli.test.ts
@@ -371,8 +371,8 @@ describe("parser", () => {
       }),
     ).toThrowError(expectedErrMsg)
   })
-  it("should not throw if an optional string is set to false", async () => {
-    expect(() => parse(["--cert=false"])).not.toThrowError()
+  it("should ignore optional strings set to false", async () => {
+    expect(parse(["--cert=false"])).toEqual({})
   })
 })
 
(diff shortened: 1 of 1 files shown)


code-asher (approved):

test/unit/node/cli.test.ts
```diff
@@ -371,6 +371,9 @@ describe("parser", () => {
       }),
     ).toThrowError(expectedErrMsg)
   })
+  it("should not throw if an optional string is set to false", async () => {
+    expect(() => parse(["--cert=false"])).not.toThrowError()
```
code-asher, line 375: Should we check that it is not set in the result?  Something like:

```typescript
expect(parse(["--cert=false"])).toEqual({})
```

jsjoeio: Yeah, that's a better assertion IMO. Good callout!

PR #4911, opened 2022-02-25
chore: resources definition for initContainer

krisek:
Fixes helm chart/deployment failing on namespaces with ResourceQuota defined


jsjoeio:
cc @Matthew-Beckett or @jawnsy

Issue #4874, opened 2022-02-16
[Feat]: Add VS Code version number to --version output


jawnsy:
## What is your suggestion?

The `code-server` version output shows the version of `code-server` itself:

```
$ code-server --version
4.0.2 5cdfe74686aa73e023f8354a9a6014eb30caa7dd
```

It is possible to get the corresponding (embedded) VS Code/code-oss-dev version using the `--json` output:

```
$ code-server --version --json
{"codeServer":"4.0.2","commit":"5cdfe74686aa73e023f8354a9a6014eb30caa7dd","vscode":"1.63.0"}
```

It would be convenient to also have the VSCode version in the --version output.

## Why do you want this feature?

This makes the VS Code version more discoverable, which is useful when evaluating the versions of extensions that are compatible.

## Are there any workarounds to get this functionality today?

Yes, users can run `code-server --version --json` and look at the output. But JSON output is not as human-readable.

## Are you interested in submitting a PR for this?

Sure, would like a pointer to the code though.

jsjoeio:
Fantastic idea! This is where the code lives in `entry.ts`:

https://github.com/coder/code-server/blob/94f378c1960a2b8df4da70fb900f6cfec0191982/src/node/entry.ts#L40-L54

I don't think we have a file yet called `entry.test.ts` so you could create one if you're comfortable adding a test as well (if not, I'm happy to).

Otherwise, you could add another test to our standalone release here:

https://github.com/coder/code-server/blob/94f378c1960a2b8df4da70fb900f6cfec0191982/ci/build/test-standalone-release.sh

PR #4920, opened 2022-02-26
feat: add version string functions to constants

jawnsy:
Introduce helper functions for getting human- and machine-readable
version strings from the constants package, and cover it in unit
tests.

This is a first step to resolving #4874.

### DIFF
--- a/src/node/constants.ts
+++ b/src/node/constants.ts
@@ -32,7 +32,7 @@ export const httpProxyUri =
 // getVersionString returns a human-readable version string suitable
 // for outputting to the console.
 export function getVersionString(): string {
-  return [ version, commit ].join(" ")
+  return [version, commit].join(" ")
 }
 
 // getVersionJsonString returns a machine-readable version string
--- a/src/node/constants.ts
+++ b/src/node/constants.ts
@@ -23,7 +23,24 @@ export const version = pkg.version || "development"
 export const commit = pkg.commit || "development"
 export const rootPath = path.resolve(__dirname, "../..")
 export const vsRootPath = path.join(rootPath, "vendor/modules/code-oss-dev")
+export const codeVersion = require(path.join(vsRootPath, "package.json")).version
 export const tmpdir = path.join(os.tmpdir(), "code-server")
 export const isDevMode = commit === "development"
 export const httpProxyUri =
   process.env.HTTPS_PROXY || process.env.https_proxy || process.env.HTTP_PROXY || process.env.http_proxy
+
+// getVersionString returns a human-readable version string suitable
+// for outputting to the console.
+export function getVersionString(): string {
+  return [ version, commit ].join(" ")
+}
+
+// getVersionJsonString returns a machine-readable version string
+// suitable for outputting to the console.
+export function getVersionJsonString(): string {
+  return JSON.stringify({
+    codeServer: version,
+    commit,
+    vscode: codeVersion,
+  })
+}
--- a/src/node/constants.ts
+++ b/src/node/constants.ts
@@ -23,7 +23,7 @@ export const version = pkg.version || "development"
 export const commit = pkg.commit || "development"
 export const rootPath = path.resolve(__dirname, "../..")
 export const vsRootPath = path.join(rootPath, "vendor/modules/code-oss-dev")
-export const codeVersion = require(path.join(vsRootPath, "package.json")).version
+export const codeVersion = require(path.join(vsRootPath, "package.json")).version || "development"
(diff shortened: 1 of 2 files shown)


src/node/constants.ts
```diff
@@ -23,7 +23,24 @@ export const version = pkg.version || "development"
...
 export const vsRootPath = path.join(rootPath, "vendor/modules/code-oss-dev")
+export const codeVersion = require(path.join(vsRootPath, "package.json")).version
 export const tmpdir = path.join(os.tmpdir(), "code-server")
 export const isDevMode = commit === "development"
 export const httpProxyUri =
   process.env.HTTPS_PROXY || process.env.https_proxy || process.env.HTTP_PROXY || process.env.http_proxy
+
+// getVersionString returns a human-readable version string suitable
+// for outputting to the console.
+export function getVersionString(): string {
```
jawnsy, line 34: exporting these functions means we can test in constants.test.ts, and then the entry.ts code just needs to call these (and thus does not need to be tested)

this is part 1, the second part will change this to return the code-server version info and change the entry.ts code to call these functions. I split these up to streamline review.

src/node/constants.ts
```diff
@@ -23,7 +23,24 @@ export const version = pkg.version || "development"
 export const commit = pkg.commit || "development"
 export const rootPath = path.resolve(__dirname, "../..")
 export const vsRootPath = path.join(rootPath, "vendor/modules/code-oss-dev")
+export const codeVersion = require(path.join(vsRootPath, "package.json")).version || "development"
```
jawnsy, line 26: The e2e tests don't seem to like this. Thoughts/ideas?

```
$ /home/runner/work/code-server/code-server/test/node_modules/.bin/playwright test
Error: Cannot find module '/home/runner/work/code-server/code-server/vendor/modules/code-oss-dev/package.json'
```

src/node/constants.ts
jsjoeio: nit: if you use the TSDoc block syntax, IDEs will show this information on hover.

image


```typescript
/**
 * getVersionJsonString returns a machine-readable version string
 * suitable for outputting to the console.
 */
```

src/node/constants.ts
```diff
@@ -23,7 +23,24 @@ export const version = pkg.version || "development"
...
 export const httpProxyUri =
   process.env.HTTPS_PROXY || process.env.https_proxy || process.env.HTTP_PROXY || process.env.http_proxy
+
+// getVersionString returns a human-readable version string suitable
+// for outputting to the console.
+export function getVersionString(): string {
+  return [version, commit].join(" ")
+}
+
+// getVersionJsonString returns a machine-readable version string
```
jsjoeio, line 38: same as above^

test/unit/node/constants.test.ts
```diff
@@ -24,12 +31,46 @@ describe("constants", () => {
...
+    })
+
+    it("should return a machine-readable version string", () => {
+      expect(constants.getVersionJsonString()).toStrictEqual(
+        JSON.stringify({
+          codeServer: mockPackageJson.version,
+          commit: mockPackageJson.commit,
+          vscode: mockCodePackageJson.version,
+        }),
+      )
```
jsjoeio, line 73: Nice test additions 👏

src/node/constants.ts
jsjoeio: Hmm... my guess would be that it doesn't like that we're using `require` here and it expects that to be a module? 🤔

That may be what lead me to create [`getPackageJson`](https://github.com/coder/code-server/blob/7d0610320f6aeca26cc591a74315aa099d304dcd/src/node/constants.ts#L8-L17).

There's probably something funky that Playwright does where it requires all the modules before running the tests. And since it can't find a module, it complains. 

Maybe @code-asher knows how to work around this

jsjoeio (approved):
Woo! Thank you @jawnsy 🎉

src/node/constants.ts
jawnsy: Great tip, will update!

src/node/constants.ts
code-asher: Using `getPackageJson` for `VS Code`'s `package.json` as well makes sense to me.

Maybe we are not packaging the `package.json` as part of the release artifact.

jsjoeio: > Using getPackageJson for VS Code's package.json as well makes sense to me.

Oh duh...didn't even put two-and-two together. Yeah you could use it here.

jawnsy: switched to `getPackageJson`, hopefully that fixes it :crossed_fingers:

PR #4932, opened 2022-03-01
refactor: remove folder/workspace from vsCodeCliArgs

jsjoeio:
Since we handle this in the vscode.ts route, we no longer need to pass it to VS
Code as a CLI arg since it's deprecated on that side.

This adds logic to our `vscode` route handler to handle both folder and workspace paths passed in via the CLI

Fixes an issue related to https://github.com/coder/vscode/pull/45

Blocked by https://github.com/coder/vscode/pull/49

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -3,15 +3,7 @@ import { promises as fs } from "fs"
 import yaml from "js-yaml"
 import * as os from "os"
 import * as path from "path"
-import {
-  canConnect,
-  generateCertificate,
-  generatePassword,
-  humanPath,
-  paths,
-  isNodeJSErrnoException,
-  isFile,
-} from "./util"
+import { canConnect, generateCertificate, generatePassword, humanPath, paths, isNodeJSErrnoException } from "./util"
 
 const DEFAULT_SOCKET_PATH = path.join(os.tmpdir(), "vscode-ipc")
 
@@ -448,7 +440,7 @@ export interface DefaultedArgs extends ConfigArgs {
   "extensions-dir": string
   "user-data-dir": string
   /* Positional arguments. */
-  _: []
+  _: string[] | []
 }
 
 /**
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -770,25 +770,9 @@ export const shouldOpenInExistingInstance = async (args: UserProvidedArgs): Prom
  * Convert our arguments to VS Code server arguments.
  */
 export const toVsCodeArgs = async (args: DefaultedArgs): Promise<CodeServerLib.ServerParsedArgs> => {
-  let workspace = ""
-  let folder = ""
-  if (args._.length) {
-    const lastEntry = path.resolve(args._[args._.length - 1])
-    const entryIsFile = await isFile(lastEntry)
-    if (entryIsFile && path.extname(lastEntry) === ".code-workspace") {
-      workspace = lastEntry
-    } else if (!entryIsFile) {
-      folder = lastEntry
-    }
-    // Otherwise it is a regular file.  Spawning VS Code with a file is not yet
-    // supported but it can be done separately after code-server spawns.
-  }
-
   return {
     "connection-token": "0000",
     ...args,
-    workspace,
-    folder,
     "accept-server-license-terms": true,
     /** Type casting. */
     help: !!args.help,
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -440,7 +440,7 @@ export interface DefaultedArgs extends ConfigArgs {
   "extensions-dir": string
   "user-data-dir": string
   /* Positional arguments. */
-  _: string[] | []
+  _: string[]
 }
 
 /**
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -57,14 +57,12 @@ export class CodeServerRouteWrapper {
           workspace: lastOpened.workspace,
         })
       } else if (req.args._.length > 0) {
-        if (req.args._.length) {
-          const lastEntry = path.resolve(req.args._[req.args._.length - 1])
-          const entryIsFile = await isFile(lastEntry)
-          if (entryIsFile && path.extname(lastEntry) === ".code-workspace") {
-            workspace = lastEntry
-          } else if (!entryIsFile) {
-            folder = lastEntry
-          }
+        const lastEntry = path.resolve(req.args._[req.args._.length - 1])
+        const entryIsFile = await isFile(lastEntry)
+        if (entryIsFile && path.extname(lastEntry) === ".code-workspace") {
+          workspace = lastEntry
+        } else if (!entryIsFile) {
+          folder = lastEntry
         }
       }
       return redirect(req, res, to, {
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -42,7 +42,7 @@ export class CodeServerRouteWrapper {
       const lastOpened = settings.query || {}
       // This flag disables the last opened behavior
       const IGNORE_LAST_OPENED = req.args["ignore-last-opened"]
-      const HAS_LAST_OPENED_FOLDER_OR_WORKSPACE = (!WORKSPACE_WAS_CLOSED && lastOpened.folder) || lastOpened.workspace
+      const HAS_LAST_OPENED_FOLDER_OR_WORKSPACE = lastOpened.folder || lastOpened.workspace
       const HAS_FOLDER_OR_WORKSPACE_FROM_CLI = req.args._.length > 0
       const to = self(req)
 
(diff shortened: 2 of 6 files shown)


src/node/cli.ts
```diff
@@ -434,7 +426,7 @@ export interface DefaultedArgs extends ConfigArgs {
   "extensions-dir": string
   "user-data-dir": string
   /* Positional arguments. */
-  _: []
+  _: string[] | []
```
code-asher, line 429: Haha I never noticed this was just a plain array type.  We could probably leave this at just `string[]` if we wanted!

src/node/cli.ts
```diff
@@ -751,25 +743,9 @@ export const shouldOpenInExistingInstance = async (args: UserProvidedArgs): Prom
  * Convert our arguments to VS Code server arguments.
  */
 export const toVsCodeArgs = async (args: DefaultedArgs): Promise<CodeServerLib.ServerParsedArgs> => {
```
code-asher, line 745: I think we will need to remove `workspace` and `folder` from the VS Code side because right now they are required arguments.

src/node/routes/vscode.ts
```diff
@@ -33,27 +34,43 @@ export class CodeServerRouteWrapper {
...
-    const { query } = await req.settings.read()
-    if (query) {
-      // Ew means the workspace was closed so clear the last folder/workspace.
-      if (req.query.ew) {
-        delete query.folder
-        delete query.workspace
-      }
+    let folder = undefined
+    let workspace = undefined
+    const to = self(req)
```
code-asher, line 39: We could move these into the lower `if` block since they are not used outside it.

src/node/routes/vscode.ts
```diff
@@ -33,27 +34,43 @@ export class CodeServerRouteWrapper {
...
+    const settings = await req.settings.read()
+    const lastOpened = settings.query || {}
+
+    // Ew means the workspace was closed so clear the last folder/workspace.
+    if (req.query.ew) {
+      delete lastOpened.folder
+      delete lastOpened.workspace
+    }
```
code-asher, line 47: I suppose we could move this into the `if` block below as well.

src/node/routes/vscode.ts
```diff
@@ -33,27 +34,43 @@ export class CodeServerRouteWrapper {
...
+        (lastOpened.folder || lastOpened.workspace) &&
         !req.args["ignore-last-opened"] // This flag disables this behavior.
       ) {
-        const to = self(req)
         return redirect(req, res, to, {
-          folder: query.folder,
-          workspace: query.workspace,
+          folder: lastOpened.folder,
+          workspace: lastOpened.workspace,
         })
```
code-asher, line 58: We could do this to share the redirect block below:

```suggestion
        folder = lastOpened.folder
        workspace = lastOpened.workspace
```

src/node/routes/vscode.ts
```diff
@@ -33,27 +34,43 @@ export class CodeServerRouteWrapper {
...
+        if (req.args._.length) {
+          const lastEntry = path.resolve(req.args._[req.args._.length - 1])
+          const entryIsFile = await isFile(lastEntry)
+          if (entryIsFile && path.extname(lastEntry) === ".code-workspace") {
+            workspace = lastEntry
+          } else if (!entryIsFile) {
+            folder = lastEntry
+          }
+        }
       }
```
code-asher, line 69: The top two if statements are basically the same check so we could get rid of one:

```suggestion
      } else if (req.args._.length > 0) {
        const lastEntry = path.resolve(req.args._[req.args._.length - 1])
        const entryIsFile = await isFile(lastEntry)
        if (entryIsFile && path.extname(lastEntry) === ".code-workspace") {
          workspace = lastEntry
        } else if (!entryIsFile) {
          folder = lastEntry
        }
      }
 ```

File: test/unit/node/routes/vscode.test.ts [hunk 7]
```diff
@@ -52,52 +42,25 @@ describe("vscode", () => {
     }
   })
 
-  /**
-   * Get the workbench config from the provided response.
-   */
-  const getConfig = async (resp: Response): Promise<WorkbenchConfig> => {
```
code-asher, line 58: Pretty stoked to see this gone :smile:

test/unit/node/routes/vscode.test.ts
```diff
@@ -136,6 +99,42 @@ describe("vscode", () => {
...
+    await fs.writeFile(workspace, "")
+    codeServer = await integration.setup(["--auth=none", workspace], "")
+
+    let resp = await codeServer.fetch("/", undefined)
+
+    expect(resp.status).toBe(200)
+    const url = new URL(resp.url)
+    expect(url.search).toBe(`?workspace=${workspace}`)
+    await resp.text()
+  })
```
code-asher, line 113: This is an interesting case.  While it is the  same test as the one above the description is different and it does bring attention to a different thing even if the code is identical.

Should we keep it or try renaming the other?  For example we could rename the other to something like `should redirect to the passed in workspace using human-readable query` which covers both the redirect case and the human-readable case.

test/unit/node/routes/vscode.test.ts
```diff
@@ -136,6 +99,42 @@ describe("vscode", () => {
...
+    const folder = await tmpdir(testName)
+    codeServer = await integration.setup(["--auth=none", folder], "")
+
+    let resp = await codeServer.fetch("/", undefined)
+
+    expect(resp.status).toBe(200)
+    const url = new URL(resp.url)
+    expect(url.search).toBe(`?folder=${folder}`)
+    await resp.text()
+  })
```
code-asher, line 136: Same comment as above.

code-asher (changes requested):
A much cleaner paradigm!

src/node/routes/vscode.ts
code-asher: Actually it occurs to me we only need to read settings when there are no query parameters too, which could be a bit of an optimization.

src/node/cli.ts
code-asher: This is the only blocker btw, the rest of my comments are suggestions/optional.

src/node/cli.ts
jsjoeio: Yeah, it was causing issues telling me it was a tuple and I couldn't access it 😂 I'll remove `[]`

test/unit/node/routes/vscode.test.ts
jsjoeio: I'll rename it to the other - good suggestion!

src/node/cli.ts
jsjoeio: Whoops! That's why `yarn lint` is unhappy with me. Good call!

jsjoeio:
Argh, this won't pass CI until I remove those args in the VS Code side, merge and then update here.

src/node/routes/vscode.ts
```diff
@@ -33,25 +37,43 @@ export class CodeServerRouteWrapper {
...
+      const HAS_FOLDER_OR_WORKSPACE_FROM_CLI = req.args._.length > 0
+      const to = self(req)
+
+      let folder = undefined
+      let workspace = undefined
+
+      if (WORKSPACE_WAS_CLOSED) {
+        delete lastOpened.folder
+        delete lastOpened.workspace
       }
```
code-asher, line 55: Can be deleted since we skip this block when this is true

jsjoeio: ooo more code to delete - nice catch!

code-asher (approved):

src/node/routes/vscode.ts
```diff
@@ -33,25 +37,38 @@ export class CodeServerRouteWrapper {
...
-      if (req.query.ew) {
-        delete query.folder
-        delete query.workspace
-      }
+    if (NO_FOLDER_OR_WORKSPACE_QUERY && !WORKSPACE_WAS_CLOSED) {
+      const settings = await req.settings.read()
+      const lastOpened = settings.query || {}
+      // This flag disables the last opened behavior
+      const IGNORE_LAST_OPENED = req.args["ignore-last-opened"]
+      const HAS_LAST_OPENED_FOLDER_OR_WORKSPACE = (!WORKSPACE_WAS_CLOSED && lastOpened.folder) || lastOpened.workspace
```
code-asher, line 45: `ew` actually affects both folders and workspaces, I guess we could call it `FOLDER_OR_WORKSPACE_WAS_CLOSED` or something.

```suggestion
      const HAS_LAST_OPENED_FOLDER_OR_WORKSPACE = lastOpened.folder || lastOpened.workspace
```

jsjoeio: image

Is this commit suggestion right? I would assume I only need to change the name?

code-asher: Well you can remove it because it is always true (I edited my comment too late to say that)

code-asher: imo no need to change the name unless you want to

code-asher (approved):
Great refactor!

PR #4943, opened 2022-03-03
fix: re-enable trivvy docker scan

### LINKED ISSUE #4903: [Chore]: re-add trivvy scan to scan Docker image
We used to have this and lost it somehow. We need to re-add it.

jsjoeio:
This PR re-enables the trivy Docker scan as part of our CI pipeline. With this, we will have trivy do a security check on our Docker image tagged `latest` so that we can catch any security vulnerabilities.

## Testing

Example scan: https://github.com/coder/code-server/runs/5428919398?check_suite_focus=true

## Previous Context

@code-asher I were chatting about this. We want the scan to run on `latest` instead of the local build since the local build will use the newest dependencies. Instead, we want to check what's in production and ensure that it has no issues. We'll run this new workflow nightly (once a day) and check for any issues. This way, we can stay on top of the Docker image and fix things sooner.

Fixes #4903

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -464,7 +464,6 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v3
       - name: Run Trivy vulnerability scanner in repo mode
-        #Commit SHA for v0.0.17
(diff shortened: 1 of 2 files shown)


code-asher (approved):
Awesome!

PR #4957, opened 2022-03-07
chore: add permissions trivy-docker

jsjoeio:
Looks like I didn't give this workflow the right permissions: https://github.com/coder/code-server/runs/5432361797?check_suite_focus=true

I am not sure if `actions` needs `write` permissions or if `security` needs `write` permissions so I added it to both.

Fixes N/A

### DIFF
--- a/.github/workflows/trivy-docker.yaml
+++ b/.github/workflows/trivy-docker.yaml
@@ -25,7 +25,7 @@ on:
   workflow_dispatch:
 
 permissions:
-  actions: none
+  actions: write
   checks: none
   contents: read
   deployments: none
@@ -33,7 +33,7 @@ permissions:
   packages: none
   pull-requests: none
   repository-projects: none
-  security-events: none
+  security-events: write
   statuses: none
 
 # Cancel in-progress runs for pull requests when developers push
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy-docker.yaml
```diff
@@ -25,15 +25,15 @@ on:
...
+  actions: write
   checks: none
   contents: read
   deployments: none
   issues: none
   packages: none
   pull-requests: none
   repository-projects: none
-  security-events: none
+  security-events: write
```
jawnsy, line 36: I think this is the only one needed, it's documented here: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#example-workflow-for-sarif-files-generated-outside-of-a-repository

jsjoeio: Ahhh nice find!

.github/workflows/trivy-docker.yaml
```diff
@@ -25,15 +25,15 @@ on:
   workflow_dispatch:
 
 permissions:
-  actions: none
+  actions: write
```
jsjoeio, line 28: Thanks to @jawnsy, we don't need this. 

```suggestion
  actions: none
```

code-asher (approved):

PR #4968, opened 2022-03-09
docs: update docs preview

BrunoQuaresma:
Updates:
- No need to deploy a website preview
- Supports commit hashes in the URL
- Probably it will work good with forks

### DIFF
--- a/.github/workflows/docs-preview.yaml
+++ b/.github/workflows/docs-preview.yaml
@@ -22,7 +22,7 @@ permissions:
 jobs:
   preview:
     name: Docs preview
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/docs-preview.yaml
+++ b/.github/workflows/docs-preview.yaml
@@ -17,89 +17,26 @@ permissions:
   security-events: none
   statuses: none
 
-# Cancel in-progress runs for pull requests when developers push
-# additional changes, and serialize builds in branches.
-# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 
 jobs:
   preview:
     name: Docs preview
-    runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-preview.yaml
```diff
@@ -17,89 +17,26 @@ permissions:
   security-events: none
   statuses: none
 
-# Cancel in-progress runs for pull requests when developers push
-# additional changes, and serialize builds in branches.
-# https://docs.github.com/en/actions/using-jobs/using-concurrency#example-using-concurrency-to-cancel-any-in-progress-job-or-run
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```
jsjoeio, line 25: Wouldn't we want to keep this so that it cancels in-progress docs previews when devs push?

.github/workflows/docs-preview.yaml
```diff
@@ -17,89 +17,26 @@ permissions:
...
 
 jobs:
   preview:
     name: Docs preview
-    runs-on: ubuntu-20.04
-    environment: CI
-    # Only run if PR comes from base repo
-    # Reason: forks cannot access secrets and this will always fail
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
```
jsjoeio, line 25: Do we need latest? Otherwise, I think we should keep `ubuntu-20.04` since @jawnsy has suggested pinning these in the past IIRC.

jsjoeio (changes requested):
Looks good in general! 🎉

A couple questions + you'll need to fix the formatting

.github/workflows/docs-preview.yaml
BrunoQuaresma: We don't have to worry about that because this is not deploying or building anything anymore it is just a question of sharing the URL with the sha code on it.

.github/workflows/docs-preview.yaml
BrunoQuaresma: Not at all, I will change this.

.github/workflows/docs-preview.yaml
jsjoeio: Ah okay! Sweet.

jsjoeio (approved):
Looks good to me! Thanks @BrunoQuaresma 🎉

PR #4971, opened 2022-03-09
feat(testing): add test for app > listen

### LINKED ISSUE #4913: [Testing]: test handleArgsSocketCatchError is called on unlink error
We don't have coverage on L41 in `app.ts`:

https://github.com/coder/code-server/blob/main/src/node/app.ts#L41

We should be able to test this by:
- exporting `listen`
- passing in a directory instead of a file for `socket`

These are the errors that `unlink` can throw: https://man7.org/linux/man-pages/man2/unlink.2.html#ERRORS

jsjoeio:
This PR adds a new test to handle test when `fs.unlink` throws inside `listen`.

Fixes #4913

### DIFF
--- a/src/node/app.ts
+++ b/src/node/app.ts
@@ -22,7 +22,7 @@ export interface App extends Disposable {
   server: http.Server
 }
 
-const listen = (server: http.Server, { host, port, socket, "socket-mode": mode }: ListenOptions) => {
+export const listen = (server: http.Server, { host, port, socket, "socket-mode": mode }: ListenOptions) => {
   return new Promise<void>(async (resolve, reject) => {
     server.on("error", reject)
 
--- a/test/unit/node/app.test.ts
+++ b/test/unit/node/app.test.ts
@@ -3,7 +3,7 @@ import { promises } from "fs"
 import * as http from "http"
 import * as https from "https"
 import * as path from "path"
-import { createApp, ensureAddress, handleArgsSocketCatchError, handleServerError } from "../../../src/node/app"
+import { createApp, ensureAddress, handleArgsSocketCatchError, handleServerError, listen } from "../../../src/node/app"
 import { OptionalString, setDefaults } from "../../../src/node/cli"
 import { generateCertificate } from "../../../src/node/util"
 import { clean, mockLogger, getAvailablePort, tmpdir } from "../../utils/helpers"
@@ -219,7 +219,7 @@ describe("handleArgsSocketCatchError", () => {
     expect(logger.error).toHaveBeenCalledWith(errorMessage)
   })
 
-  it("should not log an error if its a iNodeJS.ErrnoException", () => {
+  it("should not log an error if its a NodeJS.ErrnoException", () => {
     const error: NodeJS.ErrnoException = new Error()
     error.code = "ENOENT"
 
@@ -248,5 +248,36 @@ describe("handleArgsSocketCatchError", () => {
 
     expect(logger.error).toHaveBeenCalledTimes(1)
     expect(logger.error).toHaveBeenCalledWith(error)
+  })
+})
+
+describe("listen", () => {
+  let tmpDirPath: string
+  let mockServer: http.Server
+
+  const testName = "listen"
+
+  beforeEach(async () => {
+    await clean(testName)
+    mockLogger()
+    tmpDirPath = await tmpdir(testName)
+    mockServer = http.createServer()
+  })
+
+  afterEach(() => {
+    mockServer.close()
+    jest.clearAllMocks()
+  })
+
+  it.only("should log an error if a directory is passed in instead of a file", async () => {
+    const errorMessage = "EPERM: operation not permitted, unlink"
+    const port = await getAvailablePort()
+    const mockArgs = { port, host: "0.0.0.0", socket: tmpDirPath }
+    try {
+      await listen(mockServer, mockArgs)
+    } catch (error) {}
+
+    expect(logger.error).toHaveBeenCalledTimes(1)
+    expect(logger.error).toHaveBeenCalledWith(expect.stringContaining(errorMessage))
   })
 })


test/unit/node/app.test.ts
```diff
@@ -250,3 +250,34 @@ describe("handleArgsSocketCatchError", () => {
...
+    tmpDirPath = await tmpdir(testName)
+    mockServer = http.createServer()
+  })
+
+  afterEach(() => {
+    mockServer.close()
+    jest.clearAllMocks()
+  })
+
+  it.only("should log an error if a directory is passed in instead of a file", async () => {
```
jsjoeio, line 272:
```suggestion
  it("should log an error if a directory is passed in instead of a file", async () => {
```

code-asher (approved):
Test looks great.  I wonder if we should actually throw on any non-ENOENT error here instead of just logging because otherwise we get:

```
[2022-03-10T19:08:49.832Z] error EISDIR: illegal operation on a directory, unlink '/home/coder'
[2022-03-10T19:08:49.835Z] error listen EADDRINUSE: address already in use /home/coder
```

Which feels to me not quite correct, `/home/coder` is not already in use, it is just not valid because it is a directory.

jsjoeio:
> Which feels to me not quite correct, /home/coder is not already in use, it is just not valid because it is a directory.

Great suggestion! I like correct better. I'll update the code and the test. (test first of course 😉 )

src/node/app.ts
```diff
@@ -22,39 +22,33 @@ export interface App extends Disposable {
...
     } else {
       // [] is the correct format when using :: but Node errors with them.
       server.listen(port, host.replace(/^\[|\]$/g, ""), onListen)
     }
+    if (socket && mode) {
+      await fs.chmod(socket, mode).catch((err) => {
+        util.logError(logger, "socket chmod", err)
+      })
+    }
   })
```
code-asher, line 52: I think this may need to go outside the promise since we need to `chmod` after the server is finished listening otherwise the socket may not have been created yet.  Something like this although I not sure I quite have it right, kinda hard to edit in GitHub :stuck_out_tongue: 

```suggestion
  })
  if (socket && mode) {
    await fs.chmod(socket, mode).catch((err) => {
      util.logError(logger, "socket chmod", err)
    })
  }
```

src/node/app.ts
```diff
@@ -137,7 +131,8 @@ export const handleServerError = (resolved: boolean, err: Error, reject: (err: E
  * test this logic more easily.
  */
 export const handleArgsSocketCatchError = (error: any) => {
+  console.log("hello")
```
code-asher, line 134: Left over console.log

test/unit/node/constants.test.ts
```diff
@@ -20,9 +21,14 @@ describe("constants", () => {
     }
 
     beforeAll(() => {
+      jest.clearAllMocks()
```
code-asher, line 24: We have `clearAllMocks` in the after callback, do we need it here as well?

test/unit/node/constants.test.ts
```diff
@@ -106,8 +112,13 @@ describe("constants", () => {
     }
 
     beforeAll(() => {
-      jest.mock("../../../package.json", () => mockPackageJson, { virtual: true })
-      jest.mock("../../../vendor/modules/code-oss-dev/package.json", () => mockCodePackageJson, { virtual: true })
+      jest.clearAllMocks()
```
code-asher, line 115: Same question

src/node/app.ts
jsjoeio: left over? No I wanted that there 😜

test/unit/node/constants.test.ts
jsjoeio: Hmm...I am thinking out loud. We need one or the other but probably not both 🤔 I think since we do the `clear(testname)` pattern before tests, we should clear the mocks in the `beforeAll` and not the after callback. Thoughts?

code-asher: I like that!

code-asher: Also happy to just leave it as-is for now, not a big deal either way was mostly curious.

jsjoeio: No no, if I leave it and merge, I won't ever do it lol let's do it right the first time!

jsjoeio: Consistency is key so keep asking these questions 👏🏼

code-asher: ahahahaha fair fair

src/node/app.ts
```diff
@@ -22,40 +22,37 @@ export interface App extends Disposable {
...
-
       server.listen(socket, onListen)
     } else {
       // [] is the correct format when using :: but Node errors with them.
       server.listen(port, host.replace(/^\[|\]$/g, ""), onListen)
     }
   })
+
+  // NOTE@jsjoeio: we need to chmod after the server is finished
+  // listening. Otherwise, teh socket may not have been created yet.
```
jawnsy, line 50: tiny typo here, no big deal though
```suggestion
  // listening. Otherwise, the socket may not have been created yet.
```

src/node/app.ts
```diff
@@ -22,40 +22,37 @@ export interface App extends Disposable {
...
+  if (socket && mode) {
+    await fs.chmod(socket, mode).catch((err) => {
+      util.logError(logger, "socket chmod", err)
+    })
```
code-asher, line 54: Sorry for not catching this earlier, we should let this error throw so code-server will exit:

```suggestion
    await fs.chmod(socket, mode)
```

jsjoeio: Nice last-minute-typical-super-asher-catch! 🔥

code-asher (approved):

src/node/app.ts
jsjoeio: You:
image

code-asher: woah how did you find a photo of me, I thought that was impossible

jsjoeio: Wayback Machine. Found it on your old freelancing portfolio site

code-asher: hahahahahaha :joy:

PR #4972, opened 2022-03-09
feat(ci): publish dev builds to @coder/code-server-pr

### LINKED ISSUE #4924: [Feat]: publish dev builds to @coder/code-server-pr
## What is your suggestion?

Currently, we publish dev builds to npm under the same code-server name and tag it with a PR like `code-server@4569`. Instead, we should publish dev builds to a separate npm namespace: `@coder/code-server-pr`.

### How to do this?

We'll need to modify this logic

https://github.com/coder/code-server/blob/main/ci/steps/publish-npm.sh#L111-L118

Basically, we need to add a line there that replaces `"name": "code-server"` with `"name": "@coder/code-server-pr"`. That should be it. We shouldn't have to add a new NPM_TOKEN since they're not scoped to packages.

As @jawnsy brought up, this is cluttering the tags under the code-server package and makes it look ugly but also makes it hard for users to see _actual_ versions.

image

The current workaround is to just leave it as is. 

Yes! But also open to mentor others who might be interested in doing this :)

jsjoeio:
This PR modifies the publish to npm workflow for development (commits to PRs and PRs) to use a different package name: `@coder/code-server-pr` as to not clutter the version list under the `code-server` name [on npm](https://www.npmjs.com/package/code-server). 

## Notes
- uses the [`npm pkg` CLI command](https://docs.npmjs.com/cli/v7/commands/npm-pkg) to modify `package.json` name

Fixes #4924

TODOs
- [ ] refactor to use `jq` instead of `npm`


.github/workflows/ci.yaml
```diff
@@ -205,6 +205,15 @@ jobs:
...
           path: release-npm-package
 
+      # NOTE@jsjoeio - we need to make sure we're using
+      # v7 or higher of the npm CLI because it's used
+      # in the yarn publish:npm script in development builds
+      # to modify package.json name.
+      - name: Install Node & npm v8
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.14.0"
```
jsjoeio, line 215: We don't need this specific version, but I think it's better to use a specific version so we it's consistent and we don't get any surprises from it changing (if we were to only specify "16")

jsjoeio: We also only need this in development, which is why I only changed it in `ci.yaml` and not the other places where we use `yarn publish:npm`

jawnsy: agreed! makes sense

ci/steps/publish-npm.sh
```diff
@@ -112,9 +112,15 @@ main() {
...
       NPM_VERSION="$VERSION-$PR_NUMBER-$COMMIT_SHA"
+      DEV_PACKAGE_NAME="@coder/code-server-pr"
       # This means the npm version will be tagged with "<pr number>"
       # and installed when a user runs `yarn install code-server@<pr number>`
       NPM_TAG="$PR_NUMBER"
+      # Use the development package name
+      # This is so we don't clutter the code-server versions on npm
+      # with development versions.
+      # NOTE: Requires npm Version 7.x or higher
+      npm pkg set name="$DEV_PACKAGE_NAME"
```
jawnsy, line 123: does this modify the package.json? neat!

jawnsy (approved):

.github/workflows/ci.yaml
jsjoeio: @jawnsy P.S. this is something I _specifically_ learned from you! ♥️

ci/steps/publish-npm.sh
jsjoeio: yessir! I guess they introduce it in v7, I think? Makes it a looot easier modifying the `package.json` vs. manually trying to modify with `jq` or `sed` 😂

code-asher (approved):
:tada:

jsjoeio:
In theory, this _should_ work 🤔 But when I search

It does publish but it's using the old name 🤔 
image

I think I know why. This line:

https://github.com/coder/code-server/blob/ed0a2a22e89c95b3ec21ceb86163bf8a58d5dc53/ci/steps/publish-npm.sh#L123

Needs to be after the 

https://github.com/coder/code-server/blob/ed0a2a22e89c95b3ec21ceb86163bf8a58d5dc53/ci/steps/publish-npm.sh#L133

Because the `package.json` has to be in the current path from where you run `npm pkg`. I think the best approach is to set the default name to `code-server` then always set it with `npm pkg` (but open to other ideas). Something like:

```sh
DEV_PACKAGE_NAME="code-server"
if [[ "$NPM_ENVIRONMENT" == "development" ]]; then
  DEV_PACKAGE_NAME="@coder/code-server-pr"
fi

pushd release
npm pkg set name="$DEV_PACKAGE_NAME"
npm version "$NPM_VERSION"
popd
```

Though if we do that, then we'll need the v7/v8 `npm` CLI in any place we use this script 🤔 Which is fine but I'll need to update it in the `npm-brew.yaml` file.

ci/steps/publish-npm.sh
```diff
@@ -51,6 +51,15 @@ main() {
     exit 1
   fi
 
+  # TODO@jsjoeio
+  # Check that we're using at least v7 of npm CLI
+  if ! command -v npm &> /dev/null; then
```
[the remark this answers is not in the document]
code-asher: Hmm could be tough since we would need to do semver checks.  It might be better to just run the `npm config set` command but on failure echo a message saying something like "failed to set package.json name, are you on v7 or higher?"

code-asher:
The Brew workflow will not trigger this `else` branch right?  Since `NPM_ENVIRONMENT` will be `PRODUCTION`.

ci/steps/publish-npm.sh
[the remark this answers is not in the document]
jsjoeio: That's a great alternative!

jsjoeio:
> Although the npm set is a great QOL improvement, maybe we should just use jq since we already use it elsewhere hahaha

I'm all about those QOL improvements though 😛 Plus less tools/maintenance overhead only having to worry about `npm`

jsjoeio:
> The Brew workflow will not trigger this else branch right? Since NPM_ENVIRONMENT will be PRODUCTION.

Exactly!

jsjoeio:
```text
error Couldn't publish package: "https://registry.yarnpkg.com/@coder%2fcode-server-pr: You must sign up for private packages"
info Visit https://yarnpkg.com/en/docs/cli/publish for documentation about this command.
```

Looks like I may need to use a different npm token 🤔

ci/steps/publish-npm.sh
```diff
@@ -51,6 +51,27 @@ main() {
...
+    echo "Are you sure you have it installed?"
+    echo "Please check again with: npm -v"
+    echo "And re-run the script."
+    exit 1
+  fi
+
+  # NOTE@jsjoeio - this needs to run inside the release dir
+  # where the package.json for code-server is.
+  pushd release
+  if npm pkg get name && [ $? -eq 1 ]; then
```
jsjoeio, line 66: @code-asher how's this?

code-asher:
> Plus less tools/maintenance overhead only having to worry about `npm`

I think this is the main (only?) argument in favor, but since we already have `jq` as a dep it weakens the argument in this particular case

jsjoeio:
> I think this is the main (only?) argument in favor, but since we already have jq as a dep it weakens the argument in this particular case

Fair point! If I use `jq`, I don't have to use npm v7 or above, so that'd save time in the install steps. And we're all in favor of faster workflows

ci/steps/publish-npm.sh
code-asher: I think this ends up using the exit code of `[` so it never enters the `if`.

I tested with:

```
bash -c 'if npm pkg get name && [ $? -eq 1 ]; then echo hello ; fi'
```

This should work:

```
if ! npm pkg get name ; then
```

We might also want to redirect the output because `npm` outputs a whole bunch of junk:

```
if ! npm pkg get name > /dev/null ; then
```

code-asher:
I think either way is not a big deal, the only pain point I can think of is if I have to release manually (like what happened last release :cry:) then I would need to update node/npm so I can run it then downgrade it again (since code-server needs a lower version).

jsjoeio:
> I think either way is not a big deal, the only pain point I can think of is if I have to release manually (like what happened last release 😢) then I would need to update node/npm so I can run it then downgrade it again (since code-server needs a lower version).

Well technically you could upgrade `npm` only with `npm i -g npm@7`. Using a newer version of `npm` wouldn't affect your use of code-server, but agreed, it is one extra step and we should avoid it if we can!

code-asher:
> Well technically you could upgrade `npm` only with `npm i -g npm@7`. Using a newer version of `npm` wouldn't affect your use of code-server, but agreed, it is one extra step and we should avoid it if we can!

:exploding_head: Good point!  That sounds like a pretty reasonable workaround IMO

jsjoeio:
> 🤯 Good point! That sounds like a pretty reasonable workaround IMO

Well...you convinced me to use `jq` so I'm changing it 🤣 In all honesty though, you're right in that we should limit dependencies and using `jq` would be a smaller code change, which I think we're all in favor of. No biggie.

code-asher:
hahahahahaha well here is a heart for you <3

I look forward to the day we can remove `jq` tbh, maybe by writing the scripts in TypeScript or calling out to Node or something

ci/steps/publish-npm.sh
code-asher: I know it no longer matters because we are using `jq` but there must be something else at play.  I tried changing `-eq 1` to `eq 0` and it still does not enter the `if` block.  What is bash doing to me?

code-asher: ohhhhhh right because `npm pkg get name` exits with 1 so it aborts the `&&`

jsjoeio: ohhh that's really strange 🤔 LOL well makes me glad we didn't go with this solution because I probably would have introduced a bug. I wonder if @deansheather knows

jsjoeio: Ahhh  that makes sense!

jsjoeio: Er, I knew that 😉

code-asher: ahahahahaha just keeping my sharp, I appreciate it!  :joy:

jsjoeio:
@code-asher bump

code-asher (approved):

PR #4979, opened 2022-03-10
feat: relaunch on SIGUSR2

code-asher:
This is because Node uses SIGUSR1 to enable the debug listener so even
if you just want to restart code-server you end up enabling the debug
listener as well.

Opted to leave the SIGUSR1 handler in to avoid breaking existing
workflows even though it does mean even if you only want to enable the
debug listener you will end up restarting code-server as well.  We could
consider removing it after a transition phase.


jsjoeio (approved):
Nice fix! What prompted this btw?

code-asher:
This has been mostly an internal thing so I have just lived with the debugger turning on but today I was asked how to restart code-server without rebuilding a workspace and it felt bad that I had to say "here is how...but the debugger will turn on" so I decided to fix it real swift.  :smile:

code-asher:
It has been on my mind for a very long time, should have done this ages ago.

jsjoeio:
> decrease coverage by 0.12%

😧

code-asher:
lol would probably be good to add a test for relaunching

jsjoeio:
bump @code-asher what should we do about this?

code-asher:
I was hoping to add tests although it will not actually increase coverage I guess because we have to spawn a separate process to test it.

jsjoeio:
@code-asher not to necrobump this, but what should we do with this?

code-asher:
Best options right now are probably to merge as-is or close; unlikely I can add tests right now :cry:

jsjoeio:
I'm okay with merging as is if you are!

PR #4995, opened 2022-03-15
fix(ci): correctly download npm/docker artifacts

### LINKED ISSUE #4949: [Chore]: fix npm workflow
> I think the `download-artifact` action is trying to download from that workflow rather than the workflow that ran on the version branch maybe?

> I suppose this could be why Anmol used a manual script for downloading instead; I tried looking for an option to download from a specific branch but am not seeing it (edited) 

> Really unfortunate though...maybe we can think of a new architecture where we can use download-artifact

See here: https://github.com/coder/code-server/runs/5424884882?check_suite_focus=true

jsjoeio:
## Context

This PR fixes the npm and docker workflows that runs on a release to correctly download artifacts so it can publish it to the npm and Docker registries. 

https://github.com/coder/code-server/blob/main/.github/workflows/npm-brew.yaml#L21-L37

### Problem

The default [`download-artifact`](https://github.com/actions/download-artifact) assumes that you're downloading artifacts from the running workflow. The problem is [`npm-brew.yaml`](https://github.com/coder/code-server/blob/main/.github/workflows/npm-brew.yaml#L21-L37) runs on [`released` events](https://github.com/coder/code-server/blob/main/.github/workflows/npm-brew.yaml#L8-L9).

Example: maintainer creates release branch `v4.1.0` ➡️ opens PR ➡️ team approves PR ➡️ maintainer merges PR into `main` ➡️ `npm` job from `npm-brew.yaml` and the `docker.yaml` workflows kick off. 

Therefore this workflow is not associated with any PR or any commit, but rather a GitHub event. 

### Solution

Download `npm-release` artifact and `release-packages` from last successful run where branch is named `v0.0.0` (with actual version) using this community GitHub Action [`dawidd6/action-download-artifact`](https://github.com/dawidd6/action-download-artifact).

#### Other considerations

_Why not use latest successful CI run for `main`?_

There would be a race condition since by the time the release PR gets merged and `main` finishes running, this npm workflow would already have run, so the last successful CI run for `main` wouldn't be 1:1 with the release branch merge commit.

_Why not use the same branch name each time like `release`?_

@vapurrmaid and @code-asher brought up a good point. If we do this, then we have no way to differentiate between a 4.1.0 and a 4.2.0 release workflow since they both use the same branch name. This is important because there are times we need to republish releases like on Docker and we'd want to use the same build. 

## Testing Plan

1. Fork `cdr/code-server`
2. Checkout this branch: https://github.com/jsjoeio/code-server/pull/2
3. Merge into `main` 
4. Create new branch called `release`
5. Get CI to run and pass.
6. Merge branch
7. Create new GitHub release
8. See if `npm-brew` runs and download-artifact works as expected: https://github.com/jsjoeio/code-server/runs/5562662740?check_suite_focus=true#step:3:16 

Fixes #4949

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -197,9 +197,12 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/download-artifact@v3
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@v2
         id: download
         with:
+          branch: release
+          workflow_conclusion: completed
           name: "npm-package"
           path: release-npm-package
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -198,7 +198,7 @@ jobs:
           fetch-depth: 0
 
       - name: Download artifact
-      - uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v3
         id: download
         with:
           name: "npm-package"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -198,11 +198,9 @@ jobs:
           fetch-depth: 0
 
       - name: Download artifact
-        uses: dawidd6/action-download-artifact@v2
+      - uses: actions/download-artifact@v3
         id: download
         with:
-          branch: release
-          workflow_conclusion: completed
           name: "npm-package"
           path: release-npm-package
 
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -35,15 +35,15 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
-      - name: Get version branch
-        id: version-branch
-        run: echo "::set-output name=name::v$(jq .version package.json)"
+      - name: Get version
+        id: version
+        run: echo "::set-output name=version::$(jq .version package.json)"
 
       - name: Download artifact
         uses: dawidd6/action-download-artifact@v2
         id: download
         with:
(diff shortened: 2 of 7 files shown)


jsjoeio:
Happy I tested this. The current approach didn't work 🤔

image

https://github.com/jsjoeio/code-server/runs/5562477153?check_suite_focus=true#step:3:11

vapurrmaid (commented):
I wonder if there's a way that it can match branches like `release-*`:

```yaml
with:
    branch: release-*
```

so that it doesn't rely on creating and deleting the same `release` branch over and over again.

🤔

jsjoeio:
> I wonder if there's a way that it can match branches like `release-*`:
> 
> ```yaml
> with:
>     branch: release-*
> ```
> 
> so that it doesn't rely on creating and deleting the same `release` branch over and over again.
> 
> 🤔

Doesn't seem like it from what I can tell, but there is an open PR from January: https://github.com/dawidd6/action-download-artifact/pull/107 but it's only for artifact names :/

.github/workflows/ci.yaml
```diff
@@ -197,9 +197,12 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/download-artifact@v3
+      - name: Download artifact
+        uses: dawidd6/action-download-artifact@v2
         id: download
         with:
+          branch: release
+          workflow_conclusion: completed
```
jsjoeio, line 205: Uh whoops...this was added to the wrong workflow. This is supposed to be on the `npm-brew.yaml` workflow

jsjoeio (commented):
Need to make changes

code-asher (approved):
:tada: :tada: :tada:

PR #4996, opened 2022-03-16
refactor: checkout homebrew-core in action instead of script

### LINKED ISSUE #4950: [Chore]: fix brew workflow
https://github.com/coder/code-server/runs/4974830516?check_suite_focus=true#step:5:142

jsjoeio:
This fixes the `brew-bump.sh` script by using the correct logic. It also moves a `git clone` line from the script to a step in the `homebrew` job.

## Testing Plan

1. Fork `coder/code-server`
2. hard-code `VERSION` to `4.1.0-4970-6afc87b46dfa70adc0afe164a89662aac1a0b0cd` (otherwise it will fail) and disable `npm` job
3. Push up this change to a new branch
4. Add `HOMEBREW_GITHUB_API_TOKEN` as secret to fork (using my own token).
5. Manually dispatch action
6. See that it opens a PR upstream in homebrew-core

✅ tested this with @cdr-oss account on Coder workspace
✅ tested with @cdrci account and it worked: https://github.com/Homebrew/homebrew-core/pull/97235 (oops, should have just used the `--dry-run` option)

## Other Notes
🔑 Updated the `HOMEBREW_GITHUB_API_TOKEN` under secrets with a PAT generated by @deansheather from the @cdrci account which has the scopes `repo` and `workflow`.

Fixes #4950

### DIFF
--- a/.github/workflows/npm-brew.yaml
+++ b/.github/workflows/npm-brew.yaml
@@ -49,11 +49,19 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
 
-      - uses: actions/checkout@v3
+      - name: Checkout code-server
+        uses: actions/checkout@v3
+
+      - name: Checkout coder/homebrew-core
+        uses: actions/checkout@v3
+        with:
+          repository: coder/homebrew-core
+
       - name: Configure git
         run: |
           git config user.name github-actions
           git config user.email github-actions@github.com
+
       - name: Bump code-server homebrew version
         env:
           HOMEBREW_GITHUB_API_TOKEN: ${{secrets.HOMEBREW_GITHUB_API_TOKEN}}
--- a/.github/workflows/npm-brew.yaml
+++ b/.github/workflows/npm-brew.yaml
@@ -52,10 +52,10 @@ jobs:
       - name: Checkout code-server
         uses: actions/checkout@v3
 
-      - name: Checkout coder/homebrew-core
+      - name: Checkout cdrci/homebrew-core
         uses: actions/checkout@v3
         with:
-          repository: coder/homebrew-core
+          repository: cdrci/homebrew-core
 
       - name: Configure git
         run: |
--- a/.github/workflows/npm-brew.yaml
+++ b/.github/workflows/npm-brew.yaml
@@ -63,6 +63,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: cdrci/homebrew-core
+          path: homebrew-core
 
       - name: Configure git
         run: |
--- a/ci/steps/brew-bump.sh
+++ b/ci/steps/brew-bump.sh
@@ -6,8 +6,8 @@ main() {
   GITHUB_USERNAME="cdrci"
   UPSTREAM_USERNAME_AND_REPO="Homebrew/$REPO"
   # Only sourcing this so we get access to $VERSION
-  source ./lib.sh
-  source ./steps-lib.sh
+  source ./ci/lib.sh
+  source ./ci/steps-lib.sh
 
   echo "Checking environment variables"
 
--- a/ci/steps/brew-bump.sh
+++ b/ci/steps/brew-bump.sh
@@ -7,7 +7,7 @@ main() {
   UPSTREAM_USERNAME_AND_REPO="Homebrew/$REPO"
   # Only sourcing this so we get access to $VERSION
   source ./ci/lib.sh
-  source ./ci/steps-lib.sh
+  source ./ci/steps/steps-lib.sh
 
   echo "Checking environment variables"
 
--- a/ci/steps/brew-bump.sh
+++ b/ci/steps/brew-bump.sh
@@ -89,16 +89,6 @@ main() {
   # https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L18
   # local output
   brew bump-formula-pr --version="${VERSION}" code-server --no-browse --no-audit
-
-  # Clean up and remove homebrew-core
-  popd
-  rm -rf "$REPO"
-
-  # Make sure $REPO is removed
-  if directory_exists "$REPO"; then
-    echo "rm -rf $REPO failed."
-    ls -la
-  fi
 }
 
 main "$@"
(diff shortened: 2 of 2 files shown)


jsjoeio:
## Notes

Ran out of time with asher but stopped at the step where we push latest changes to fork

ci/steps/brew-bump.sh
```diff
@@ -77,36 +70,33 @@ main() {
...
+  rm -rf "$REPO"
 
-  # Make sure homebrew-core is removed
-  if directory_exists "homebrew-core"; then
-    echo "rm -rf homebrew-core failed."
+  # Make sure $REPO is removed
+  if directory_exists "$REPO"; then
+    echo "rm -rf $REPO failed."
     ls -la
   fi
```
code-asher, line 101: I am thinking we might want to remove this, for example if I ran it locally could be weird that my clone gets removed, plus since we clone outside this script I think any cleanup could happen outside the script as well, but in this case it will get automatically cleaned up when the job ends.

ci/steps/brew-bump.sh
```diff
@@ -77,36 +70,33 @@ main() {
...
-  if ! output=$(brew bump-formula-pr --version="${VERSION}" code-server --no-browse --no-audit 2>&1); then
-    if [[ $output == *"Duplicate PRs should not be opened"* ]]; then
-      echo "$VERSION is already submitted"
-    else
-      echo "$output"
-      exit 1
-    fi
-  fi
+  # local output
+  brew bump-formula-pr --version="${VERSION}" code-server --no-browse --no-audit
```
code-asher, line 91: With this change we can no longer distinguish between a failure and the PR already existing but I am not sure this matters.  I think the question to ask is if we re-run this script and the version is already pushed do we want the job to be marked red or green?

.github/workflows/npm-brew.yaml
```diff
@@ -56,11 +56,20 @@ jobs:
...
 
-      - uses: actions/checkout@v3
+      - name: Checkout code-server
+        uses: actions/checkout@v3
+
+      - name: Checkout cdrci/homebrew-core
+        uses: actions/checkout@v3
+        with:
+          repository: cdrci/homebrew-core
+          path: homebrew-core
```
code-asher, line 66: Clever!  I like this over cloning in the script.

ci/steps/brew-bump.sh
```diff
@@ -2,10 +2,12 @@
...
 
 main() {
-  cd "$(dirname "$0")/../.."
+  REPO="homebrew-core"
+  GITHUB_USERNAME="cdrci"
+  UPSTREAM_USERNAME_AND_REPO="Homebrew/$REPO"
   # Only sourcing this so we get access to $VERSION
   source ./ci/lib.sh
-  source ./ci/steps/steps-lib.sh
+  source ./ci/steps-lib.sh
```
code-asher, line 10: Is this change correct?  Looks like the script is under the `steps` directory.

jsjoeio:
> question: With this change we can no longer distinguish between a failure and the PR already existing but I am not sure this matters. 

Agreed.  The original intent of that block was to prevent duplicate PRs since `yarn` wouldn't do that for us.

> I think the question to ask is if we re-run this script and the version is already pushed do we want the job to be marked red or green?

My thinking is that is should be marked as green. Reason? Because running this job again and having the version already in upstream isn't a failure and shouldn't be marked as red. It should only be red if it fails to push a version (meaning that version _never_ gets to Homebrew upstream).

In a world with more options, I would do this:
- failed to publish version? red.
- published version? green.
- skipped since version already published? yellow. (netural)

Red should be used to signal failures.

But to counter myself...we should match the error code of `brew bump-formula-pr` in which this case it would be `1` or red. This is the output:
```shell
➜  ~ brew bump-formula-pr --version="4.0.1" code-server --no-browse --no-audit  
Error: These pull requests may be duplicates:
code-server 4.0.1 https://github.com/Homebrew/homebrew-core/pull/92582
Duplicate PRs should not be opened. Use --force to override this error
```

Hmmm going back to the other side I feel like this should be a warning, not an error. I don't know. As you can tell, I can understand both sides. I can go either way.

code-asher:
> Agreed. The original intent of that block was to prevent duplicate PRs since `yarn` wouldn't do that for us.

A little different than that I think, it just checks whether `brew-formula-pr` aborted because the PR was a duplicate so we could exit with 0 instead of 1 (to be green instead of red).  The prevention itself was always delegated to `bump-formula-pr`.

> Hmmm going back to the other side I feel like this should be a warning, not an error. I don't know. As you can tell, I can understand both sides. I can go either way.

Pretty much the same thoughts going through my head.

I think...ultimately if the action goes red we have to take a look at it.  If we see it already uploaded then we have nothing to do, in which case we just wasted our time.  So maybe green would be the least frustrating.  I definitely agree if there was some kind of warning state that would be ideal.  Or if we could somehow make use of skipped steps.  But that would require we check beforehand whether the PR was already submitted in a previous step and if so skip the actual publish step.

jsjoeio:
> I think...ultimately if the action goes red we have to take a look at it. If we see it already uploaded then we have nothing to do, in which case we just wasted our time. **So maybe green would be the least frustrating**. I definitely agree if there was some kind of warning state that would be ideal. .

I agree, I think for now, I'll add back the check to do that. 

> Or if we could somehow make use of skipped steps. But that would require we check beforehand whether the PR was already submitted in a previous step and if so skip the actual publish step

Yeah, I think for now, it's okay but we could revisit this if we think we need that.

jsjoeio:
FWIW, I did look at the Homebrew docs. `brew info --json --formula code-server` returns JSON like this:

(omitted some info)
````json
[
  {
    "name": "code-server",
    "full_name": "code-server",
    "tap": "homebrew/core",
    "desc": "Access VS Code through the browser",
    "license": "MIT",
    "homepage": "https://github.com/cdr/code-server",
    "versions": {
      "stable": "4.1.0",
      "head": null,
      "bottle": true
  }
]
```

We could get the latest published version but we'd have to check for a PR somehow. Seems like a lot of work for not a lot of benefit so I think we'll go with make green if PR already open.

jsjoeio:
@code-asher ready for another review!

code-asher (approved):

jsjoeio:
Blocked by https://github.com/coder/code-server/pull/5006

PR #4997, opened 2022-03-17
chore: move to patches

code-asher:
Still a bit to go but opening the PR now for hype.

Closes https://github.com/coder/code-server/issues/3462
Closes https://github.com/coder/code-server/issues/4687
Fixes https://github.com/coder/code-server/issues/4826 (I changed the update check code a bit so this is now fixed)
Fixes https://github.com/coder/code-server/issues/4994


code-asher:
This is not a straight conversion from our existing patch, I streamlined a bunch of things and removed some things that become unnecessary in 1.64.

code-asher:
I think I got everything.  Let me know if any patches appear to be missing.

src/node/main.ts
```diff
@@ -23,28 +23,38 @@ export const shouldSpawnCliProcess = (args: UserProvidedArgs): boolean => {
   )
 }
 
+export interface OpenCommandPipeArgs {
```
jsjoeio, line 26: docs: can we add a description to this? Not sure what this is for without digging further

jsjoeio: Like I assume some "Open Command" but no idea what this is actually used for

src/node/routes/vscode.ts
```diff
@@ -1,19 +1,30 @@
...
+  handleServerError(err: Error): void
+  dispose(): void
+}
+
+export type CreateServer = (address: string | net.AddressInfo | null, args: CodeArgs) => Promise<IServerAPI>
+
 export class CodeServerRouteWrapper {
   /** Assigned in `ensureCodeServerLoaded` */
-  private _codeServerMain!: CodeServerLib.IServerAPI
+  private _codeServerMain!: IServerAPI
```
jsjoeio, line 26: question: not your code but why do we have an exclamation point in this name?

ci/build/build-release.sh
```diff
@@ -89,16 +89,39 @@ bundle_vscode() {
...
+    "dataFolderName": ".code-server",
+    "win32MutexName": "codeserver",
+    "licenseUrl": "https://github.com/cdr/code-server/blob/main/LICENSE.txt",
+    "win32DirName": "code-server",
+    "win32NameVersion": "code-server",
+    "win32AppUserModelId": "coder.code-server",
+    "win32ShellNameShort": "c&ode-server",
+    "darwinBundleIdentifier": "com.coder.code.server",
+    "linuxIconName": "com.coder.code.server",
+    "reportIssueUrl": "https://github.com/cdr/code-server/issues/new",
```
jsjoeio, line 93:
```suggestion
    "reportIssueUrl": "https://github.com/coder/code-server/issues/new",
```

src/node/routes/vscode.ts
```diff
@@ -92,17 +91,6 @@ export class CodeServerRouteWrapper {
   }
 
   private $proxyRequest: express.Handler = async (req, res, next) => {
-    // We allow certain errors to propagate so that other routers may handle requests
```
jsjoeio, line 84: deleted code is the best kind of code! 👏🏼

src/node/routes/vscode.ts
```diff
@@ -34,6 +34,32 @@ export class CodeServerRouteWrapper {
 
   //#region Route Handlers
 
+  private manifest: express.Handler = async (req, res, next) => {
```
jsjoeio, line 37: Nice work!

.github/workflows/ci.yaml
```diff
@@ -110,6 +110,12 @@ jobs:
           fetch-depth: 0
           submodules: true
 
+      - name: Install quilt
+        run: sudo apt update && sudo apt install quilt
```
jsjoeio, line 114: question: Do we need `-y` here?

.gitignore
```diff
@@ -8,12 +8,17 @@ release-packages/
...
-vendor/modules
-node-*
 /plugins
 /lib/coder-cloud-agent
 .home
 coverage
 **/.DS_Store
+
+# Code packages itself here.
+/lib/vscode-reh-web-linux-x64/
```
jsjoeio, line 18: question: are there other `vscode-reh-web-*` packages we need to ignore?

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "lib/vscode"]
 	path = lib/vscode
-	url = https://github.com/coder/vscode
+	url = https://github.com/microsoft/vscode
```
jsjoeio, line 3: sooo excited for this 😄

ci/build/build-packages.sh
```diff
@@ -50,6 +50,11 @@ release_nfpm() {
 
   export NFPM_ARCH
 
+  # Code deletes some files from the extension node_modules directory which
+  # leaves broken symlinks in the corresponding .bin directory.  nfpm will fail
+  # on these broken symlinks so clean them up.
```
jsjoeio, line 55: beautiful documentation ✨

docs/CONTRIBUTING.md
```diff
@@ -78,26 +81,44 @@ we'll guide you.
...
+7. Open a PR.
+
+### Patching Code
+
+0. You can go through the patch stack with `quilt push` and `quilt pop`.
+1. Create a new patch (`quilt new {name}.diff`) or use an existing patch.
+2. Add the file(s) you are patching (`quilt add [-P patch] {file}`). A file
+   **must** be added before you make changes to it.
+3. Make your changes. Patches do not need to be independent of each other but
+   each patch must result in a working code-server without any broken in-between
```
jsjoeio, line 116: Very glad you added the "each patch must result in a working code-server without any broken in-between" 👏🏼

docs/CONTRIBUTING.md
```diff
@@ -193,99 +214,46 @@ code-server running locally. In CI, this is taken care of for you.
...
+
+Microsoft eventually made the server open source and we were able to reduce our
+changes significantly. Some time later we moved back to a submodule and patches
+(managed by `quilt` this time instead of the mega-patch).
+
+As the web portion of Code continues to mature, we'll be able to shrink and
+possibly eliminate our patches. In the meantime, upgrading the Code version
+requires us to ensure that our changes are still applied correctly and work as
+intended. In the future, we'd like to run Code unit tests against our builds to
+ensure that features work as expected.
```
jsjoeio, line 249: Beautiful prose ✨ I really enjoyed reading this lol

patches/insecure-notification.diff
```diff
@@ -0,0 +1,57 @@
+Add a notification when accessing code-server in an insecure context
+
+This is done because otherwise when things like the clipboard do not work users
+may think code-server is broken.  Ideally there would be a notification at the
+point where these things are used instead of this though.
+
+To test access over something like an HTTP domain or an IP address (not
```
jsjoeio, line 7: Love that you have the "to test" parts in some of these 👏🏼

src/node/routes/index.ts
```diff
@@ -129,6 +129,11 @@ export const register = async (app: App, args: DefaultedArgs): Promise<Disposabl
     express.static(rootPath, {
       cacheControl: commit !== "development",
       fallthrough: false,
+      setHeaders: (res, path, stat) => {
+        if (path.endsWith("/serviceWorker.js")) {
+          res.setHeader("Service-Worker-Allowed", "/")
+        }
+      },
```
jsjoeio, line 136: question: what's this for?

jsjoeio (changes requested):
I think overall, this looks awesome! Most of my comments are non-blocking, but a few things we need to change for this to be ready.

I'm also going to test this locally and follow the docs to make sure it sounds good

Thanks for taking this on! 🙌🏼

src/node/cli.ts
```diff
@@ -761,10 +765,27 @@ export const shouldOpenInExistingInstance = async (args: UserProvidedArgs): Prom
   return undefined
 }
 
+/**
+ * Arguments for running Code's server.
```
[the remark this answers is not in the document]
code-asher: This is `Code` as in `Code OSS`, I am trying to avoid `VS Code` since that is the off-limits brand name.

src/node/cli.ts
```diff
@@ -761,10 +765,27 @@ export const shouldOpenInExistingInstance = async (args: UserProvidedArgs): Prom
...
+ */
+export interface CodeArgs extends UserProvidedCodeArgs {
```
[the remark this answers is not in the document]
code-asher: I wanted to do that but was concerned we would get it confused with `code-server` haha.  I really wish they had named it something other than `Code`.

src/node/cli.ts
```diff
@@ -761,10 +765,27 @@ export const shouldOpenInExistingInstance = async (args: UserProvidedArgs): Prom
...
+  "accept-server-license-terms"?: boolean
+  "connection-token"?: string
+  help: boolean
+  port: string
+  version: boolean
```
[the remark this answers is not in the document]
code-asher: I have no idea unfortunately.  Here are the matching types on Code's side:

https://github.com/microsoft/vscode/blob/8eb7d49eaacdac7495c4504feb103ad2d3f8f8e2/src/vs/server/node/serverEnvironmentService.ts#L194-L195

Actually `port` is not required so I should change that (I think it used to be).  Looks like I should add `compatibility` as well.

src/node/main.ts
code-asher: Will do.

src/node/cli.ts
```diff
@@ -774,6 +774,7 @@ export interface CodeArgs extends UserProvidedCodeArgs {
   help: boolean
   port: string
   version: boolean
+  "without-connection-token"?: boolean
```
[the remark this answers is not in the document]
code-asher: According to my tests it does!

.github/workflows/ci.yaml
code-asher: Huh, I guess not!  My guess is that it detects it is not an interactive terminal so it skips the question.

.gitignore
code-asher: Ah whoops good point.  This should be a wildcard.

docs/CONTRIBUTING.md
code-asher: Gotta have a working `git bisect`!!

docs/CONTRIBUTING.md
code-asher: To be fair most of this was written by oxy I think.

patches/insecure-notification.diff
code-asher: Glad you think so!  I wanted to have that on all of them but I missed some.

src/node/routes/index.ts
code-asher: The service worker is served from `/_static/out/browser/serviceWorker.js` so by default it is not allowed to register a scope above `/_static/out/browser`.  This lets it register on the root.

code-asher: Will add a comment.

src/node/main.ts
code-asher: Added a comment.  It would be cool if we could import this type directly from Code without having to import the code itself which will mess with the `out` directory structure.  I wonder if we can do this with `import type` but I have not tested it yet.  For now we can duplicate the types I suppose.

code-asher: Actually I am going to put a link to the type on the Code side instead.

src/node/cli.ts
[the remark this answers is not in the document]
code-asher: Oh yeah I misread your comment.  These are *not* for code-server, they are for (VS) Code's server.

src/node/routes/vscode.ts
code-asher: Looks like because we create it on-demand so it might be undefined but we want to use it in a few places without having to keep checking if it is undefined.

IMO this is incorrect usage, we should put it behind a function that checks if it is undefined and throws if so.

code-asher: I think we should make it illegal via eslint actually https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-non-null-assertion.md

code-asher: We could also use `?` instead if it is not important to error.

src/node/cli.ts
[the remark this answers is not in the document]
jsjoeio: ahhhh i thought that was a typo. nevermind!

jsjoeio:
Re: failing CI checks
- [non-blocking] Docs preview: @BrunoQuaresma any ideas why this is failing on forks? Might have to re-introduce skip 
- Code scanning - not something new with this code so I will dimiss
- codecov - 0.29% decrease. I don't want to block this so @code-asher what if you follow-up with some new tests in a separate PR? (just to cover difference)

jsjoeio (approved):
Nice work!!!

PR #5010, opened 2022-03-22
feat: Add a shrinkwrap file to the NPM release

### LINKED ISSUE #4927: [Feat]: Align the dependencies for binary and NPM artifact releases
This issue is to continue the discussion from #4918 regarding the different release methods of code-server, and how NPM artifacts work.

## What is your suggestion?
* Be in a state where the builds and installs are both deterministic when it comes to dependencies.
* Avoid having the dependencies being figured out at install time of the NPM package.

Right now, there's inconsistency on what dependencies might get used for the binaries generated (because they get generated after installing and building the package - which is an action that does respect the `yarn.lock` file) and the NPM artifacts (because the lockfiles are not published, nor the lockfiles for dependencies gets respected even if published).

This causes problems of dependency drifts like what was seen in #4900 - where releases might use different versions or worse, something working on release stops working because of newer versions being published.

Not for the end-user. Specific versions can be pinned under the `package.json` file in this repo, but not something controlled by the end-user.

Yes

edvincent:
Fixes #4927

Tried multiple ideas/approaches, this is the least worst I could find. Happy to amend/change things, and at least this will get the discussion going.

Here's a diff from an npm install before the change (left - installing from the public repos) and after (right - installing from the tar.gz generated by the build): https://editor.mergely.com/8tpzCXG6/

As expected, some versions have been bumped down, to actually follow what the yarn.lock had:
* `rotating-file-stream@3.0.3` => `rotating-file-stream@3.0.0` as per the [yarn.lock](https://github.com/coder/code-server/blob/be727871f6f7f0af48e9e3d0039d59aa46c855b5/yarn.lock#L3627)
* `argon2@0.28.5` => `argon2@0.28.4` as per the [yarn.lock](https://github.com/coder/code-server/blob/be727871f6f7f0af48e9e3d0039d59aa46c855b5/yarn.lock#L698)

For this approach, I'm introducing two new devDependencies:
* `synp`, which means we can seed the shrinkwrap from the yarn.lock - rather than maintaining two different ones.
* `json`, to do some simplifications on the shrinkwrap file. Effectively, this is similar to `jq` - but a bit simpler for this use-case. And also, opens the door to not have to require `jq` from being installed, as this is a pure NPM package that can be declared as a dependency. Might check later if I can easily migrate every `jq` command to it.

I tried using `npm shrinkwrap` directly, but the current NPM and NodeJS versions being used seem to choke when `resolutions` is used to force newer versions of transitive dependencies we don't depend on (they generate a package-lock.json without a version...). Not even sure why those resolutions are needed...?

### DIFF
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -55,6 +55,17 @@ bundle_code_server() {
 EOF
   ) > "$RELEASE_PATH/package.json"
   rsync yarn.lock "$RELEASE_PATH"
+
+  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we seed
+  # an npm-shrinkwrap file from our yarn lockfile and the current node-modules installed.
+  synp --source-file yarn.lock
+  npm shrinkwrap
+  # HACK@edvincent: The shrinkwrap file will contain the devDependencies, which by default
+  # are installed if present in a lockfile. To avoid every user having to specify --production
+  # to skip them, we carefully remove them from the shrinkwrap file.
+  json -f npm-shrinkwrap.json -I -e "Object.keys(this.dependencies).forEach(dependency => { if (this.dependencies[dependency].dev) { delete this.dependencies[dependency] } } )"
+  mv npm-shrinkwrap.json "$RELEASE_PATH"
+
   rsync ci/build/npm-postinstall.sh "$RELEASE_PATH/postinstall.sh"
 
   if [ "$KEEP_MODULES" = 1 ]; then
--- a/ci/steps/publish-npm.sh
+++ b/ci/steps/publish-npm.sh
@@ -59,11 +59,11 @@ main() {
   # This string is used to determine how we should tag the npm release.
   # Environment can be one of three choices:
   # "development" - this means we tag with the PR number, allowing
-  # a developer to install this version with `yarn add code-server@<pr-number>`
+  # a developer to install this version with `npm install code-server@<pr-number>`
   # "staging" - this means we tag with `beta`, allowing
-  # a developer to install this version with `yarn add code-server@beta`
+  # a developer to install this version with `npm install code-server@beta`
   # "production" - this means we tag with `latest` (default), allowing
-  # a developer to install this version with `yarn add code-server@latest`
+  # a developer to install this version with `npm install code-server@latest`
   if ! is_env_var_set "NPM_ENVIRONMENT"; then
     echo "NPM_ENVIRONMENT is not set. Determining in script based on GITHUB environment variables."
 
@@ -96,7 +96,7 @@ main() {
   if [[ "$NPM_ENVIRONMENT" == "production" ]]; then
     NPM_VERSION="$VERSION"
     # This means the npm version will be published as "stable"
-    # and installed when a user runs `yarn install code-server`
+    # and installed when a user runs `npm install code-server`
     NPM_TAG="latest"
   else
     COMMIT_SHA="$GITHUB_SHA"
@@ -107,7 +107,7 @@ main() {
     if [[ "$NPM_ENVIRONMENT" == "staging" ]]; then
       NPM_VERSION="$VERSION-beta-$COMMIT_SHA"
       # This means the npm version will be tagged with "beta"
-      # and installed when a user runs `yarn install code-server@beta`
+      # and installed when a user runs `npm install code-server@beta`
       NPM_TAG="beta"
     fi
 
@@ -117,7 +117,7 @@ main() {
       NPM_VERSION="$VERSION-$PR_NUMBER-$COMMIT_SHA"
       PACKAGE_NAME="@coder/code-server-pr"
       # This means the npm version will be tagged with "<pr number>"
-      # and installed when a user runs `yarn install code-server@<pr number>`
+      # and installed when a user runs `npm install code-server@<pr number>`
       NPM_TAG="$PR_NUMBER"
     fi
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -4,7 +4,7 @@
 
 - [install.sh](#installsh)
   - [Detection reference](#detection-reference)
-- [yarn, npm](#yarn-npm)
+- [npm](#npm)
 - [Standalone releases](#standalone-releases)
 - [Debian, Ubuntu](#debian-ubuntu)
 - [Fedora, CentOS, RHEL, SUSE](#fedora-centos-rhel-suse)
@@ -19,7 +19,7 @@
 - [Uninstall](#uninstall)
   - [install.sh](#installsh-1)
   - [Homebrew](#homebrew)
-  - [yarn, npm](#yarn-npm-1)
+  - [npm](#npm-1)
   - [Debian, Ubuntu](#debian-ubuntu-1)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -87,17 +87,16 @@ _exact_ same commands presented in the rest of this document.
 
   - Ensure that you add `~/.local/bin` to your `$PATH` to run code-server.
 
-- For FreeBSD, code-server will install the [npm package](#yarn-npm) with `yarn`
-  or `npm`.
+- For FreeBSD, code-server will install the [npm package](#npm) with `npm`.
 
 - If you're installing code-server onto architecture with no releases,
-  code-server will install the [npm package](#yarn-npm) with `yarn` or `npm`
+  code-server will install the [npm package](#npm) with `npm`
   - We currently offer releases for amd64 and arm64.
-  - The [npm package](#yarn-npm) builds the native modules on post-install.
+  - The [npm package](#npm) builds the native modules on post-install.
 
-## yarn, npm
+## npm
 
-We recommend installing with `yarn` or `npm` when:
+We recommend installing with `npm` when:
 
 1. You aren't using a machine with `amd64` or `arm64`.
 1. You are installing code-server on Windows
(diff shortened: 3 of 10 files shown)


jsjoeio:
> npm ERR! peer dep missing: @google-cloud/logging@^4.5.2, required by @coder/logger@1.1.16

I do see this in the diff 🤔 But it seems like the build and the tests are passing so I'm not 100% if we need to worry about it or not. Any ideas on why that's happening? 

> Not even sure why those resolutions are needed...?

If this is referring to the `resolutions` key in the `package.json`, it's for security reasons that we use it. Even though we don't depend on the dependencies directly, this ensures that when folks install code-server, they're not installing packages with security vulnerabilities. (at least that's my understanding)

ci/build/build-release.sh
```diff
@@ -55,6 +55,17 @@ bundle_code_server() {
...
   rsync yarn.lock "$RELEASE_PATH"
+
+  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we seed
+  # an npm-shrinkwrap file from our yarn lockfile and the current node-modules installed.
+  synp --source-file yarn.lock
+  npm shrinkwrap
+  # HACK: The shrinkwrap file will contain the devDependencies, which by default
+  # are installed if present in a lockfile. To avoid every user having to specify --production
+  # to skip them, we carefully remove them from the shrinkwrap file.
+  json -f npm-shrinkwrap.json -I -e "Object.keys(this.dependencies).forEach(dependency => { if (this.dependencies[dependency].dev) { delete this.dependencies[dependency] } } )"
```
jsjoeio, line 66: great inline docs 👏🏼

ci/steps/publish-npm.sh
```diff
@@ -59,11 +59,11 @@ main() {
   # This string is used to determine how we should tag the npm release.
   # Environment can be one of three choices:
   # "development" - this means we tag with the PR number, allowing
-  # a developer to install this version with `yarn add code-server@<pr-number>`
+  # a developer to install this version with `npm install -g code-server@<pr-number>`
```
jsjoeio, line 62: I think in this file, we don't need `-g` since before we were telling folks to install as a dependency (notice we don't use `yarn global add`). Guessing this was a find and replace though

docs/install.md
```diff
@@ -19,7 +19,7 @@
 - [Uninstall](#uninstall)
   - [install.sh](#installsh-1)
   - [Homebrew](#homebrew)
-  - [yarn, npm](#yarn-npm-1)
+  - [npm](#npm-1)
```
jsjoeio, line 22: Will this new heading `#npm-1` work? I didn't know that was a thing.

docs/npm.md
```diff
@@ -20,6 +20,11 @@ If you're installing code-server via `npm`, you'll need to install additional
 dependencies required to build the native modules used by VS Code. This article
 includes installing instructions based on your operating system.
 
+> **WARNING**: Do not use Yarn to install code-server. Unlike NPM, it does not respect
```
jsjoeio, line 23:
```suggestion
> **WARNING**: Do not use `yarn` to install code-server. Unlike `npm`, it does not respect
```

docs/npm.md
```diff
@@ -20,6 +20,11 @@ If you're installing code-server via `npm`, you'll need to install additional
...
+> lockfiles for distributed applications. It will instead use the latest version
+> available at installation time - which might not be the one used for a given
+> code-server release, and [might lead to unexpected behavior](https://github.com/coder/code-server/issues/4927).
```
jsjoeio, line 26: Nice addition 👏🏼

jsjoeio (commented):
Nice work and thank you for doing this! 👏🏼

I'll defer the approval to @code-asher

ci/steps/publish-npm.sh
edvincent: Ah damn yes, my bad! Great catch!

docs/install.md
edvincent: I don't see why not? This is the auto-generated one, and because there are two anchors named the same - one gets the `1` appended.

That was already happening before:
* https://coder.com/docs/code-server/latest/install#yarn-npm
* https://coder.com/docs/code-server/latest/install#yarn-npm-1

I didn't manage to figure out how to serve the documentation locally so couldn't really test... Any tips/guide on how to validate it?

ci/build/build-release.sh
edvincent: Yeah, explaining the why/context - especially on codebases with multiple different contributors - is something I think is extremely important. Thanks for the kudos ;D

ci/build/build-release.sh
```diff
@@ -55,6 +55,17 @@ bundle_code_server() {
 EOF
   ) > "$RELEASE_PATH/package.json"
   rsync yarn.lock "$RELEASE_PATH"
+
+  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we seed
+  # an npm-shrinkwrap file from our yarn lockfile and the current node-modules installed.
+  synp --source-file yarn.lock
+  npm shrinkwrap
+  # HACK: The shrinkwrap file will contain the devDependencies, which by default
```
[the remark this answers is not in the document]
edvincent: Hahahaha you mean - easier to know who to blame? ;D But yeah makes sense.

Thoughts on the overall idea/hack of removing devDependencies? I think NPM 7/8 will fix this, but with NPM 6 it would force people to use the `--production`... So IMO it's a worthwile hack?

docs/install.md
jsjoeio: Ah! I misread the diff and thought you manually added it. Sadly, docs preview doesn't work on forks since GH Actions won't let you share secrets with forks. At least we haven't had any luck.

Looks good to me - disregard! Sorry for the noise!

docs/npm.md
edvincent: :blush:

ci/build/build-release.sh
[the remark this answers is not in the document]
jsjoeio: LOLOL hey you said it not me 😂

I mean...it is hacky but it sounds like it's the best we can do without forcing people to use NPM 7/8. And since we're still forcing people to use Node v14, I don't think we should require NPM 7/8 otherwise you'd have to install Node v14 then upgrade NPM which sounds painful from a UX perspective. 

Would love to get @code-asher's thoughts though.

edvincent:
> I do see this in the diff 🤔 But it seems like the build and the tests are passing so I'm not 100% if we need to worry about it or not. Any ideas on why that's happening?

Since I started playing around with code-server (around 3.9.1 IIRC), I've always had this error showing - yet everything always works well.

My understanding [from the code](https://github.com/coder/ts-logger/blob/master/src/extender.ts) is that it works because it's technically an optional dependency of `@coder/logger` if you wanna log to GCP - so if it's not there, nothing bad happens. And that the behavior of `peerDependencies` is being leveraged to make it optional?

@code-asher can likely confirm?

I guess the nicer way to fix this is to make that extender (to log to GCP) a separate package, to avoid pulling the `@google-cloud/logging` package by default.

> If this is referring to the `resolutions` key in the `package.json`, it's for security reasons that we use it. Even though we don't depend on the dependencies directly, this ensures that when folks install code-server, they're not installing packages with security vulnerabilities. (at least that's my understanding)

Yep that makes sense actually now that you say it. I think NPM 6 is choking on lockfiles/shrinkwraps because they are in `resolutions` without being in `dependencies`.

jsjoeio:
bump @code-asher

code-asher:
> Might check later if I can easily migrate every jq command to it.

That would be brilliant.

> My understanding [from the code](https://github.com/coder/ts-logger/blob/master/src/extender.ts) is that it works because it's technically an optional dependency of `@coder/logger` if you wanna log to GCP - so if it's not there, nothing bad happens. And that the behavior of `peerDependencies` is being leveraged to make it optional?
> 
> @code-asher can likely confirm?

Yup.

> I guess the nicer way to fix this is to make that extender (to log to GCP) a separate package, to avoid pulling the `@google-cloud/logging` package by default.

tbh I am tempted to remove the GCP code from the logger entirely since we do not even use it anywhere :stuck_out_tongue:

docs/install.md
```diff
@@ -228,14 +227,13 @@ You can install code-server using the [Helm package manager](https://coder.com/d
 
 ## Windows
 
-We currently [do not publish Windows releases](https://github.com/coder/code-server/issues/1397). We recommend installing code-server onto Windows with [`yarn` or `npm`](#yarn-npm).
+We currently [do not publish Windows releases](https://github.com/coder/code-server/issues/1397). We recommend installing code-server onto Windows with [ `npm`](#npm).
```
code-asher, line 230:
```suggestion
We currently [do not publish Windows releases](https://github.com/coder/code-server/issues/1397). We recommend installing code-server onto Windows with [`npm`](#npm).
```

code-asher (approved):
Nice work!

Something we might want to consider in the future is to shrinkwrap Code's dependencies as well.

code-asher:
One thought just occurred to me, I am thinking we probably need to change the `yarn` commands to `npm` in the post-install script since we removed `yarn` as a dependency from the docs.

https://github.com/coder/code-server/blob/18e19d29e6ad858206b84176ad287e22c51da4cb/ci/build/npm-postinstall.sh#L92-L109

Which I suppose means we need to generate shrinkwraps for the Code package.json files sooner rather than later.

jsjoeio:
> Which I suppose means we need to generate shrinkwraps for the Code package.json files sooner rather than later.

Nice catch! Should we open an issue for this? Happy to do so, just lmk.

code-asher:
I  will stop being lazy and do it. :laughing:

code-asher:
https://github.com/coder/code-server/issues/5056

PR #5014, opened 2022-03-22
release: 4.2.0

### LINKED ISSUE #4942: release: v4.2.0
## Checklist

- [x] Assign to next release manager
- [ ] Close previous release milestone
- [ ] Create next release milestone
- [ ] Associate issue with next release milestone

jsjoeio:
This PR is to generate a new release of `code-server` at `v4.2.0`

## Screenshot

image

## TODOs

Follow "Publishing a release" steps in `ci/README.md`

- [ ] publish release and merge PR
- [ ] update the AUR package

Fixes #4942


ci/helm-chart/Chart.yaml
```diff
@@ -20,4 +20,4 @@ version: 2.2.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 4.1.0
+appVersion: 4.2.0
```
code-asher, line 23: I think we are supposed to bump `version` in this file as well.

code-asher (approved):
One spot I think we need to update, otherwise looks good!

ci/helm-chart/Chart.yaml
jsjoeio: ah! that must not be part of the release prep script. nice catch!

benz0li:
@jsjoeio Help > About reports `code-server: vUnknown`.

version

jsjoeio:
This looks like a bug: https://github.com/coder/code-server/pull/5014#issuecomment-1076026582

Do you mind opening a bug report for that? You can leave most of it blank (I know how to reproduce).

> @jsjoeio I have noticed, that the Assets for v4.2.0 are ~90 MB smaller than the ones for v4.1.0.
>
> ℹ️ There are no source map files in v4.2.0, although being pointed to in the JS files (e.g. lib/vscode/out/vs/loader.js).
>
> Should I open an issue regarding this?

Great question! I'm not sure if that was intentional 🤔 Could have missed a patch potentially. Any ideas @code-asher ?

code-asher:
Definitely not intentional.  I bet it is due to using Code's packaging task instead of cobbling it together ourselves.  They might be removing the source maps because they host them elsewhere.  So we might have to patch that removal out.

jsjoeio:
Ahhh gotcha. We can track here: https://github.com/coder/code-server/issues/5026

Thanks @benz0li for reporting!

PR #5039, opened 2022-03-29
fix: minor build fixes

code-asher:
See commits for list of fixes.

### DIFF
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -77,7 +77,10 @@ EOF
 
 bundle_vscode() {
   mkdir -p "$VSCODE_OUT_PATH"
-  rsync ./lib/vscode-reh-web-*/ "$VSCODE_OUT_PATH"
+
+  # - Some extensions have a .gitignore which excludes their built source from
+  #   the npm package so exclude any .gitignore files.
+  rsync -avh --exclude .gitignore ./lib/vscode-reh-web-*/ "$VSCODE_OUT_PATH"
 
   # Add the commit, date, our name, links, and enable telemetry. This just makes
   # telemetry available; telemetry can still be disabled by flag or setting.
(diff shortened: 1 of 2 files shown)


ci/build/build-release.sh
```diff
@@ -77,7 +77,12 @@ EOF
 
 bundle_vscode() {
   mkdir -p "$VSCODE_OUT_PATH"
-  rsync ./lib/vscode-reh-web-*/ "$VSCODE_OUT_PATH"
+
+  # - Some extensions have a .gitignore which excludes their built source from
+  #   the npm package so exclude any .gitignore files.
+  # - Exclude Node as we will add it ourselves for the standalone and will not
+  #   need it for the npm package.
+  rsync -avh --exclude .gitignore --exclude /node ./lib/vscode-reh-web-*/ "$VSCODE_OUT_PATH"
```
jsjoeio, line 85: Will this fix https://github.com/coder/code-server/issues/5019?

patches/last-opened.diff
```diff
@@ -10,7 +10,7 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 +++ code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
-@@ -350,19 +350,6 @@ class WorkspaceProvider implements IWork
+@@ -410,19 +410,6 @@ class WorkspaceProvider implements IWork
```
jsjoeio, line 13: Does this fix that weird space thing we were seeing?

jsjoeio (approved):
Might be worth adding a small description to the PR as well

patches/last-opened.diff
code-asher: I am not sure if we are talking about the same thing but this fixes the warning output by `quilt push -a` since the patch is slightly off.

jsjoeio: yes, exactly!

ci/build/build-release.sh
code-asher: I am not seeing a `.gitignore` in the `typescript-language-features` extension so probably not unfortunately :cry: 

Will comment more on that issue; planning on getting through all my GitHub notifs today

PR #5042, opened 2022-03-30
fix(workflows): docs-preview and npm job

jsjoeio:
This includes the `GITHUB_TOKEN` on the `docs-preview.yaml` workflow, which should fix it on forks I believe. I also added a step to the `npm` job to comment instructions for how to test the PR locally.

Fixes N/A


code-asher (approved):
Handy!

PR #5055, opened 2022-03-31
feat: add option for disabling file downloads

### LINKED ISSUE #2426: Add option for disabling file downloads
Being able to easily download might be an issue in some situations. See some comments on https://github.com/cdr/code-server/issues/1386.

Maybe it also makes sense to prevent the HTTP server from serving these files as well so you can't just `curl` them.

jsjoeio:
WIP

Fixes #2426

## Notes

<details>
  <summary>Click to expand</summary>
  
 Gonna put my notes in here

</details>


patches/disable-downloads.diff
```diff
@@ -0,0 +1,150 @@
...
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/workbench/services/environment/common/environmentService.ts
++++ code-server/lib/vscode/src/vs/workbench/services/environment/common/environmentService.ts
+@@ -44,6 +44,9 @@ export interface IWorkbenchEnvironmentSe
+ 	readonly filesToOpenOrCreate?: IPath[] | undefined;
+ 	readonly filesToDiff?: IPath[] | undefined;
+ 
++	// --- code-server
++	readonly isEnabledFileDownloads: boolean;
++
```
jsjoeio, line 110: I added it to the common `environmentService` since this seems to be the `environmentService` used in `contextkeys` [here](https://github.com/microsoft/vscode/blob/main/src/vs/workbench/browser/contextkeys.ts#L78) but that leads to type errors since I didn't add it to the other `environmentService`s. See here:
https://github.com/coder/code-server/runs/5796104643?check_suite_focus=true#step:11:17

So either:
1. I need to update the other `environmentService`s
2. I need to take a different approach
3. something else

src/node/cli.ts
```diff
@@ -556,6 +562,10 @@ export async function setDefaults(cliArgs: UserProvidedArgs, configArgs?: Config
   const proxyDomains = new Set((args["proxy-domain"] || []).map((d) => d.replace(/^\*\./, "")))
   args["proxy-domain"] = Array.from(proxyDomains)
 
+  if (!args["disable-file-downloads"]) {
```
jsjoeio, line 565: ❓ I wasn't sure where to put this. Happy to move if it should be set somewhere else.

code-asher: I think we can leave this out, for boolean CLI flags we assume not being set is the same as false.

src/node/cli.ts
```diff
@@ -157,6 +158,10 @@ export const options: Options<Required<UserProvidedArgs>> = {
       "Disable update check. Without this flag, code-server checks every 6 hours against the latest github release and \n" +
       "then notifies you once every week that a new release is available.",
   },
+  "disable-file-downloads": {
+    type: "boolean",
+    description: "Disable file downloads from Code. Defaults to true.",
```
code-asher, line 163: It looks like the default is to not disable downloads, right?  (Which I think is the right move.)

```suggestion
    description: "Disable file downloads from Code.",
```

test/e2e/models/CodeServer.ts
```diff
@@ -134,7 +134,7 @@ export class CodeServer {
       })
 
       proc.on("close", (code) => {
-        const error = new Error("code-server closed unexpectedly")
+        const error = new Error("code-server closed unexpectedly. Try running with LOG_LEVEL=debug to see more info.")
```
code-asher, line 137: Helpful!

test/e2e/downloadsDisabled.test.ts
```diff
@@ -0,0 +1,26 @@
+import * as path from "path"
```
code-asher, line 1: Thoughts on putting this in the same file as the other download test?  My thinking is that they test the same code path so they should go together.

patches/disable-downloads.diff
```diff
@@ -0,0 +1,150 @@
...
++import { SplitEditorsVertically, InEditorZenModeContext, ActiveEditorCanRevertContext, ActiveEditorGroupLockedContext, ActiveEditorCanSplitInGroupContext, SideBySideEditorActiveContext, AuxiliaryBarVisibleContext, SideBarVisibleContext, PanelAlignmentContext, PanelMaximizedContext, PanelVisibleContext, ActiveEditorContext, EditorsVisibleContext, TextCompareEditorVisibleContext, TextCompareEditorActiveContext, ActiveEditorGroupEmptyContext, MultipleEditorGroupsContext, EditorTabsVisibleContext, IsCenteredLayoutContext, ActiveEditorGroupIndexContext, ActiveEditorGroupLastContext, ActiveEditorReadonlyContext, EditorAreaVisibleContext, ActiveEditorAvailableEditorIdsContext, DirtyWorkingCopiesContext, EmptyWorkspaceSupportContext, EnterMultiRootWorkspaceSupportContext, HasWebFileSystemAccess, IsFullscreenContext, OpenFolderWorkspaceSupportContext, RemoteNameContext, VirtualWorkspaceContext, WorkbenchStateContext, WorkspaceFolderCountContext, PanelPositionContext, IsEnabledFileDownloads } from 'vs/workbench/common/contextkeys';
```
code-asher, line 86: Is there something missing from the patch?  I am not seeing `IsEnabledFileDownloads` in `common/contextkeys`.

test/e2e/downloadsDisabled.test.ts
jsjoeio: I thought splitting out would help but that's anecdotal. I'll move it!

patches/disable-downloads.diff
jsjoeio: Hmm sounds like that didn't get added to the patch. Let me try adding again

test/e2e/downloadsDisabled.test.ts
code-asher: Ohh very interesting point though, does Jest only run individual files in parallel?  I wonder if we can make it run `describe` blocks in parallel if so.

patches/disable-downloads.diff
jsjoeio: Ahh... there are two `contextkeys` files. I didn't add one of them. Fixed.

src/node/cli.ts
jsjoeio: Hmm...does that mean I need to update the tests then? My thinking was if it's not passed, we need to set this to `false`

But we don't do this for `help` which is another boolean flag with default set to `false`. Do I need to set it in `toCodeArgs` then?

test/e2e/downloadsDisabled.test.ts
jsjoeio: Well this is being run by Playwright since it's an e2e test. It sounds like tests in the same file are not parallelized by default: https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file

> By default, tests in a single file are run in order. If you have 
many
 independent tests in a single file, you might want to run them in parallel with [test.describe.parallel(title, callback)](https://playwright.dev/docs/api/class-test#test-describe-parallel).

patches/disable-downloads.diff
jsjoeio: @code-asher thoughts on this?

src/node/cli.ts
code-asher: Yup, you would need to update the tests.  We should not need it in `toCodeArgs`, we can do the same as with `disable-update-check`.  `disable-file-downloads` already allows undefined in `serverEnvironmentService` so no need to do anything.

code-asher: fwiw I think undefined semantically makes more sense than false because the user did not provide it (rather than them explicitly providing false).

test/e2e/downloadsDisabled.test.ts
code-asher: Ooo that looks promising.

patches/disable-downloads.diff
code-asher: Definitely will comment, I just wanted to get the missing context key to see how it all ties together and what can be done

code-asher: All right so since the context key is only used in the browser I think keeping this value in only the browser environment service makes the most sense.

To do that:

1. Remove `isEnabledFileDownloads` from this file (`common/environmentService`)
2. In `browser/contextkeys` instead of using `IWorkbenchEnvironmentService` use `IBrowserWorkbenchEnvironmentService` (replace in the constructor and the import).
3. Add `isEnabledFileDownloads: boolean` to `IBrowserWorkbenchEnvironmentService`

At this point the native environment service extends the browser environment service for some reason...seems like a bug to me so we could change it to not do that.  Or if that feels sketchy we could just add a `?` to `isEnabledFileDownloads`.

If we do the second option we also will need `=== true` to the `set` call in `browser/contextkeys` since it does not like the undefined.

code-asher: Although...maybe to minimize patching we could just add a `?` to this line here (plus the `=== true`)

code-asher: That seems to be how they do `remoteAuthority` which is web-only so even if that is not the most technically correct it does follow an existing pattern.

code-asher: nvm `remoteAuthority` is used in native as well, ignore me :stuck_out_tongue: 

But maybe for now we just go the minimal patch route and if we were to upstream we could do it more "properly"

jsjoeio: > At this point the native environment service extends the browser environment service for some reason...seems like a bug to me so we could change it to not do that

Yeah, that's what I was super confused about. Okay, I'm going to go this route! Thanks for elaborating and providing some steps on how we should tackle this 🙌🏼

test/e2e/downloadsDisabled.test.ts
jsjoeio: I couldn't figure out the parallelism stuff. I may have misunderstood it from the beginning. I believe that it's for parallelizing `test`s within a `describe` block. But maybe it can be used for [parallelizing test suites as well](https://stackoverflow.com/questions/70100608/how-to-run-test-suite-in-parallel-but-not-the-single-tests)?

Either way, I'll move tests into one file. Then later if we have evidence the e2e tests are running slowly and we want to speed them up, we can look into it then.

patches/disable-downloads.diff
```diff
@@ -0,0 +1,183 @@
...
++		@IBrowserWorkbenchEnvironmentService private readonly environmentService: IBrowserWorkbenchEnvironmentService,
+ 		@IEditorService private readonly editorService: IEditorService,
+ 		@IEditorResolverService private readonly editorResolverService: IEditorResolverService,
+ 		@IEditorGroupsService private readonly editorGroupService: IEditorGroupsService,
+@@ -194,6 +194,9 @@ export class WorkbenchContextKeysHandler
+ 		this.auxiliaryBarVisibleContext = AuxiliaryBarVisibleContext.bindTo(this.contextKeyService);
+ 		this.auxiliaryBarVisibleContext.set(this.layoutService.isVisible(Parts.AUXILIARYBAR_PART));
+ 
++		// code-server
++		IsEnabledFileDownloads.bindTo(this.contextKeyService).set(this.environmentService.isEnabledFileDownloads ?? true)
```
code-asher, line 129: The perfect use case for `??`, nice!

code-asher (dismissed):
Looks great!

I am not sure if we should try this now or if this is good enough but we might want to disable the download command itself because it is possible to bind the download command to a keybinding and then just use that to download files.

patches/disable-downloads.diff
jsjoeio: I couldn't remember the name and had to google it 😂 ([nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) for future me)

code-asher: Ahaha I never bothered remembering the name :laughing:

src/node/cli.ts
```diff
@@ -542,6 +542,10 @@ export async function setDefaults(cliArgs: UserProvidedArgs, configArgs?: Config
     args.password = process.env.PASSWORD
   }
 
+  if (process.env.CS_DISABLE_FILE_DOWNLOADS) {
```
jsjoeio, line 545: ✍️ This check only looks for the existence of the env var. As long as it is a truthy value (i.e. _not_ ""), it will set to `true`. Open to other approaches as well.

src/node/cli.ts
```diff
@@ -556,6 +560,7 @@ export async function setDefaults(cliArgs: UserProvidedArgs, configArgs?: Config
   delete process.env.PASSWORD
   delete process.env.HASHED_PASSWORD
   delete process.env.GITHUB_TOKEN
+  delete process.env.CS_DISABLE_FILE_DOWNLOADS
```
code-asher, line 563: I think in this case we actually let it stay.  We delete the others because they are secrets.

src/node/cli.ts
code-asher: Hmmm maybe we should do an explicit `true` and/or `1`, like `process.env.CS_DISABLE_FILE_DOWNLOADS === "true"`, just because it feels weird if you did `CS_DISABLE_FILE_DOWNLOADS=false code-server` and it still got enabled :stuck_out_tongue:

jsjoeio: Hahaha I knew there was a hole in this 🙈 Yeah, I like explicitness! I'll go with your suggestion

src/node/cli.ts
jsjoeio: Ah got it. I wasn't sure if we deleted all but since this isn't sensitive no need to delete.

code-asher: Plus if they run code-server in code-server that will mean the sub-code-server will also get downloads disabled which is probably the desired behavior (of course they could just unset the env var to get around that).

src/node/cli.ts
```diff
@@ -542,7 +542,7 @@ export async function setDefaults(cliArgs: UserProvidedArgs, configArgs?: Config
     args.password = process.env.PASSWORD
   }
 
-  if (process.env.CS_DISABLE_FILE_DOWNLOADS) {
+  if (process.env.CS_DISABLE_FILE_DOWNLOADS === "1") {
```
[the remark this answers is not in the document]
jsjoeio: I think it's okay for now. If they come back and say `"true"` we could add that as `||` right?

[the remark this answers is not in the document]
code-asher: True, sgtm

PR #5068, opened 2022-04-05
fix: Exclude .deb files from release Docker image

### LINKED ISSUE #4112: Smaller docker image for production?
Hello,
is it possible to create a smaller docker image for production purposes?
I would be happy to make a contribution.

dwahler:
Currently, the `codercom/code-server` Docker image includes a copy of the release's .deb package files, even though they're only used in an intermediate layer. What's more, it includes packages from every architecture, not just the one being released.

This patch fixes this by updating the Dockerfile to use a multi-stage build. First, the packages are copied from the local build context into a temporary directory in a "scratch" build stage. Then, this directory is mounted into the main build, so that the appropriate .deb file can be installed without ever being actually stored in the image.

In my testing (with the 4.2.0 packages for linux-amd64) this reduces the total compressed and uncompressed size by a fairly substantial amount:

| | Before | After | Change |
| --- | ---: | ---: | ---: |
| compressed | 637 MB | 268 MB | -58% |
| uncompressed | 1094 MB | 724 MB | -34% |

Partially fixes #4112

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -1,3 +1,8 @@
+# syntax=docker/dockerfile:experimental
+
+FROM scratch AS packages
+COPY release-packages/code-server*.deb /tmp/
+
 FROM debian:11
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


dwahler:
I guess the "semantic pull request" check is failing because it wants a prefix on the PR title, but I'm not sure what would be most appropriate. It's not really a "feature", so maybe "perf"?

code-asher:
Good question.  `perf` seems good to me although we do not have that as an option so we would need to add it.

https://github.com/coder/code-server/blob/5bc26e90cbf17c9e79d5646fd93838fc0082d9ab/.github/semantic.yaml#L37-L66

code-asher (approved):
Those are some excellent savings.  The multi-stage build is clever!  Thanks for the improvement!

jsjoeio:
Echoing Asher, those are some huge improvements. Thank you so much for doing this!

PR #5071, opened 2022-04-07
fix: Generate shrinkwraps for the bundled vscode

### LINKED ISSUE #5056: [Bug]: Use npm instead of yarn in postinstall script

### OS/Web Information

- Web Browser: n/a
- Local OS: n/a
- Remote OS: n/a
- Remote Architecture: n/a
- `code-server --version`: development

### Steps to Reproduce

1. `yarn build release`
2. Uninstall yarn
3. `cd release && npm install`

### Expected

Install completes successfully.

### Actual

Although I have not actually done this (uninstalling yarn is a pain), the postinstall script uses yarn so it will probably error that yarn cannot be found.

### Logs

_No response_

### Screenshot/Video

_No response_

### Does this issue happen in VS Code?

- [X] I am using HTTPS.

### Notes

To fix:

1. Replace `yarn` with `npm` in `npm-postinstall.sh`
2. Generate shrinkwrap files for:
    1. `lib/vscode/package.json`
    2. `lib/vscode/extensions/package.json`

edvincent:
Fixes #5056 #5174

> **Why are we removing `synp` after all?**

Because it only takes care of putting in the package-lock/shrinkwrap file the first level of dependencies, not the nested dependencies which we have from the yarn lockfile.

See the contents of https://unpkg.com/browse/code-server@4.2.0-5047-90c9adcded65d10004a9db71d0717cd90565baa3/npm-shrinkwrap.json, and how for example, `@mapbox/node-pre-gyp` is listed as a dependency of `argon2` - but is not present anywhere in the shrinkwrap file.

> **Why are we adding packages from `resolutions` in `devDependencies` too?**

What initially drove me to use `synp` was the fact that the `npm shrinkwrap` command would generate a lockfile that wasn't usable. Trying to install the package with that lockfile would fail with the following exception:

```
Installing Code dependencies...
[..]
npm ERR! Cannot read property 'requires' of undefined
```

This was actually due to `npm shrinkwrap` (in NPM 6) choking with some packages not being anywhere in the lockfile but being present in the `package.json`.

So I simply tracked down their dependency tree, saw they were coming from devDependencies, and added them there - which now prevents the `npm shrinkwrap`-generated lockfile from breaking, and allows us to remove `synp`. This doesn't change the size of the dependency tree, and should be able to be removed when there are higher versions of NPM used.

> **Why aren't we using `npm ci` to replace `yarn --frozen-lockfile`?**

NPM 6's `npm ci` doesn't deal very well with optional dependencies, which the bundled vscode needs for some Windows dependencies (namely, `windows-process-tree` and `@vscode/windows-registry`):
* The `npm shrinkwrap` command doesn't add them to the lockfile, because the CI runs from Linux, so it doesn't install them - and the commands generates the lockfile based on what's in `node_modules`.
* If we hacked the lockfile and added them there manually (with `"optional": true`), the install on a Linux machine would fail - because NPM 6's `npm ci` doesn't respect them from an `npm-shrinkwrap.json` file.
* If we don't add them and use `npm ci`, they are not installed - even on Windows machines.

What I settled for was using `npm install --production` (for now), that will always respect the lockfile version if an entry is there, and will try to get versions not listed there. This effectively means that **those two packages, on windows, won't be following the `yarn.lock` that is being used in the development package**, but that seemed less worse.

*Note: Another idea was to hackily add the packages to the lockfile + enhance the `postinstall.sh` script to conditionally use `--no-optional` if we're running on a non-Windows machine - which skips those optional dependencies. I wasn't convinced by either the hacking around of the lockfile nor this conditional... But happy to hear thoughts?*

> **Why don't we need the "Remove the `devDependencies`" hack anymore?**

Because we are now generating the shrinkwrap file after stuff has been installed with `yarn --production`, which means `npm shrinkwrap` will only list the production dependencies.

> **Why are we removing `json` after all?**
Because we don't need to remove the `devDependencies` anymore - which needed some more complex JSON mangling. Now we simply need to remove a line, which can be done with `sed`.

But also, because doing it with `json` gets a bit tricky, as `json` would need to be listed as an actual dependency (rather than `devDependency`) of code-server and vscode too to be able to be used... Not worth it.

Switching the `jq` commands to `json` might still be worth it, but no need to conflate the commits.

> **Why no more `yarn pack` artifact?**
Realized I was trying to fix/conflate too many things into one PR, so will likely create a different issue to discuss this one.

### DIFF
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -18,6 +18,8 @@ main() {
   VSCODE_SRC_PATH="lib/vscode"
   VSCODE_OUT_PATH="$RELEASE_PATH/lib/vscode"
 
+  create_shrinkwraps
+
   mkdir -p "$RELEASE_PATH"
 
   bundle_code_server
@@ -55,15 +57,6 @@ bundle_code_server() {
 EOF
   ) > "$RELEASE_PATH/package.json"
   rsync yarn.lock "$RELEASE_PATH"
-
-  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we seed
-  # an npm-shrinkwrap file from our yarn lockfile and the current node-modules installed.
-  synp --source-file yarn.lock
-  npm shrinkwrap
-  # HACK@edvincent: The shrinkwrap file will contain the devDependencies, which by default
-  # are installed if present in a lockfile. To avoid every user having to specify --production
-  # to skip them, we carefully remove them from the shrinkwrap file.
-  json -f npm-shrinkwrap.json -I -e "Object.keys(this.dependencies).forEach(dependency => { if (this.dependencies[dependency].dev) { delete this.dependencies[dependency] } } )"
   mv npm-shrinkwrap.json "$RELEASE_PATH"
 
   rsync ci/build/npm-postinstall.sh "$RELEASE_PATH/postinstall.sh"
@@ -105,11 +98,44 @@ bundle_vscode() {
     "$VSCODE_SRC_PATH/package.json" > "$VSCODE_OUT_PATH/package.json"
 
   rsync "$VSCODE_SRC_PATH/remote/yarn.lock" "$VSCODE_OUT_PATH/yarn.lock"
+  mv "$VSCODE_SRC_PATH/remote/npm-shrinkwrap.json" "$VSCODE_OUT_PATH/npm-shrinkwrap.json"
 
   # Include global extension dependencies as well.
   rsync "$VSCODE_SRC_PATH/extensions/package.json" "$VSCODE_OUT_PATH/extensions/package.json"
   rsync "$VSCODE_SRC_PATH/extensions/yarn.lock" "$VSCODE_OUT_PATH/extensions/yarn.lock"
+  mv "$VSCODE_SRC_PATH/extensions/npm-shrinkwrap.json" "$VSCODE_OUT_PATH/extensions/npm-shrinkwrap.json"
   rsync "$VSCODE_SRC_PATH/extensions/postinstall.mjs" "$VSCODE_OUT_PATH/extensions/postinstall.mjs"
+}
+
+create_shrinkwraps() {
+  # yarn.lock or package-lock.json files (used to ensure deterministic versions of dependencies) are
+  # not packaged when publishing to the NPM registry.
+  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we create
+  # an npm-shrinkwrap.json file from the currently installed node_modules. This ensures the versions used
+  # from development (that the yarn.lock guarantees) are also the ones installed by end-users.
+  # These will include devDependencies, but those will be ignored when installing globally (for code-server), and
+  # because we use --omit=dev when installing vscode.
+
+  # We first generate the shrinkwrap file for code-server itself - which is the current directory
+  create_shrinkwrap_keeping_yarn_lock
+
+  # Then the shrinkwrap files for the bundled VSCode
+  pushd "$VSCODE_SRC_PATH/remote/"
+  create_shrinkwrap_keeping_yarn_lock
+  popd
+
+  pushd "$VSCODE_SRC_PATH/extensions/"
+  create_shrinkwrap_keeping_yarn_lock
+  popd
+}
+
+create_shrinkwrap_keeping_yarn_lock() {
+  # HACK@edvincent: Generating a shrinkwrap alters the yarn.lock which we don't want (with NPM URLs rather than the Yarn URLs)
+  # But to generate a valid shrinkwrap, it has to exist... So we copy it to then restore it
+  cp yarn.lock yarn.lock.temp
+  npm shrinkwrap
+  cp yarn.lock.temp yarn.lock
+  rm yarn.lock.temp
 }
 
 main "$@"
--- a/ci/build/npm-postinstall.sh
+++ b/ci/build/npm-postinstall.sh
@@ -110,7 +110,7 @@ main() {
     echo "Failed to download cloud agent; --link will not work"
   fi
 
-  if ! vscode_yarn; then
+  if ! vscode_install; then
     echo "You may not have the required dependencies to build the native modules."
     echo "Please see https://github.com/coder/code-server/blob/main/docs/npm.md"
     exit 1
@@ -123,17 +123,44 @@ main() {
   fi
 }
 
-vscode_yarn() {
+install_with_yarn_or_npm() {
+  # NOTE@edvincent: We want to keep using the package manager that the end-user was using to install the package.
+  # This also ensures that when *we* run `yarn` in the development process, the yarn.lock file is used.
+  case "${npm_config_user_agent-}" in
+    yarn*)
+      if [ -f "yarn.lock" ]; then
+        yarn --production --frozen-lockfile --no-default-rc
+      else
+        echo "yarn.lock file not present, not running in development mode. use npm to install code-server!"
+        exit 1
+      fi
+      ;;
+    npm*)
+      if [ -f "yarn.lock" ]; then
+        echo "yarn.lock file present, running in development mode. use yarn to install code-server!"
+        exit 1
+      else
+        npm install --omit=dev
+      fi
+      ;;
+    *)
+      echo "Could not determine which package manager is being used to install code-server"
+      exit 1
+      ;;
+  esac
+}
+
+vscode_install() {
   echo 'Installing Code dependencies...'
   cd lib/vscode
-  yarn --production --frozen-lockfile --no-default-rc
+  install_with_yarn_or_npm
 
   symlink_asar
   symlink_bin_script remote-cli code code-server
   symlink_bin_script helpers browser browser .sh
 
   cd extensions
-  yarn --production --frozen-lockfile
+  install_with_yarn_or_npm
 }
 
 main "$@"
--- a/package.json
+++ b/package.json
@@ -59,13 +59,11 @@
     "eslint-import-resolver-typescript": "^2.5.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-prettier": "^4.0.0",
-    "json": "^11.0.0",
     "prettier": "^2.2.1",
     "prettier-plugin-sh": "^0.12.0",
     "shellcheck": "^1.0.0",
     "stylelint": "^13.0.0",
     "stylelint-config-recommended": "^5.0.0",
-    "synp": "^1.9.10",
     "ts-node": "^10.0.0",
     "typescript": "^4.6.2"
   },
@@ -108,8 +106,7 @@
     "semver": "^7.1.3",
     "split2": "^4.0.0",
     "ws": "^8.0.0",
-    "xdg-basedir": "^4.0.0",
-    "yarn": "^1.22.4"
+    "xdg-basedir": "^4.0.0"
   },
   "bin": {
     "code-server": "out/node/entry.js"
--- a/yarn.lock
+++ b/yarn.lock
@@ -587,11 +587,6 @@
     "@typescript-eslint/types" "5.23.0"
     eslint-visitor-keys "^3.0.0"
 
-"@yarnpkg/lockfile@^1.1.0":
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz#e77a97fbd345b76d83245edcd17d393b1b41fb31"
-  integrity sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==
-
 JSONStream@^1.3.5:
   version "1.3.5"
   resolved "https://registry.yarnpkg.com/JSONStream/-/JSONStream-1.3.5.tgz#3208c1f08d3a4d99261ab64f92302bc15e111ca0"
@@ -731,11 +726,6 @@ argparse@^2.0.1:
   resolved "https://registry.yarnpkg.com/argparse/-/argparse-2.0.1.tgz#246f50f3ca78a3240f6c997e8a9bd1eac49e4b38"
   integrity sha512-8+9WqebbFzpX9OR+Wa6O29asIogeRMzcGtAINdpMHHyAg10f05aSFVBbcEqGf/PXw1EjAZ+q2/bEBg3DvurK3Q==
 
-arr-union@^3.1.0:
-  version "3.1.0"
-  resolved "https://registry.yarnpkg.com/arr-union/-/arr-union-3.1.0.tgz#e39b09aea9def866a8f206e288af63919bae39c4"
-  integrity sha1-45sJrqne+Gao8gbiiK9jkZuuOcQ=
-
 array-flatten@2.1.1:
   version "2.1.1"
   resolved "https://registry.yarnpkg.com/array-flatten/-/array-flatten-2.1.1.tgz#426bb9da84090c1838d812c8150af20a8331e296"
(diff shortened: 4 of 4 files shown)


edvincent:
So the failure seems to be because of some dependencies missing in the `node_modules` from the original install... Which my guess is because something is failing to be installed. Will be playing around with it (expect a bunch of updates to this PR to "test" the CI) and try to find the culprit.

code-asher:
GitHub seems to have ignored my email response so I will paste it below, apologies if it does eventually come in and the comment gets duplicated:

----

I love the detailed description!

> What is used to install is completely meaningless to the end-user. If
> only, by not using yarn even when invoked with yarn, we can guarantee
> the deterministic versions for the bundled vscode. Thoughts?

My first thought when I mentioned this was that the user might have
`yarn` installed but not `npm` so switching the package manager from
under the user could cause installation failures (`npm: command not found` 
or something like that).

Buuuut...is it likely someone has Node installed but not npm?  I know
some distros package them separately but it still seems unlikely.

The second thought I had was that using the same package manager
throughout the entire installation process /feels/ more correct to me.
I doubt it would cause issues in practice but it seems like it would be
less of a surprise ("Wait why is it calling npm when I ran yarn?").

.github/workflows/ci.yaml
```diff
@@ -110,6 +110,9 @@ jobs:
           fetch-depth: 0
           submodules: true
 
+      - name: Install development tools
+        run: sudo apt install -y build-essential pkg-config g++ libx11-dev libxkbfile-dev libsecret-1-dev
```
edvincent, line 114: I suspect this was the culprit of the failures: because `npm shrinkwrap` generates the shrinkwrap file from the existing `node_modules` folder (to extract the versions that had been installed), if an optional module failed to install (e.g. because some build dependencies like X11 etc.. were missing) - it wouldn't be in the shrinkwrap file.

Unfortunately, looks like I need some more approvals again to be able to run the workflows and validate this... @jsjoeio @code-asher if you can unblock the run(s) that'd be awesome!

ci/build/build-release.sh
```diff
@@ -153,4 +145,34 @@ EOF
...
+  # to skip them, we carefully remove them from the shrinkwrap file.
+  json -f npm-shrinkwrap.json -I -e "Object.keys(this.dependencies).forEach(dependency => { if (this.dependencies[dependency].dev) { delete this.dependencies[dependency] } } )"
+  mv npm-shrinkwrap.json "$RELEASE_PATH"
+
+  # Then the shrinkwrap files for the bundled VSCode
+  # We don't need to remove the devDependencies for these because we control how it's installed - and
+  # as such we can force the --production flag
+  cd lib/vscode/
+  npm shrinkwrap
+  json -f npm-shrinkwrap.json -I -e "Object.keys(this.dependencies).forEach(dependency => { if (this.dependencies[dependency].dev) { delete this.dependencies[dependency] } } )"
```
code-asher, line 168: My bad; I should have made a note of this earlier; we recently switching to using `lib/vscode/remote/package.json` since it turns out the upper `package.json` is for desktop VS Code and one in `remote` is for the web version (which is what code-server is running).

Could we generate the shrinkwraps directly in `$RELEASE_PATH`?  That way it is always using the right `package.json` no matter what we end up copying into it.

edvincent: Yup let me try that. For now mostly trying to get it to build more than anything, so no probs on other things.

edvincent:
Ok new tentative out, with:
1. Using `npm` or `yarn`, depending on what was initially used.
1. Generating the shrinkwraps when generating the release, rather than initially.
1. Removing `@parcel/node-addon-api` from the shrinkwrap, as it's a folder in `node_modules` which is actually not a package and was making the whole process choke.

(1) and (2) together (should) also help the other issue I was trying to track down, which happens because the tests were failing for the standalone release as it wasn't installing some of the devDependencies that we were removing too early from the shrinkwrap file.

edvincent:
FWIW, I'm playing around in https://github.com/edvincent/code-server/pull/1#partial-pull-merging to be able to run the CI/CD and pinpoint some of the culprits. To avoid the flood and needing approvals for the workflows to run. Sorry it's taking so long, but trying to do it the best possible way this time 💪

edvincent:
Ok, I think I got it this time! Moved to generate the shrinkwraps in the `release:standalone` command, which basically allows not having to know which `package.json` is being copied etc...

Also updated the overview of this PR to cover some of the reasoning.

Thoughts? I think I still need to add something so that the NPM publish CI step calls this release:standalone too?

code-asher:
Awesome work!  I love your comments; very thorough and made it super easy to understand the rationale.

> I think I still need to add something so that the NPM publish CI step calls this release:standalone too?

I think leaving it in `build-release.sh` is the right move.  We use the result of that script in both the NPM publish step and standalone step.

code-asher:
Ohh I see in your PR comment you are doing it there because it runs after `yarn --production`.  Hmm...

edvincent:
Yep.. We need the `node_modules` to be populated to be able to generate the shrinkwrap file...

The other solution would be to still do it in `build-release.sh`, which means we'd need to `yarn install` in `build-release.sh`. That would address this `TODO` https://github.com/coder/code-server/blob/main/.github/workflows/ci.yaml#L240-L242 - and the NPM release would move to be a `yarn pack` (rather than a raw `tar` command). But I'm not sure how would having the `node_modules` folder from one arch would affect another arch - for the cross-compile sections...

Unless we do it in `build-release.sh`, with a `yarn install`, but then generate two type of artifacts - the current one (for the other platforms) and the current one...

Thoughts?

code-asher:
Maybe we could just add `yarn --production` to `build-release.sh` and it would not take too much longer since it only needs to remove files (but I have no idea how yarn works internally so maybe this takes longer for some reason).

code-asher:
Oh right we have no Node modules at all at that stage, we would need to remove that whole `KEEP_MODULES` behavior and just keep the modules (in addition to then running `yarn --production` at the end if I understand correctly).

code-asher:
> But I'm not sure how would having the node_modules folder from one arch would affect another arch

This is exactly why I have avoided doing that todo so far lol

edvincent:
Today there is never a `yarn` in `build-release.sh` - so a `yarn --production` won't be removing files, but also installing stuff that was never installed.

The only `yarn` that gets done is at the root of the repo, not inside the `release` folder no?

EDIT: Yep, what you just said now. Happy to try and take a stab at that if it seems to be the best.

code-asher:
> Today there is never a yarn in build-release.sh

Ah yup it is done in a previous step and then this step has an option (`KEEP_MODULES`) for whether to also copy the Node modules or skip them.

code-asher:
Yeah I say we go ahead with this plan.

edvincent:
Damn it's actually a bit more nuanced than just removing the `KEEP_MODULES` behavior... If we run `yarn --production`, it would also install some of the things we were excluding in the `rsync` like the `node` executable, the `node_modules.ara` and the `lib/coder-cloud-agent`. Which means the `tar -czf` command in the `ci.yaml` would need to replicate/hard-code some of those excludes...

The other solution would be to come back to the original idea of generating the shrinkwrap files from the code-server repo itself, and copy them into release (after removing the devDependencies from it). i.e. we'd need to cd into `lib/vscode/remote/` - rather than relying on the fact we copied the right `package.json`.

BTW - Right now, that `KEEP_MODULES` never seems used?

code-asher:
Er wait I might have misunderstood.  We do already run `yarn --production` so those things exist in the standalone already but now the npm package will include them when it did not before maybe?  In that case maybe we add to the npmignore.

edvincent:
No, I haven't forgotten about this/you ❤️ life got in the way...

I'm back trying to fix this, sent a new approach PR (which gets a bit easier thanks to NodeJS 16!). Still a WIP and needs some tweaks, just sharing for awareness. Testing in https://github.com/edvincent/code-server/actions/runs/2883629793

HOWEVER, something that is a bit of a concern to me now seems to be a bug in NPM itself... Despite the shrinkwrap file, installing with the `global` flag still doesn't seem to respect them 😭

edvincent:
FYI - opened https://github.com/npm/cli/issues/5325 to track that nasty bug about the shrinkwrap file not being respected.

I'll still continue with the PR here (need to clean-up some of the code/options, I used `--no-default-rc` everywhere but supposed to only be for one of the folders) and we likely will be able to merge it - as the result will be the same as the current state: every install installs the latest version. But as soon as the bug upstream will be fixed, stuff will start working with no action :p

Any thoughts on the approach proposed in the latest code?

ci/build/build-standalone-release.sh
```diff
@@ -28,8 +28,31 @@ main() {
   ln -s "./lib/node" "$RELEASE_PATH/node"
 
   pushd "$RELEASE_PATH"
-  yarn --production --frozen-lockfile
+  yarn --frozen-lockfile
+  create_shrinkwraps
   popd
 }
 
+create_shrinkwraps() {
```
code-asher, line 36: The `build-standalone-release.sh` script takes the npm package generated in a previous step, installs deps, then packages the result so we can upload a standalone build complete with deps to GitHub.  Assuming I have that right (sometimes I get confused with all the steps) generating a shrinkwrap here would not get uploaded to npm.

So I think we will have to do this in `build-release.sh` which is where we create and upload the npm package artifact that eventually gets uploaded to npm.

And then here we would install with `npm` instead of `yarn`.

ci/build/npm-postinstall.sh
```diff
@@ -123,17 +123,44 @@ main() {
   fi
 }
 
-vscode_yarn() {
+install_with_yarn_or_npm() {
+  # NOTE@edvincent: We want to keep using the package manager that the end-user was using to install the package.
+  # This also ensures that when *we* run `yarn` in the development process, the yarn.lock file is used.
```
code-asher, line 128: This is dope!

ci/build/npm-postinstall.sh
```diff
@@ -123,17 +123,44 @@ main() {
...
+  case "${npm_config_user_agent-}" in
+    yarn*)
+      if [ -f "yarn.lock" ]; then
+        yarn --production --frozen-lockfile --no-default-rc
+      else
+        echo "yarn.lock file not present, not running in development mode. use npm should be used to install code-server!"
```
code-asher, line 134:
```suggestion
        echo "yarn.lock file not present, not running in development mode. use npm to install code-server!"
```

ci/build/build-standalone-release.sh
```diff
@@ -28,8 +28,31 @@ main() {
...
+  # not packaged when publishing to the NPM registry.
+  # To ensure deterministic dependency versions (even when code-server is installed with NPM), we create
+  # an npm-shrinkwrap.json file from the currently installed node_modules. This ensures the versions used
+  # from development (that the yarn.lock guarantees) are also the ones installed by end-users.
+
+  # We first generate the shrinkwrap file for code-server itself - from being in $RELEASE_PATH
+  npm shrinkwrap
+
+  # Then the shrinkwrap files for the bundled VSCode
+  # We don't need to remove the devDependencies for these because we control how it's installed - and
```
code-asher, line 47: > We don't need to remove the devDependencies for these

That makes sense, does that mean we should be removing them for the root dependencies?  It looks like we are including dev deps right now in the root shrinkwrap (I might be missing something!).

code-asher (commented):
Welcome back from life!  :laughing:

This looks really good.

ci/build/build-standalone-release.sh
edvincent: Ah damn yes, now I think I remember being bugged/blocked by something like that in the past... Gonna double-check that now!

ci/build/build-standalone-release.sh
edvincent: So the expectation is that when installing something with `--global` infers only keeping the prod dependencies. So rather than manipulating the shrinkwrap to remove them, we should be clear to leave them there and let `npm` do its thing.

That being said, there seems to be another nasty bug with that... https://github.com/npm/cli/issues/4323

But I'm not sure that removing them from the shrinkwrap won't simply cause NPM to try to download them anyway... Let me play around!

code-asher: Ohhhh that makes sense!  Definitely agreed that we should just let `npm` do its thing.

ci/build/npm-postinstall.sh
edvincent: Ah yes good catch, thanks!

edvincent:
Ok... This is what it looks like when doing it from the release script... Which is indeed where it's needed, as the tgz we use to publish are the artifacts from it - not from release standalone. Confirmed from the artifacts the shrinkwrap files and their content from downloading the build artifacts.

It looks a bit worse because shrinkwrapping tries to alter the `yarn.lock` - so we have to keep it and replace it again. But seems to be the least-worst way - the other being doing a `release:standalone` before publishing, which would need to install again all the development 
dependencies etc...

ci/build/build-standalone-release.sh
edvincent: > And then here we would install with `npm` instead of `yarn`.

Can I suggest to avoid doing too many things in the same PR if that works for you:
* (This PR) Just adding the `npm-shrinkwrap.json` file - which basically should not change anything else in the development/CI workflow.
* Another PR to change the `release:standalone` and other CI scripts to start using `npm ci`
* Another PR (when everything works) to amend the doc to recommend `npm` (which can be done last, as using yarn will still always work).

code-asher: I like the incremental approach!

ci/build/build-standalone-release.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+#!/usr/bin/env bash
```
code-asher, line 2: Looks like an extra hashbang snuck in.
```suggestion
```

code-asher (approved):
:tada:

ci/build/build-standalone-release.sh
edvincent: Yup - already sending an amend completely removing any diff on that file as not needed anymore.

code-asher: Awesome~

ci/build/npm-postinstall.sh
```diff
@@ -123,17 +123,48 @@ main() {
...
 
-vscode_yarn() {
+install_with_yarn_or_npm() {
+  # NOTE@edvincent: We want to keep using the package manager that the end-user was using to install the package.
+  # This also ensures that when *we* run `yarn` in the development process, the yarn.lock file is used.
+  case "${npm_config_user_agent-}" in
+    yarn*)
+      if [ -f "yarn.lock" ]; then
+        if [[ $PWD/ = */lib/vscode/ ]]; then
+          yarn --production --frozen-lockfile --no-default-rc
```
edvincent, line 133: This is just replicating the current behavior introduced in https://github.com/coder/code-server/pull/5135/commits/7579dbbb06ba6c4b9a3e9ad94cdccfde15fcb5c7

@jsjoeio Remember why it's needed for `lib/vscode` but not for `lib/vscode/extensions`? Happy to add a comment to explain why this is needed, likely useful :D

code-asher: My memory is fuzzy so @jsjoeio can confirm but I think it was because of one or more of the following:

- The `.yarnrc` (which targets Electron) is in `lib/vscode` so it might not affect `lib/vscode/extensions` (idk if yarn travels upwards looking for a `.yarnrc` but it might).
- Even if the `.yarnrc` does affect `lib/vscode/extensions`, that directory has no native modules so it would have no influence anyway

That said, there is probably no harm in always including `--no-default-rc` for all our `yarn` calls.

edvincent: > That said, there is probably no harm in always including `--no-default-rc` for all our `yarn` calls.

Reverting to that then! (Which will fix my bad `bash` syntax that just caused a failed build too).

Reasoning makes sense. And yes, `yarn` does travel the parent directories looking for other `.yarnrc` files - and merges them - as per https://classic.yarnpkg.com/en/docs/yarnrc/

code-asher: Aha that makes sense, thanks for the link!

jsjoeio: I should have added a comment because my memory is fuzzy too 😅 Sounds like @code-asher nailed it though.

code-asher:
Everything looks good to me so I am going to merge it in.  Thanks a million!

jsjoeio:
@edvincent really can't thank you enough - it's rare someone from the community takes on a hard problem, takes a break because life, then comes back to finish it 💪🏼 We appreciate you a ton!

edvincent:
Anytime! I like giving back to the community too, especially for stuff I use daily! Plus I got to learn a LOT about NPM's dependency management, which I enjoyed. And thank to both of you for bearing with me during the (long) process.

You can also thank @bpmct and the nice t-shirt he sent me couple of months ago, it definitely played a role in reminding me I still had this PR pending every time I wore it. 👀

Will still send a follow-up PR (likely next week) for the documentation to remind folks to use `npm` rather than `yarn`.

edvincent:
Also, whilst confirming the fix of https://github.com/coder/code-server/issues/5174#event-7234295102, realized that the bug https://github.com/npm/cli/issues/5325 is actually less of an issue here - because it doesn't seem to be buggy when installing it from a remote repo!

**Which means... There now is a way to install `code-server` with fully-deterministic dependencies, which will be the same whether `npm` or a packaged version is used. 🎉🎉**

code-asher:
Awesome news!!

code-asher:
I love the detailed description!

> What is used to install is completely meaningless to the end-user. If
> only, by not using yarn even when invoked with yarn, we can guarantee
> the deterministic versions for the bundled vscode. Thoughts?

My first thought when I mentioned this was that the user might have
`yarn` installed but not `npm` so switching the package manager from
under the user could cause installation failures (`npm: command not
found` or something like that).

Buuuut...is it likely someone has Node installed but not npm?  I know
some distros package them separately but it still seems unlikely.

The second thought I had was that using the same package manager
throughout the entire installation process /feels/ more correct to me.
I doubt it would cause issues in practice but it seems like it would be
less of a surprise ("Wait why is it calling npm when I ran yarn?").

code-asher:
Ignore this lol GitHub seems to finally be posting comments I made via email months ago

PR #5079, opened 2022-04-09
docs: Update ios.md

rileycalhoun:
Fixes #
NodeJS and NPM say "Illegal Instruction" if repositories are not downgraded


jsjoeio (approved):
Thank you for doing this!

PR #5097, opened 2022-04-13
fix: skip flaky update test

jsjoeio:
I've noticed this test fail multiple times. I think we should skip it until we set aside time to rewrite it.

Evidence: https://github.com/coder/code-server/runs/5999223188?check_suite_focus=true#step:12:13

Fixes N/A

### DIFF
--- a/test/unit/node/update.test.ts
+++ b/test/unit/node/update.test.ts
@@ -122,7 +122,8 @@ describe("update", () => {
     spy = []
   })
 
-  it("should get the latest", async () => {
+  // TODO@jsjoeio: rewrite this test so it's not flaky
+  it.skip("should get the latest", async () => {
     version = "2.1.0"
 
     const p = provider()


code-asher (approved):
Will approve but IMO we should not skip tests

jsjoeio:
> Will approve but IMO we should not skip tests

I agree! This is a temporary fix. We shouldn't add tests that are flaky but sometimes we don't realize until it's already merged. I will make sure this tracked and addressed in the next version/sprint!

jsjoeio:
Ironically, another test in that suite failed 🤦🏼‍♂️

https://github.com/coder/code-server/runs/6012334860?check_suite_focus=true#step:12:13

Should I delete the entire suite?

jsjoeio:
https://github.com/jest-community/eslint-plugin-jest also seems relevant to enforcing good testing rules (i.e. no skipping tests)

code-asher:
Ah sorry I missed your response!  The Slack integration has betrayed me and I am only now going through emails.

This is a long-standing one.  It has gone through at least one cycle of being enabled and disabled but no one has ever sat down to figure out why it actually flakes.

I think the appropriate response is to keep running the tests until they pass, as frustrating as that is, and make fixing the flake a priority.

One thing we should do is add some output around the actual values being compared since just seeing `true` and `false` is not very helpful in determining why the test is failing.

jsjoeio:
Ahhh...Okay that makes sense. I'm going to close this for now.

PR #5103, opened 2022-04-14
fix: webviews failing to load the iframe HTML

code-asher:
Code added in 1.65.0 broke webviews when you are hosting them from the
same domain.

Really should add an e2e test as well but I do not have the time right now.

### DIFF
--- a/patches/webview.diff
+++ b/patches/webview.diff
@@ -6,11 +6,14 @@ self-hosted.
 When doing this CSP will block resources (for example when viewing images) so
 add 'self' to the CSP to fix that.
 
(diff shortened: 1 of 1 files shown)


jsjoeio:
> Really should add an e2e test as well but I do not have the time right now.

Don't worry, I will take care of that later! (Or make tickets and either of us can do it, planning to do it for all patches).

jsjoeio (approved):
Appreciate you updating the description/comment as well! 🙌🏼

jsjoeio (approved):
Appreciate your thoroughness!

PR #5115, opened 2022-04-18
chore: provide details when `update.checked` tests fail

code-asher:
Using the toBe* functions will let us know what the actual values are
rather than just telling us true does not equal false.

Hopefully when this flakes again this helps us devise a fix.


jsjoeio (approved):
Great fix!

PR #5122, opened 2022-04-21
feat: add tests for node/heart.ts

jsjoeio:
This PR adds tests to increase coverage for `src/node/heart.ts`.

Related to https://github.com/coder/code-server/issues/5121


src/node/heart.ts
```diff
@@ -55,3 +45,20 @@ export class Heart {
...
 }
+
+/**
+ * Helper function for the heartbeatTimer.
+ *
+ * If heartbeat is active, call beat. Otherwise do nothing.
+ *
+ * Extracted to make it easier to test.
+ */
+export async function heartbeatTimer(isActive: Heart["isActive"], beat: Heart["beat"]) {
```
jsjoeio, line 56: `isActive` is a private readonly property so that's why I had to pass in individual functions instead of `heart`.

test/unit/node/heart.test.ts
```diff
@@ -0,0 +1,99 @@
+import { readFile, writeFile } from "fs/promises"
+import { logger } from "@coder/logger"
+import { Heart, heartbeatTimer } from "../../../src/node/heart"
+import { clean, mockLogger, tmpdir } from "../../utils/helpers"
+
+function mockIsActive(resolveTo: boolean): () => Promise<boolean> {
+  return () => new Promise((resolve, reject) => setTimeout(() => resolve(resolveTo), 100))
+}
```
jsjoeio, line 8: Couldn't decide if this should be defined in the `describe` block of the `Heart` test suite or outside. By being outside, it's only defined once and easier to read the suite (but the suite only runs once anyway 🤷🏼‍♂️).

test/unit/node/heart.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(logger.warn).toHaveBeenCalled()
+  })
+  it("should be active after calling beat", () => {
+    heart.beat()
+
+    const isAlive = heart.alive()
+    expect(isAlive).toBe(true)
+  })
+  it("should not be active after dispose is called", () => {
+    heart = new Heart(`${testDir}/shutdown.txt`, mockIsActive(true))
```
jsjoeio, line 63: Don't need this

src/node/heart.ts
```diff
@@ -33,17 +33,7 @@ export class Heart {
...
-        .then((active) => {
-          if (active) {
-            this.beat()
-          }
-        })
-        .catch((error) => {
-          logger.warn(error.message)
-        })
-    }, this.heartbeatInterval)
+    this.heartbeatTimer = setTimeout(async () => await heartbeatTimer(this.isActive, this.beat), this.heartbeatInterval)
```
code-asher, line 36: Optional: this is equivalent to

```suggestion
    this.heartbeatTimer = setTimeout(() => heartbeatTimer(this.isActive, this.beat), this.heartbeatInterval)
```

minus an extra async wrapper

src/node/heart.ts
code-asher: I do think we should long-term consider testing this holistically instead since what we really want to know is if the heart works as a whole but we can look into it later.

test/unit/node/heart.test.ts
code-asher: Outside makes sense to me :+1: 

Could we `return Promise.resolve(resolveTo)`?  Might be nice to avoid an extra 100ms wait each time.  Or the `jest.fn().mockResolvedValue` you use below, that looks pretty dope.

test/unit/node/heart.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    // Check that the heart wrote to the heartbeatFilePath and overwrote our text
+    const fileContentsAfterBeat = await readFile(pathToFile, { encoding: "utf8" })
+    expect(fileContentsAfterBeat).not.toBe(text)
+  })
+  it("should log a warning when given an invalid file path", async () => {
+    heart = new Heart(`fakeDir/fake.txt`, mockIsActive(false))
+    heart.beat()
+    // HACK@jsjoeio - beat has some async logic but is not an async method
+    // Therefore, we have to create an artificial wait in order to make sure
+    // all async code has completed before asserting
```
code-asher, line 51: This looks good to me for now.  Maybe long-term we can make `beat` `async`, at least up to the file write.

test/unit/node/heart.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    const pathToFile = `${testDir}/file.txt`
+    await writeFile(pathToFile, text)
+    const fileContents = await readFile(pathToFile, { encoding: "utf8" })
+    expect(fileContents).toBe(text)
+
+    heart = new Heart(pathToFile, mockIsActive(true))
+    heart.beat()
+    // Check that the heart wrote to the heartbeatFilePath and overwrote our text
+    const fileContentsAfterBeat = await readFile(pathToFile, { encoding: "utf8" })
+    expect(fileContentsAfterBeat).not.toBe(text)
```
code-asher, line 44: One thing we might consider testing is that the timestamp updated (this is how to tell if code-server is active) rather than the contents (or in addition to the contents but the contents can technically be anything so idk if they need to be tested).

code-asher (approved):
:tada:

test/unit/node/heart.test.ts
jsjoeio: Yes! I briefly looked into it for this PR but since it happens in a request, I wasn't sure if changing it [here](https://github.com/coder/code-server/blob/a1af9e2a5646df1a587f73fbe1455ca28824aeee/src/node/routes/index.ts#L55-L60) to make the handler `async` would causes so I didn't.

test/unit/node/heart.test.ts
jsjoeio: I don't know why I didn't think of that lol good call!

test/unit/node/heart.test.ts
jsjoeio: Oh as in write a test that makes a call to the `/healthz` endpoint? I went with the simple approach for this one but agree, that would be worth testing later as well!

test/unit/node/heart.test.ts
code-asher: Ah yeah I think the move would be to only `await` in tests.  For other uses we can leave them as-is (since we do not want to delay the request to write to the heartbeat file).

test/unit/node/heart.test.ts
code-asher: Ah nah I mean something like (untested):

```typescript
const before = Date.now()
...
// Make sure the modified timestamp was updated.
const newStat = await fs.stat(pathToFile)
expect(newStat.modified).toBeGreaterThan(before)
```

Or something like that

test/unit/node/heart.test.ts
jsjoeio: Ah gotcha! I did use `await` when I was first writing the tests but TS said it would have no effect lol but I guess that's because it couldn't infer the callback logic? I'm not sure lol

test/unit/node/heart.test.ts
jsjoeio: OH! I was wondering how you could do something like that. Let me play around with that.

jsjoeio: It worked! Had to modify slightly but you got me 90% of the way there 😎

```typescript
import { stat } from "fs/promises"
const before = Date.now()
heart.beat()
// Make sure the modified timestamp was updated.
const status = await fs.stat(pathToFile)
expect(status.mtimeMs).toBeGreaterThan(before)
```

test/unit/node/heart.test.ts
code-asher: Ahh yup we would need to make `beat` `async` first and then add the `await` in the tests.

test/unit/node/heart.test.ts
code-asher: Awesome!  :tada:

test/unit/node/heart.test.ts
code-asher: But I think the current workaround is chill

test/unit/node/heart.test.ts
code-asher: It runs so fast we might need greater than or equal lol

jsjoeio: I haven't had coffee yet so a little slow but realized that as well. Super odd? Getting before and after seems like the best solution to me. Thanks for calling that out!

code-asher: It could be that async issue, we might be issuing the stat before the write completes

jsjoeio: Oh right! Well...should I add that HACK I used into this test as well? I did run it locally and it passed, but it could be flakey.

code-asher: That could work, it is weird though because the content check passes so it does seem like the write finished.  :thinking:

jsjoeio: Astute observation! I almost forgot about that. Okay well if CI passes, I'm going to merge. If it fails, I'll add the hack.

PR #5131, opened 2022-04-26
refactor: add timeout for race condition in heart test

jsjoeio:
@code-asher was right. https://github.com/coder/code-server/pull/5122 introduced a flakey test due to async logic in `beat`. Added an artificial timeout in test to ensure no race conditions and test always passes.

Had to re-run Build job here: https://github.com/coder/code-server/actions/runs/2228063603

Fixes N/A

### DIFF
--- a/test/unit/node/heart.test.ts
+++ b/test/unit/node/heart.test.ts
@@ -38,12 +38,16 @@ describe("Heart", () => {
 
     heart = new Heart(pathToFile, mockIsActive(true))
     heart.beat()
+    // HACK@jsjoeio - beat has some async logic but is not an async method
+    // Therefore, we have to create an artificial wait in order to make sure
+    // all async code has completed before asserting
+    await new Promise((r) => setTimeout(r, 100))
     // Check that the heart wrote to the heartbeatFilePath and overwrote our text
     const fileContentsAfterBeat = await readFile(pathToFile, { encoding: "utf8" })
     expect(fileContentsAfterBeat).not.toBe(text)
     // Make sure the modified timestamp was updated.
     const fileStatusAfterEdit = await stat(pathToFile)
-    expect(fileStatusAfterEdit.mtimeMs).toBeGreaterThan(fileStatusBeforeEdit.mtimeMs)
+    expect(fileStatusAfterEdit.mtimeMs).toBeGreaterThanOrEqual(fileStatusBeforeEdit.mtimeMs)
   })
   it("should log a warning when given an invalid file path", async () => {
     heart = new Heart(`fakeDir/fake.txt`, mockIsActive(false))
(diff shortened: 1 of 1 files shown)


test/unit/node/heart.test.ts
```diff
@@ -38,12 +38,16 @@ describe("Heart", () => {
...
+    // Therefore, we have to create an artificial wait in order to make sure
+    // all async code has completed before asserting
+    await new Promise((r) => setTimeout(r, 100))
     // Check that the heart wrote to the heartbeatFilePath and overwrote our text
     const fileContentsAfterBeat = await readFile(pathToFile, { encoding: "utf8" })
     expect(fileContentsAfterBeat).not.toBe(text)
     // Make sure the modified timestamp was updated.
     const fileStatusAfterEdit = await stat(pathToFile)
-    expect(fileStatusAfterEdit.mtimeMs).toBeGreaterThan(fileStatusBeforeEdit.mtimeMs)
+    expect(fileStatusAfterEdit.mtimeMs).toBeGreaterThanOrEqual(fileStatusBeforeEdit.mtimeMs)
```
code-asher, line 50: It occurs to me that if we allow it to be equal then technically we are not testing anything because this will pass if the file is not modified.  Any idea if this still flakes if we remove the equal?

code-asher: Maybe what we can do is set the modified time explicitly before the heartbeat and then do the comparison?

https://nodejs.org/api/fs.html#filehandleutimesatime-mtime

jsjoeio: Oh good point. Let me remove the Equal and check. I also like the idea of explicitly setting. That gives us more control.

test/unit/node/heart.test.ts
```diff
@@ -33,17 +33,26 @@ describe("Heart", () => {
     const pathToFile = `${testDir}/file.txt`
     await writeFile(pathToFile, text)
     const fileContents = await readFile(pathToFile, { encoding: "utf8" })
-    const fileStatusBeforeEdit = await stat(pathToFile)
+    // Explicitly set the modified time to 0 so that we can check
+    // that the file was indeed modified after calling heart.beat().
+    // This works around any potential race conditions.
+    const fileHandle = await open(pathToFile, "r+")
+    await fileHandle.utimes(0, 0)
```
code-asher, line 40: Whoops, sorry I linked you to the wrong section that makes you open a file handle.  I think you might have to close this as well?  But it might be easier to this one: https://nodejs.org/api/fs.html#fspromisesutimespath-atime-mtime

jsjoeio: Oh, no, that's my bad. I assumed you could only access `utimes` after opening the file and didn't bother checking the rest of the docs. Yeah, that would leak since I don't have logic to close it (I've run into that with Deno).

Nice!!! Our codebase thanks you for your attention-to-detail 😂

test/unit/node/heart.test.ts
```diff
@@ -1,5 +1,5 @@
 import { logger } from "@coder/logger"
-import { readFile, writeFile, stat } from "fs/promises"
+import { readFile, writeFile, stat, open, utimes } from "fs/promises"
```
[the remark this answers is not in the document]
jsjoeio: Oh whoops! VS Code should remove those for you 😛 Removing then merging

PR #5133, opened 2022-04-26
fix: listening on IPv6 address not possible

### LINKED ISSUE #1582: Listening on INADDR6_ANY not possible
Hi,

when starting the server while trying to listen on INADDR6_ANY, also known as "::", the server errors out. It's not accepting a simple "::". According to RFC4038 the addresses should be placed in brackets, even when not using a port, so the "::" turns into "[::]", doesn't work either.

After switching to the --bind-addr parameter the situation got worse, because now, using "::" together with a port isn't possible anymore. When following RFC4038 I'm neither able to use "[::]", nor "[::]:1234".

Same happens with INADDR6_LOOPBACK, e.g. "::1" or "[::1]" or "[::1]:1234".

KR,

G.

awfulcooking:
Wrap IPv6 addresses in square brackets for URL in ensureAddress.

Fixes #1582 (regression)

Thanks ;-)

### DIFF
--- a/src/node/app.ts
+++ b/src/node/app.ts
@@ -94,7 +94,8 @@ export const ensureAddress = (server: http.Server, protocol: string): URL | stri
   }
 
   if (typeof addr !== "string") {
-    return new URL(`${protocol}://${addr.address}:${addr.port}`)
+    const host = addr.family === "IPv6" ? `[${addr.address}]` : addr.address
+    return new URL(`${protocol}://${host}:${addr.port}`)
   }
 
   // If this is a string then it is a pipe or Unix socket.
--- a/test/unit/node/app.test.ts
+++ b/test/unit/node/app.test.ts
@@ -152,10 +152,20 @@ describe("ensureAddress", () => {
   it("should throw and error if no address", () => {
     expect(() => ensureAddress(mockServer, "http")).toThrow("Server has no address")
   })
-  it("should return the address if it exists", async () => {
-    mockServer.address = () => "http://localhost:8080/"
+  it("should return the address if it's a string", async () => {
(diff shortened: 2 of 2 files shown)


src/node/app.ts
```diff
@@ -94,7 +94,8 @@ export const ensureAddress = (server: http.Server, protocol: string): URL | stri
   }
 
   if (typeof addr !== "string") {
-    return new URL(`${protocol}://${addr.address}:${addr.port}`)
+    const host = addr.family === "IPv6" ? `[${addr.address}]` : addr.address
```
jsjoeio, line 97: Love how simple this is 👏🏼

test/unit/node/app.test.ts
```diff
@@ -152,10 +152,20 @@ describe("ensureAddress", () => {
...
+  })
+  it("should construct URL with an IPv4 address", async () => {
+    mockServer.address = () => ({ address: "1.2.3.4", port: 5678, family: "IPv4" })
+    const address = ensureAddress(mockServer, "http")
+    expect(address.toString()).toBe(`http://1.2.3.4:5678/`)
+  })
+  it("should construct URL with an IPv6 address", async () => {
+    mockServer.address = () => ({ address: "a:b:c:d::1234", port: 5678, family: "IPv6" })
+    const address = ensureAddress(mockServer, "http")
+    expect(address.toString()).toBe(`http://[a:b:c:d::1234]:5678/`)
```
jsjoeio, line 168: appreciate you adding more tests as well 👏🏼

jsjoeio (approved):
This looks great to me! Thank you so much 👏🏼

I'll let @code-asher give the final ✅

code-asher (approved):

PR #5142, opened 2022-04-28
refactor(heart): bind class methods and make beat async

jsjoeio:
While implementing https://github.com/coder/code-server/pull/5122, I didn't `bind` the class methods, which meant passing them to another function them calling `this` in them would result in `undefined`. 

I also cleaned up the hack I used before with some async logic.

## References
- https://stackoverflow.com/a/67497368/3015595

Partially related to https://github.com/coder/code-server/pull/5131


src/node/heart.ts
```diff
@@ -20,20 +23,28 @@ export class Heart {
...
     if (typeof this.heartbeatTimer !== "undefined") {
       clearTimeout(this.heartbeatTimer)
     }
-    this.heartbeatTimer = setTimeout(() => heartbeatTimer(this.isActive, this.beat), this.heartbeatInterval)
+    this.heartbeatTimer = setTimeout(async () => {
+      try {
+        await heartbeatTimer(this.isActive, this.beat)
+      } catch (error: any) {
+        logger.warn(error.message)
+      }
```
code-asher, line 46: `heartbeatTimer` can never throw so we could leave this as-is since the catch will never trigger

src/node/heart.ts
```diff
@@ -20,20 +23,28 @@ export class Heart {
...
-  public beat(): void {
+  public async beat(): Promise<void> {
     if (this.alive()) {
       return
     }
 
     logger.trace("heartbeat")
-    fs.writeFile(this.heartbeatPath, "").catch((error) => {
+    try {
+      await fs.writeFile(this.heartbeatPath, "")
```
code-asher, line 33: Thinking about some more, we might want to return this promise at the very end to avoid concurrency issues.

For example if `beat` was called multiple times subsequent calls will think it is not alive because the original call has not yet updated `lastHeartbeat') and they will all try writing to the file and start their own timers

jsjoeio: Ah, good call!

src/node/heart.ts
jsjoeio: Nice catch!

code-asher (approved):

PR #5175, opened 2022-05-06
fix: Docker push

code-asher:
The action actually does not support wildcards but it does support * for
downloading all artifacts.


jsjoeio:
Related: https://github.com/robinraju/release-downloader/issues/422

jsjoeio (approved):

PR #5194, opened 2022-05-11
Allow user Entrypoint scripts

### LINKED ISSUE #5177: [Feat]: Dockerfile entrypoint.d
## What is your suggestion?
Add logic to the entrypoint to load user provided entrypoint scripts.

This would allow for us to setup the workspace beyond the base image without maintaining individual images per purpose.
For example, we could call to install a language or runtime dependency, or even install code-server plugins.

Manually run the scripts via the code-server terminal.

Possibly.

djarbz:
Fixes #5177 

I created an ENV for easy changing of the `entrypoint.d` location in case it is ever wanted/needed to move it out of the home directory. This also would allow for the end user to set an ENVIRONMENT variable to specify a custom mount path as the entrypoint code is dynamic.


ci/release-image/entrypoint.sh
```diff
@@ -18,4 +18,13 @@ if [ "${DOCKER_USER-}" ]; then
...
 fi
 
+# Allow users to have scripts run on container startup to prepare workspace.
+# https://github.com/coder/code-server/issues/5177
+chmod u+x ${ENTRYPOINTD}/*.sh
+sudo chown -R ${USER} ${ENTRYPOINTD}/*
+for f in ${ENTRYPOINTD}/*.sh; do
+  echo "Running Entrypoint: ${f}"
+  bash "${f}"
+done
```
code-asher, line 28: Should we check that the directory exists first?  I see we have a `mkdir -p` in the Dockerfile but my thinking is that if someone mounts the entire home directory then they may not have that directory.  If we add the check I suppose we may not need the `mkdir -p` either.

ci/release-image/entrypoint.sh
```diff
@@ -18,4 +18,13 @@ if [ "${DOCKER_USER-}" ]; then
   fi
 fi
 
+# Allow users to have scripts run on container startup to prepare workspace.
+# https://github.com/coder/code-server/issues/5177
+chmod u+x ${ENTRYPOINTD}/*.sh
+sudo chown -R ${USER} ${ENTRYPOINTD}/*
```
code-asher, line 24: Do you think we could omit these?  I think it may be reasonable to let the user assume responsibility for having the correct bits on their scripts (but let me know if there is an edge case I am missing).

ci/release-image/entrypoint.sh
```diff
@@ -18,4 +18,13 @@ if [ "${DOCKER_USER-}" ]; then
...
+for f in ${ENTRYPOINTD}/*.sh; do
+  echo "Running Entrypoint: ${f}"
+  bash "${f}"
```
code-asher, line 27: Any thoughts on if we could run the script directly instead of through bash?  That way users can use whatever interpreter they want (sh, zsh, or even something entirely different) via the shebang.  Then we could do `${ENTRYPOINTD}/*` as well.

Also I think if any of these scripts exit with non-zero it will cause the entrypoint to exit (due to the `set -e`) so we might want to catch them so they do not prevent code-server from starting (`if ! "${f}" ; then echo "${f} exited with non-zero"; fi`) or something like that.

But then again maybe we do want these to cause a hard exit?  I could go either way personally.  Any thoughts?  @jsjoeio as well.

ci/release-image/Dockerfile
```diff
@@ -42,6 +42,11 @@ RUN ARCH="$(dpkg --print-architecture)" && \
 COPY ci/release-image/entrypoint.sh /usr/bin/entrypoint.sh
 RUN --mount=from=packages,src=/tmp,dst=/tmp/packages dpkg -i /tmp/packages/code-server*$(dpkg --print-architecture).deb
 
+# Allow users to have scripts run on container startup to prepare workspace.
+# https://github.com/coder/code-server/issues/5177
+ENV ENTRYPOINTD=${HOME}/entrypoint.d
```
code-asher, line 47: Nice!

ci/release-image/entrypoint.sh
```diff
@@ -18,4 +18,13 @@ if [ "${DOCKER_USER-}" ]; then
   fi
 fi
 
+# Allow users to have scripts run on container startup to prepare workspace.
+# https://github.com/coder/code-server/issues/5177
+chmod u+x ${ENTRYPOINTD}/*.sh
+sudo chown -R ${USER} ${ENTRYPOINTD}/*
+for f in ${ENTRYPOINTD}/*.sh; do
```
code-asher, line 25:
```suggestion
for f in "${ENTRYPOINTD}"/*.sh; do
```

code-asher (commented):
Thank you!

ci/release-image/entrypoint.sh
djarbz: Yes, that would probably be a better solution.

ci/release-image/entrypoint.sh
djarbz: Probably not, plus sourcing the script with the `bash` prefix does not require the `x` bit to be set.

ci/release-image/entrypoint.sh
djarbz: That might not be the worst idea, then people could also include say binary to execute as well.
This would require that the end user does ensure that the `x` bit is set.

My personal preference would be to hard exit, otherwise, unless we are watching the logs, we wouldn't know that they failed to run.

jsjoeio: There are two points here.

Point 1: running script directly. Agree with @code-asher - we should let users decide which interpreter they want via the shebang.

Point 2: catch scripts exits. From a user standpoint, I imagine I would expect code-server to not start if my entrypoint script had an issue so I think I vote for the hard exit.

jsjoeio:
You'll want to run `yarn fmt` again to fix CI ICYDK.

jsjoeio:
Thank you for adding this! This is an awesome improvement! 🎉

ci/release-image/entrypoint.sh
```diff
@@ -20,11 +20,11 @@ fi
...
 # Allow users to have scripts run on container startup to prepare workspace.
 # https://github.com/coder/code-server/issues/5177
-chmod u+x ${ENTRYPOINTD}/*.sh
-sudo chown -R ${USER} ${ENTRYPOINTD}/*
-for f in "${ENTRYPOINTD}"/*.sh; do
-  echo "Running Entrypoint: ${f}"
-  bash "${f}"
-done
+if [ -d "${ENTRYPOINTD}" ]; then
+    for f in $(find "${ENTRYPOINTD}" -type f -executable); do
```
code-asher, line 24: Ooo this is slick, I like it.

djarbz:
> You'll want to run `yarn fmt` again to fix CI ICYDK.

So I did that, I think correctly, I don't have experience with npm and yarn.
Looks like it is still complaining about it?

code-asher:
We should drop the commit that updates prettier and the lockfiles since those are unrelated changes and seem to be causing Trivy to freak out. :laughing:

djarbz:
> We should drop the commit that updates prettier and the lockfiles since those are unrelated changes and seem to be causing Trivy to freak out. 😆

I'm with you on that, unfortunately I am anything but a git expert and I'm not quite sure what I need to do to drop that commit.

code-asher:
I think the easiest way is probably to do `git revert 5ca347f36155ec731587c1ed8437bca332c76693` and then push the resulting commit.

code-asher (approved):
:tada: :tada: :tada:

jsjoeio (approved):

djarbz:
Thanks for working with me on this one!

jneuhauser:
I'm not sure if I'm using this enhancement correctly, but the following part in my Dockerfile created the `entrypoint.d` directory in `/` instead of `/home/coder`.

```
FROM codercom/code-server:latest
COPY entrypoint.d/* ${ENTRYPOINTD}/
```

For me there is not env `HOME` defined in the docker build process and therefore `ENTRYPOINTD` is always set to `/entrypoint.d`.

djarbz:
Interesting, I personally overwrite the directory to root anyways for my needs. I see that `HOME` is set inside the VSCode shell, but that may be an artifact from that particular shell.

As a fix, I might propose that we hard-code the path.  
`ENV ENTRYPOINTD=/home/coder/entrypoint.d`

In your particular case, I would probably want it outside the home directory anyways in case an end user decides to bind mount the home dir.

jneuhauser:
You are absolutely right that it would be better to hardcode it, but I would prefer `/entrypoint.d` so that mounting a volume/directory to `/home/coder` remains possible.

For me it was a bit confusing that it is defined like `ENV ENTRYPOINTD=${HOME}/entrypoint.d`.

So to not confuse users, I would suggest changing it to `ENV ENTRYPOINTD=/entrypoint.d`.

code-asher:
If you are in control of the Docker image could you just write your own `RUN` commands?  My understanding was that `entrypoint.d` was specifically for end users who have no control over the image (thus you want it in a writable place that gets mounted).  Quite possibly I am misunderstanding the use case though.

PR #5203, opened 2022-05-13
test: replace ts-jest with @swc-node/jest

F3n67u:
Fixes https://github.com/coder/code-server/issues/4403

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
   },
   "jest": {
     "transform": {
-      "^.+\\.ts$": "<rootDir>/test/node_modules/ts-jest"
+      "^.+\\.ts$": "<rootDir>/test/node_modules/@swc-node/jest"
     },
     "testEnvironment": "node",
     "testPathIgnorePatterns": [
--- a/test/package.json
+++ b/test/package.json
@@ -4,6 +4,7 @@
   "devDependencies": {
     "@jest-mock/express": "^1.4.5",
     "@playwright/test": "^1.16.3",
+    "@swc-node/jest": "^1.5.2",
     "@types/jest": "^27.0.2",
     "@types/jsdom": "^16.2.13",
     "@types/node-fetch": "^2.5.8",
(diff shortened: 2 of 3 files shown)


jsjoeio (commented):
Thank you for doing this! You're helping a ton and we appreciate it :D 

Two questions:
1. any chance you can update your description with some stats: how long did unit tests take before and how much faster are they? 
2. do we need to add a separate typechecking step? i originally chose `ts-jest` since it does typechecking on our tests. are we losing that now and need to add elsewhere?

F3n67u:
From ci stat:

## @swc-node/jest:
```
Test Suites: 20 passed, 20 total
Tests:       207 passed, 207 total
Snapshots:   0 total
Time:        15.393 s
Ran all test suites.
Done in 20.30s.
```
https://github.com/coder/code-server/runs/6426594467?check_suite_focus=true#step:12:76

## ts-jest
```
Test Suites: 20 passed, 20 total
Tests:       207 passed, 207 total
Snapshots:   0 total
Time:        21.505 s
Ran all test suites.
Done in 27.95s.
```
https://github.com/coder/code-server/runs/6426871269?check_suite_focus=true#step:12:76

total unit test time is from 28s -> 20s，but because we lose the type check of `ts-jest`, I don't think the 8s is worth it.

our unit test suite is very small size, maybe we could reconsider switch to `@swc-node/jest` when our unit test grow to big enough.

@jsjoeio what do you think?

jsjoeio:
@F3n67u thank you for taking the time to add that! Hmm...you're probably right. I thought it would be a greater savings to be honest. I agree though, might not make sense to add if we lose typechecking then.

Thoughts @code-asher?

code-asher:
I think we should keep the type checking even if it does make things slower.

jsjoeio:
Agreed.  I think it that case, let's close this PR and that issue. Is that okay with you @F3n67u? Thank you for doing this though! Really appreciate it.

F3n67u:
> Agreed. I think it that case, let's close this PR and that issue. Is that okay with you @F3n67u? Thank you for doing this though! Really appreciate it.

yep. lets close this pr.

PR #5219, opened 2022-05-23
fix regex for manifest route

marshall-orevillestudios:
I noticed that the regex would match other strings than manifest.json such as manifest-json or something-manifest.json

### DIFF
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -170,7 +170,7 @@ export class CodeServerRouteWrapper {
 
   constructor() {
     this.router.get("/", this.ensureCodeServerLoaded, this.$root)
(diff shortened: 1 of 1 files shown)


jsjoeio:
> I noticed that the regex would match other strings than manifest.json such as manifest-json or something-manifest.json

I'm guessing that wasn't intentional but let's double check with @code-asher.

code-asher:
I think this was because there used to be a `webmanifest.json` link?  But as far as I can tell everything is `manifest.json` now.  Could we get rid of the regex entirely and just do `/manifest.json`?

jsjoeio (changes requested):
Do you mind addressing @code-asher's feedback? Then we can merge this

jsjoeio:
@code-asher ready for review

code-asher (approved):

PR #5243, opened 2022-06-06
feat: add additional test for handlePassswordValidation

jsjoeio:
This adds an additional test for `handlePassowrdValidation` to ensure the correct behavior happens when passing in an invalid `passwordMethod`. Although it shouldn't happen, the new test ensurs that `default` case in the switch statement has coverage.

Related to #5153


BrunoQuaresma (approved):
Code LGTM

PR #5255, opened 2022-06-10
Allow setting priorityClassName via helm chart

timbrd:
Adds an option to the helm chart to set a [pod priority class](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass).


jsjoeio:
@Matthew-Beckett can I get your help reviewing this? 🙏🏼

ci/helm-chart/values.yaml
```diff
@@ -33,6 +33,8 @@ podAnnotations: {}
 podSecurityContext: {}
   # fsGroup: 2000
 
+priorityClassName: ""
```
Matthew-Beckett, line 36: The if statement which evaluates if this is populated or not in deployment.yaml will evaluate to true and populate this invalid null string by default won't it?

Should this be commented out by default?

Matthew-Beckett (commented):
Not sure if that default empty string will stroll evaluate true on build unless commented out, need to test this properly in a cluster.

ci/helm-chart/values.yaml
Matthew-Beckett: Never mind docs say empty strings evaluate false.

https://helm.sh/docs/chart_template_guide/control_structures/

jsjoeio: TIL! So do you think this is good to merge?

Matthew-Beckett (approved):
I'm so sorry this has taken so long, I could make a whole load of excuses for the delay but none would make it okay.

I am so sorry @timbrd, thanks for your contribution ❤️.

I've tested this now and I'm happy for merge!

jsjoeio (approved):

timbrd:
@jsjoeio @Matthew-Beckett Is there anything more which has to be done or will the PR land shortly?

Matthew-Beckett:
> @jsjoeio @Matthew-Beckett Is there anything more which has to be done or will the PR land shortly?

Based on that merge I'd say no... 😅

PR #5263, opened 2022-06-13
chore: update Code to 1.68

### LINKED ISSUE #5165: [Chore]: upgrade Code to 1.67

code-asher:
Closes #5165 

I had a bit of trouble going straight to 1.68 (had some weird webview issues) so I went to 1.67 first then 1.68 and then it went smoothly.  Not sure if that is a coincidence but seems like updating one version at a time might be the way to go.

### DIFF
--- a/patches/display-language.diff
+++ b/patches/display-language.diff
@@ -91,7 +91,7 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.html
 +						return cb(undefined, result)
 +					}
 +					const path = nlsConfig['vs/nls']._resolvedLanguagePackCoreLocation + "/" + bundle.replace(/\//g, "!") + ".nls.json"
-+					fetch(`{{VS_BASE}}/vscode-remote-resource?path=${encodeURIComponent(path)}`)
++					fetch(`{{WORKBENCH_WEB_BASE_URL}}/vscode-remote-resource?path=${encodeURIComponent(path)}`)
 +						.then((response) => response.json())
 +						.then((json) => {
 +							bundles[bundle] = json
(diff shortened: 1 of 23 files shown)


MitchTalmadge:
Thanks for doing this, I am eagerly awaiting 1.68 myself. I saw that this is marked as a draft, is there anything that you need help with?

MitchTalmadge:
Hmm yeah that's a tough problem indeed. It would suck to lose additional language support that we have added, but it also sucks to have to replace a core feature. I'm not sure what can be done for option 1, but if it's not feasible, I would prefer option 2 to losing languages in option 3.

code-asher:
I think this is good for testing and PR review.  I went with option 2 for the display language so we have maintained the status quo.

patches/display-language.diff
```diff
@@ -1,17 +1,28 @@
 Add display language support
 
-This likely needs tweaking if we want to upstream.
+We can remove this once upstream supports all language packs.
```
jsjoeio, line 3: 👏🏼 Thanks for updating this!

jsjoeio (approved):
Nice work! Impressed you were able to update two versions. All looks good to me 👍🏼

code-asher:
Ran into a couple things when testing the full build:

- Some /vscode-remote-resource requests are returning 404s, seems the commit is somehow missing so it defaults to `dev` but then that fails because the actual commit is not `dev`
- There is another parentOrigin check in the web worker HTML, it does not seem to actually trigger validation but I should probably patch it anyway

code-asher:
@jsjoeio Three new commits fixing the aforementioned problems plus one spot where I could use the new base URL instead of the old one (both work for now but might as well use the new one).

code-asher:
OK one more problem with our parent origin patch, on non-Firefox browsers the CSP breaks because there is a nonce so I need to update those.  Give me a second...

jsjoeio (approved):

xiongsongsong:
when will the new version be released?

Issue #5265, opened 2022-06-14
[Feat]: add OpenRC support


akr2002:
## What is your suggestion?
Theoretically it should be possible to install code-server on any machine that supports npm and yarn. 

Currently the docs only cover installation of code-server on systemd-based distributions, with the exception of Termux. While writing an OpenRC script is trivial, the user is required to edit it in order to get it working under the user's permissions rather than root.

## Why do you want this feature?
I want to run it on non-systemd distros. 

## Are there any workarounds to get this functionality today?
Install manually and write an OpenRC script.

## Are you interested in submitting a PR for this?
Yep. I managed to run it on Artix Linux. Expect a PR soon.

PR #5266, opened 2022-06-14
Add OpenRC support

GNUxeava:
Address #5265


GNUxeava:
I think we should have the script as a separate file instead of putting it in the markdown file. That way the user can simply use wegt to download the script instead of copy-pasting the stuff.

jsjoeio:
This is out of my wheelhouse. cc @oxy or @code-asher any thoughts?

My preference would be to not make this a new docs file and instead put it under https://github.com/coder/code-server/blob/main/docs/install.md as it's own section. I also wouldn't be in favor of adding a new script because then we'd have to maintain the docs and the script.

code-asher:
Maybe we can take the same path we do with the Helm stuff and delegate maintenance and testing of this to others.

If we are cool with that then splitting out the script to make life easier for users sounds chill to me.

I also like the idea of keeping it all in `install.md`.  Out of scope for this PR but I think we might want to move some other files into there as well like `android.md`, `ios.md`, `termux.md`, and `npm.md`?  Maybe even `helm.md`.

jsjoeio:
> Maybe we can take the same path we do with the Helm stuff and delegate maintenance and testing of this to others.
> If we are cool with that then splitting out the script to make life easier for users sounds chill to me.

Cool! @GNUxeava you can add a script. Would you be okay reviewing any PRs that touch the docs/code you add? We'd make you a `CODEOWNER`?

GNUxeava:
Fine by me

jsjoeio:
Sweet! @GNUxeava okay to get this in, I think we need to: 

- [ ] move this section to `install.md`
- [ ] move `bash` script into it's own script under `/ci/dev` (right? @code-asher)
- [ ] add yourself as under `CODEOWNERS` and point to new script

GNUxeava:
I squashed some commits; hopefully it didn't break anything

jsjoeio:
Weird...it says there is an error since you don't have write access 🤔 

image

I think this is fine for now. We can always edit later if needed.

jsjoeio (approved):

PR #5275, opened 2022-06-15
refactor: move integration tests to Jest

jsjoeio:
## Description

This PR refactors our integration tests to be written in TypeScript and run with Jest instead of bash. It also adds a helper [inspired by the Next.js team](https://github.com/vercel/next.js/blob/canary/test/lib/next-test-utils.js#L139) to make it easy to test code-server commands.

## Motivation

We currently only have [one integration test](https://github.com/coder/code-server/blob/main/ci/build/test-standalone-release.sh) and it's not easy to write new ones because Bash and it's all in one file. 

This should make it easier to write integration tests, specifically around the CLI. This is also prep work for https://github.com/coder/code-server/issues/5031

### Before

Here is what our integration tests looked like:
```bash
#!/usr/bin/env bash
set -euo pipefail

# Make sure a code-server release works. You can pass in the path otherwise it
# will use release-standalone in the current directory.
#
# This is to make sure we don't have Node version errors or any other
# compilation-related errors.
main() {
  cd "$(dirname "${0}")/../.."

  local EXTENSIONS_DIR
  EXTENSIONS_DIR="$(mktemp -d)"

  local path=${1:-./release-standalone/bin/code-server}

  echo "Testing standalone release in $path."

  # NOTE: using a basic theme extension because it doesn't update often and is more reliable for testing
  "$path" --extensions-dir "$EXTENSIONS_DIR" --install-extension wesbos.theme-cobalt2
  local installed_extensions
  installed_extensions="$("$path" --extensions-dir "$EXTENSIONS_DIR" --list-extensions 2>&1)"
  # We use grep as wesbos.theme-cobalt2 may have dependency extensions that change.
  if ! echo "$installed_extensions" | grep -q "wesbos.theme-cobalt2"; then
    echo "Unexpected output from listing extensions:"
    echo "$installed_extensions"
    exit 1
  fi

  echo "Standalone release works correctly."
}

main "$@"
```

### After

This is what they'll look like moving forward:
```typescript
import { clean, tmpdir } from "../utils/helpers"
import { runCodeServerCommand } from "../utils/integration"

describe("--install-extension", () => {
  const testName = "installExtension"
  let tempDir: string
  let setupFlags: string[]

  beforeEach(async () => {
    await clean(testName)
    tempDir = await tmpdir(testName)
    setupFlags = ["--extensions-dir", tempDir]
  })
  it("should install an extension", async () => {
    const extName = "wesbos.theme-cobalt2"
    await runCodeServerCommand([...setupFlags, "--install-extension", extName], {})
    const { stdout } = await runCodeServerCommand([...setupFlags, "--list-extensions"], {
      stdout: "log",
    })
    expect(stdout).toContain(extName)
  })
})
```

Fixes N/A


docs/CONTRIBUTING.md
```diff
@@ -188,9 +188,7 @@ We use these to test anything related to our scripts (most of which live under `
 
 ### Integration tests
 
-These are a work in progress. We build code-server and run a script called
-[test-standalone-release.sh](../ci/build/test-standalone-release.sh), which
-ensures that code-server's CLI is working.
+These are a work in progress. We build code-server and run tests with `yarn test:integration`, which ensures that code-server's binary works.
```
jsjoeio: nit: this should be wrapped

.github/workflows/ci.yaml
```diff
@@ -288,8 +288,11 @@ jobs:
       - name: Build standalone release
         run: source scl_source enable devtoolset-9 && yarn release:standalone
 
-      - name: Sanity test standalone release
-        run: yarn test:standalone-release
+      - name: Install test dependencies
+        run: yarn install --ignore-scripts && cd test && yarn install
+
+      - name: Run integration tests on standalone release
+        run: yarn test:integration
```
jsjoeio, line 295: NOTE: I had to use `--ignore-scripts` because otherwise `yarn install` fails if you don't checkout the repo with submodules. On the Linux job that runs on CentOS7, it didn't have the latest `git` so this one-liner was the best alternative. It ensures we have the right TS/Jest dependencies to run the `yarn test:integration` step.

.github/workflows/installer.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         run: ./install.sh
 
       - name: Test code-server
-        run: yarn test:standalone-release code-server
+        run: CODE_SERVER_PATH="code-server" yarn test:integration
```
jsjoeio, line 37: Since we install code-server in these steps, we need to tell the integration tests to look at `code-server` and use the CLI/binary.

ci/dev/test-unit.sh
```diff
@@ -30,7 +30,7 @@ main() {
   # We must keep jest in a sub-directory. See ../../test/package.json for more
   # information. We must also run it from the root otherwise coverage will not
   # include our source files.
-  CS_DISABLE_PLUGINS=true ./test/node_modules/.bin/jest "$@"
+  CS_DISABLE_PLUGINS=true ./test/node_modules/.bin/jest "$@" --testRegex "./test/unit/.*ts" --testPathIgnorePatterns "./test/unit/node/test-plugin"
```
jsjoeio, line 33: NOTE: since we run `jest` from the root, it grabs the `package.json.jest` config and uses that for running tests. By adding these flags here, we can use the same Jest config for both unit & integration tests and then pass additional flags as needed.

test/integration/installExtension.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    await clean(testName)
+    tempDir = await tmpdir(testName)
+    setupFlags = ["--extensions-dir", tempDir]
+  })
+  it("should install an extension", async () => {
+    const extName = "wesbos.theme-cobalt2"
+    await runCodeServerCommand([...setupFlags, "--install-extension", extName], {})
+    const { stdout } = await runCodeServerCommand([...setupFlags, "--list-extensions"], {
+      stdout: "log",
+    })
```
jsjoeio, line 19: NOTE: this is the same test we had previously. The only down-side is it's subject to network failures. We could add `test/integration/fixtures` and upload a `.vsix` and use that instead if desired.

test/integration/installExtension.test.ts
```diff
@@ -0,0 +1,22 @@
...
+    expect(stdout).toContain(extName)
+  })
```
jsjoeio, line 21: This might fail on macOS due to timeouts so you may need to increase it here. 
```suggestion
  }, 10000)
```

test/utils/integration.ts
```diff
@@ -29,3 +30,87 @@ export async function setup(argv: string[], configFile?: string): Promise<httpse
 
   return new httpserver.HttpServer(server)
 }
+
+type RunCodeServerCommandOptions = {
+  stderr?: "log"
+  stdout?: "log"
+  ignoreFail?: boolean
+}
+interface ErrorWithMoreInfo extends Error {
```
jsjoeio, line 39: Open to better names lol

test/integration/installExtension.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  beforeEach(async () => {
+    await clean(testName)
+    tempDir = await tmpdir(testName)
+    setupFlags = ["--extensions-dir", tempDir]
+  })
+  it("should install an extension", async () => {
+    const extName = "wesbos.theme-cobalt2"
+    await runCodeServerCommand([...setupFlags, "--install-extension", extName], {})
+    const { stdout } = await runCodeServerCommand([...setupFlags, "--list-extensions"], {
```
jsjoeio, line 17: After thinking about this more, I don't think we should use `--list-extensions` to verify `--install-extension` passed. Reasons:
1. we're essentially saying `--install-extension` relies on `--list-extensions` in order to pass, which doesn't mimic real-world usage
2. spawns a code-server instance which increases run time for this test

I think instead we should check the `tempDir` for the extension directory using `fs.promises`

code-asher: That makes sense to me.  The original intent of this test was actually just to make sure there were no native module errors and installing/listing extensions was the easiest way to trigger those errors.  I do not remember if install alone triggers the errors or if both were required to trigger the error so if we do add a check to `tempDir` we might want to keep both (maybe put the list part in a separate test).

jsjoeio: that sounds ideal to me! I'll do that then re-request a review

test/utils/runCodeServerCommand.ts
```diff
@@ -0,0 +1,86 @@
...
+      stdout: string
+      stderr: string
+    }
+  | ErrorWithMoreInfo
+
+/**
+ *
+ * A helper function for integration tests to run code-server commands.
+ */
+export async function runCodeServerCommand(argv: string[], options: RunCodeServerCommandOptions = {}): Promise<CSCmd> {
```
code-asher, line 27: We could use `exec` (either directly in the tests or with a thin wrapper here) which I think does everything this function does (it even has stdout/stderr on thrown errors):

```js
const { stdout, stderr } = await util.promisify(cp.exec)(CODE_SERVER_PATH, argv)
```

docs/CONTRIBUTING.md
code-asher, line 191: I am not sure exactly how to word this but I think we should avoid `binary` since code-server is a bunch of loose files rather than a single binary.

Maybe...`ensures that code-server builds work on their respective platforms`?

code-asher (commented):
Happy to see this move into our standard testing framework!

.github/workflows/ci.yaml
```diff
@@ -288,8 +288,11 @@ jobs:
       - name: Build standalone release
         run: source scl_source enable devtoolset-9 && yarn release:standalone
 
-      - name: Sanity test standalone release
-        run: yarn test:standalone-release
+      - name: Install test dependencies
+        run: SKIP_SUBMODULE_DEPS=1 yarn install
+
+      - name: Run integration tests on standalone release
+        run: yarn test:integration
```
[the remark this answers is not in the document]
jsjoeio: Due to some issues with TypeScript/Jest, we have some dependencies at the root level and some in `/test`. When I first set up Jest, I tried to keep them all in `test` and couldn't get it to work. I agree though, we should revisit this later.

test/integration/fixtures/wesbos.theme-cobalt2-2.1.6/.vsixmanifest
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
jsjoeio: Great suggestion! Sounds like a better approach to me.

package.json
```diff
@@ -18,8 +18,8 @@
     "release:github-assets": "./ci/build/release-github-assets.sh",
     "release:prep": "./ci/build/release-prep.sh",
     "test:e2e": "VSCODE_IPC_HOOK_CLI= ./ci/dev/test-e2e.sh",
-    "test:standalone-release": "./ci/build/test-standalone-release.sh",
```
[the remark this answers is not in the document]
jsjoeio: Ah! Yes, that file is outdated lol I'll update.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
code-asher: Yeah I think we would just copy those dependencies into `test/package.json`.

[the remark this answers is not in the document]
code-asher: Oh do you mean you already tried moving them into `test/package.json`?  Weird.  Yeah we can revisit later.

test/utils/runCodeServerCommand.ts
code-asher: Yeah, you could keep this exact same function but just replace all the contents with `return util.promisify(cp.exec)(CODE_SERVER_PATH, argv)`.  Just seems unfortunate to reimplement a builtin! :)

.github/workflows/ci.yaml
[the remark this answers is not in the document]
jsjoeio: Yes! But maybe with a fresh set of eyes, it will work now. Worth a shot! I'll open an issue.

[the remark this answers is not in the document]
jsjoeio: https://github.com/coder/code-server/issues/5291

test/utils/runCodeServerCommand.ts
jsjoeio: Note to self, i think `argv` must be a string: `exec('"/path/to/test file/test.sh" arg1 arg2');`

```typescript
  const { stdout, stderr } = await promisify(exec)(`${CODE_SERVER_COMMAND} ${argv.join(" ")}`)
```

code-asher: Ohhh yeah you are right it does not accept an array.  :facepalm:

jsjoeio: Dude, I just removed so much code. I went with the thin wrapper approach. Nice!

test/integration/listExtensions.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  beforeEach(async () => {
+    await clean(testName)
+    tempDir = await tmpdir(testName)
+    setupFlags = ["--extensions-dir", tempDir]
+    const extensionFixture = path.resolve(`./test/integration/fixtures/${vsixFileName}`)
+    // Make folder because this is where we'll move the extension
+    const pathToUnpackedExtension = `${tempDir}/${extName}-${extVersion}`
+    const tempPathToUnpackedExtension = `${tempDir}/${extName}-temp}`
+    await mkdir(path.resolve(pathToUnpackedExtension))
```
code-asher, line 23: I think you could omit the `mkdir` (and the import) since `rename` does not require the destination to exist.  Actually I thought this would error since the Node docs say:

> In the case that newPath already exists, it will be overwritten. If there is a directory at newPath, an error will be raised instead.

But apparently that is not true...it happily overwrites directories too.  :shrug: 

```suggestion
```

test/integration/listExtensions.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    await extract(extensionFixture, { dir: tempPathToUnpackedExtension })
+    await rename(`${tempPathToUnpackedExtension}/extension`, pathToUnpackedExtension)
```
code-asher, line 25: We should use `path` utilities any time we are constructing paths:

```suggestion
    await rename(path.join(tempPathToUnpackedExtension, "extension", pathToUnpackedExtension))
	
```

test/utils/runCodeServerCommand.ts
```diff
@@ -0,0 +1,13 @@
+import { exec } from "child_process"
+import { promisify } from "util"
+
+/**
+ *
+ * A helper function for integration tests to run code-server commands.
+ */
+export async function runCodeServerCommand(argv: string[]): Promise<{ stdout: string; stderr: string }> {
+  const CODE_SERVER_COMMAND = process.env.CODE_SERVER_PATH || "/var/tmp/coder/code-server/bin/code-server"
```
code-asher, line 9: Should we default to `path.resolve("../../release-standalone/bin/code-server")`?  Reason being the current path will only work in v1 of Coder.

test/integration/listExtensions.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  let setupFlags: string[]
+
+  beforeEach(async () => {
+    await clean(testName)
+    tempDir = await tmpdir(testName)
+    setupFlags = ["--extensions-dir", tempDir]
+    const extensionFixture = path.resolve(`./test/integration/fixtures/${vsixFileName}`)
+    // Make folder because this is where we'll move the extension
+    const pathToUnpackedExtension = `${tempDir}/${extName}-${extVersion}`
+    const tempPathToUnpackedExtension = `${tempDir}/${extName}-temp}`
```
code-asher, line 22:
```suggestion
    const pathToUnpackedExtension = path.join(tempDir, `${extName}-${extVersion}`)
    const tempPathToUnpackedExtension = path.join(tempDir, `${extName}-temp`)
```

code-asher (approved):

test/utils/runCodeServerCommand.ts
jsjoeio: Oh yeah, I was first writing this in v1 hence the default value. Good call.

PR #5285, opened 2022-06-21
Allow more comprehensive affinity config

strowk:
This PR changes helm template to allow more comprehensive configuration in affinity. Instead of treating affinity as a static map, it expects a string passed there, which would be then processed as a template and therefore resolve any links, such as f.e. {{ include "code-server.name" . }} . 
This then allows to configure something like this:

```yaml
affinity: |
  podAntiAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: app.kubernetes.io/name
              operator: In
              values:
              - {{ include "code-server.name" . }}
          topologyKey: kubernetes.io/hostname
```

Above allows to properly spread pods over nodes without hardcoding any labels in values file. 

This approach is somewhat common in helm ecosystem, f.e. see here:
https://github.com/codecentric/helm-charts/blob/31f44ca368cef7585760edc8f22c011cab1be284/charts/keycloakx/templates/statefulset.yaml#L184

The change is breaking though, so I am not 100 percent sure this has to be included, I am sending this as a suggestion and input for discussion.


jsjoeio:
cc @bpmct @Matthew-Beckett

bpmct (commented):
This seems good to me, I did some digging and this does indeed look like a common practice. I also think it's fine to push a breaking change given we don't publish our helm charts anywhere and they are manually pulled.

However, will defer to @Matthew-Beckett to review and approve.

Matthew-Beckett (approved):
LGTM!

PR #5296, opened 2022-06-24
chore: update codecov to 3.7.1

jsjoeio:
Fixes #


kylecarbs (approved):

jsjoeio:
Argh... so a couple things:
- this vulneraility report shows a patched version which does not exist: https://github.com/advisories/GHSA-mh2h-6j8q-x246
- codecov deprecated their node uploader which means the true solution is to migrate away from it and use the [new uploader ](https://about.codecov.io/blog/introducing-codecovs-new-uploader/)

I will open a separate PR and then close this one.

PR #5317, opened 2022-07-02
Update display-language.diff

bbbboom:
The requested "vscode-remote-resource" is not under the "{{WORKBENCH_WEB_BASE_URL}}" path, It needs to access the upper level path.

Fixes #


code-asher (approved):
Thank you and nice find!!

PR #5332, opened 2022-07-11
fix(terminal): use window.location to work when served NOT on root path

jsjoeio:
## Description

We made a change in `workbench-dev.html` but forgot to make it in `workbench.html`. This fixes the Integrated Terminal to work when code-server is served NOT on the root path (i.e. accessed at `/vscode` instead of `/`).

Fixes https://github.com/coder/code-server/issues/5321


oxy (approved):
+1!

jsjoeio:
Tested this locally behind Caddy and it worked 🎉

1. `yarn init -y`
2. `yarn add @coder/code-server-pr@5332`
3. `yarn code-server --auth none`
4. `caddy run` with this:
```Caddyfile
http://localhost:8082/ide/* {
  uri strip_prefix /ide
  reverse_proxy 127.0.0.1:8080
}
```

image

PR #5346, opened 2022-07-15
fix: output lost after hotswapping

code-asher:
I think the pipe was closing the other streams when the child stream
closed so instead just write it one way.

This was making it quite difficult to debug code-server when doing live
edits in Coder.


jsjoeio (approved):
This is one of those intricate pieces of the codebase where I'm really happy you know what's going wrong and how to fix  😂 Thanks for doing this!

Issue #1315, opened 2020-01-29
Implement VS Code remote extensions


nhooyr:
- [ ] Container
- [ ] SSH

Related https://github.com/cdr/code-server/issues/1032

LukasMasuch:
In case this issue is about fixing the remote extensions to work with code-server, this related issue can be closed: #892. I would really like to see the SSH extension to work within code-server and would be happy to test it once there is any fix.

sr229:
@nhyoor if this interests you, there's a remote subfolder in the main vscode folder, wanna ~~abuse~~ reuse that?

frank-dspeed:
The remote extension does some checks as far as i know i decompiled it some times.

nhooyr:
We cannot and will not decompile the extension, that is not valid per the license terms.

Just-Insane:
Is there any further plan to implement these extensions?

kirscheGIT:
> Is there any further plan to implement these extensions?

Would also like to know, if you have still have plans to implement this. Thanks!

gyzerok:
I currently have a need for Remote - SSH extension and unfortunately the official one doesn't work for me (which is unlikely to be solved upstream in the near future). So I am considering of implementing it myself.

At least initially I would like to have it as simple as possible. Without any fancy terminal forwarding or something like that. Just make it so that the work with the filesystem and all the language servers are running on the vm and UI is running in the vscode. Similar two what we now have with code-server in the browser, but without browser version problems (such as cmd+w for example).

Would you have some pointers as to where to start investigating from inside vscode source?

code-asher:
At the moment I don't have a good idea of the full scope of what'll be required to make this work but it'll probably have to do with using and/or implementing the remote authority stuff.

There's a `workspace.registerRemoteAuthorityResolver` in the proposed API which might be a good starting point.

JayDoubleu:
building from vscode-1.48.0 branch + adding https://github.com/cdr/code-server/pull/1779/commits/9fb33dce3d8f4854163c978ef02bdd06f1983722 , as well as adding ```ms-vscode-remote.remote-containers```  to  ```extensionAllowedProposedApi```.

This gets the remote extensions installed however it is still complaining with the error:
```
[ms-vscode-remote.remote-containers]: View container 'remote' requires 'enableProposedApi' turned on to be added to 'Remote'.
```
Tried executing via below.
```
code-server --enable-proposed-api ms-vscode-remote.remote-containers
[...]

PR #5360, opened 2022-07-20
fix: invoking code-server in integrated terminal

### LINKED ISSUE #5335: [Bug]: Cannot invoke `code-server` from integrated terminal

### OS/Web Information

- Web Browser: Firefox 102
- Local OS: Win10
- Remote OS: Ubuntu
- Remote Architecture: 86
- `code-server --version`: 4.5.0 (4.4.0)

### Steps to Reproduce

1. docker run codeserver:latest
2. try to install extensions via code-server `code-server --install-extension ms-python`
3. asd

### Expected

Installation of Extension

### Actual

Console throws error
`/usr/lib/code-server/lib/vscode/bin/remote-cli/code-server: 12: /usr/lib/code-server/lib/vscode/node: not found`

### Logs

_No response_

### Screenshot/Video

_No response_

- [X] I am using HTTPS.

### Notes

_No response_

code-asher:
Fixes #5335 

This fixes the path to Node in upstream's scripts then links the right one based on the platform.

There is also some removal of duplicate code where we symlink the asar twice; turns out we do not need this since the standalone release runs yarn which will run the postinstall which will add the link.

If anyone wishes to test manually open up the integrated terminal and run `code-server <file>`.  You can try things like `code-server --wait <file>` as well which should work.

Possibly in a separate PR we should think about using the same for external terminals then remove our bespoke code.

Thanks to @fritterhoff for linking to where this happens, made this very easy to fix!

Fixes https://github.com/coder/code-server/issues/2110

Follow-up TODOs
- [x] open issue for external terminal usage (aka `--wait` outside code-server)


code-asher:
Decided to move this back to a draft.  I am going to look into making their script work because that would be the more maintainable solution.

benz0li:
> I opted to point it to our own bin directory instead of Code's.

Doesn't this simply add the same path twice? I.e.

```zsh
$ echo $PATH
/opt/code-server/bin:/opt/code-server/bin:...
```

jsjoeio:
Untested, but maybe we could add an e2e test for this too. Something like:

```typescript
// bin.test.ts

import * as cp from "child_process"
import * as path from "path"
import util from "util"
import { clean, tmpdir } from "../utils/helpers"
import { describe, expect, test } from "./baseFixture"

describe("Bin", true, [], {}, () => {
  const testName = "bin"
  test.beforeAll(async () => {
    await clean(testName)
  })

  test("should give access to code-server in terminal", async ({ codeServerPage }) => {
    const tmpFolderPath = await tmpdir(testName)
    const tmpFile = path.join(tmpFolderPath, "pipe")

    const command = `mkfifo '${tmpFile}' && cat '${tmpFile}'`
    const exec = util.promisify(cp.exec)
    const output = exec(command, { encoding: "utf8" })

    // Open terminal and type in value
    await codeServerPage.focusTerminal()

    await codeServerPage.page.waitForLoadState("load")
    await codeServerPage.page.keyboard.type(`code-server --help`)
    await codeServerPage.page.keyboard.press("Enter")
    // It may take a second to process
    await codeServerPage.page.waitForTimeout(1000)

    const { stdout } = await output
    expect(stdout).toMatch(/usage: code-server/)
  })
})
```

code-asher:
@jsjoeio that is awesome, I actually intended on adding a test for this so that code will be very helpful!

benz0li:
@code-asher What about https://github.com/coder/code-server/issues/5335#issuecomment-1191075397?

code-asher:
I think the ideal solution is going to be to fix upstream's `code-server` script so that it can find Node.  This means it will use their script instead of ours but long-term that is probably a good move.  For example things like `--wait` will just work.  Thoughts?

fritterhoff:
> I think the ideal solution is going to be to fix upstream's `code-server` script so that it can find Node. This means invocations on the integrated terminal will use their script instead of ours but long-term that is probably a good move. For example things like `--wait` will just work. Thoughts?

From my point of view fixing/patching the script would make more sense and should be better maintanable than patching the actual source code of vs code?

benz0li:
IMHO there should be no file `/opt/code-server/lib/vscode/bin/remote-cli/code-server`. Reposting https://github.com/coder/code-server/issues/5335#issuecomment-1191075397 here:

> @code-asher In code-server v4.4.0/v4.3.0 there is `/opt/code-server/lib/vscode/bin/remote-cli/code-oss`.  
> 👉 `/opt/code-server/lib/vscode/bin/remote-cli` is prepended to `PATH`, too; but there is `code-oss` instead of `code-server` at the said path.
> 
> ---
> 
> My workaround is renaming `/opt/code-server/lib/vscode/bin/remote-cli/code-server` to `/opt/code-server/lib/vscode/bin/remote-cli/code-oss` and replacing all occurrences of `code-server` with `code-oss` in
> 
> 1. `/opt/code-server/lib/vscode/bin/remote-cli/code-oss`
> 2. `/opt/code-server/lib/vscode/bin/helpers/browser.sh`
> 
> ---
> 
> code-server v4.4.0:
> 
> ```shell
> $ tree /opt/code-server/lib/vscode/bin
> /opt/code-server/lib/vscode/bin
> ├── code-server-oss
> ├── helpers
> │   └── browser.sh
> └── remote-cli
>     └── code-oss
> 
> 2 directories, 3 files
> 
> $ cat /opt/code-server/lib/vscode/bin/code-server-oss 
> #!/usr/bin/env sh
> #
> # Copyright (c) Microsoft Corporation. All rights reserved.
> #
> 
> case "$1" in
>         --inspect*) INSPECT="$1"; shift;;
> esac
> 
> ROOT="$(dirname "$(dirname "$(readlink -f "$0")")")"
> 
> "$ROOT/node" ${INSPECT:-} "$ROOT/out/server-main.js" "$@"
> 
> $ cat /opt/code-server/lib/vscode/bin/helpers/browser.sh 
> #!/usr/bin/env sh
> #
> # Copyright (c) Microsoft Corporation. All rights reserved.
> #
> ROOT=$(dirname "$(dirname "$(dirname "$0")")")
> 
> APP_NAME="code-oss"
> VERSION="1.66.2"
> COMMIT=","
> EXEC_NAME="code-oss"
> CLI_SCRIPT="$ROOT/out/server-cli.js"
> "$ROOT/node" "$CLI_SCRIPT" "$APP_NAME" "$VERSION" "$COMMIT" "$EXEC_NAME" "--openExternal" "$@"
> 
> $ cat /opt/code-server/lib/vscode/bin/remote-cli/code-oss 
> #!/usr/bin/env sh
> #
> # Copyright (c) Microsoft Corporation. All rights reserved.
> #
> ROOT=$(dirname "$(dirname "$(dirname "$0")")")
> 
> APP_NAME="code-oss"
> VERSION="1.66.2"
> COMMIT=","
> EXEC_NAME="code-oss"
> CLI_SCRIPT="$ROOT/out/server-cli.js"
> "$ROOT/node" "$CLI_SCRIPT" "$APP_NAME" "$VERSION" "$COMMIT" "$EXEC_NAME" "$@"
> ```

---

Does `/opt/code-server/lib/vscode/bin/*` come into play with `code-server` at all?

And yes, `"$ROOT/node"` does not work; but `"$ROOT/../node"` would.

code-asher:
Yeah renaming the upstream script would cause `code-server` invocations to use our script instead (for users that have it on their path) but since upstream's script does everything ours does and more it might make sense to just use theirs.

The only thing theirs does not support is launching a new instance of code-server (it only supports interacting with the current instance of code-server) but I have no idea if that is a common use in the integrated terminal.

benz0li:
> Yeah renaming the upstream script would cause `code-server` invocations to use our script instead (for users that have it on their path) but since upstream's script does everything ours does and more it might make sense to just use theirs.

`/opt/code-server/lib/vscode/bin/code-server-oss` is actually meant to start a new server.

> The only thing theirs does not support is launching a new instance of code-server (it only supports interacting with the current instance of code-server) but I have no idea if that is a common use in the integrated terminal.

I have never started a new instance of code-server using the integrated terminal.

benz0li:
`/opt/code-server/bin/code-server` is the equivalent to `/opt/code-server/lib/vscode/bin/code-server-oss` and not `/opt/code-server/lib/vscode/bin/remote-cli/code-oss`.

The latter named `/opt/code-server/lib/vscode/bin/remote-cli/code-server` in `code-server`, creating a name conflict.

benz0li:
Also reposting https://github.com/coder/code-server/issues/5335#issuecomment-1189936620 here:

> The way `code-server` is released(/run?), it behaves like a _remote installation_. This ~might be~ is the reason, why `/opt/code-server/lib/vscode/bin/remote-cli` is prepended to `PATH`.

This will also generate a settings conflict should you ever consider resolving https://github.com/coder/code-server/issues/1315.  
ℹ️ In VS Code it is not possible to start another remote connection from within a remote connection.

code-asher:
> `/opt/code-server/bin/code-server` is the equivalent to `/opt/code-server/lib/vscode/bin/code-server-oss` and not `/opt/code-server/lib/vscode/bin/remote-cli/code-oss`.

`./bin/code-server` is like `./lib/vscode/bin/code-server-oss` when ran from an external terminal (spawns a server) but it is like `./lib/vscode/bin/remote-cli/code-oss` when ran in an integrated terminal (interacts with the existing code-server).  code-server switches behavior based on the `VSCODE_IPC_HOOK_CLI` environment variable which is set in the integrated terminal.

For example here is a test with `./lib/vscode/bin/remote-cli/code-oss`:

```
$ code-server ~/test
/var/tmp/coder/code-server/lib/vscode/bin/remote-cli/code-server: 12: /var/tmp/coder/code-server/lib/vscode/node: not found
$ sed -i 's/ROOT\/node/ROOT\/..\/..\/node/' /var/tmp/coder/code-server/lib/vscode/bin/remote-cli/code-server
$ code-server ~/test
```

This opened the file `~/test` in my current instance of code-server, which is what `./bin/code-server` would have done and is what I think we want the `code-server` invocation to always do in the integrated terminal.

code-asher:
Sorry for being dense; how does this change anything with https://github.com/coder/code-server/issues/1315?  Our script and upstream's are more or less doing the same thing (run Node and interact with code-server via the socket stored in `VSCODE_IPC_HOOK_CLI`).  The extensions should work regardless of what is in the path right?

code-asher:
The VS Code remote extensions.  Remote - SSH, remote - containers, and remote WSL I think.  Although that issue is only about implementing open-source versions of the first two (SSH and containers).

benz0li:
> Sorry for being dense; how does this change anything with https://github.com/coder/code-server/issues/1315?

Not because of `/opt/code-server/lib/vscode/bin/remote-cli` but `Settings`; This looks somehow confusing:

Settings

But it seems to work correctly. `Workspace Settings` override `Remote Settings` (`~/.local/share/code-server/Machine/settings.json`) override `User Settings` (`~/.local/share/code-server/User/settings.json`).
👉 User Settings being the settings of the _local_ Machine – in VS Code terms.

If I remember correctly: When changing `User Settings` in **older versions** of code-server it always printed _Also modified in: Remote_, displaying the same value in `Remote Settings` as in `User Setting`.  
ℹ️ This might have changed as soon as you have moved to upstream `microsoft/vscode`.

> The VS Code remote extensions. Remote - SSH, remote - containers, and remote WSL I think.

👍 

> Although that issue (#1315) is only about implementing open-source versions of the first two (SSH and containers).

Yes, of course.

benz0li:
@code-asher Thank you for your explanations.

code-asher:
Ahh right I see what you mean!  Yeah the settings situation has been confusing.  Upstream `User` settings were changed to be stored in the browser so now they are local to the client and separate from remote but we patch them to use `~/.local/share/code-server/User/settings.json` which makes them no longer local...definitely a weird situation.  We did that for backwards compatibility but I wonder if we should remove that patch and recommend `~/.local/share/code-server/Machine/settings.json` for seeding settings.  Definitely something to keep in mind as we try to align ourselves with upstream.

benz0li:
> Ahh right I see what you mean!

🙇 

> We did that for backwards compatibility but I wonder if we should remove that patch and recommend `~/.local/share/code-server/Machine/settings.json` instead for seeding settings on the server.

For me the current patch is fine.

---

Cross references: https://github.com/coder/code-server/issues/2274, https://github.com/coder/code-server/issues/4609

ci/build/build-vscode.sh
```diff
@@ -6,12 +6,25 @@ set -euo pipefail
...
 
+delete-bin-script() {
+  rm "lib/vscode-reh-web-linux-x64/bin/$1"
+}
+
+copy-bin-script() {
+  local script="$1"
+  local dest="lib/vscode-reh-web-linux-x64/bin/$script"
+  cp "lib/vscode/resources/server/bin/$script" "$dest"
+  sed -i "s/@@VERSION@@/$(vscode_version)/g" "$dest"
```
jsjoeio, line 17: reminder: may want to look into making sure it works on both macOS and Linux: https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

jsjoeio: > The command (note the lack of space between -i and '' and the -e to make it work on new versions of Mac and on GNU):

```shell
sed -i'.original' -e 's/old_link/new_link/g' *
```

jsjoeio (approved):
We reviewed this together in person at the off-site. All looks good!

PR #5385, opened 2022-07-29
define extraInitContainers

masterkain:
otherwise custom values will not be picked up by Flux/Helm

### DIFF
--- a/ci/helm-chart/values.yaml
+++ b/ci/helm-chart/values.yaml
@@ -147,6 +147,25 @@ extraContainers: |
 #  - name: DOCKER_DRIVER
 #    value: "overlay2"
 
+extraInitContainers: |
+# - name: customization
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Thank you for the PR!  Looks good to me.  I see we have the same in the documentation but seems like a good idea to me to have it in the config as well.

masterkain:
not sure what's going on but `extraInitContainers` still isn't picked up by the deployment (using git chart version with the merged change), needs further investigation

masterkain:
ok, that might be this https://github.com/coder/code-server/blob/main/ci/helm-chart/templates/deployment.yaml#L35-L36

`initContainers` is wrapped in a `volumePermissions ` and `persistence ` conditional, so if it is not specified `extraInitContainers` won't make it

PR #5396, opened 2022-08-01
Correct trial length

1submarine:
Fixes #
Trial length in requirements.md was mistakenly at 1 year.


code-asher (approved):
Thank you for the correction!

PR #5415, opened 2022-08-04
Logging improvement: PROXY-DOMAIN may be a subdomain + domain

ltomes:
This logging is confusing when a user passes a full `<subdomain>.<domain>` as proxy-domain.
As an example if PROXY-DOMAIN = `code.test.com`
Before the logging output was:
```
[2022-08-04T16:15:11.829Z] info    - Proxying the following domain:
[2022-08-04T16:15:11.830Z] info      - *.code.test.com
```
After:
```
[2022-08-04T16:15:11.829Z] info    - Proxying the following domain:
[2022-08-04T16:15:11.830Z] info      - code.test.com
```

This was confusing when attempting to setup code-server on a reverse proxy with a service like [cloudflare tunnels](https://www.cloudflare.com/products/tunnel/).

This is likely not the best fix. If preferred, I can do some additional work here to determine if the string is just a domain, or a subdomain+domain, and switch logging behavior based on that condition, and update the PR accordingly.

### DIFF
--- a/src/node/main.ts
+++ b/src/node/main.ts
@@ -153,7 +153,7 @@ export const runCodeServer = async (
 
   if (args["proxy-domain"].length > 0) {
     logger.info(`  - ${plural(args["proxy-domain"].length, "Proxying the following domain")}:`)
-    args["proxy-domain"].forEach((domain) => logger.info(`    - *.${domain}`))
+    args["proxy-domain"].forEach((domain) => logger.info(`    - ${domain}`))
   }
 
   if (args.link) {


jsjoeio:
@code-asher I think you're the best person to review this so assigning to you

code-asher (commented):
I like the existing `*.` because code-server will proxy for example:

8080.code.test.com
8081.code.test.com

But it will *not* proxy code.test.com (it will just show code-server).

Also there will needs to be wildcard DNS records and certificates available at *.code.test.com.

I might be overthinking this though.  Could you elaborate on the specific confusion you ran into?

ltomes:
> Also there will needs to be wildcard DNS records and certificates available at *.code.test.com.

That was helpful/I had a misconception, I think we should close this PR. 

I just tested running my container without `PROXY-DOMAIN` set, and it worked properly.
I had been under the impression `PROXY-DOMAIN` was working with my reverse proxy but clearly it's not in use as you stated. If I only set the domain, that interferes with my reverse proxy (as expected).

PR #5422, opened 2022-08-07
Patches 1.70

fritterhoff:
Hello :)

I started playing around with the new version 1.70. Some patches must be adjusted...

* The telemetry will need a larger fix - I simply made it a nop client for my first tests
* The storage service maybe must be checked ...
* I assume the language fix is still required but I am not sure if the fix still works. The previously new file ` languagePacks.ts` now contains the following fragment(?):
```
export class WebLanguagePacksService extends LanguagePackBaseService {
	// Web doesn't have a concept of language packs, so we just return an empty array
	getInstalledLanguages(): Promise<ILanguagePackItem[]> {
		return Promise.resolve([]);
	}
}
```

Feel free to close the PR if my changes don't help at all 😉


fritterhoff:
For sure, some further work will be required... Esp. fixing the telemetry seems to be a bit cumberstone

code-asher:
Thanks for getting us started!  I will hopefully have time later this week to finish it out.  :crossed_fingers:

fritterhoff:
FYI: I just rebased my branch on the normal base branch. Sadly the CI needs an approval again 😉

jsjoeio:
I'm going to try adding a label to this PR which should auto-rebase (I think) and even deploy your PR to npm. Let's give it a try.

fritterhoff:
Alright. Maybe it would help if a dummy PR created by me gets merged... According to GitHub only first time contributors require approval 🤣

fritterhoff:
From my point of view the e2e tests did fail due to an internal error and not due to my changes?

jsjoeio:
Yes, looks like an internal GitHub Actions error:

image

code-asher:
Awesome work @fritterhoff, I implemented the telemetry and everything else looks great (I just added back some comments and one other tweak).

Going to do a bit more testing then open this as ready for review.

code-asher:
Display language is not working for me, maybe there is some conflict between our patch and the stuff upstream has been adding.  Ideally we would use upstream's but they mentioned their implementation is not quite ready so we might need to look into how to cut it out for now.

fritterhoff:
> Awesome work @fritterhoff, I implemented the telemetry and everything else looks great (I just added back some comments and made two other small tweaks not really related to the update).
> 
> Going to do a bit more testing then open this as ready for review.

Great thanks for the feedback and sorry about the comments... in a first step I was happy that I had a working version 1.70 😉 

> Display language is not working for me, maybe there is some conflict between our patch and the stuff upstream has been adding. Ideally we would use upstream's but they mentioned their implementation is not quite ready so we might need to look into how to cut it out for now.

I'm not sure what the expected behavior is or what isn't working here? I only tried to install a language pack and that was working 😁 So feel free to give me a hint what got broken in the new version 😉

code-asher:
Here is the behavior I see (I ran from the latest 
linux-amd64.tar.gz artifact from CI):

- Launch code-server
- Extensions > Search "Japanese"
- Click "Set display language"
- Click "Restart"
- Still in English

Interestingly I just tried with Spanish and it worked.  I also 
tried Chinese but it did *not* work.  Which language pack did you 
try?

It is also interesting that the button says "set display language" 
now instead of "install" like it used to.

Lastly, when I run "configure display language" only English shows 
for me.  It is not showing all the language packs I have 
installed.

fritterhoff:
Hi, I tried the German language pack. That worked fine. I'll check it again later and update my comment

fritterhoff:
So spanish and German work at least in the Installation fine. Chinese does not work.
Setting via the command menu does not work at all...

jsjoeio:
1.70 might have broken our language patch :( at least that's what we were discussing offline. Thank you for testing those and reporting back!

fritterhoff:
Well ok. I sort of expected something like that 😔

code-asher:
I should have time to look into it this week if no one else gets 
to it first.

code-asher:
Language issues should be sorted, I will do a final check on the artifact CI builds then merge up.

code-asher (approved):

patches/unique-db.diff
```diff
@@ -26,23 +26,25 @@ Index: code-server/lib/vscode/src/vs/workbench/services/storage/browser/storageS
...
+@@ -67,7 +68,11 @@ export class BrowserStorageService exten
  					return `global-${this.profileStorageProfile.id}`;
  				}
  			case StorageScope.WORKSPACE:
 -				return this.payload.id;
++				// Add a unique ID based on the current path for per-workspace databases.
++				// This prevents workspaces on different machines that share the same domain
++				// and file path from colliding (since it does not appear IndexedDB can be
++				// scoped to a path) as long as they are hosted on different paths.
 +				return this.payload.id + '-' + hash(location.pathname.toString().replace(/\/$/, "")).toString(16);
```
jsjoeio, line 38: Will this commit affect the migration/state collisions e2e test we have?

patches/disable-downloads.diff
```diff
@@ -108,7 +108,7 @@ Index: code-server/lib/vscode/src/vs/workbench/browser/contextkeys.ts
  import { Schemas } from 'vs/base/common/network';
  import { WebFileSystemAccess } from 'vs/platform/files/browser/webFileSystemAccess';
  import { IProductService } from 'vs/platform/product/common/productService';
-+import { IBrowserWorkbenchEnvironmentService } from '../services/environment/browser/environmentService';
++import { IBrowserWorkbenchEnvironmentService } from 'vs/workbench/services/environment/browser/environmentService';
```
jsjoeio, line 111: Nice catch! I don't know why i didn't do that before lol thanks for fixing.

patches/display-language.diff
```diff
@@ -312,3 +314,19 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/localization/electron-san
...
++++ code-server/lib/vscode/src/vs/base/node/languagePacks.js
+@@ -73,7 +73,10 @@
+ 		function getLanguagePackConfigurations(userDataPath) {
+ 			const configFile = path.join(userDataPath, 'languagepacks.json');
+ 			try {
+-				return nodeRequire(configFile);
++				// This must not use Node's require otherwise it will be cached forever.
++				// Code can get away with this since the process actually restarts but
++				// that is not currently the case with code-server.
++				return JSON.parse(fs.readFileSync(configFile, "utf8"));
```
jsjoeio, line 329: Nice fix! How did you figure this out?

patches/unique-db.diff
code-asher: Nope it operates the same way; things just moved around a little bit.

patches/display-language.diff
code-asher: I actually fixed this before but it was lost when we switched to the fork so basically I was stepping through the code until suddenly I remembered this issue.

The first time around it just involved stepping through the code one step at time to figure out why the language object was empty until I arrived at this line and realized it would never update since requires are cached.  I am not sure where that knowledge came from, probably read it somewhere or maybe it was from when we implemented tarfs which hijacks require.

PR #5434, opened 2022-08-09
feat: add --help integration test

### LINKED ISSUE #5369: [Chore]: Investigate integration test flakes
## Description

For some reason, the `installExtesion` test has been flaking out and we're not sure why:
https://github.com/coder/code-server/runs/7458796120?check_suite_focus=true#step:9:13

## Solution

Replace it with another test like `--help` which fails if used on the wrong platform 

For example, running `--help` with the Linux release on macOS results in this error: 
image

jsjoeio:
## Description

Previously, we added the `--install-extension` test to ensure that native modules worked as expected with the binary.

However, that relies on a network connection (see https://github.com/microsoft/vscode/issues/142812#issuecomment-1194155328). Instead, I propose this new integration test which checks the `--help` flag
which fails if used on the wrong platform.

Fixes #5369

### DIFF
--- a/test/integration/help.test.ts
+++ b/test/integration/help.test.ts
@@ -1,0 +1,13 @@
+import { runCodeServerCommand } from "../utils/runCodeServerCommand"
+
+// NOTE@jsjoeio
+// We have this test to ensure that native modules
+// work as expected. If this is called on the wrong
+// platform, the test will fail.
+describe("--help", () => {
+  it("should list code-server usage", async () => {
+    const expectedOutput = "Usage: code-server [options] [path]"
+    const { stdout } = await runCodeServerCommand(["--help"])
+    expect(stdout).toMatch(expectedOutput)
+  }, 20000)
+})


test/integration/help.test.ts
```diff
@@ -0,0 +1,13 @@
+import { runCodeServerCommand } from "../utils/runCodeServerCommand"
+
+// NOTE@jsjoeio
+// We have this test to ensure that native modules
+// work as expected. If this is called on the wrong
+// platform, the test will fail.
```
jsjoeio, line 6: 🙅🏼‍♂️ I know we generally avoid comments like this and prefer putting this info in commits. But I feel like this should be front and center so that future developers don't accidentally delete this test.

I can be convinced otherwise though.

code-asher: Yeah this seems like a reasonable comment to me because it is not self-apparent why we test `--help`.  

One way around that would be to change the wording of the test, maybe something like `describe("native-modules")` and `test("it should trigger native modules by running --help)` so the reasoning is baked into the code but IMO either way is good (I probably would have gone the comment route as well).

code-asher (approved):
Nice!  We can remove the list/install extensions tests since now right?  (Or should we just leave them in anyway?)

test/integration/help.test.ts
jsjoeio: I hadn't thought of going with the test name like that! Good point.

Cool! Glad we're in agreement. Happy to change later if we change our minds as well!

PR #5442, opened 2022-08-10
[Security Issue] Avoid Exposing Ports without Authentication 

rafaelcalpena:
Hi, I have recently started using `code-server` and the proxy feature is very handy, however I realized that anyone can access proxied ports on public domains. The `authenticated` call seems to be working correctly, but the proxy function does not wait asynchronously for the authentication result. 

I believe this is a high-priority issue with consequences such as disclosure of unwanted data and potential priviledge escalation on the system. I have not tested my fix, but only minor changes were necessary, so it should be ready to go.


code-asher (approved):
Thank you!! This is definitely a big deal.  @jsjoeio we should probably try releasing soon with this fix.

jsjoeio:
Big thank you for finding and fixing @rafaelcalpena!

> we should probably try releasing soon with this fix.

Agreed!

rafaelcalpena:
Thanks @code-asher and @jsjoeio for merging and releasing this!

PR #5451, opened 2022-08-12
fix: socket disconnections behind reverse proxies

code-asher:
Fixes https://github.com/coder/code-server/issues/5378


jsjoeio (approved):
Very happy to see a fix for this. This seems like one that should be submitted upstream potentially.

code-asher:
> This seems like one that should be submitted upstream potentially.

Definitely agree.

PR #5490, opened 2022-08-22
fix: Update publish.yaml

jsjoeio:
Fixes #


code-asher (approved):
Man yaml so is hard to check by hand, I completely missed this.

jsjoeio:
> Man yaml so is hard to parse visually, I completely missed this.

Agreed! This one has happened to me many times before.

PR #5494, opened 2022-08-23
chore(testing): change timeouts, failures for e2e tests

### LINKED ISSUE #5435: [Chore]: Speed up e2e tests
## What is your suggestion?

To speed up our e2e tests using one of the two options:

### Option 1 - one test per file

> By default, tests in a single file are run in order. If you have many independent tests in a single file, you might want to run them in parallel with [test.describe.configure([options])](https://playwright.dev/docs/api/class-test#test-describe-configure).

They won't parallelize multiple tests in a single file so we could split tests out to individual files.

[Source](https://playwright.dev/docs/test-parallel#parallelize-tests-in-a-single-file)

### Option 2 - figure out how to use configure

Supposedly there is a method: `test.describe.configure({ mode: 'parallel' });`. However, we extend `test` and it doesn't seem to exist when I try to use it locally. I've only managed to get `test.describe.parallel` semi-working.

I want to speed up CI and cut down on e2e run times. 

image

No

Yes

jsjoeio:
- chore(e2e): update playwright timeouts
- chore(e2e): update maxFailures to 5 in CI
- refactor: duplicate onLine to test/utils/helpers
- Revert "refactor: duplicate onLine to test/utils/helpers"

Fixes #5435


test/playwright.config.ts
```diff
@@ -10,10 +10,10 @@ import path from "path"
 //   yarn test:e2e --grep login       # Run tests matching "login"
 const config: PlaywrightTestConfig = {
   testDir: path.join(__dirname, "e2e"), // Search for tests in this directory.
-  timeout: 60000, // Each test is given 60 seconds.
+  timeout: 30000, // Each test is given 25 seconds.
```
code-asher, line 13:
```suggestion
  timeout: 30000, // Each test is given 30 seconds.
```

Although we might want to remove the comment since otherwise we have to update it every time we change the timeout and doing the math is pretty easy.  We could do `// In milliseconds.` in case the unit is not obvious.

code-asher (approved):

test/playwright.config.ts
jsjoeio: Good idea, I'm guessing I copied that from Playwright. I'll update.

jsjoeio:
It appears that this is having the reverse effect. I'm not seeing the improvements I'd like with this approach so closing for now.

PR #5495, opened 2022-08-23
feat: add displayLang test

### LINKED ISSUE #5367: [Feat]: add e2e test for display-language patch
## What is your suggestion?

Add an e2e test for our [display-language patch](https://github.com/coder/code-server/blob/main/patches/display-language.diff).

To prevent a regression like [this](https://github.com/coder/code-server/issues/5318) which happened in 4.5.0

Manually testing

Yes.

jsjoeio:
This PR adds an e2e test to ensure our language patch is working as expected.

[7b3af0074186a31b540173c75758b033.webm](https://user-images.githubusercontent.com/3806031/186280990-d5d43c1e-cd81-4301-b25a-dabda9687b89.webm)

NOTE to self: won't pass with dev builds because lang. patch only works with prod builds. Might need to conditionally run this test. 

## Notes

When you install a language pack, two things happen:
- the extension is unzipped to `--extensions-dir` (expected, same as other extensions)
- it adds a file called `languagepacks.json` to `~/.local/share/code-server` (see [example gist](https://gist.github.com/jsjoeio/94fa0202b2204342e5a225987a6e6a76))

That means in order for this test to pass consistently (without relying on a network connection), we need to add both of those things in global setup or in the test setup.

To test locally and just this test, I run this command:
```shell
CODE_SERVER_TEST_ENTRY="/usr/local/Cellar/code-server/4.6.0/libexec" yarn test:e2e --grep "load code-server in Spanish" --quiet
```

Fixes #5367


test/e2e/models/CodeServer.ts
```diff
@@ -88,6 +88,13 @@ export class CodeServer {
       }),
       "utf8",
     )
+
+    const extensionsDir = path.join(__dirname, "../extensions")
+
+    await fs.writeFile(
+      path.join(dir, "languagepacks.json"),
+      `{"es":{"hash":"8d919a946475223861fa0c62665a4c50","extensions":[{"extensionIdentifier":{"id":"ms-ceintl.vscode-language-pack-es","uuid":"47e020a1-33db-4cc0-a1b4-42f97781749a"},"version":"1.70.0"}],"translations":{"vscode":"${extensionsDir}/ms-ceintl.vscode-language-pack-es-1.70.0/translations/main.i18n.json"},"label":"español"}}`,
```
code-asher, line 96: We could use `JSON.stringify` here for readability.

test/e2e/models/CodeServer.ts
```diff
@@ -88,6 +88,13 @@ export class CodeServer {
       }),
       "utf8",
     )
+
+    const extensionsDir = path.join(__dirname, "../extensions")
+
+    await fs.writeFile(
```
code-asher, line 94: Should we leave a comment here saying this is a workaround until we fix the file not generating automatically?

test/e2e/extensions/ms-ceintl.vscode-language-pack-es-1.70.0/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "vscode-language-pack-es",
```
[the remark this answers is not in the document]
code-asher: We should make sure anything we merge can be found in the open source.  https://github.com/microsoft/vscode-loc/tree/main/i18n/vscode-language-pack-es  (could also start from scratch and add our own).  For example `__metadata` is specific to Microsoft's marketplace and should probably get axed.

code-asher (approved):

test/e2e/models/CodeServer.ts
jsjoeio: That sounds like a better approach! I'll refactor

test/e2e/models/CodeServer.ts
jsjoeio: Yes! Otherwise, we will surely forget 😂

jsjoeio:
> We should make sure anything we merge can be found in the open source. https://github.com/microsoft/vscode-loc/tree/main/i18n/vscode-language-pack-es (could also start from scratch and create our own)

Good call. I'll double-check that and/or explore starting from scratch.

I'm going to move this back to draft and address changes.

code-asher (approved):

PR #5510, opened 2022-08-26
fix: propagate execArgv

code-asher:
This allows us to use `--prof` to profile code-server.


jsjoeio (approved):
Sounds like this will allow us to better debug extensions? Nice!

PR #5518, opened 2022-08-29
docs: remove references to installing with yarn in favor of npm

edvincent:
Follow-up of https://github.com/coder/code-server/pull/5071#discussion_r950597128 to update the docs.

Only left references to `yarn` commands for the development process.


docs/termux.md
```diff
@@ -104,7 +103,7 @@ you will get node version `v16.15.0`
 5. Now install code-server
 
 ```sh
-yarn global add code-server
+npm install --global code-server
 ```
```
edvincent, line 107: Should this simply be a link to https://coder.com/docs/code-server/latest/npm - rather than re-explaining the process?

jsjoeio: That makes sense to me. Either works - I'll leave that up to you!

docs/install.md
```diff
@@ -346,12 +344,12 @@ brew remove code-server
...
 ```
 
-### yarn, npm
+### npm
 
 To remove the code-server global module, run:
 
 ```shell
-yarn global remove code-server
+npm -g uninstall code-server
```
jsjoeio, line 352: Since we're using `--global` elsewhere:

```suggestion
npm --global uninstall code-server
```

jsjoeio (approved):
Woo! Thanks for doing this!

docs/install.md
edvincent: Ah yes! Plus I actually forgot to remove the "or" section just below. Updated to `--global` + removing that section.

docs/termux.md
edvincent: I like consistency and less duplication... Latest amended commit just redirects to existing guide.

jsjoeio (approved):

edvincent:
Actually... Where's my head. Forgot about the `--unsafe-perm` that `npm` requires... Updating coming up!

jsjoeio:
> Actually... Where's my head. Forgot about the --unsafe-perm that npm requires... Updating coming up!

Nice catch! Remind me again, why do we need this?

edvincent:
> Nice catch! Remind me again, why do we need this?

https://github.com/coder/code-server/blob/main/ci/build/npm-postinstall.sh#L95-L101

Now, is it actually needed? Not sure. Definitely something I'm happy to look into in the future.

jsjoeio:
I honestly can't remember. @code-asher might know. Probably fine to add for now then we can revert/fix if needed.

edvincent:
> I think we only need it if someone is installing code-server with root because NPM will drop permissions in the post install scripts making the code-server install fail.

So we should add a check for the user running the script, and only show the warning if it's `root`? Happy to play around with that and send a separate PR for that in the next couple of days.

code-asher:
Yeah I think that would be ideal!!

PR #5519, opened 2022-08-30
refactor: drop db migration patch

jsjoeio:
This drops the patch and also the tests (which were flakey and our longest running e2e tests).

- refactor: remove database migration patch & test
- chore: refresh patches

Fixes https://github.com/coder/code-server/issues/5482

### DIFF
--- a/patches/telemetry.diff
+++ b/patches/telemetry.diff
@@ -104,6 +104,13 @@ Index: code-server/lib/vscode/src/vs/workbench/services/telemetry/browser/teleme
 -				sendErrorTelemetry: this.sendErrorTelemetry,
 -			};
 -			this.impl = this._register(new BaseTelemetryService(config, configurationService, productService));
+-
+-			if (getTelemetryLevel(configurationService) !== TelemetryLevel.NONE) {
+-				// If we cannot fetch the endpoint it means it is down and we should not send any telemetry.
+-				// This is most likely due to ad blockers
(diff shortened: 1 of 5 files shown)


patches/telemetry.diff
```diff
@@ -104,6 +104,13 @@ Index: code-server/lib/vscode/src/vs/workbench/services/telemetry/browser/teleme
 -				sendErrorTelemetry: this.sendErrorTelemetry,
 -			};
 -			this.impl = this._register(new BaseTelemetryService(config, configurationService, productService));
+-
```
jsjoeio, line 107: @code-asher I refreshed all the patches. Does this look okay to you?

code-asher: Looks like the blocks changed a little, probably due to using a different diff algorithm, but I think it should be fine!  Looks like I am using `patience` (you can see what you have by running `git config diff.algorithm` in the code-server repo).

code-asher (approved):

patches/telemetry.diff
jsjoeio: Uh...looks like I'm not using one? 😅 Reminds me of when I first started at Coder and Anmol/I were using different algorithms.

image

Okay I will leave as is for now then and merge this!

code-asher: Ah that probably means it is using the default (`myers`).

PR #5550, opened 2022-09-09
fix(ci): specify ref in aur publish job

jsjoeio:
I've noticed the last two times the `aur` job runs and opens a PR into `code-server-aur`, there are conflicts. It seems like it's not getting the latest commits. Based on [this](https://github.com/actions/checkout/issues/439#issuecomment-830862188), I think specifying `ref` may help.

I do not see any downsides to adding this even if this doesn't fix it.


code-asher (approved):
Ahh that makes sense!

PR #5560, opened 2022-09-15
feat: add release workflow

jsjoeio:
# Description

This PR is Part 1 / ? of the release workflow refactor. This adds a new workflow called `release.yaml` which creates a draft GitHub release and uploads artifacts from `main`. It assumes you've bumped the version in the `package.json` already.

## Changes
- feat(ci): add draft release workflow
- refactor: delete old release-github workflows
- fixup! refactor: delete old release-github workflows

## Testing Plan

I tested this on my fork and was able to use it successfully.

Run: https://github.com/jsjoeio/code-server/actions/runs/3064111088

image

Related to https://github.com/coder/code-server/issues/2817


.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  draft:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      # TODO: delete our release scripts and the ci/lib.sh funcs they use
```
jsjoeio, line 24:
```suggestion
```

.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
...
+          workflow: ci.yaml
+          workflow_conclusion: completed
+          check_artifacts: true
+          name: release-packages
+          path: ./release-packages
+
+      - uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          discussion_category_name: "📣 Announcements"
```
jsjoeio, line 39: I wasn't able to test this since I don't have Discussions on my fork (not sure if possible). I don't think this should cause any issue though 🤞🏼

.github/workflows/release.yaml
```diff
@@ -0,0 +1,41 @@
+name: Draft release
+
+on:
+  workflow_dispatch:
```
jsjoeio, line 4: I decided for the first version of this to not add on new tags but we can add that in next version. Happy to be convinced to add it here though.

code-asher: Makes sense to me!

.github/workflows/release.yaml
```diff
@@ -0,0 +1,40 @@
...
+          check_artifacts: true
+          name: release-packages
+          path: ./release-packages
+
+      - uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          discussion_category_name: "📣 Announcements"
+          files: ./release-packages/*
+          body_path: ci/build/release-notes.txt
```
code-asher, line 40: We can probably remove this since we will just overwrite this when copying from the changelog right?  I wonder if we can somehow get fancy and automatically extract the last block from the changelog.

docs/MAINTAINING.md
```diff
@@ -137,43 +137,19 @@ changelog](https://github.com/emacs-mirror/emacs/blob/master/etc/NEWS).
...
-   to build.
-1. Run `yarn release:github-assets` to download the `release-packages` artifact.
-   They will upload them to the draft release.
-1. Run some basic sanity tests on one of the released packages (pay special
-   attention to making sure the terminal works).
-1. Publish the release and merge the PR. CI will automatically grab the
+1. Summarize the major changes in the `CHANGELOG.md`
+1. Download CI artifacts and make sure code-server works locally.
+1. Merge PR and wait for CI build on `main` to finish.
+1. Go to GitHub Actions > Draft release > Run workflow off `main`. CI will automatically upload the artifacts to the release.
```
code-asher, line 149: So much better already!

code-asher (approved):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Download artifacts
+        uses: dawidd6/action-download-artifact@v2
+        id: download
+        with:
+          branch: main
```
code-asher, line 28: Actually do we need the branch?  We could just use whatever artifacts are on the branch we set the workflow to target when dispatching it I think?  That would let us release off different branches too if for some reason we needed to do that (like if we needed to backport a fix).

.github/workflows/release.yaml
jsjoeio: Yeah that's true. I'll delete `release-notes.txt` too since it's already in the changelog.

> I wonder if we can somehow get fancy and automatically extract the last block from the changelog.

Haha that would be awesome! Probably some script step before this one that parses `CHANGELOG` -> grabs relevant block -> outputs to file -> we use that.

.github/workflows/release.yaml
jsjoeio: I guess not! That's a good point so I guess I'll change this to `${{ github.ref }}`

PR #5563, opened 2022-09-16
refactor(ci): fix fetch-depth and add some caching

jsjoeio:
This adds caching for the `node_modules` in platform steps, `quilt` and fixes the `fetch-depth` so checkouts are faster.

Also renames `ci.yaml` -> `build.yaml`.

Fixes N/A


.github/workflows/build.yaml
```diff
@@ -42,17 +41,17 @@ jobs:
         uses: azure/setup-helm@v3.3
 
       - name: Fetch dependencies from cache
-        id: cache-yarn
+        id: cache-root-node-mods
```
jsjoeio, line 44: 🧹 I thought this name was more clear

.github/workflows/build.yaml
```diff
@@ -42,17 +41,17 @@ jobs:
...
+          path: "node_modules"
+          key: yarn-build-${{ hashFiles('yarn.lock') }}
           restore-keys: |
             yarn-build-
 
       - name: Install dependencies
-        if: steps.cache-yarn.outputs.cache-hit != 'true'
-        run: yarn --frozen-lockfile
+        if: steps.cache-root-node-mods.outputs.cache-hit != 'true'
+        run: SKIP_SUBMODULE_DEPS=1 yarn --frozen-lockfile
```
jsjoeio, line 54: 🧹 I realized we don't need the submodules checked out to run `fmt` or `lint` so I think we can skip installing submodule deps here.

jsjoeio:
Ah, right, we have `node_modules` in `/test` so I need to cache those too.

.github/workflows/build.yaml
```diff
@@ -40,9 +37,14 @@ jobs:
 
       - name: Install helm
         uses: azure/setup-helm@v3.3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
jsjoeio, line 41: Fixes warning from `helm`. Token needed for v3 and later.

.github/workflows/build.yaml
```diff
@@ -40,9 +37,14 @@ jobs:
...
+
+      - name: Install helm kubeval plugin
+        run: helm plugin install https://github.com/instrumenta/helm-kubeval
```
jsjoeio, line 44: We need this for our `lint.sh` to actually lint the Helm chart.

.github/workflows/build.yaml
```diff
@@ -398,7 +404,17 @@ jobs:
...
         run: yarn release:standalone
 
+      - name: Fetch dependencies from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "**/node_modules"
+          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            yarn-build-
```
code-asher, line 414: Same comment as previous except `yarn-build-macos-`.

.github/workflows/build.yaml
```diff
@@ -262,8 +262,18 @@ jobs:
...
         run: source scl_source enable devtoolset-9 && yarn release:standalone
 
+      - name: Fetch dependencies from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "**/node_modules"
+          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            yarn-build-
```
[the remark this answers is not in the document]
code-asher: Alternatively we could remove just the `npm install` line and put that in the yaml.  Or make `yarn release:standalone` take an argument for whether to install.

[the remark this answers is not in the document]
jsjoeio: Good thinking! I'm going to add this to my todo list since I didn't get to it here. I'm thinking we do the approach you suggested and move the platform steps out of `build.yaml` so maybe I can tackle this with that!

PR #5572, opened 2022-09-19
feat(ci): refactor build workflow

jsjoeio:
This PR does some major refactoring to our `build.yaml` workflow, specifically the `prebuild` step. The two themes of this PR are:
- parallelization
- fail-fast

It splits up the prebuild jobs - lint, format, etc - and runs them in separate jobs so they can run in parallel. It also adds a new step to each which cancels the entire workflow if any of them fail. This way, we can fail-fast and not waste developer time or CI minutes. Plus, it allows us to run these prebuild checks _in parallel_ to building code-server which means faster feedback loop.

Fixes N/A


.github/workflows/build.yaml
```diff
@@ -35,6 +35,41 @@ jobs:
         with:
           node-version: "16"
 
+      - name: Fetch dependencies from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "node_modules"
```
jsjoeio, line 42: ✍🏼 Since these jobs like `fmt` don't require any dependencies besides the root ones, I left the caching to only these `node_modules`. Happy to change back if desired.

jsjoeio, line 42: ✍🏼 Since these jobs like `fmt` don't require any dependencies besides the root ones, I left the caching to only these `node_modules`. Happy to change back if desired.

.github/workflows/build.yaml
```diff
@@ -43,30 +78,150 @@ jobs:
...
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v23.2
+        with:
+          files: |
+            **.ts
+            **.tsx
```
jsjoeio, line 147: ❓We don't have any `.tsx` files so we could remove this technically

.github/workflows/build.yaml
```diff
@@ -43,30 +78,150 @@ jobs:
...
+      - name: Lint TypeScript files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: yarn lint:ts
+
+      - name: Fail workflow
+        if: failure()
+        uses: andymckay/cancel-action@0.2
+
+  typecheck:
+    name: Check TypeScript types
```
jsjoeio, line 181: ✍🏼 If we decide to move away from `ts-jest` we can modify this job to also typecheck `/test`

.github/workflows/build.yaml
```diff
@@ -43,30 +78,150 @@ jobs:
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v23.2
+        with:
+          files: |
+            ./src/**/*
```
jsjoeio, line 193: 🤔 I might be able to change this to `./src` but not sure

ci/dev/helm.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
jsjoeio, line 1: I was going to add this as a `yarn` script but then it'd require the CI job to install Node just to have `yarn` so I opted for a separate bash script

ci/dev/test-integration.sh
```diff
@@ -24,6 +24,13 @@ main() {
     path="$CODE_SERVER_PATH"
   fi
 
+  if [[ ! -f "test/integration/test-plugin/out/index.js" ]]; then
```
jsjoeio, line 27: Small optimization: only build test-plugin if not already built.

test/integration/test-plugin/src/index.ts
```diff
@@ -1,22 +1,26 @@
+// TODO@jsjoeio - how do I fix this?
+// @ts-ignore - we know code-server exists
```
jsjoeio, line 2: 🤔 not sure why this wasn't happening before I move it. Maybe I need to `tsconfig.json` ignore the test plugin? Any ideas?

test/unit/node/app.test.ts
```diff
@@ -107,18 +107,6 @@ describe("createApp", () => {
     app.dispose()
   })
 
-  it("should change the file mode of a socket", async () => {
```
jsjoeio, line 110: Flakey test so decided we should remove it. Happy to add back if desired.

.github/workflows/build.yaml
code-asher: I think if we do this we should make the postinstall script not install the extra dependencies in `test` otherwise it will do that every time since they will not be cached.

code-asher: I am not sure but it might be a little less efficient overall to do it this way because we would have one cache with node_modules, another cache with node_modules plus test/node_modules.

code-asher: Also if the cache contents are different I am not sure they should share the same fallback restore keys.  Sometimes it would have node_modules, sometimes it would also include test/node_modules.

.github/workflows/build.yaml
```diff
@@ -43,30 +78,150 @@ jobs:
...
+          restore-keys: |
+            yarn-build-
+
+      - name: Install dependencies
+        if: steps.changed-files.outputs.any_changed == 'true' && steps.cache-node-modules.outputs.cache-hit != 'true'
+        run: SKIP_SUBMODULE_DEPS=1 yarn --frozen-lockfile
+
+      - name: Run tsc on TypeScript files
+        if: steps.changed-files.outputs.any_changed == 'true'
+        run: yarn typecheck
```
code-asher, line 217: I wonder if because we typecheck here we can skip typechecks on the build step.

ci/dev/test-integration.sh
code-asher: Could this cause issues when developing?  For example making a change to the plugin but then it does not rebuild.

Actually I think we can skip this because `make` is supposed to avoid rebuilding if not necessary.

ci/dev/test-integration.sh
```diff
@@ -24,6 +24,13 @@ main() {
...
+    echo "Building test plugin"
+    pushd test/integration/test-plugin
+    make -s out/index.js
```
code-asher, line 30: Make seems to be failing in CI, might need to install buildtools?

package.json
```diff
@@ -26,8 +26,11 @@
     "publish:docker": "./ci/steps/docker-buildx-push.sh",
     "_audit": "./ci/dev/audit.sh",
     "fmt": "./ci/dev/fmt.sh",
-    "lint": "./ci/dev/lint.sh",
+    "lint:helm": "./ci/dedv/helm.sh",
```
code-asher, line 29: Typo I think `dedv` -> `dev`

ci/dev/helm.sh
code-asher: Would it make sense to put `helm kubeval ci/helm-chart` directly in the yaml?

test/unit/node/app.test.ts
code-asher: It seems odd to me that this flakes so it could be an actual bug.  IMO we leave tests even if flaky.

test/integration/test-plugin/src/index.ts
code-asher: Might need to update the path here: https://github.com/coder/code-server/blob/f9b3bda9d5ce0e3185b7dfd9205615871302e1d5/test/integration/test-plugin/tsconfig.json#L47

.github/workflows/build.yaml
code-asher: Yeah I never know if these actions treat directories recursively or not haha

ci/dev/helm.sh
jsjoeio: Good idea! I doubt either of us have run this locally and then we have one less script to maintain.

ci/dev/test-integration.sh
jsjoeio: Ah TIL! Cool, I'l remove it then.

test/unit/node/app.test.ts
jsjoeio: Good call. I'll add back in.

.github/workflows/build.yaml
```diff
@@ -35,6 +35,41 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v23.2
+        with:
+          files: |
+            ci/helm-chart/**
+
       - name: Install helm
```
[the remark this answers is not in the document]
jsjoeio: 🔥 didn't even think of that, good call!

ci/dev/test-integration.sh
jsjoeio: I saw that! Super weird because I see this: 

```
          yum install -y devtoolset-9-{make,gcc,gcc-c++} jq rsync python3
```

Online though it says to do: 
```
yum groupinstall "Development Tools"
```

If it happens again, I'll try that.

package.json
jsjoeio: Nice catch! Fixed

.github/workflows/build.yaml
jsjoeio: Ahhh okay. Good points. I think I'll change back to how it was before. I do want to revisit the caching in a follow-up PR because `setup-node` supports caching out of the box, so no need to do it manually. When I do that, we can revisit the idea of splitting up caches.

.github/workflows/build.yaml
```diff
@@ -35,6 +35,41 @@ jobs:
...
+      - name: Fetch dependencies from cache
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: "node_modules"
+          key: yarn-build-${{ hashFiles('yarn.lock') }}
+          restore-keys: |
+            yarn-build-
+
+      - name: Install dependencies
```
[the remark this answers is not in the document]
jsjoeio: I think it should be okay. From the testing I've done it hasn't seemed to be a problem (though I haven't tried changing a dependency). I try that now though and see.

.github/workflows/build.yaml
jsjoeio: I like that! Help me out here...where would we skip the typecheck on the build step?

.github/workflows/build.yaml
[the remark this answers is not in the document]
jsjoeio: https://github.com/coder/code-server/actions/runs/3093663725

Seems like it's okay as is to me. What do you think?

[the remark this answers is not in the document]
code-asher: Only one of them ran since the rest have no changed files so it is hard to tell if there will be a problem in cases where multiple of them will runp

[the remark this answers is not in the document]
code-asher: If they did all run though it looks like it would take about an extra two minutes (30s per cache restore and install) and idk about the cache upload, some of them might fail or it time out or maybe overwrite each other.

.github/workflows/build.yaml
code-asher: I think we would add `--transpileOnly` to the `tsc` command.

code-asher: Oh wait this might not exist yet https://github.com/microsoft/TypeScript/issues/29651

code-asher: Yeah nevermind but we can do it if we switch to something like esbuild.

.github/workflows/build.yaml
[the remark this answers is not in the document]
jsjoeio: Hmm...talking out loud for a second, these are the jobs running in parallel:
- "Format with Prettier"
- "Lint helm chart"
- "Lint shell files"
- "Lint TS files"
- "Check TypeScript types"

3 rely on `node-modules` - prettier and two TS jobs. 30s per cache restore x 3 = 1.5 mins worst-case scenario if the change affects all 3. Example of that would be change to `*.ts` file _and_ a change to a `yarn.lock`. I don't think `yarn.lock` changes too often but worth testing. I'll add a test commit for this.

> The cache is scoped to the key and branch. The default branch cache is available to other branches.

Didn't realize this. Since they all use the same key, they could overwrite each other. We could wait until we have evidence that problem exists before we preemptively solve for it?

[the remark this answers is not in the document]
code-asher: We should test that these workflows actually run when changes are made.  It sounds like they might not be?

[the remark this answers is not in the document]
code-asher: > We could wait until we have evidence that problem exists before we preemptively solve for it?

IMO this is something we can easily test now before merging.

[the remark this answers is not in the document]
jsjoeio: Dude, you the man! Thanks for checking that conflict. Great news 🎉

Okay let me take these action items and address them so we can move this forward:
- [ ] rename `scripts.yml` to `scripts.yaml`
- [ ] move lint scripts to that workflow
- [ ] use `apt` so we don't need node
- [ ] fix typecheck step based on your run
- [ ] fix lint typescript step not finding changes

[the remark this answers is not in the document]
code-asher: First three definitely optional, totally down to leave as-is if you wanna get this PR merged ASAP.

For the typecheck script actually I am curious for the reasoning because we would already be doing it in the build step.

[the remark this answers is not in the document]
jsjoeio: > For the typecheck script actually I am curious for the reasoning because we would already be doing it in the build step

Originally added it for two reasons:
1. because i was going to remove it from our tests and move away from `ts-jest`
2. faster feedback instead of waiting for the Build step (but probably doesn't happen much faster)

I don't think those are good enough reasons now that I'm writing this though so I'll remove it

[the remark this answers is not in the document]
code-asher: Also I was editing my comment like crazy, let me know if you have any thoughts on the cancel action portion.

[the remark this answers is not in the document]
code-asher: >  because i was going to remove it from our tests and move away from ts-jest
>  faster feedback instead of waiting for the Build step (but probably doesn't happen much faster)

Very fair points, we should definitely revisit a separate typecheck step when we investigate alternate build tools.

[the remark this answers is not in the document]
jsjoeio: > And it could be nice to verify if the build at least works even if you have, for example, a helm error (to avoid the pattern where you only see one error from a job at a time and have to keep waiting to see each error rather than see them all at once).

I see what you mean. Wouldn't you end up having to rebuild regardless though after you push a fix for the helm linting error? My thinking was at least you'd see CI in a "done" state faster and be able to fix it but maybe not. I could go both ways.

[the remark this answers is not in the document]
jsjoeio: Also if you think about it...the "cancel workflow" kinda mimics what we had before when it was called "Prebuild". "Build" only ran if "Prebuild" passed so it was essentially the same flow (not saying we have to keep it but wanted to mention).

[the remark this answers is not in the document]
jsjoeio: I can't reproduce your TS error locally 🤔


```shell
❯ yarn tsc -p tsconfig.json --noEmit --skipLibCheck
yarn run v1.22.18
$ /Users/jp/dev/oss/code-server/node_modules/.bin/tsc -p tsconfig.json --noEmit --skipLibCheck
✨  Done in 1.77s.
```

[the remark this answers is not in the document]
jsjoeio: oh wow...so i was looking at our current [`lint.sh`](https://github.com/coder/code-server/blob/main/ci/dev/lint.sh) and it has:
```shell
tsc --noEmit --skipLibCheck
```

_that's_ why I added the typecheck step 🤦🏼‍♂️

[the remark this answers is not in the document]
jsjoeio: @code-asher I think it's ready for another review.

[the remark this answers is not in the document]
code-asher: > the "cancel workflow" kinda mimics what we had before when it was called "Prebuild"

True!  I think the main difference is that the cancel action seems to fail on forks and prevent cache uploads.

>  Wouldn't you end up having to rebuild regardless though after you push a fix for the helm linting error?

Yup, but my thinking is that once a check shows a red X you have two choices:

1. Wait for the rest of CI to see if anything else went wrong
2. Immediately fix and push the lint error

Having the option of the first is nice because depending on the timing of the parallel steps you could run into something like:

Ah I have a Helm error.  The workflow was canceled so no idea if there are any other errors.  I will push up the Helm fix, wait for the build again.  Ah I have a Typescript lint error, I will fix that.  Ah this time I have a build error.

Canceling CI automatically forces the second option.  And without the automatic cancel you can still take the second option if you want and the cancel will happen naturally since we have that cancel thing when you re-push.

> I can't reproduce your TS error locally

Weird...maybe my branch is just borked somehow.  I can try again.

> that's why I added the typecheck step

Ohhhhhhhhhh dude I had no idea that was even in there!  That explains it.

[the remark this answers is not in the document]
jsjoeio: > Canceling CI automatically forces the second option. And without the automatic cancel you can still take the second option if you want and the cancel will happen naturally since we have that cancel thing when you re-push.

Ah, that's a good point! Okay I'm on board. Let's remove it.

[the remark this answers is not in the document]
jsjoeio: Should be good now 👍🏼

.github/workflows/build.yaml
```diff
@@ -35,38 +35,95 @@ jobs:
...
+          node-version: "16"
+
       - name: Fetch dependencies from cache
+        if: steps.changed-files.outputs.any_changed == 'true'
         id: cache-node-modules
         uses: actions/cache@v3
         with:
           path: "**/node_modules"
-          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
+          key: yarn-build-${{ hashFiles('**yarn.lock') }}
```
code-asher, line 113: Was the / removal here intentional? (Same for the other keys.)

.github/workflows/build.yaml
```diff
@@ -417,8 +482,8 @@ jobs:
         if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: SKIP_SUBMODULE_DEPS=1 yarn install
 
-      - name: Run integration tests on standalone release
-        run: yarn test:integration
+      - name: Run native module tests on standalone release
+        run: yarn test:native
```
code-asher, line 486: I like the rename.

.github/workflows/build.yaml
jsjoeio: Bad search and replace 🤦🏼‍♂️

jsjoeio: Fixed

code-asher (commented):
Looks good to me, I am going to run it on my branch again to double check the changed files stuff then will approve.

code-asher:
The installation tests are failing but I think it is due to an older change, not this PR.  I will push up a PR after this merges to fix it.

I am getting lint and build errors but I am not sure why...maybe unrelated to this PR?  Investigating.

The Typescript lint check seems to run every time now?

code-asher:
False alarm on the lint step, I thought the changed files worked from commit to next commit but I think it compares to `main` each time.  So I think it is working.

I think we might need `fetch-depth: 2` though for when the branch gets merged otherwise the checks would not run on main (only on the PR) based on:

> IMPORTANT: For push events you need to include fetch-depth: 0 OR fetch-depth: 2 depending on your use case.

jsjoeio:
> So I think it is working.

Nice!

> I think we might need fetch-depth: 2

Ah interesting. If understanding right, you want me to add that to every step in `build.yaml` where we checkout code-server?

code-asher:
> If understanding right, you want me to add that to every step in build.yaml where we checkout code-server?

Only on the steps with the `changed-files` action (so `lint-ts` and `lint-helm`).

jsjoeio:
> Only on the steps with the changed-files action (so lint-ts and lint-helm).

Done!

code-asher (approved):
My build issues seem to have disappeared so maybe it was a cache issue?  If it crops up again we can take another look but I think we are good to go.  Nicely done!

PR #5578, opened 2022-09-22
fix: installer tests

code-asher:
Also rename from yml to yaml.


jsjoeio (approved):
Thanks for fixing this!

PR #5595, opened 2022-09-27
fix: fix can't run it as expected in freebsd

### LINKED ISSUE #5594: [Bug]: install.sh does not work as expected on freebsd

### OS/Web Information

- Web Browser: N/A
- Local OS: N/A
- Remote OS: freebsd
- Remote Architecture: x86_64
- `code-server --version`: Not yet installed

### Steps to Reproduce

1. Run install.sh as a normal user

### Expected

It should be installed successfully or something

### Actual

```
$ ./install.sh 
FreeBSD 13.1-RELEASE-p2
Installing latest from npm.

./install.sh: NPM_PATH: not found
Installing with npm.

+ su - -c 'npm install -g code-server --unsafe-perm'
Password:
su: only root may use -c
```

### Logs

_No response_

### Screenshot/Video

_No response_

### Notes

_No response_

unknowndevQwQ:
In freebsd, su -c expects a login class argument instead of a command, if -c is preceded by a username, then -c and the arguments that follow will be passed as shell arguments

Fixes #5594

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -556,7 +556,7 @@ sudo_sh_c() {
   elif command_exists sudo; then
     sh_c "sudo $*"
   elif command_exists su; then
-    sh_c "su - -c '$*'"
+    sh_c "su root -c '$*'"
   else
     echoh
     echoerr "This script needs to run the following command as root."


unknowndevQwQ:
The question I'm confused about is why the -l/- option is added to the su command, what is it to avoid?

code-asher (approved):
I am not sure if there was a good reason for the `-` but I think it is safe to remove.

Thank you for the fix!

jsjoeio:
You can ignore the security audit failure since it's unrelated to your PR.

PR #5600, opened 2022-09-29
feat: add doas support

### LINKED ISSUE #5599: [Feat]: add doas support
## What is your suggestion?
Suggestions? Sorry, I don't understand.
Some users may not use sudo but doas
Maybe there is, but it's ugly
~/.local/bin/sudo:
```
#!/bin/sh
doas "${@}"
```
y

unknowndevQwQ:
Some people may have some reason to drop sudo and switch to doas

Fixes #5599

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -553,6 +553,8 @@ sh_c() {
 sudo_sh_c() {
   if [ "$(id -u)" = 0 ]; then
     sh_c "$@"
+  elif command_exists doas; then
+    sh_c "doas $*"
   elif command_exists sudo; then
     sh_c "sudo $*"
   elif command_exists su; then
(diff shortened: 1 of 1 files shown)


jsjoeio:
I'm not familiar with `doas`. I did a quick search and it seems like it should be fine. 

Are there any potential side effects of this?

code-asher (commented):
Seems chill to me, could we add `doas` to the echo a few lines lower?  So we would have `Please install doas, sudo, or su`.

unknowndevQwQ:
> 



> Seems chill to me, could we add `doas` to the echo a few lines lower? So we would have `Please install doas, sudo, or su`.

sure

code-asher (approved):

PR #5602, opened 2022-09-29
refactor: run security on package.json changes

jsjoeio:


.github/workflows/security.yaml
```diff
@@ -4,8 +4,8 @@ on:
   push:
     branches: [main]
   pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [main]
+    paths:
+      - "package.json"
```
code-asher, line 8: Should we add to main as well?  Not a huge deal but it does kinda give a bad look to have seemingly failing CI on main.

jsjoeio: Oh yeah good call!

code-asher (approved):

PR #5605, opened 2022-09-29
refactor: update prettier and doctoc

jsjoeio:
This PR refactors our `prettier` and `doctoc` tasks.


.github/workflows/build.yaml
```diff
@@ -22,34 +22,47 @@ concurrency:
...
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
       - name: Checkout repo
         uses: actions/checkout@v3
 
+      - name: Run prettier with actionsx/prettier
+        uses: actionsx/prettier@v2
+        with:
+          args: --check --loglevel=warn .
```
jsjoeio, line 36: The downside to this action is we can't specify the version of Prettier. It's significantly faster than our previous Prettier step so I think that's worth switching. Happy to change back and use something like `npx` if desired.

.github/workflows/build.yaml
```diff
@@ -22,34 +22,47 @@ concurrency:
...
+
+  doctoc:
```
jsjoeio, line 38: Now that I'm adding this...do we really need to run `doctoc` in CI? Seems like something that would only be run locally, right? Should I remove?

.github/workflows/build.yaml
```diff
@@ -22,34 +22,47 @@ concurrency:
...
-          path: "**/node_modules"
-          key: yarn-build-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            yarn-build-
-
-      - name: Install dependencies
-        if: steps.cache-node-modules.outputs.cache-hit != 'true'
-        run: SKIP_SUBMODULE_DEPS=1 yarn --frozen-lockfile
+      - name: Install doctoc
+        run: yarn global add doctoc@2.0.0
```
jsjoeio, line 61: Didn't want to have to install all the modules to use `doctoc` so went with this approach.

docs/guide.md
```diff
@@ -89,11 +91,10 @@ we recommend using another method, such as [Let's Encrypt](#let-encrypt) instead
    using [mutagen](https://mutagen.io/documentation/introduction/installation)
    to do so. Once you've installed mutagen, you can port forward as follows:
 
-   ```console
+   ```shell
```
jsjoeio, line 94: Probably not necessary but changed since I've seen that more often

docs/guide.md
```diff
@@ -89,11 +91,10 @@ we recommend using another method, such as [Let's Encrypt](#let-encrypt) instead
...
    # This is the same as the above SSH command, but it runs in the background
    # continuously. Be sure to add `mutagen daemon start` to your ~/.bashrc to
    # start the mutagen daemon when you open a shell.
-
```
jsjoeio, line 96: Prettier didn't like this so I removed it

.github/workflows/build.yaml
code-asher: I think we want to because it catches when people have not ran it.

code-asher: Or at least that is how we should set it up if we have not already.

.github/workflows/build.yaml
code-asher: I think this is OK, probably there is not a huge amount of difference in Prettier versions and we do not change code-server code very often anyway.

Long term it does seem worth somehow making it use the same version since this has the potential to be frustrating since every time you push to CI it is more or less a gamble whether Prettier passes.

code-asher: Maybe what we should do is pin our package.json version to whatever version the action uses.  That way we have parity with CI and local.

code-asher: Dude you already did that, I had not gone through the PR yet lol.  Nice!  Works for me.

code-asher: I guess we could still go out of date if the action updates though.  Ah well.

.github/workflows/build.yaml
code-asher: Ah yup looks like we do so I think this is good.

.github/workflows/build.yaml
jsjoeio: I can pin the action too! If the action does update, Rennovate will tell us and then i can update prettier to match

ci/release-image/Dockerfile
```diff
@@ -29,15 +29,15 @@ RUN sed -i "s/# en_US.UTF-8/en_US.UTF-8/" /etc/locale.gen \
...
+  && echo "coder ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
 
-RUN ARCH="$(dpkg --print-architecture)" && \
-    curl -fsSL "https://github.com/boxboat/fixuid/releases/download/v0.5/fixuid-0.5-linux-$ARCH.tar.gz" | tar -C /usr/local/bin -xzf - && \
-    chown root:root /usr/local/bin/fixuid && \
-    chmod 4755 /usr/local/bin/fixuid && \
-    mkdir -p /etc/fixuid && \
-    printf "user: coder\ngroup: coder\n" > /etc/fixuid/config.yml
+RUN ARCH="$(dpkg --print-architecture)" \
+  && curl -fsSL "https://github.com/boxboat/fixuid/releases/download/v0.5/fixuid-0.5-linux-$ARCH.tar.gz" | tar -C /usr/local/bin -xzf - \
```
code-asher, line 36: Big fan of this.

code-asher (approved):

jsjoeio:
I want to make a couple minor changes so converting to draft

.github/workflows/build.yaml
code-asher: Similar to prettier we might one day want to make sure this uses the pinned version in the lockfile (or pin both) to avoid desync between CI and local.

.github/workflows/build.yaml
code-asher: Oh snap good point.

code-asher (approved):

PR #5614, opened 2022-10-02
Update the cloudflare tunnel description.

ELXIAGHT:
Fixes #

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -417,20 +417,20 @@ sudo passwd {user} # replace user with your code-server user
 
 [![Cloudflared](https://img.shields.io/badge/Cloudflared-E4863B?style=for-the-badge&logo=cloudflare&logoColor=ffffff)](https://github.com/cloudflare/cloudflared)
 
-1.  Install [cloudflared](https://github.com/cloudflare/cloudflared#installing-cloudflared) on your local computer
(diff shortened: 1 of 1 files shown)


docs/guide.md
```diff
@@ -417,20 +417,20 @@ sudo passwd {user} # replace user with your code-server user
...
 
-1.  Install [cloudflared](https://github.com/cloudflare/cloudflared#installing-cloudflared) on your local computer
-2.  Then go to `~/.ssh/config` and add the following:
+1.  Install [cloudflared](https://github.com/cloudflare/cloudflared#installing-cloudflared) on your local computer and remote server
+2.  Then go to `~/.ssh/config` and add the following on your local computer:
 
 ```shell
 Host *.trycloudflare.com
 HostName %h
-User root
```
jsjoeio, line 426: Hmm...could this have unintended consequences? @code-asher

ELXIAGHT: I actually prefer to use a non-root user login in the local configuration file for security purposes. Also the username entered in vscode should be the same as the user in the file.

code-asher: Good point, this should be the user that is running code-server and not root.  Looks good to me.  :+1:

jsjoeio (approved):

PR #5624, opened 2022-10-06
fix: add handle for resolveExternalUri

### LINKED ISSUE #1936: asExternalUri doesn't work
Hi Guys,

I had written a new extension which renders Graphs from matplotlib into vscode webview component, it works great on vscode desktop, but I encounter problems when trying to work with it in code-server.

I cannot state what exactly the problem is but here is the workflow:
1. my vscode extension starts and listens to requests for plot viewing on port 11111 (POST request)
2. matplotlib custom backend sends a request to the extension with the url to open in webview
3. I use vscode.env.asExternalUri tunneling in order to open the url in a webview and embed the generated address in an Iframe.
4. Usually by now the webview with the webpage would have been loaded, but instead I get and error openning localhost refuse to connect

Screen Shot 2020-08-02 at 11 50 37

Any ideas what went wrong?

- Web Browser: Chrome
- Local OS: Ubuntu
- Remote OS: MacOS
- Remote Architecture: x86_64
- `code-server --version`: v3.4.1

jsjoeio:
This PR fixes `asExternalUri` to use code-server's built-in proxy or `VSCODE_PROXY_URI` if it's been set. 

Todos
- [x] handle path-based proxies
- [x] add e2e test for it
- [x] add e2e test for `VSCODE_PROXY_URI`
- [x] support `VSCODE_PROXY_URI`

Fixes #1936


patches/proxy-uri.diff
```diff
@@ -98,3 +98,66 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
+ 	}
+-	const config: IWorkbenchConstructionOptions & { folderUri?: UriComponents; workspaceUri?: UriComponents; callbackRoute: string } = JSON.parse(configElementAttribute);
++	const config: IWorkbenchConstructionOptions & { folderUri?: UriComponents; workspaceUri?: UriComponents; callbackRoute: string } = { ...JSON.parse(configElementAttribute), remoteAuthority: location.host }
+ 
+ 	// Create workbench
+ 	create(document.body, {
+@@ -506,6 +508,32 @@ function doCreateUri(path: string, query
+ 		} : undefined,
+ 		workspaceProvider: WorkspaceProvider.create(config),
+ 		urlCallbackProvider: new LocalStorageURLCallbackProvider(config.callbackRoute),
```
jsjoeio, line 133: Everything above this comment was not made by me but by other patches. I could move my changes into a new patch if we think that's better. Open to either!

patches/proxy-uri.diff
```diff
@@ -98,3 +98,66 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
+-		credentialsProvider: config.remoteAuthority ? undefined : new LocalStorageCredentialsProvider() // with a remote, we don't use a local credentials provider
++		credentialsProvider: config.remoteAuthority ? undefined : new LocalStorageCredentialsProvider(), // with a remote, we don't use a local credentials provider
++		resolveExternalUri: (uri: URI): Promise<URI> => {
++			// TODO@jsjoeio - how do we make it relative to work when hosted on subpath?
++			const baseUrl = `${window.location.protocol}//${window.location.host}`
```
jsjoeio, line 138: Wondering if there's a better way to determine the `baseUrl`. `relativeRoot` is something we use but it's not available on the browser side AFAIK.

patches/proxy-uri.diff
```diff
@@ -98,3 +98,66 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
++			let resolvedUri = uri
++
++			// NOTE@jsjoeio - this isn't exhaustive
++			// also doesn't handle if not http or https i.e. ws
++			const localhostMatch = uri.toString().includes("localhost")
```
jsjoeio, line 143: I tried `extractLocalHostUriMetaDataForPortMapping` but was having trouble getting it to work for all `localhost` situations. Might be able to revisit.

jsjoeio:
> Hunk #2 FAILED at 486.
> Hunk #*** FAILED at 497.
> Hunk #4 succeeded at 508 (offset 1 line).
> 2 out of 4 hunks FAILED -- rejects in file lib/vscode/src/vs/code/browser/workbench/workbench.ts
> Patch patches/proxy-uri.diff does not apply (enforce with -f)

Hmm...wondering if I need to refresh other patches or something. Any ideas?

patches/proxy-uri.diff
code-asher: Was it for the no port scenarios?  If it returns undefined for those IMO we just return unchanged.

jsjoeio: Yes, exactly!

> If it returns undefined for those IMO we just return unchanged

That means `http://localhost` and `https://localhost` would not be transformed to use `/proxy/<80 | 443>`. Sounds like we're okay with that so I'll update 👍🏼

patches/proxy-uri.diff
code-asher: We need to resolve against the path in case code-server is hosted at a sub-path which means we have to get `proxyEndpointTemplate` in here somehow as it contains the information needed to resolve relatively.  Then we can do:

```
resolved = new URL(proxyEndpointTemplate.replace('{port}', port), window.location.href)
```

patches/proxy-uri.diff
code-asher: If we do need to modify it we could patch the original `extractLocalHostUriMetaDataForPortMapping`?

patches/proxy-uri.diff
```diff
@@ -98,3 +98,66 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
  		mergeEnvironments(env, allowedEnvFromConfig);
+Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/code/browser/workbench/workbench.ts
++++ code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
+@@ -17,6 +17,7 @@ import { isFolderToOpen, isWorkspaceToOp
+ import { create, ICredentialsProvider, IURLCallbackProvider, IWorkbenchConstructionOptions, IWorkspace, IWorkspaceProvider } from 'vs/workbench/workbench.web.main';
+ import { posix } from 'vs/base/common/path';
+ import { ltrim } from 'vs/base/common/strings';
++import { relativeRoot } from 'vs/server/node/webClientServer';
```
code-asher, line 109: This might have been left over from experimentation.  You are right it is not available on browser (you can only use things in `common` or `browser`).  It would not work anyway as it needs the server-side request.  This is actually why we have `proxyEndpointTemplate` to pass the relative root to the browser!

patches/proxy-uri.diff
code-asher: It looks like this patch now contains part of `base-path.diff`?  Not completely sure.

patches/proxy-uri.diff
code-asher: Yeah I think that is chill because it gives us parity with upstream!  :tada:

patches/proxy-uri.diff
jsjoeio: Super helpful! Thank you!

patches/proxy-uri.diff
```diff
@@ -98,3 +98,39 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
++		credentialsProvider: config.remoteAuthority ? undefined : new LocalStorageCredentialsProvider(), // with a remote, we don't use a local credentials provider
++		resolveExternalUri: (uri: URI): Promise<URI> => {
++			let resolvedUri = uri
++			const localhostMatch = extractLocalHostUriMetaDataForPortMapping(uri)
++
++			if (localhostMatch) {
++				let baseUrl = `${window.location.protocol}//${window.location.host}`
++				if (config.productConfiguration && config.productConfiguration.proxyEndpointTemplate) {
++					baseUrl = new URL(config.productConfiguration.proxyEndpointTemplate.replace('{{port}}', localhostMatch.port.toString()), window.location.href).toString()
++				}
```
code-asher, line 127: I think if the production configuration does not exist we should error because we have no way of knowing what the real base is.  In the enterprise product for example this will end up being the wrong URI.

jsjoeio:
Oops..looks like `proxyEndpointTemplate` adds `/proxy/port` for you.

image

patches/proxy-uri.diff
```diff
@@ -98,3 +98,39 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/terminal/common/terminalE
...
++				// Use code-server's built in /proxy/<port>
++				resolvedUri = URI.parse(`${baseUrl}/proxy/${localhostMatch.port}`)
```
code-asher, line 129: We can skip the `URI.parse` because `proxyEndpointTemplate` already includes `/proxy/{{port}}`.  For example it could be something like `../../proxy/{{port}}` then the `new URL` line will transform this into `https://mydomain.com/some/base/path/proxy/8080`.

code-asher (commented):
Awesome work!

patches/proxy-uri.diff
jsjoeio: Ah got it. okay I'll throw an error here.

patches/proxy-uri.diff
jsjoeio: Updated!

jsjoeio:
Looks good now! (ignore expected, hard-coded)

image

code-asher (commented):
Looks great to me!  Do you want an approval now and merge or are you going to add the domain-based proxy in this same PR so I should approve after that is done?

patches/proxy-uri.diff
code-asher: Just realized we do need the `URI.parse` just not the inner part but you already figured that out.  My bad!

jsjoeio:
> are you going to add the domain-based proxy in this same PR so I should approve after that is done?

I'll work on that here then mark PR ready when all is said and done. Thanks for the early review!

jsjoeio:
It appears the display language e2e test is failing. I'm going to see if I can reproduce locally.

ksylvan:
Would you like me to try testing this?

jsjoeio:
@ksylvan that would be super helpful 🙏🏼

patches/service-worker.diff
```diff
@@ -65,3 +50,18 @@ Index: code-server/lib/vscode/src/vs/workbench/browser/client.ts
...
+--- code-server.orig/lib/vscode/src/vs/server/node/webClientServer.ts
++++ code-server/lib/vscode/src/vs/server/node/webClientServer.ts
+@@ -319,6 +319,10 @@ export class WebClientServer {
+ 				updateEndpoint: !this._environmentService.args['disable-update-check'] ? base + '/update/check' : undefined,
+ 				logoutEndpoint: this._environmentService.args['auth'] && this._environmentService.args['auth'] !== "none" ? base + '/logout' : undefined,
+ 				proxyEndpointTemplate: process.env.VSCODE_PROXY_URI ?? base + '/proxy/{{port}}',
++				serviceWorker: {
++					scope: vscodeBase + '/',
++					path: base + '/_static/out/browser/serviceWorker.js',
++				},
```
jsjoeio, line 64: Looks like this patch update was just moving this down in the `.diff` file.

test/e2e/extensions/ms-ceintl.vscode-language-pack-es-1.70.0/package.json
```diff
@@ -1,32 +1,38 @@
...
+				]
+			}
+		]
+	},
+	"__metadata": {
+		"id": "47e020a1-33db-4cc0-a1b4-42f97781749a",
+		"publisherDisplayName": "MS-CEINTL",
+		"publisherId": "0b0882c3-aee3-4d7c-b5f9-872f9be0a115",
+		"isPreReleaseVersion": false
+	}
```
jsjoeio, line 37: I noticed VS Code added this after running the test so I decided to commit it.

test/e2e/models/CodeServer.ts
```diff
@@ -327,7 +327,7 @@ export class CodeServerPage {
     const selector = "text=test extension loaded"
     this.codeServer.logger.debug("Waiting for test extension to load...")
 
-    await this.page.waitForSelector(selector)
+    await this.page.waitForSelector(selector, { timeout: 5000 })
```
jsjoeio, line 330: gonna drop this change

test/playwright.config.ts
```diff
@@ -8,6 +8,7 @@ import path from "path"
 //   yarn test:e2e --workers 1        # Run with one worker
 //   yarn test:e2e --project Chromium # Only run on Chromium
 //   yarn test:e2e --grep login       # Run tests matching "login"
+//   PWDEBUG=1 yarn test:e2e          # Run Playwright inspector
```
jsjoeio, line 11: saves time from having to look this up each time 😂

jsjoeio:
@code-asher ready for another review!

test/e2e/extensions/ms-ceintl.vscode-language-pack-es-1.70.0/package.json
```diff
@@ -28,5 +28,11 @@
         ]
       }
     ]
+  },
+  "__metadata": {
+    "id": "47e020a1-33db-4cc0-a1b4-42f97781749a",
+    "publisherDisplayName": "MS-CEINTL",
+    "publisherId": "0b0882c3-aee3-4d7c-b5f9-872f9be0a115",
+    "isPreReleaseVersion": false
```
code-asher, line 36: Was this necessary to fix something?  Might be sketchy to include stuff that comes from the closed-source marketplace.

code-asher (approved):
Nicely done!  I only have a comment about the added metadata.

test/playwright.config.ts
code-asher: 100% lol same reason I added the other comments I am always forgetting the flags

test/e2e/extensions/ms-ceintl.vscode-language-pack-es-1.70.0/package.json
code-asher: Oh somehow missed your comment where you said VS Code added it.  In that case seems fine to me!

ericzhucode:
I found there's a conflict with url displayed in open external uri dialog.
If you open an external uri for the first time then the domain of it will be resolved to cache in map data structure.
Like https://www.w3schools.com/c/c_structs.php
image

For the next time if you open a different uri with the same domain
Like https://github.com/coder/code-server/issues
it will be resolved as the last uri which has been cached.
image

The uri displayed here is not correct
Hope there's a solutaion for this. Thank you!

PR #5640, opened 2022-10-12
Preinstall flyte and publish to Union ECR

andrewwdye:
This change
* Preinstalls pip, flytekit, flytectl, flytesnacks, and getting started code in a custom code server image
* Publishes the image to the Union public ECR


andrewwdye:
Mistakenly merged back to coder

PR #5645, opened 2022-10-13
style: remove a potentially redundant comma

unknowndevQwQ:
noting


jsjoeio:
Thank you @unknowndevQwQ! I think this is okay actually and we do not want to remove the comma.

PR #5673, opened 2022-10-19
jsjoeio/tunnel

jsjoeio:
- feat: enable ports panel in proxy-uri patch
- feat: add tunnelProvider in proxy-uri patch


patches/proxy-uri.diff
```diff
@@ -115,22 +122,22 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
...
+@@ -511,6 +512,39 @@ function doCreateUri(path: string, query
  		} : undefined,
  		workspaceProvider: WorkspaceProvider.create(config),
  		urlCallbackProvider: new LocalStorageURLCallbackProvider(config.callbackRoute),
 -		credentialsProvider: config.remoteAuthority ? undefined : new LocalStorageCredentialsProvider() // with a remote, we don't use a local credentials provider
 +		credentialsProvider: config.remoteAuthority ? undefined : new LocalStorageCredentialsProvider(), // with a remote, we don't use a local credentials provider
 +		resolveExternalUri: (uri: URI): Promise<URI> => {
 +			let resolvedUri = uri
-+			const localhostMatch = extractLocalHostUriMetaDataForPortMapping(uri)
++			const localhostMatch = extractLocalHostUriMetaDataForPortMapping(resolvedUri)
```
jsjoeio, line 137: cleanup: should have been using `resolvedUri` here

patches/proxy-uri.diff
```diff
@@ -115,22 +122,22 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
...
 +
 +			if (localhostMatch) {
-+				if (config.productConfiguration && config.productConfiguration.proxyEndpointTemplate) {
++				if ((resolvedUri.authority !== location.host) && config.productConfiguration && config.productConfiguration.proxyEndpointTemplate) {
```
jsjoeio, line 140: H/T to @code-asher the `resolvedUri.authority !== location.host` part 🎉 Clicking the globe icon sends the proxied URI here so we skip modifying it if authority and host match.

patches/proxy-uri.diff
```diff
@@ -139,6 +146,37 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
...
+Index: code-server/lib/vscode/src/vs/workbench/contrib/remote/browser/remoteExplorer.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/workbench/contrib/remote/browser/remoteExplorer.ts
++++ code-server/lib/vscode/src/vs/workbench/contrib/remote/browser/remoteExplorer.ts
+@@ -73,7 +73,7 @@ export class ForwardedPortsView extends 
+ 			this.contextKeyListener = undefined;
+ 		}
+ 
+-		const viewEnabled: boolean = !!forwardedPortsViewEnabled.getValue(this.contextKeyService);
++		const viewEnabled: boolean = true;
```
jsjoeio, line 179: This makes the Ports panel always visible which seems like what we want.

patches/proxy-uri.diff
code-asher, line 140: This will cause it to throw an error saying it is unable to resolve due to the product configuration being missing right?  We might need to move it up a bit.

```
  if (localhostMatch && resolvedUri.authority !== location.host) {
				if (config.productConfiguration && config.productConfiguration.proxyEndpointTemplate) {
```

patches/proxy-uri.diff
```diff
@@ -139,6 +146,37 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
...
 +			// If not localhost, return unmodified
 +			return Promise.resolve(resolvedUri)
++		},
++		tunnelProvider: {
++			tunnelFactory: (tunnelOptions: TunnelOptions, tunnelCreationOptions: TunnelCreationOptions) => {
++				const onDidDispose: Emitter<void> = new Emitter();
++				let isDisposed = false;
++				return Promise.resolve({
++					remoteAddress: tunnelOptions.remoteAddress,
++					//The complete local address(ex. localhost:1234)
```
code-asher, line 156: nit: we might not need this comment, the code I think makes it obvious it is the local address and the format it takes and I think "complete" in this context is ambiguous anyway (I would expect that to mean absolute and include the protocol).  I think we just copied this from elsewhere anyway.

code-asher (commented):
Very exciting :tada: 

For some reason I am not seeing the ports panel (tried `npm` and the CI artifact).  :thinking: 

```
$ code-server --auth none
[2022-10-21T15:48:23.934Z] info  code-server 4.7.1-5673-3c17798eb377432f160f3aef7cad61bb2a26a83f 3c17798eb377432f160f3aef7cad61bb2a26a83f
```

patches/proxy-uri.diff
jsjoeio: Good call. I'll move it up.

patches/proxy-uri.diff
jsjoeio: Yes! It was copied. I'll remove it.

jsjoeio:
> For some reason I am not seeing the ports panel (tried npm and the CI artifact). 🤔

Hmm...do you have it hidden by chance? @bpmct and I did test it in a v2 workspace (Linux) and we saw it. I'm not sure what else might cause it to be missing 🤔

bpmct:
> > For some reason I am not seeing the ports panel (tried npm and the CI artifact). 🤔
> 
> Hmm...do you have it hidden by chance? @bpmct and I did test it in a v2 workspace (Linux) and we saw it. I'm not sure what else might cause it to be missing 🤔

here are the steps we did: https://github.com/coder/coder/compare/main...dogfood-code-server-ports

code-asher (approved):
Apparently I was having a caching issue.  Works for me now!

code-asher:
We might want to edit `webClientServer.ts` and add a trailing slash since most of the time that will be needed:
```
				proxyEndpointTemplate: process.env.VSCODE_PROXY_URI ?? base + '/proxy/{{port}}/',
```

jsjoeio:
Good point! I'll do that and then merge

codecae:
FWIW:

I have been experiencing potential issues with this PR on systems in which the code-server is running behind another proxy.  Websocket connections are constantly being disconnected and reconnected at a high rate making the UI inaccessible and rendering the application unusable.

I will open another issue for this, but wanted to share in this thread as well.

jsjoeio:
> Having an optional kill switch for this feature might be beneficial for folks like myself who would prefer to opt out so we can stay current with code-server, otherwise!

I didn't consider that when I merged this - great point! I'll keep an eye out for that issue.

PR #5730, opened 2022-10-28
fix: merge in master from coder/code-server-aur

jsjoeio:
This ensures that the branch is up-to-date with coder/code-server-aur
before opening a PR.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -97,6 +97,12 @@ jobs:
           token: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
           ref: "master"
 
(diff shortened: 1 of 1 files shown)


code-asher (approved):

code-asher:
Er wait when it uses the checkout action should it not already be the right branch?

code-asher:
https://github.com/coder/code-server/blob/main/.github/workflows/publish.yaml#L93-L98

jsjoeio:
> Er wait when it uses the checkout action should it not already be latest master?

That's what I thought too! 🤔 Oh but it's because it's `cdrci`'s fork which never get's updated. That's why we have to do this

PR #5740, opened 2022-10-31
feat: add test for markdown webview

jsjoeio:
This adds a test to ensure webviews work by rendering the Markdown Preview.

### DIFF
--- a/test/e2e/webview.test.ts
+++ b/test/e2e/webview.test.ts
@@ -1,0 +1,30 @@
+import { promises as fs } from "fs"
+import * as path from "path"
+import { describe, test, expect } from "./baseFixture"
+
+describe("Webviews", ["--disable-workspace-trust"], {}, () => {
+  test("should preview a Markdown file", async ({ codeServerPage }) => {
+    // Create Markdown file
+    const heading = "Hello world"
+    const dir = await codeServerPage.workspaceDir
+    const file = path.join(dir, "text.md")
+    await fs.writeFile(file, `# ${heading}`)
+    await codeServerPage.openFile(file)
+
+    // Open Preview
+    await codeServerPage.executeCommandViaMenus("Markdown: Open Preview to the Side")
+    // Wait for the iframe to open and load
+    await codeServerPage.page.waitForTimeout(2000)
+    await codeServerPage.waitForTab(`Preview ${file}`)
+
+    let totalCount = 0
+    for (const frame of codeServerPage.page.frames()) {
+      // Check for heading in frames
+      const count = await frame.locator(`text=${heading}`).count()
+      totalCount += count
+    }
+
+    // One in the file and one in the preview
+    expect(totalCount).toBe(2)
+  })
+})
--- a/test/e2e/webview.test.ts
+++ b/test/e2e/webview.test.ts
@@ -14,17 +14,15 @@ describe("Webviews", ["--disable-workspace-trust"], {}, () => {
     // Open Preview
     await codeServerPage.executeCommandViaMenus("Markdown: Open Preview to the Side")
     // Wait for the iframe to open and load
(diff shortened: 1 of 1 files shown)


test/e2e/webview.test.ts
```diff
@@ -0,0 +1,30 @@
...
+    const dir = await codeServerPage.workspaceDir
+    const file = path.join(dir, "text.md")
+    await fs.writeFile(file, `# ${heading}`)
+    await codeServerPage.openFile(file)
+
+    // Open Preview
+    await codeServerPage.executeCommandViaMenus("Markdown: Open Preview to the Side")
+    // Wait for the iframe to open and load
+    await codeServerPage.page.waitForTimeout(2000)
+    await codeServerPage.waitForTab(`Preview ${file}`)
```
jsjoeio, line 18: The iframe takes a second to load which is why I have this timeout here. Not the greatest but unsure if there's a better option.

test/e2e/webview.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    let totalCount = 0
+    for (const frame of codeServerPage.page.frames()) {
+      // Check for heading in frames
+      const count = await frame.locator(`text=${heading}`).count()
+      totalCount += count
+    }
+
+    // One in the file and one in the preview
+    expect(totalCount).toBe(2)
```
jsjoeio, line 28: I was having trouble figuring out how to look for elements within the `iframe`. In the doc examples, they expect the iframe to have an `id` which you can use to locate the frame.

One thing surprising to me is when I inspect the DOM inside code-server, I see the `src` for the `iframe` is the `pre/index.html` but when i log the `frame.url` in this for loop, I see two frames and one is the `webWorkerExtensionHostIframe.html` so a little confused. I guess `url` and `src` are not the same?

Anyways, this test locates the heading text and counts the selectors then adds it to the total. Total should be two since in file and one in the preview. I don't love this approach so I'm open to other ideas

test/e2e/webview.test.ts
code-asher: If we use the frame locator maybe Playwright will wait for it to appear like with other selectors?

```
page.frameLocator('#my-frame').locator(...);
```

https://playwright.dev/docs/api/class-framelocator

test/e2e/webview.test.ts
code-asher: I bet there are at least some class names we could use.

test/e2e/webview.test.ts
jsjoeio: I didn't try it using `"iframe"` as the selector. Good idea!

test/e2e/webview.test.ts
jsjoeio: Ahh that makes sense. Okay, I'll try the `frameLocator` and see if that works. Thanks man!

test/e2e/webview.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    await codeServerPage.executeCommandViaMenus("Markdown: Open Preview to the Side")
+    // Wait for the iframe to open and load
+    await codeServerPage.waitForTab(`Preview ${file}`)
+
+    // It's an iframe within an iframe
+    // so we have to do .frameLocator twice
+    const renderedText = await codeServerPage.page
+      .frameLocator("iframe.webview.ready")
+      .frameLocator("#active-frame")
+      .locator("text=Hello world")
```
jsjoeio, line 24: This was the hardest part. I didn't realize it was an `<iframe>` inside of an `<iframe>` but some rubber-ducking with @code-asher helped me figure it out.

code-asher (approved):
:tada: :tada: :tada:

PR #5752, opened 2022-11-03
Improve getting started Coder CTA

kylecarbs:
This wasn't very standout-ish before, and I think it's wise for us to experiment with directing users to Coder.

image


patches/getting-started.diff
```diff
@@ -19,32 +19,89 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/welcomeGettingStarted/bro
...
+-				reset(leftColumn, startList.getDomElement(), recentList.getDomElement());
+-				reset(rightColumn, gettingStartedList.getDomElement());
++				reset(leftColumn, startList.getDomElement(), recentList.getDomElement(), gettingStartedList.getDomElement());
++				reset(rightColumn, gettingStartedCoder);
+ 				recentList.setLimit(5);
+ 			}
+ 			else {
+ 				this.container.classList.add('noWalkthroughs');
+-				reset(leftColumn, startList.getDomElement());
++				reset(leftColumn, startList.getDomElement(), gettingStartedCoder);
```
jsjoeio, line 101: Should we gate these column changes behind
```typescript
if (this.contextService.contextMatchesRules(IsEnabledCoderGettingStarted))
```
so they don't affect users who disable this?

kylecarbs: Good point!

kylecarbs: Fixed!

jsjoeio (approved):
Looks way better!

PR #5753, opened 2022-11-04
refactor: get version dynamically

jsjoeio:
# Description 

This PR refactors the release process to dynamically get the release version from the workflow inputs. This means we don't have to open a PR into `main` that modifies the `package.json`. It also means we'll need to manually modify the `CHANGELOG` and the Helm chart after doing a release.

## Todos

- [x] remove global `VERSION`
- [x] set anywhere we need it
- [x] standardize input to always include v and then strip when not needed
- [x] update release docs

## Screenshot

image

## Testing Plan

- [x] test build flow  - published successfully to npm [here](https://www.npmjs.com/package/@coder/code-server-pr/v/0.0.0-5753-9c704e6bc864b0e7511226fab77a5424ecf31b4d), successful build CI here (todo link)
- [x] test release flow: run [draft release manually](https://github.com/coder/code-server/actions/runs/3397555429/jobs/5649824113), and double-check assets
- [x] test publish flow: copy changes to fork and test `publish` [there](https://github.com/jsjoeio/code-server/actions/runs/3397622429/jobs/5649948994). uses the correct input


jsjoeio:
First test: draft workflow succeeds. And when you download an asset, it shows the `package.json` has been updated and the `product.json`:

image

image

However, the assets have the wrong version in their name 🤔

image

I'm not sure what the root cause of this problem is though.

jsjoeio:
Okay figured out the issue. we get `VERSION` from the root `package.json`. I need to refactor to be a function `version()` which takes an optional json path to get the version from.

.github/workflows/publish.yaml
```diff
@@ -24,21 +29,23 @@ jobs:
...
       - name: Download npm package from release artifacts
         uses: robinraju/release-downloader@v1.5
         with:
           repository: "coder/code-server"
-          tag: v${{ steps.version.outputs.version }}
+          tag: ${{ github.event.inputs.version || github.ref_name }}
           fileName: "package.tar.gz"
           out-file-path: "release-npm-package"
 
+      - name: Get and set VERSION
```
code-asher, line 40: For a second I forgot why we needed to do this.  :laughing:   We might want to leave a comment that this exists just so we can strip out the `v` (same elsewhere this is used).

.github/workflows/release.yaml
```diff
@@ -265,8 +296,24 @@ jobs:
...
+      - name: Modify version
+        run: |
+          echo "Updating version in root package.json"
+          npm version --prefix release ${{ inputs.version || github.ref_name }}
+
+          echo "Updating version in lib/vscode/product.json"
+          tmp=$(mktemp) 
+          jq '.codeServerVersion = "${{ inputs.version || github.ref_name }}"' release/lib/vscode/product.json > "$tmp" && mv "$tmp" release/lib/vscode/product.json
+
+      # TODO@jsjoeio - will it throw error if it already exists?
```
code-asher, line 311: I tested and it seems it will override so we are good here :+1:

ci/build/build-vscode.sh
```diff
@@ -47,6 +47,7 @@ main() {
   # as it is a submodule.
   export VSCODE_DISTRO_COMMIT
   VSCODE_DISTRO_COMMIT=$(git rev-parse HEAD)
+  VERSION=$(jq -r .version ../../package.json)
```
code-asher, line 50: Will this not be set by the workflow?

.github/workflows/release.yaml
```diff
@@ -265,8 +296,24 @@ jobs:
...
+        run: tar -xzf package.tar.gz
+
+      - name: Modify version
+        run: |
+          echo "Updating version in root package.json"
+          npm version --prefix release ${{ inputs.version || github.ref_name }}
+
+          echo "Updating version in lib/vscode/product.json"
+          tmp=$(mktemp) 
+          jq '.codeServerVersion = "${{ inputs.version || github.ref_name }}"' release/lib/vscode/product.json > "$tmp" && mv "$tmp" release/lib/vscode/product.json
```
code-asher, line 309: I think we need to strip the `v` on these.

.github/workflows/publish.yaml
```diff
@@ -147,19 +161,21 @@ jobs:
...
           repository: "coder/code-server"
-          tag: v${{ steps.version.outputs.version }}
+          tag: v${{ env.VERSION }}
           fileName: "*.deb"
           out-file-path: "release-packages"
 
       - name: Publish to Docker
         run: yarn publish:docker
         env:
+          VERSION: ${{ steps.version.outputs.version }}
```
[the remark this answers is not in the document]
jsjoeio: Oh yes! Nice catch.

.github/workflows/publish.yaml
jsjoeio: Gooooood idea!

ci/build/build-vscode.sh
jsjoeio: AFAICT, we only build Code as part of the "Build code-server" job, which only runs in `build.yaml` which runs automatically so we have to grab it from the `package.json` here.

Is that right? Is there a better way to do this?

.github/workflows/release.yaml
jsjoeio: Nice catch!

ci/build/build-vscode.sh
code-asher: Ahhh I saw `VERSION: "0.0.0"` and assumed that was for the build job but it was for the npm publish job.  I would be down to set `VERSION: "0.0.0"` for the build job as well.  That way the version is always set the same way (always by env var from a workflow).

code-asher:
Should we also change the `package.json` version to be `development` or `0.0.0` or something like that?

ci/build/build-vscode.sh
jsjoeio: Good idea. Okay I'll do `VERSION: "0.0.0"`

jsjoeio:
> Should we also change the package.json version to be development or 0.0.0 or something like that?

Yeah, good call. I'll update it to `0.0.0`

jsjoeio:
Ready for another review 👀

jsjoeio:
e2e tests failing because it's using a cached version of Code. bumped cache key in secrets.

ci/build/build-vscode.sh
code-asher, line 50: We still need to delete this line otherwise the version we set in the workflow gets overridden!

jsjoeio, line 50:
```suggestion
```

.github/workflows/publish.yaml
```diff
@@ -73,6 +91,8 @@ jobs:
     timeout-minutes: 10
     env:
       GH_TOKEN: ${{ secrets.HOMEBREW_GITHUB_API_TOKEN }}
+      VERSION: ${{ github.event.inputs.version || github.ref_name }}
```
code-asher, line 94: Do we need this line since we set version in another step?

code-asher (approved):

.github/workflows/publish.yaml
jsjoeio: I think so? Don't we need to set it in every step?

jsjoeio: Hmm double-checking..i see what you mean now

jsjoeio: Ah, you were right - nice catch. Removed!

PR #5772, opened 2022-11-10
release: 4.9.0

jsjoeio:
Published as a release candidate which can be installed via:

```shell
curl -fsSL https://code-server.dev/install.sh | sh -s -- --edge --standalone
```

### DIFF
--- a/test/unit/node/wrapper.test.ts
+++ b/test/unit/node/wrapper.test.ts
@@ -12,7 +12,7 @@ describe("wrapper", () => {
     const p = new ChildProcess()
     // our ChildProcess isn't exported
     // and shouldn't be for a test so surpressing TS error.
-    // @ts-expect-error
+    // @ts-expect-error - see above
     expect(isChild(p)).toBe(false)
   })
 })
--- a/test/unit/node/wrapper.test.ts
+++ b/test/unit/node/wrapper.test.ts
@@ -1,5 +1,4 @@
-import { ChildProcess } from "child_process"
-import { ParentProcess, isChild } from "../../../src/node/wrapper"
+import { ChildProcess, ParentProcess, isChild } from "../../../src/node/wrapper"
 
 describe("wrapper", () => {
   describe("isChild", () => {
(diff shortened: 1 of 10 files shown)


jsjoeio:
Terminal is not working as shared first here: https://github.com/coder/code-server/pull/5769#issuecomment-1311343428

Need to fix this before releasing.

jsjoeio:
Oddly enough...when I run `yarn watch` locally, the terminal is working 🤔

image

I'm going to test installing this via npm locally next.

jsjoeio:
Works on npm too 🤔 

image

Maybe it's only affecting the non-npm builds? I'll try creating a standalone release locally.

jsjoeio:
How I'm testing standalone release:
1. download npm artifact into `code-server` repo locally (on `main` branch) from [here](https://github.com/coder/code-server/actions/runs/3586520335)
2. unzip
3. run `yarn release:standalone`
4. run `node ./release-standalone --port 3000`

Still same result  🤔
image

jsjoeio:
Tested with `yarn package` like so:
1. `VERSION="4.9.0-joe" yarn package`
2. `tar -xzf ./release-packages/code-server-4.9.0-joe-macos-amd64.tar.gz`
3. `node ./code-server-4.9.0-joe-macos-amd64 --port 3000`

And still no issues 🤔
image

Only other thing that comes to mind is creating a draft release which will build artifacts in CI, then I can download them and test those (in case something in CI is being funky).

jsjoeio:
Testing here: https://github.com/coder/code-server/actions/runs/3587010753

jsjoeio:
Hmm...I couldn't run code-server from in a Coder OSS workspace 🤔 I'm not sure why: 
image

I think the next best option is to publish a release candidate 2 and have others test.

jsjoeio:
I'm not sure why unit tests are suddenly failing 🤔 They're passing locally for me. Wonder if it's a caching issue with node_modules or something.

jsjoeio:
see https://github.com/adobe/fetch/pull/318#issuecomment-1306070259

jsjoeio:
Release candidates have been tested and I've tested locally so I'm confident we can release this now. cc @code-asher

test/unit/node/wrapper.test.ts
```diff
@@ -0,0 +1,18 @@
+import { ChildProcess } from "child_process"
```
jsjoeio, line 1: Unrelated to release. Added to keep code coverage from decreasing and CI ✅

PR #5860, opened 2022-12-09
chore(nfpm): bump to latest version

ghuntley:
Double check I'm seeing this right here but the latest is lower in number...

Latest - https://github.com/goreleaser/nfpm/releases/tag/v2.22.2
The version we use - https://github.com/goreleaser/nfpm/releases/tag/v2.3.1


jsjoeio (approved):
Nice catch!

PR #5911, opened 2022-12-20
refactor: move vscode tests to e2e

jsjoeio:
Upstream moved away from `require` in some parts but did some other funky stuff in other parts which is now causing issues with Jest. So we decided to move these tests to e2e instead.


.github/workflows/build.yaml
```diff
@@ -139,6 +139,49 @@ jobs:
         if: steps.changed-files.outputs.any_changed == 'true'
         run: yarn lint:ts
 
+  test-unit:
```
jsjoeio, line 142: Since we no longer need a build to run unit tests, it can now run in parallel with our other pre-build jobs 🎉

test/e2e/models/CodeServer.ts
```diff
@@ -119,7 +119,7 @@ export class CodeServer {
   private async spawn(): Promise<CodeServerProcess> {
     // This will be used both as the workspace and data directory to ensure
     // instances don't bleed into each other.
-    const dir = await this.createWorkspace()
+    let dir = await this.createWorkspace()
```
jsjoeio, line 122: we can undo this 
```suggestion
    const dir = await this.createWorkspace()
```

test/e2e/models/CodeServer.ts
```diff
@@ -128,6 +128,8 @@ export class CodeServer {
         path.join(dir, "extensions"),
         "--auth",
         "none",
+        // The last argument is the workspace to open.
```
jsjoeio, line 131:
```suggestion
        // The workspace to open.
```

test/e2e/models/CodeServer.ts
```diff
@@ -128,6 +128,8 @@ export class CodeServer {
...
+        ...(this.args.includes("--ignore-last-opened") ? [] : [dir]),
```
jsjoeio, line 132: This change allows us to _not_ include the test workspace dir when we pass the `--ignore-last-opened` flag.

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  })
+
+  test("should redirect to the passed in folder using human-readable query", async ({ codeServerPage }) => {
+    const url = new URL(codeServerPage.page.url())
+    expect(url.pathname).toBe("/")
+    expect(url.search).toBe(`?folder=${CODE_FOLDER_DIR}`)
+  })
+})
+
+describe(
```
jsjoeio, line 74: I broke some out into their own describe blocks with the hopes that it would be faster than one giant describe block.

test/unit/node/routes/vscode.test.ts
```diff
@@ -1,137 +0,0 @@
...
-  it("should do nothing when nothing is passed in", async () => {
-    codeServer = await integration.setup(["--auth=none"], "")
-
-    const resp = await codeServer.fetch("/", undefined)
-
-    expect(resp.status).toBe(200)
-    const url = new URL(resp.url)
-    expect(url.search).toBe("")
-    await resp.text()
-  })
```
jsjoeio, line 114: We lost this test because even when you don't pass anything in, it still takes you to the test workspace dir _unless_ you pass in `--ignore-last-opened` (which I added a test for) so I think it's okay.

.github/workflows/build.yaml
code-asher: :tada: :tada: :tada:

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      const html = await codeServerPage.page.innerHTML("html")
+      switch (route) {
+        case "/":
+        case "/vscode/":
+          expect(html).toMatch(/src="\.\/[a-z]+-[0-9a-z]+\/static\//)
+          break
+        case "/vscode":
+          expect(html).toMatch(/src="\.\/vscode\/[a-z]+-[0-9a-z]+\/static\//)
+          break
+      }
```
code-asher, line 32: Just a thought (not something we need to do in this PR or even ever since this works just fine): now that we are in a proper browser instead of scraping the HTML we could possibly intercept requests to make sure assets are loading from the right spot.

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    expect(url.search).toBe(`?workspace=${workspace}`)
+  })
+})
+
+const CODE_WORKSPACE_DIR = process.env.CODE_WORKSPACE_DIR || ""
+describe("VS Code Routes with code-workspace", ["--disable-workspace-trust", CODE_WORKSPACE_DIR], {}, async () => {
+  const testName = "vscode-routes"
+  test.beforeAll(async () => {
+    await clean(testName)
+  })
```
code-asher, line 51: `tmpdir` is not used in this `describe` block so we could omit the clean since there is nothing to clean.

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    expect(url.search).toBe(`?workspace=${CODE_WORKSPACE_DIR}`)
+  })
+})
+
+const CODE_FOLDER_DIR = process.env.CODE_FOLDER_DIR || ""
+describe("VS Code Routes with code-workspace", ["--disable-workspace-trust", CODE_FOLDER_DIR], {}, async () => {
+  const testName = "vscode-routes"
+  test.beforeAll(async () => {
+    await clean(testName)
+  })
```
code-asher, line 65: Could omit this as well.

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  test("should redirect to the passed in workspace using human-readable query", async ({ codeServerPage }) => {
+    const workspace = path.join(await tmpdir(testName), "test.code-workspace")
+    await fs.writeFile(workspace, "")
+
+    const url = new URL(codeServerPage.page.url())
+    expect(url.pathname).toBe("/")
+    expect(url.search).toBe(`?workspace=${workspace}`)
+  })
```
code-asher, line 43: This says it tests a passed-in workspace but it does not seem to pass it into the code-server invocation?  I have no idea how it passes. :stuck_out_tongue:

test/e2e/routes.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  ["--disable-workspace-trust", "--ignore-last-opened"],
+  {},
+  async () => {
+    test("should not redirect", async ({ codeServerPage }) => {
+      await codeServerPage.navigate(`/`)
+      // No redirections.
+      const url = new URL(codeServerPage.page.url())
+      expect(url.pathname).toBe("/")
+      expect(url.search).toBe("")
+    })
```
code-asher, line 85: To fully test that the last opened is being ignored we should open something then navigate again and make sure it is not opening again (just like the test below).

test/e2e/routes.test.ts
jsjoeio: Good call! I'll add a comment to the code for now.

test/e2e/routes.test.ts
jsjoeio: Lol I think I accidentally wrote that test twice. Deleting one of them now.

test/e2e/routes.test.ts
jsjoeio: Good point!

code-asher (approved):

PR #5915, opened 2022-12-22
chore: update aur job in publish.yaml

jsjoeio:
This uses the `env.VERSION` which should fix the issue with version being blank.

Fixes N/A

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -137,12 +137,14 @@ jobs:
       - name: Open PR
         # We need to git push -u otherwise gh will prompt
         # asking where to push the branch.
+        env:
+          VERSION: ${{ env.VERSION }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -137,12 +137,14 @@ jobs:
       - name: Open PR
         # We need to git push -u otherwise gh will prompt
         # asking where to push the branch.
+        env:
+          VERSION: ${{ env.VERSION }}
```
code-asher, line 141: If `VERSION` is already in `env` do we need to set it on `env` here?

code-asher (approved):

.github/workflows/publish.yaml
jsjoeio: It's a pattern I've been following because I thought I had to but to be honest, it may be redundant 🤔 I feel like I had to do this to make sure it's there but haven't tested in a while!

PR #5936, opened 2023-01-03
chore: update eslint-plugin-import to latest

jsjoeio:
Fixes https://github.com/coder/code-server/actions/runs/3823551292/jobs/6504840578


code-asher:
I think we can update this without a resolution which might be better because if modules update their version of json5 to something greater we will force it back down again with the resolution (this probably applies to most of our resolutions actually now that I think about it).

jsjoeio:
Ah okay, I'll do that instead in this PR.

code-asher (approved):
Awesome!

PR #5947, opened 2023-01-05
feat: add i18n in login page

### LINKED ISSUE #5919: [Feat]: Add i18n to the login page
## What is your suggestion?
Add i18n to the login page

I want to display the login page in a different language.

No. Use the `welcome-text` parameter only to translate partial content to what language I want on the login page.

Yes. Something like this:
image

zhaozhiming:
Fixes #5919


tsconfig.json
```diff
@@ -22,7 +22,8 @@
       "./test/node_modules/@types",
       "./lib/vscode/src/vs/server/@types"
     ],
-    "downlevelIteration": true
+    "downlevelIteration": true,
+    "resolveJsonModule": true
```
jsjoeio, line 26: TIL! 

https://www.typescriptlang.org/tsconfig#resolveJsonModule

test/unit/node/routes/login.test.ts
```diff
@@ -138,5 +138,16 @@ describe("login", () => {
       expect(resp.status).toBe(200)
       expect(htmlContent).toContain(`Welcome to ${appName}`)
     })
+
+    it("should return correct welcome text when lng is set to non-English", async () => {
```
jsjoeio, line 142: Thank you for adding a test 👏🏼

src/browser/pages/login.html
```diff
@@ -25,7 +25,7 @@
       <div class="card-box">
         <div class="header">
           <h1 class="main">{{WELCOME_TEXT}}</h1>
-          <div class="sub">Please log in below. {{PASSWORD_MSG}}</div>
```
jsjoeio, line 28: Part of me thinks we should keep the " " here between `{{I18N_LOGIN_BELOW}}{{PASSWORD_MSG}}` since I could see someone forgetting to add it to the `locales` file.

src/browser/pages/login.html
```diff
@@ -10,7 +10,7 @@
       http-equiv="Content-Security-Policy"
       content="style-src 'self'; script-src 'self' 'unsafe-inline'; manifest-src 'self'; img-src 'self' data:; font-src 'self' data:;"
     />
-    <title>{{APP_NAME}} login</title>
+    <title>{{APP_NAME}} {{I18N_LOGIN_TITLE}}</title>
```
jsjoeio, line 13: Re: keeping space...we do it here, so I think we should keep this pattern.

src/node/cli.ts
```diff
@@ -88,6 +88,7 @@ export interface UserProvidedArgs extends UserProvidedCodeArgs {
   verbose?: boolean
   "app-name"?: string
   "welcome-text"?: string
+  "lng"?: string
```
jsjoeio, line 91: nit: I'd vote for `lang` over `lng`. I think that's more common? Thoughts?

cc @code-asher

src/node/cli.ts
```diff
@@ -264,6 +265,13 @@ export const options: Options<Required<UserProvidedArgs>> = {
     `,
     deprecated: true,
   },
+  "lng": {
+    type: "string",
+    description: `
+      Language show on login page, more infomations to read up on
+      https://en.wikipedia.org/wiki/IETF_language_tag.
```
jsjoeio, line 272:
```suggestion
      Language to show on the login page, more info see 
      https://en.wikipedia.org/wiki/IETF_language_tag
```

src/node/routes/login.ts
```diff
@@ -28,13 +29,15 @@ export class RateLimiter {
 
 const getRoot = async (req: Request, error?: Error): Promise<string> => {
   const content = await fs.readFile(path.join(rootPath, "src/browser/pages/login.html"), "utf8")
+  const lng = req.args["lng"] || "en"
+  i18n.changeLanguage(lng)
   const appName = req.args["app-name"] || "code-server"
-  const welcomeText = req.args["welcome-text"] || `Welcome to ${appName}`
-  let passwordMsg = `Check the config file at ${humanPath(os.homedir(), req.args.config)} for the password.`
+  const welcomeText = req.args["welcome-text"] || (i18n.t("WELCOME", { app: appName }) as string)
```
jsjoeio, line 35: Is this `as string` necessary? I would think `t` returns a `string`

jsjoeio (commented):
Great work so far! Mostly a couple minor things and questions.

src/node/cli.ts
code-asher: I would actually prefer we use the existing `--locale` flag since that is what Code uses.

src/browser/pages/login.html
code-asher: Also should the app name be part of the login title?  Not sure if in some languages the app name might come after the word for `login`.

src/node/i18n/locales/zh-cn.json
```diff
@@ -0,0 +1,13 @@
+{
```
code-asher, line 1: Would you be willing to add yourself as a codeowner for this file in case we add new strings in the future?

src/node/cli.ts
jsjoeio: Good point!

src/browser/pages/login.html
zhaozhiming: ok. I will put the app name into the whole title.

src/node/i18n/locales/zh-cn.json
zhaozhiming: Yes. It's my pleasure.

src/node/cli.ts
zhaozhiming: I’m a little worried about whether it affects other places when using the existing `--locale` flag?

tsconfig.json
zhaozhiming: Yes, this is for importing the locale JSON files in typescript.

src/node/cli.ts
jsjoeio: Hmm...the only place I know is the display language but there may be others.

src/node/routes/login.ts
jsjoeio: We can always look in a followup PR. I don't want to hold you back so it's fine as is for now. Thank you for trying that!

.github/CODEOWNERS
```diff
@@ -3,3 +3,5 @@
 ci/helm-chart/ @Matthew-Beckett @alexgorbatchev
 
 docs/install.md @GNUxeava
+
+src/node/i18n/locales/zh-cn.json @zhaozhiming
```
jsjoeio, line 7: Note to self: may have to give write access...

Related: https://github.com/orgs/community/discussions/23042

jsjoeio (commented):
Looking really good! I think last thing is using `lang` instead of `lng`.

src/node/cli.ts
code-asher: So for example if someone wanted to change the language of code-server all on the command line they would have to do something like:

```
code-server --lng zh-cn --locale zh-cn
```

So it seems unfortunate to have to write the same thing twice and it would not be clear to the user why there are two separate flags.

zhaozhiming: ok. That's right. I will change to use the `--locale` flag.

.github/CODEOWNERS
zhaozhiming: I don't know much about `.github/CODEOWNERS`. Is there anything else I need to do?

src/node/cli.ts
zhaozhiming: I had put the description to the `--locale` flag.

.github/CODEOWNERS
jsjoeio: No, you're good here!

jsjoeio (approved):
Nice work! Thanks again for doing this!

zhaozhiming:
@code-asher Is there anything else I need to do in this PR?

code-asher (approved):
Thank you!!

PR #5979, opened 2023-01-19
docs: add port-forwarding

### LINKED ISSUE #5978: [Docs]: document VSCODE_PROXY_URI and port forwarding
## What is your suggestion?

see title

## How will this improve the docs?

easier for people to find: https://github.com/coder/code-server/discussions/5976

yes

jsjoeio:
Fixes #5978

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -5,6 +5,7 @@
 
 - [Expose code-server](#expose-code-server)
   - [Port forwarding via SSH](#port-forwarding-via-ssh)
+  - [Port forwarding via code-server's built-in proxy](#port-forwarding-via-code-servers-built-in-proxy)
   - [Using Let's Encrypt with Caddy](#using-lets-encrypt-with-caddy)
   - [Using Let's Encrypt with NGINX](#using-lets-encrypt-with-nginx)
   - [Using a self-signed certificate](#using-a-self-signed-certificate)
@@ -111,6 +112,18 @@ we recommend using another method, such as [Let's Encrypt](#let-encrypt) instead
 > SSH](https://developer.github.com/v3/guides/using-ssh-agent-forwarding/) and
 > [GPG agent](https://wiki.gnupg.org/AgentForwarding) to the instance to
 > securely access GitHub and sign commits without having to copy your keys.
+
+### Port forwarding via code-server's built-in proxy
+
+code-server has a proxy built-in for port-forwarding. By default, ports running
+on the same machine as code-server can be accessed at
+{current_url}/proxy/{port}. For instance, if you have code-server running on
+localhost:8080 and a Python server running on localhost:8000, you could access
+it via http://localhost:8080/proxy/8000
+
+You can also override the URL scheme for the proxy using the `VSCODE_PROXY_URI`
+environment variable. `VSCODE_PROXY_URI=https://{{port}}.kyle.dev` would forward
+an application running on localhost:3000 to https://3000.kyle.dev
 
 ### Using Let's Encrypt with Caddy
 
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -113,8 +113,6 @@ we recommend using another method, such as [Let's Encrypt](#let-encrypt) instead
 > [GPG agent](https://wiki.gnupg.org/AgentForwarding) to the instance to
 > securely access GitHub and sign commits without having to copy your keys.
 
-### Port forwarding via code-server's built-in proxy
-
 ### Using Let's Encrypt with Caddy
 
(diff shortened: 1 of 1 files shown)


code-asher (commented):
We also mention the proxy under the "Accessing web services" section further down, do you think it would make sense to add this information there instead of a new headingn/

jsjoeio:
> We also talk about the proxy under the "Accessing web services" section further down, do you think it would make sense to add this information there instead of a new heading?

Definitely! I'll revise now.

docs/guide.md
```diff
@@ -112,6 +113,8 @@ we recommend using another method, such as [Let's Encrypt](#let-encrypt) instead
 > [GPG agent](https://wiki.gnupg.org/AgentForwarding) to the instance to
 > securely access GitHub and sign commits without having to copy your keys.
 
+### Port forwarding via code-server's built-in proxy
+
```
code-asher, line 117:
```suggestion
```

docs/guide.md
```diff
@@ -300,6 +303,15 @@ If you're working on web services and want to access them locally, code-server
 can proxy to any port using either a subdomain or a subpath, allowing you to
 securely access these services using code-server's built-in authentication.
 
+By default, ports running on the same machine as code-server can be accessed at
+{current_url}/proxy/{port}. For instance, if you have code-server running on
+localhost:8080 and a Python server running on localhost:8000, you could access
+it via http://localhost:8080/proxy/8000
```
code-asher, line 309: Sorry to nit this so much, I think we might want to change this up because in this paragraph we talk about the subpath and how you can access at `/proxy/{port}`, then we talk about the subdomain version of the proxy, then we talk about the subpath again and mention `/proxy/{port}` again and I think it might not be clear why we talk about it twice.

What if we combine this with the `using a subpath` section to be something like:

```
Simply browse to `/proxy/<port>/`.  For instance, if you have code-server exposed on `domain.tld` and a Python server running on port 8080 of the same machine code-server is running on, you could access the Python server via `domain.tld/proxy/8000`.
```

And in the subdomain section maybe we do something similar and add:

```
For instance, if you have code-server exposed on `domain.tld` and a Python server running on port 8080 of the same machine code-server is running on, you could run code-server with `--proxy-domain domain.tld` and access the Python server via `http://8080.domain.tld`.
```

That way we have matching examples for both types of proxies.

docs/guide.md
```diff
@@ -300,6 +303,15 @@ If you're working on web services and want to access them locally, code-server
...
+
+You can also override the URL scheme for the proxy using the `VSCODE_PROXY_URI`
+environment variable. `VSCODE_PROXY_URI=https://{{port}}.kyle.dev` would forward
+an application running on localhost:3000 to https://3000.kyle.dev
```
code-asher, line 313: To me this makes it sound like `VSCODE_PROXY_URI` will forward your applications automatically but this is essentially a display-only change so you need to have your own proxy hosted at whatever you set this.

What about a new section like:

```
### Using your own proxy

You can make extensions and the ports panel use your own proxy by setting `VSCODE_PROXY_URI`.  For example if you set `VSCODE_PROXY_URI=https://{{port}}.kyle.dev` when an application is detected running on port 3000 of the same machine code-server is running on the ports panel will create a link to https://3000.kyle.dev instead of pointing to the built-in proxy.
```

docs/guide.md
jsjoeio: Great ideas! Let me do another version!

docs/guide.md
jsjoeio: I like how you worded this better. Good idea!

docs/guide.md
```diff
@@ -316,12 +322,29 @@ To set your domain, start code-server with the `--proxy-domain` flag:
...
+Simply browse to `/proxy/<port>/`. For instance, if you have code-server
+exposed on `domain.tld` and a Python server running on port 8080 of the same
+machine code-server is running on, you could access the Python server via
+`domain.tld/proxy/8000`.
+
+### Using your own proxy
+
+You can make extensions and the ports panel use your own proxy by setting
+`VSCODE_PROXY_URI`. For example if you set
+`VSCODE_PROXY_URI=https://{{port}}.kyle.dev` when an application is detected
```
Wireless4024, line 344: relative path like `VSCODE_PROXY_URI=./proxy/{{port}}` also work :)

jsjoeio: TIL! I'll add that. Thank you :)

jsjoeio:
Bump @code-asher

code-asher (commented):
Looking good!!

code-asher (approved):

PR #6015, opened 2023-02-12
feature: add ability to attach ports to code server

sfxworks:
This allows the code server that lives within the cluster to be reached by other services in the cluster. Say, for example, a proxy server for Minecraft.


code-asher (approved):
Thank you!

PR #6028, opened 2023-02-20
docs: fix extras dash in prefix instructions in install.sh

michaeltinsley:
Fixes #


code-asher (approved):
Nice catch and thank you!

PR #6093, opened 2023-03-21
Allow app-name configuration to change the name of the PWA

Lan-Hekary:
Fixes #
if I have multiple instances for code-server running for every user and I want to install the PWA the name currently defaults to "code-server" which overrides the desktop shortcut.
This fix allows multiple instances to be installed with customizable names.

### DIFF
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -40,6 +40,7 @@ export class CodeServerRouteWrapper {
   //#region Route Handlers
 
   private manifest: express.Handler = async (req, res, next) => {
+    const appName = req.args["app-name"] || "code-server"
     res.writeHead(200, { "Content-Type": "application/manifest+json" })
 
     return res.end(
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Thank you!

luc-languagetools:
@Lan-Hekary how does one customize the PWA app name using this code change ? is it a command line argument ?

Lan-Hekary:
Yes, there are two ways to achieve it.
The easiest way is to add it to `config.yaml`, like this ==> `app-name : "Your name"`
and then restart code-server.

The other is, like you said, to pass it as command line line argument `--app-name "Yourname"`
but in order to make this change permanently if you have it running as a service is a little complex.

For example, for a user named ubuntu:
you have to modify the service to add the argument by executing this :
`sudo systemctl edit code-server@ubuntu.service`
And type :
```
[Service]
ExecStart=
ExecStart=/usr/bin/code-server --app-name "Yourname"
```
you have to clear `ExecStart` before setting it to a new value.

you may want to add environment variables too, like Password or hashed password under [Service]
`Environment=HASHED_PASSWORD=blablabla`

See this for reference on how to generate it. 
https://github.com/coder/code-server/blob/main/docs/FAQ.md#can-i-store-my-password-hashed

PR #6154, opened 2023-04-20
Update pathProxy.ts

smalllady:
When trying to preview an HTML file with Chinese characters in its file name, the preview may fail to display correctly.

Fixes #


code-asher:
Thank you for finding this and the fix!  Just FYI I added a test and fixed the lint error, will merge once CI passes.

code-asher (approved):

sha-shrestha:
This is now causing a regression wherein if you have a url with "/" in the query pattern (i.e. example.com/proxy/1234/foo?a=/b), it now encodes it to "example.com/proxy/1234/foo?a=%2Fb" which breaks backend process parser that expects a "/" as a query parameter.

Issue #1964, opened 2020-08-12
Ability to open a file directly using a URL


AbdealiJK:
We're trying to setup a online developmenr environment for our users - and are trying code-server.

We're using it for static HTML pages for website development, and I was hoping to add a button on the HTML pages which says "Edit this page on code-server".

I was hoping to setup that up, I could provide the path / URL to code-server ... something like:
`mycodeserver.domain.com/?file=project/README.md`
And it would open that particular file in the code-server when the button is clicked..

I couldn't find any documentation or notes about whether this is possible right now

NOTE: I did see https://github.com/cdr/code-server/issues/1960 - but it didn't seem the similar

code-asher:
Yeah I don't think this is currently possible it but it seems straightforward enough to implement.

nhooyr:
So this was implemented in https://github.com/cdr/code-server/issues/1292

Will become more visible with https://github.com/cdr/code-server/issues/2179

code-asher:
I think this is for editing a file from a link on a website or something in code-server. So clicking it would launch code-server and then open that file. Actually it would be even cooler maybe if we could handle `web+code-server://` URLs and then they open in your existing code-server (or a new one if it's not open).

nhooyr:
Oh my bad. I thought we supported `?path=` for this or something along those lines @code-asher 

In regards to the scheme URL, see https://github.com/cdr/code-server/issues/1571

code-asher:
We have `?folder=` and `?workspace=` but nothing for individual files yet.

I should have linked that issue! Maybe we just roll this up into that one.

nhooyr:
I think they're different enough, will reopen this.

dkebler:
Just giving my two cents that is worthy enhancement.  A lot of doc sites themes/builders have built in opening the corresponding file in a github repo for editing from the rendered page like this https://geekdocs.de/usage/configuration/  (see `Edit this page` link).  If opening a file directly from url worked in code-server one could use this feature by just supplying the configuration values.  I've already tried and it kinda works sans being able to open the exact page.

```
  geekdocRepo: https://edit.docs.kebler.net
  geekdocEditPath: ?folder=/files/content
```

JammSpread:
I think if `code-server` were to implement this feature, the query param `folder` should be replaced with a single param named `path`. I just think it would be more straightforward and wouldn't be as confusing as having two.

dkebler:
[...]

PR #6191, opened 2023-05-05
Prefer matching editor sessions when opening files.

### LINKED ISSUE #5709: [Bug]: -r opens wrong file in wrong instance

### OS/Web Information

- Web Browser: Firefox
- Local OS: Arch
- Remote OS: Arch
- Remote Architecture: x86_64
- `code-server --version`: 4.7.1 b562d4a880f21aace51b6bd1e7ac64330d023b3a with Code 1.71.2

### Steps to Reproduce

1. Open two code-server workspaces for folders `a` and `b` in two tabs
2. In folder `a`, run `code-server -r a.php`

### Expected

The first tab opens `a/a.php`

### Actual

The second tab opens `b/a.php`

### Logs

_No response_

### Screenshot/Video

_No response_

- [X] I am using HTTPS.

### Notes

_No response_

sleexyz:
Fixes #5709

`readSocketPath` is now replaced by `VscodeSocketResolver` which will prefer editor sessions that match the files being opened.

How it works:
- Every newly created editor session now registers not just its socket path but also workspace data associated with the session, most notably the folders associated with the workspace. This data is encoded as JSON and appended to `vscode-ipc`.
- On CLI file open:
  - Read `vscode-ipc`
  - Return the first active socket, in order of "relevance". An editor session is more relevant if 1) one of the files being opened starts with one of the folders of the workspace, and otherwise if 2) the session was more recently registered to `vscode-ipc`
    - If there were any sockets that failed to connect, we write back to `vscode-ipc` with those sockets deleted. This also serves as a mechansim to prevent the file from growing unbounded.

Notes:
- We use a concatenation of newline-delimited JSON strings instead of a JSON array so we can simply append when registering an IPC socket. With pure JSON we'd have to read then write, which could introduce consistency issues.
- If there are no path matches, we fall back to most-recently created session. This matches VSCode behavior.

Future improvements:
- VSCode prefers workspace with more specific paths (i.e. `/foo/bar/` beats `/foo/` given `/foo/bar/baz.txt`). This was left out, but should not be too hard to add.

### DIFF
--- a/patches/store-socket.diff
+++ b/patches/store-socket.diff
@@ -1,4 +1,4 @@
-Store a static reference to the IPC socket
+Store the IPC socket with workspace metadata.
 
 This lets us use it to open files inside code-server from outside of
 code-server.
@@ -9,6 +9,8 @@ To test this:
 
 It should open in your existing code-server instance.
 
+When the extension host is terminated, the socket is unregistered.
+
 Index: code-server/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.ts
@@ -18,20 +20,114 @@ Index: code-server/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.
   *  Licensed under the MIT License. See License.txt in the project root for license information.
   *--------------------------------------------------------------------------------------------*/
 -
-+import { promises as fs } from 'fs';
-+import * as os from 'os'
++import * as os from 'os';
++import * as _http from 'http';
 +import * as path from 'vs/base/common/path';
  import * as performance from 'vs/base/common/performance';
  import { createApiFactoryAndRegisterActors } from 'vs/workbench/api/common/extHost.api.impl';
  import { RequireInterceptor } from 'vs/workbench/api/common/extHostRequireInterceptor';
-@@ -72,6 +74,10 @@ export class ExtHostExtensionService ext
- 		if (this._initData.remote.isRemote && this._initData.remote.authority) {
- 			const cliServer = this._instaService.createInstance(CLIServer);
- 			process.env['VSCODE_IPC_HOOK_CLI'] = cliServer.ipcHandlePath;
-+
-+			fs.writeFile(path.join(os.tmpdir(), 'vscode-ipc'), cliServer.ipcHandlePath).catch((error) => {
-+				this._logService.error(error);
-+			});
- 		}
+@@ -17,6 +19,7 @@ import { ExtensionRuntime } from 'vs/wor
+ import { CLIServer } from 'vs/workbench/api/node/extHostCLIServer';
+ import { realpathSync } from 'vs/base/node/extpath';
+ import { ExtHostConsoleForwarder } from 'vs/workbench/api/node/extHostConsoleForwarder';
++import { IExtHostWorkspace } from '../common/extHostWorkspace';
  
- 		// Module loading tricks
+ class NodeModuleRequireInterceptor extends RequireInterceptor {
+ 
+@@ -79,6 +82,52 @@ export class ExtHostExtensionService ext
+ 		await interceptor.install();
+ 		performance.mark('code/extHost/didInitAPI');
+ 
++		(async () => {
++			const socketPath = process.env['VSCODE_IPC_HOOK_CLI'];
++			if (!socketPath) {
++				return;
++			}
++			const workspace = this._instaService.invokeFunction((accessor) => {
++				const workspaceService = accessor.get(IExtHostWorkspace);
++				return workspaceService.workspace;
++			});
++			const entry = {
++				workspace,
++				socketPath
++			};
++			const message = JSON.stringify({entry});
++			const codeServerSocketPath = path.join(os.tmpdir(), 'code-server-ipc.sock');
++			await new Promise<void>((resolve, reject) => {
++				const opts: _http.RequestOptions = {
++					path: '/add-session',
++					socketPath: codeServerSocketPath,
++					method: 'POST',
++					headers: {
++						'content-type': 'application/json',
++					}
++				};
++				const req = _http.request(opts, (res) => {
++					res.on('error', reject);
++					res.on('end', () => {
++						try {
++							if (res.statusCode === 200) {
++								resolve();
++							} else {
++								reject(new Error('Unexpected status code: ' + res.statusCode));
++							}
++						} catch (e: unknown) {
++							reject(e);
++						}
++					});
++				});
++				req.on('error', reject);
++				req.write(message);
++				req.end();
++			});
++		})().catch(error => {
++			this._logService.error(error);
++		});
++
+ 		// Do this when extension service exists, but extensions are not being activated yet.
+ 		const configProvider = await this._extHostConfiguration.getConfigProvider();
+ 		await connectProxyResolver(this._extHostWorkspace, configProvider, this, this._logService, this._mainThreadTelemetryProxy, this._initData);
+Index: code-server/lib/vscode/src/vs/workbench/api/node/extensionHostProcess.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/workbench/api/node/extensionHostProcess.ts
++++ code-server/lib/vscode/src/vs/workbench/api/node/extensionHostProcess.ts
+@@ -3,6 +3,9 @@
+  *  Licensed under the MIT License. See License.txt in the project root for license information.
+  *--------------------------------------------------------------------------------------------*/
+ 
++import * as os from 'os';
++import * as _http from 'http';
++import * as path from 'vs/base/common/path';
+ import * as nativeWatchdog from 'native-watchdog';
+ import * as net from 'net';
+ import * as minimist from 'minimist';
+@@ -400,7 +403,28 @@ async function startExtensionHostProcess
+ 	);
+ 
+ 	// rewrite onTerminate-function to be a proper shutdown
+-	onTerminate = (reason: string) => extensionHostMain.terminate(reason);
++	onTerminate = (reason: string) => {
++		extensionHostMain.terminate(reason);
++
++		const socketPath = process.env['VSCODE_IPC_HOOK_CLI'];
++		if (!socketPath) {
++			return;
++		}
++		const message = JSON.stringify({socketPath});
++		const codeServerSocketPath = path.join(os.tmpdir(), 'code-server-ipc.sock');
++		const opts: _http.RequestOptions = {
++			path: '/delete-session',
++			socketPath: codeServerSocketPath,
++			method: 'POST',
++			headers: {
++				'content-type': 'application/json',
++				'accept': 'application/json'
++			}
++		};
++		const req = _http.request(opts);
++		req.write(message);
++		req.end();
++	};
+ }
+ 
+ startExtensionHostProcess().catch((err) => console.log(err));
--- a/src/node/main.ts
+++ b/src/node/main.ts
@@ -1,7 +1,6 @@
 import { field, logger } from "@coder/logger"
 import http from "http"
 import * as os from "os"
-import path from "path"
 import { Disposable } from "../common/emitter"
 import { plural } from "../common/util"
 import { createApp, ensureAddress } from "./app"
@@ -70,9 +69,8 @@ export const openInExistingInstance = async (args: DefaultedArgs, socketPath: st
     forceNewWindow: args["new-window"],
     gotoLineMode: true,
   }
-  const paths = args._ || []
-  for (let i = 0; i < paths.length; i++) {
-    const fp = path.resolve(paths[i])
+  for (let i = 0; i < args._.length; i++) {
+    const fp = args._[i]
     if (await isDirectory(fp)) {
       pipeArgs.folderURIs.push(fp)
     } else {
@@ -123,10 +121,12 @@ export const runCodeServer = async (
   const app = await createApp(args)
   const protocol = args.cert ? "https" : "http"
   const serverAddress = ensureAddress(app.server, protocol)
+  const sessionServerAddress = app.editorSessionManagerServer.address()
   const disposeRoutes = await register(app, args)
 
   logger.info(`Using config file ${humanPath(os.homedir(), args.config)}`)
   logger.info(`${protocol.toUpperCase()} server listening on ${serverAddress.toString()}`)
+  logger.info(`Session server listening on ${sessionServerAddress?.toString()}`)
 
   if (args.auth === AuthType.Password) {
     logger.info("  - Authentication is enabled")
--- a/src/node/vscodeSocket.ts
+++ b/src/node/vscodeSocket.ts
@@ -1,0 +1,206 @@
+import { logger } from "@coder/logger"
+import express from "express"
+import * as http from "http"
+import * as os from "os"
+import * as path from "path"
+import { HttpCode } from "../common/http"
+import { listen } from "./app"
+import { canConnect } from "./util"
+
+// Socket path of the daemonized code-server instance.
+export const DEFAULT_SOCKET_PATH = path.join(os.tmpdir(), "code-server-ipc.sock")
+
+export interface EditorSessionEntry {
+  workspace: {
+    id: string
+    folders: {
+      uri: {
+        path: string
+      }
+    }[]
+  }
+
+  socketPath: string
+}
+
+interface DeleteSessionRequest {
+  socketPath: string
+}
+
+interface AddSessionRequest {
+  entry: EditorSessionEntry
+}
+
+interface GetSessionResponse {
+  socketPath?: string
+}
+
+export async function makeEditorSessionManagerServer(
+  codeServerSocketPath: string,
+  editorSessionManager: EditorSessionManager,
+): Promise<http.Server> {
+  const router = express()
+
+  // eslint-disable-next-line import/no-named-as-default-member
+  router.use(express.json())
+
+  router.get("/session", async (req, res) => {
+    const filePath = req.query.filePath as string
+    if (!filePath) {
+      res.status(HttpCode.BadRequest).send("filePath is required")
+      return
+    }
+    try {
+      const socketPath = await editorSessionManager.getConnectedSocketPath(filePath)
+      const response: GetSessionResponse = { socketPath }
+      res.json(response)
+    } catch (error: unknown) {
+      res.status(HttpCode.ServerError).send(error)
+    }
+  })
+
+  router.post("/add-session", async (req, res) => {
+    const request = req.body as AddSessionRequest
+    if (!request.entry) {
+      res.status(400).send("entry is required")
+    }
+    editorSessionManager.addSession(request.entry)
+    res.status(200).send()
+  })
+
+  router.post("/delete-session", async (req, res) => {
+    const request = req.body as DeleteSessionRequest
+    if (!request.socketPath) {
+      res.status(400).send("socketPath is required")
+    }
+    editorSessionManager.deleteSession(request.socketPath)
+    res.status(200).send()
+  })
+
+  const server = http.createServer(router)
+  await listen(server, { socket: codeServerSocketPath })
+  return server
+}
+
+export class EditorSessionManager {
+  // Map from socket path to EditorSessionEntry.
+  private entries = new Map<string, EditorSessionEntry>()
+
+  addSession(entry: EditorSessionEntry): void {
+    logger.debug(`Adding session to session registry: ${entry.socketPath}`)
+    this.entries.set(entry.socketPath, entry)
+  }
+
+  getCandidatesForFile(filePath: string): EditorSessionEntry[] {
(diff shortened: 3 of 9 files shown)


patches/store-socket.diff
```diff
@@ -24,14 +24,36 @@ Index: code-server/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.
  import * as performance from 'vs/base/common/performance';
  import { createApiFactoryAndRegisterActors } from 'vs/workbench/api/common/extHost.api.impl';
  import { RequireInterceptor } from 'vs/workbench/api/common/extHostRequireInterceptor';
-@@ -72,6 +74,10 @@ export class ExtHostExtensionService ext
- 		if (this._initData.remote.isRemote && this._initData.remote.authority) {
- 			const cliServer = this._instaService.createInstance(CLIServer);
- 			process.env['VSCODE_IPC_HOOK_CLI'] = cliServer.ipcHandlePath;
-+
-+			fs.writeFile(path.join(os.tmpdir(), 'vscode-ipc'), cliServer.ipcHandlePath).catch((error) => {
```
sleexyz, line 32: I had to move this below "Module loading tricks" in order for the workspace to show up with Folders. Not familiar enough with the VSCode codebase to know why.

code-asher (commented):
Thank you for the contribution and nice work!

The test failures seem to be unrelated to your changes.  Not sure what is going on there.

src/node/main.ts
```diff
@@ -61,6 +61,11 @@ export const runCodeCli = async (args: DefaultedArgs): Promise<void> => {
   process.exit(0)
 }
 
+export const getResolvedPathsFromArgs = (args: UserProvidedArgs): string[] => {
```
code-asher, line 64: I think the test errors are from an import cycle between `main` and `cli`.  We could move this logic into `setDefaults`, we already do a bunch of manipulation on the arguments there and this would guarantee anyone trying to use the args would have properly resolved paths.

sleexyz:
Here's a product question: If two tabs for the same workspace are opened, and if the later-opened tab is closed, should we open files in the first tab? Or are we okay with the first tab being orphaned?

To rephrase in pseudocode:

```
tab1 = openEditor("/foo/")
tab2 = openEditor("/foo/")
tab2.close()
openFile("/foo/README.md") // Should this error, or open in tab1?
```

sleexyz:
> Also I think appending like this will cause us to endlessly accumulate socket paths? I know we have the deletion code but that only deletes until it finds a good socket and since the good socket will always be the latest we will never clean up the old ones. Plus that code only runs when you are trying to open a file via the CLI which some folks may never do. Changing how we write here would more or less remove the need to clean up there.

Yes, you're right, the PR right now has the issue of only deleting when opening files via CLI.

If we don't want orphan tabs (https://github.com/coder/code-server/pull/6191#issuecomment-1542592138), we cannot overwrite an entry based on some key (i.e. workspace id) as a mechanism to limit the growth of entries, as each session needs to be stored independently.

AFAIK this means we need to perform some sort of explicit "garbage collection" procedure to limit the growth of entries.

Here's my thought process into how to do this, which builds up to what I'm ultimately proposing (3).

#### 1) Garbage collect on startup

Suppose we "garbage collect" when a daemonized code-server instance starts up: iterate through all sockets and delete dead ones.

This still leaves the possibility for entries to grow unbounded if only a single code-server instance runs and never stops.

#### 2) Periodic garbage collection

Same as (1), but we garbage collect periodically as to limit entries from growing unbounded given a single code-server instance.

Note that if n daemonized code-server instances are running, we have to live with n times more garbage collections.
Read-write races are okay here since two garbage collections should return compatible if not same results.

#### 3) Periodic garbage collection by a single instance

If we're not okay with n-times more garbage collections, we can allow at most a single instance to own responsibility of garbage collection:

Same as (2), but upon garbage collection, we write a `pid, timestamp` pair to `vscode-ipc` that indicates which instance performed the GC, and when it happened.

All instances will periodically check the `vscode-ipc` file:
- If the PID of the instance matches the pid of the `pid, timestamp` pair, we garbage collect
- Otherwise, if the timestamp is beyond a certain threshold (i.e. 2.5x the gc period), we garbage collect. Note here the PID changes.
- Otherwise, we skip garbage collection.

---

@code-asher WDYT?

patches/store-socket.diff
```diff
@@ -24,14 +24,36 @@ Index: code-server/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.
...
++			}
++			const workspace = this._instaService.invokeFunction((accessor) => {
++				const workspaceService = accessor.get(IExtHostWorkspace);
++				return workspaceService.workspace;
++			});
++			const entry = {
++				workspace,
++				socketPath
++			};
++			fs.appendFile(path.join(os.tmpdir(), 'vscode-ipc'), '\n' + JSON.stringify(entry), 'utf-8');
```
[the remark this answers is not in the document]
sleexyz: > The newline-delimited JSON does seem a bit odd to me. What sort of consistency issues did you see with reading the file first compared to only writing?

I didn't encounter consistency issues in practice, but without append, one can imagine opening two editors simultaneously (e.g. chrome restores last browsing session) and a write undoing the previous write (i.e. good: `r w r w`, bad: `r r w w`), leading to an orphan session.

I too would prefer JSON, but [Newline-delimited JSON](https://en.wikipedia.org/wiki/JSON_streaming#Newline-Delimited_JSON) isn't the weirdest thing :) There are node packages for parsing them, but depending on them might be overkill...


> If we are worried about multiple things reading/writing the file at the same time maybe we could have separate files per workspace.

This would still lead to potential consistency problems within a workspace (e.g. two tabs open to the same workspace), if we don't want orphan sessions (https://github.com/coder/code-server/pull/6191#issuecomment-1542592138)

> Also I think appending like this will cause us to endlessly accumulate socket paths? I know we have the deletion code but that only deletes until it finds a good socket and since the good socket will always be the latest we will never clean up the old ones. Plus that code only runs when you are trying to open a file via the CLI which some folks may never do. Changing how we write here would more or less remove the need to clean up there.

Addressing this in https://github.com/coder/code-server/pull/6191#issuecomment-1542722947

---

I personally think removing the possibility of races here is worth using line-delimited JSON over JSON, but I wouldn't be opposed to just using JSON and live with read-write races as a theoretical possibility.

sleexyz:
There is also another possibility where we use an HTTP server instead of a file to manage editor sessions. This could be served by a daemonized code-server instance, and run under a dedicated UDS socket.
- Editor registration: Editor calls a POST handler at the socket with a JSON body.
- Editor resolution: CLI calls a GET handler.

This way, because a single server handles registration, we never have to deal with races. Also, we can just use JSON instead of new-line delimited JSON. Also, we have an API :)

To account for multiple daemonized code-server instances, we can serve under a single instance, using the mechanism detailed in https://github.com/coder/code-server/pull/6191#issuecomment-1542722947 to select the serving instance (and garbage-collecting instance).

patches/store-socket.diff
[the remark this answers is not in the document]
code-asher: Good points!  I agree with line-delimited as the better choice over races but in the case of multiple tabs open to the same workspace maybe it does not matter too much which one wins.

code-asher:
> Here's a product question: If two tabs for the same workspace are opened, and if the later-opened tab is closed, should we open files in the first tab? Or are we okay with the first tab being orphaned?

Interesting scenario!  All else being equal opening in the first tab does seem like the better option but opening the same workspace multiple times at once seems like an unusual workflow to me (maybe there is a use case I am missing though) so I personally would not be opposed to having it be orphaned if that makes things easier for us.

> We should support editor resolution across multiple daemonized code-server instances, right? Or is it sufficient to assume a single a daemonized code-server instance?

We have generally been operating under the assumption that there is a single code-server daemon per user.  I know some people run one instance per project/workspace but that is not how it was meant to be used.  That does make me realize we might want to use `~/.local/share` or something instead of `/tmp` though as multiple users could be writing their own sockets to it.

These are very well thought-out options!  If we end up deciding a tab can be orphaned if a second tab with that same workspace is opened then we could do this file-based (named by workspace ID or something) that way we would not need any garbage collection.  Actually that would eliminate a type of theoretical race the garbage collection system has: if a tab is opened as garbage collection is running after the collector reads but before it writes.

code-server is meant to be long-lived so I would avoid only garbage collecting on startup and since we expect one instance per user I think it will not be worth the effort to do any coordination between multiple instances.

I think the API idea is brilliant.  Come to think of it, we actually already use this pattern where VS Code checks for updates against an `/update` endpoint that the daemon serves so there is some prior art here.

We could probably skip the GET handler if we wanted, since it would be the same instance we could just call something directly.  But maybe a GET handler would be nice if there are external tools that want to hook into this API or something.

The only other ideas I have are to do some kind of file locking or store the socket paths via VS Code's storage provider which presumably does its own file locking.  The advantage to one of those is that the change would be more likely to be accepted upstream one day, but since this is already a feature specific to code-server going the API route seems reasonable to me.

code-asher:
So I was trying to see how native VS Code does this and I think they do it by creating predictable socket names.  For example we could patch the socket paths to be `/tmp/vscode-ipc/<hashed directory>.sock`.  If it already exists we see if it is in use and if so we append `-2` and so on until we find a free suffix to use.

Then when opening a file like `/home/coder/foo/bar/baz` we hash `/home/coder/foo/bar`, look for a match, hash `/home/coder/foo`, look for a match, and so on.  If no match we select the most recently created socket.

Presumably VS Code cleans up their own sockets so we would not have to worry about that.  Or even if they do not clean up their own sockets at least it would be an already existing problem rather than a regression.

sleexyz:
> We have generally been operating under the assumption that there is a single code-server daemon per user. I know some people run one instance per project/workspace but that is not how it was meant to be used.

Great, not having to worry about multiple instances should dramatically simplify things :)

If we combine this assumption with going the server route, we can simplify things even more by keeping the session registry entirely in memory (instead of in a file). Once the daemon dies, all the sessions end up dying too, so we'll never have sessions from previous startups.

> All else being equal opening in the first tab does seem like the better option but opening the same workspace multiple times at once seems like an unusual workflow to me (maybe there is a use case I am missing though)

If I reflect on my own usage of code-server, for most part I _converge_ to one editor session per workspace. But then sometimes I accumulate enough Chrome windows and tabs to the point that spinning up a new editor session to a workspace I might already have opened is less work than finding my original tab.

Or suppose I accidentally open the workspace again in some tab. In either case, I would not expect opening and closing the later opened tab to affect the behavior of some previously opened tab.

Here's another thing to consider: once #1964 is resolved, you could imagine use cases popping up that involve hotlinking / bookmarking files, which probably only makes sense to open in new editor sessions (without something like a Chrome extension to foreground existing tabs, but I digress). For example, I currently use code-server to jot down notes in markdown in a single workspace, and once #1964 is resolved I'd want to be able to make quick ephemeral edits to certain files by just clicking into my bookmarks bar.

FWIW VSCode will open files in an earlier-opened tab if a later-opened tab is closed.

sleexyz:
> So I was trying to see how native VS Code does this and I think they do it by creating predictable socket names (see createStaticIPCHandle). 

I traced through where the IPC path comes from in [extHostExtensionService](https://github.com/microsoft/vscode/blob/2575777b9d3cdb865d61f9f1dfaac1fea9e1bde6/src/vs/workbench/api/node/extHostExtensionService.ts) and it seems like it actually uses [`createRandomIPCHandle`](https://github.com/microsoft/vscode/blob/2575777b9d3cdb865d61f9f1dfaac1fea9e1bde6/src/vs/base/parts/ipc/node/ipc.net.ts#LL756C29-L756C29), which also makes sense if you look through the logic specified in the functions. If you [search for `createStaticIPCHandle`](https://github.com/search?q=repo%3Amicrosoft%2Fvscode%20createStaticIPCHandle&type=code) it seem to be used only in the electron case.

> For example we could patch the socket paths to be `/tmp/vscode-ipc/<hashed directory>.sock`. If it already exists we see if it is in use and if so we append `-2` and so on until we find a free suffix to use (or instead of hashing we could have a nested directory structure so `/tmp/vscode-ipc/my/workspace/ipc-2.sock` or something).

That's an interesting idea, to use the filesystem itself as some sort of registry, where the paths are esssentially registry keys. And not have to maintain a separate source of truth.

We still have an orphan tab problem that any sort of registry with overwrites would cause.

But even if we ignore that, we run into issues once we consider _multi-root workspaces_. For example, a workspace with two roots would have to be registered under two keys, aka two paths. And a socket can only live under one path.

...unless we symlinked the socket under different paths

---
What I'm thinking:

I think if we go the http server route, the registry can just live entirely in memory, which should avoid a lot of cleanup since the editor session registry is always effectively cleared on restart.
- If we're _really_ okay with orphan tabs, then we can key on [`workspace_id`](https://github.com/microsoft/vscode/blob/2575777b9d3cdb865d61f9f1dfaac1fea9e1bde6/src/vs/workbench/services/extensions/common/extensionHostProtocol.ts#LL68C4-L68C4) (which generalizes to multi-root workspaces, unlike keying on folder). If we assume people are opening the same workspaces, then we won't have to deal with garbage collection, because we are overwriting entries per-workspace.
- Otherwise, if we're concerned about memory growing unbounded, we can do something like A) a periodic gc procedure, or B) only gc when a new session is registered, or C) when a new session is registered, attempt a gc just for sessions sharing the same workspace.

code-asher:
Hmm one thing to consider is that the extension host will need to know where to send the HTTP request which I suppose means we need to send the port or socket that code-server is listening on down to the extension host.

But if we have to send a message down anyway, maybe we could send a path for it to register its socket on (rather than have it generate one randomly) then we would not need the extension host to make any HTTP calls back since we already know the socket it will use.   When the host process exits we can remove the socket.

code-asher:
Ah wait no we could not send down the socket because we need to know all the workspaces as well.

And I think sending the port/socket is not going to work either because it might be authenticated and the extension host will not have credentials.  We could probably come up with something for that but maybe it would be easier to send the IPC handle and workspaces up through `process.send()` instead.

sleexyz:
> Hmm one thing to consider is that the extension host will need to know where to send the HTTP request which I suppose means we need to send the port or socket that code-server is listening on down to the extension host.

I was thinking of just generating a UDS socket with a filename based the value of [CODE_SERVER_PARENT_PID](https://github.com/coder/code-server/blob/6745a46034c7e15203c94f571d0ab1e49b497443/src/node/wrapper.ts#L332-L341), which I checked was available in the extension host. Something like `code-server-ipc-[PID].sock`.

> One alternative idea to GC is to remove the socket when the extension host terminates, maybe through a DELETE endpoint. We could put the API calls in https://github.com/microsoft/vscode/blob/9ff83d41a3268ec578c032ae54102285b447e947/src/vs/workbench/api/node/extensionHostProcess.ts in the initialize and terminate functions maybe.

Oooh, that's not a bad idea. I like the simplicity of it, though it does require yet another patch, for arguably a pretty non-critical piece of functionality.
- Periodic GC is slightly icky (i.e. having to reason about some periodic process).
- Per-workspace GC on workspace open means needing to additionally maintain a n:1 relation of session to workspace id (on top of the existing n:m relation of session to folder path).

Let me take a first pass with the HTTP server and I'll try your suggestion with the DELETE endpoint.

src/node/vscodeSocket.ts
```diff
@@ -0,0 +1,142 @@
...
+   * Returns matching and then non-matching socket paths, sorted by most recently registered.
+   */
+  getSocketPaths(paths: string[] = []): string[] {
+    const matches = new Set<string>()
+    for (const socketPath of paths.flatMap((p) => this.getMatchingSocketPaths(p))) {
+      matches.add(socketPath)
+    }
+
+    // Sort by most recently registered, preferring matches.
+    return Array.from(this.entries.values())
```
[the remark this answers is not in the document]
code-asher: Yeah I think emulating VS Code makes the most sense here!

code-asher:
Yeah we can add another socket although we may want to avoid `$tmpdir` and do something user-based to avoid users being able to interfere with other users' instances.

code-asher:
Come to think of it, I guess we actually *have* to add another socket with a predictable file path so the external CLI can interact with it otherwise we would have to go back to storing a list of sockets in a file which is the whole thing we are trying to avoid hahaha

src/node/main.ts
sleexyz: Cool, moving this function into cli.ts fixed the tests. 

> We could move this logic into setDefaults
Done.

test/unit/node/app.test.ts
```diff
@@ -17,7 +17,7 @@ describe("createApp", () => {
   beforeAll(async () => {
     mockLogger()
 
-    const testName = "unlink-socket"
```
sleexyz, line 20: Not sure if you're aware of this but there are [maximum unix socket path lengths of ~100 chars](https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars), so I had to shorten the test name to get this working on MacOS.

Without shortening it, there's some really wonky behavior where it thinks the socket already exists, which seems to be caused by the fact we're creating two sockets now.

Spent 2/3 hours debugging before I realized the socket path lengths were the issue. I'd suggest that we error out explicitly if the socket path lengths exceed the OS maximum, or at least shorten the test tmpdir paths, lest some poor developer come across this silent failure mode again...

sleexyz (commented):
Finally got around to finishing this!

Now a daemonized instance will spin up an additional http server at `$TMPDIR/code-server.sock`.
- The extension host connects to it to register sessions on creation and deregister on termination. No garbage collection needed :)
- The CLI connects to it to get the best matching active socket

> Yeah we can add another socket although we may want to avoid `$tmpdir` and do something user-based to avoid users being able to interfere with other users' instances.

Both the extension host's `vscode-ipc-XXX.sock` files and the old `vscode-ipc` file use $TMPDIR, so perhaps this could be done a in a separate PR?

test/unit/node/app.test.ts
code-asher: Oh wow that is wild.  Good to know!  Erroring out seems very wise to me.  Sounds like the perfect nightmare of a bug.

patches/store-socket.diff
code-asher: I will check it tomorrow!

code-asher:
Absolutely stellar!  I will review the changes first thing tomorrow.  Thank you sticking through all the back and forth!

> Both the extension host's vscode-ipc-XXX.sock files and the old vscode-ipc file use $TMPDIR, so perhaps this could be done a in a separate PR?

That makes perfect sense!  :+1:

code-asher:
One thing we will have to do is update VS Code to 1.78.2 and make sure the patches still apply since mainline was updated to 1.78.2 somewhat recently.

test/unit/node/vscodeSocket.test.ts
```diff
@@ -0,0 +1,243 @@
...
+        socketPath: `${tmpDirPath}/vscode-ipc-bbb.sock`,
+      })
+      const socketPaths = manager.getCandidatesForFile("/ccc/some-file:1:1")
+      expect(socketPaths.map((x) => x.socketPath)).toEqual([
+        `${tmpDirPath}/vscode-ipc-bbb.sock`,
+        `${tmpDirPath}/vscode-ipc-aaa.sock`,
+      ])
+    })
+
+    it("does not just directly do a substring match", async () => {
```
code-asher, line 104: Ahh yeah nice catch :+1:

patches/store-socket.diff
```diff
@@ -18,20 +20,121 @@ Index: code-server/lib/vscode/src/vs/workbench/api/node/extHostExtensionService.
...
++					path: '/add-session',
++					socketPath: codeServerSocketPath,
++					method: 'POST',
++					headers: {
++						'content-type': 'application/json',
++						'accept': 'application/json'
++					}
++				};
++				const req = _http.request(opts, (res) => {
++					if (res.headers['content-type'] !== 'application/json') {
```
code-asher, line 67: Does it matter what the `content-type` is if we are not reading the response body anyway?  I mainly ask because I am seeing this error logged on every request so maybe we are not setting it but I am not sure we need to since we are not sending content anyway.

patches/store-socket.diff
code-asher: Hmm yeah mine are still empty for some reason!  I tried both a workspace file (output listed above) and opening a folder directly.  That one looks like this:

```
{"workspace":{"id":"6b5587dc","folders":[],"transient":false,"configuration":null},"socketPath":"/run/user/1002/vscode-ipc-01935aa6-3ef8-4a7c-8f7b-732ec465af53.sock"}
```

patches/store-socket.diff
sleexyz: Good catch, removed the extraneous lines.

sleexyz (commented):
Just rebased it to the latest version on main (with 1.78.2), works so far for me!

code-asher (approved):
Looks like the linter wants something formatted then we are good to go!

For some reason when I run `yarn watch` I still get no folders but I downloaded the build from CI and it works perfectly fine for me so I am going to chalk it up to something weird on my machine.

sleexyz:
Cool, fixed the formatting issue!

code-asher:
Thanks again!

PR #6198, opened 2023-05-09
fix(Install.sh): Add single qoutes to correctly create the dir

KozmikNano:
# Fixes

- Fixes an issue where creating a folder in the standalone version, it wouldn't actually create it. It just needed some single quotes.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -401,7 +401,7 @@ install_standalone() {
 
   # -w only works if the directory exists so try creating it first. If this
   # fails we can ignore the error as the -w check will then swap us to sudo.
-  sh_c mkdir -p "$STANDALONE_INSTALL_PREFIX" 2> /dev/null || true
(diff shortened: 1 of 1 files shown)


code-asher:
The single quotes will prevent `$STANDALONE_INSTALL_PREFIX` from expanding, no?  So it would be passed literally and then expanded by the subcommand instead which does not have that variable so it would be blank.

KozmikNano:
Closing, due to the fact this isnt needed

KozmikNano:
> The single quotes will prevent `$STANDALONE_INSTALL_PREFIX` from expanding, no? So it would be passed literally and then expanded by the subcommand instead which does not have that variable so it would be blank.

Yeah, I don't know what I was thinking 😅

code-asher:
Hahaha all good, shell semantics are nuts

Issue #5741, opened 2022-11-01
[Bug]: No working Terminal on FreeBSD 13.1-RELEASE-p2


ilovebsd:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### OS/Web Information

- Web Browser: Chrome
- Local OS: Windows 10
- Remote OS: FreeBSD 13.1-RELEASE-p2 Jails
- Remote Architecture: amd64
- `code-server --version`: 4.8.1

### Steps to Reproduce

1. open code-server on Chrome
2. New Notification "The connection to the terminal's pty host process is unresponsive, the terminals may stop working."
3. Repeated "Restart pty host"

### Expected

pty host must operate normally

### Actual

Terminal is not working.

### Logs

[17:00:43] ptyHost was unable to resolve shell environment Error: Unable to resolve your shell environment: Unexpected exit code from spawned shell (code 1, signal null)
    at /usr/local/lib/node_modules/code-server/lib/vscode/out/vs/server/node/server.main.js:92:2607
    at async /usr/local/lib/node_modules/code-server/lib/vscode/out/vs/server/node/server.main.js:67:23805
[IPC Library: Pty Host] The factory function of "vs/platform/terminal/node/ptyHostMain" has thrown an exception
[IPC Library: Pty Host] Error: Platform not supported
    at E (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/platform/terminal/node/ptyHostMain.js:19:811)
    at y (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/platform/terminal/node/ptyHostMain.js:19:110)
    at new E (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/platform/terminal/node/ptyHostMain.js:19:1504)
    at Object.<anonymous> (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/platform/terminal/node/ptyHostMain.js:27:17502)
    at Function.s._safeInvokeFunction (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/loader.js:5:1337)
    at Function.s._invokeFactory (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/loader.js:5:1522)
    at s.complete (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/loader.js:5:1801)
    at s._onModuleComplete (/usr/local/lib/node_modules/code-server/lib/vscode/out/vs/loader.js:6:1112)
[...]

PR #6233, opened 2023-05-23
Update FreeBSD section of npm.md

rowheel:
Partially addresses #5741

### DIFF
--- a/docs/npm.md
+++ b/docs/npm.md
@@ -79,8 +79,7 @@ Proceed to [installing](#installing)
 ## FreeBSD
 
 ```sh
-pkg install -y git python npm-node16 pkgconf
-pkg install -y libinotify
+pkg install -y git python npm-node16 pkgconf libsecret
 ```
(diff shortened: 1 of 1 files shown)


code-asher:
Thank you!  Do we not need `libinotify`?  And how does adding `libsecret` affect the terminal?  If `libsecret` is missing Code is supposed to fall back to an in-memory keystore I think.

rowheel:
> Thank you! Do we not need `libinotify`? And how does adding `libsecret` affect the terminal? If `libsecret` is missing Code is supposed to fall back to an in-memory keystore I think.

As far as I can recall from our observation, yes `libinotify` didn't seem to be mandatory requirement, however, missing `libsecret` was actually causing compile failures.

I'll post a detailed log later.  It's been a while since I worked on this issue, so I'll have to restore the FreeBSD VM snapshot, conduct the experiment again and collect any necessary log.

PR #6282, opened 2023-06-21
Set session socket into environment variable

code-asher:
While I was at it I added a CLI flag to override the default.  I also swapped the default to --user-data-dir.

The value is set on an environment variable so it can be used by the extension host similar to VSCODE_IPC_HOOK_CLI.

Continuation of a fix for https://github.com/coder/code-server/issues/6275.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,16 @@ Code v99.99.999
 
 Code v1.79.2
 
+### Fixed
+
+- Fix being unable to launch multiple instances of code-server for different
+  users.
+
+### Added
+
+- `--session-socket` CLI flag to configure the location of the session socket.
+  By default it will be placed in `--user-data-dir`.
+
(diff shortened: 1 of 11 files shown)


code-asher:
Going to add an e2e test as well since I think we can do that fairly easily.

benz0li:
@code-asher FYI Release v4.14.1-rc.1 (which does not include this change) work just fine.

code-asher:
Thank you!  I will make one more RC then I can make a new release with the bugfix.

code-asher:
@sleexyz just a ping in case you felt like checking this over.  Apologies for unilaterally continuing the fix; I was a bit anxious to finish it because I had just released the RC and I figured we should add an e2e test and did not want to saddle you with that (the e2e scaffolding can be difficult to work with).  Please let me know if anything looks dumb.

Going to merge for now so I can start the next RC tonight but if something needs to be fixed we can always do another RC.

sleexyz:
Looks great! Thanks for the heads up, and sorry for the oversight 😅

PR #6349, opened 2023-07-21
Add --disable-proxy option

ryanbrainard:
Fixes # https://github.com/coder/code-server/issues/6348

This change adds a `disable-proxy` option (set by `--disable-proxy`  flag, `CS_DISABLE_PROXY=1`  or
`CS_DISABLE_PROXY=true`, or `disable-proxy` config), which disables the proxy routes to forwarded ports, including
the domain and path proxy routes over HTTP and WebSockets.

Note, it does not patch VS Code itself to block the auto port forwarding, but that can be disabled with `remote.autoForwardPorts=false`. Generally, these two features would be used together, but I didn't want to tightly couple them.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -34,6 +34,7 @@
 - [Are there community projects involving code-server?](#are-there-community-projects-involving-code-server)
 - [How do I change the port?](#how-do-i-change-the-port)
 - [How do I hide the coder/coder promotion in Help: Getting Started?](#how-do-i-hide-the-codercoder-promotion-in-help-getting-started)
+- [How do I disable the proxy?](#how-do-i-disable-the-proxy)
 - [How do I disable file download?](#how-do-i-disable-file-download)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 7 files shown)


code-asher (approved):
Thank you!  Implementation and tests look flawless.

> Generally, these two features would be used together, but I didn't want to tightly couple them.

Good call, there are cases where they are used separately so this is perfect.

ryanbrainard:
@code-asher Thanks for the quick merge! Any idea when this change will be released?

code-asher:
We generally have a monthly release cadence, usually as soon as we merge the latest Code release from upstream.  This PR just barely missed the previous release so if we wait it would be released next month but I am not opposed to doing another release in the meantime if it seems urgent.

ryanbrainard:
@code-asher It's not super urgent, but since this is related to security/authentication, it's something we'd like to be able to deploy soon. If it's not too much trouble, doing another release would be appreciated.

code-asher:
No problem, seems reasonable to me.  Our usual pattern is to put out an RC and then promote the release the next day if nothing goes wrong so I will go ahead and do that now.

ryanbrainard:
@code-asher awesome, thank you!

PR #6382, opened 2023-08-09
Update Caddyfile syntax

encg:
Added missing curly brackets in Caddyfile configuration example.


code-asher (approved):
Yeah I think braces are only optional if there is only one site block in the file.  The braces make it more flexible for sure.  Thank you!

PR #6385, opened 2023-08-11
Update Code to 1.81.1

benz0li:
Ping @code-asher

### DIFF
--- a/patches/base-path.diff
+++ b/patches/base-path.diff
@@ -255,15 +255,7 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 +++ code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
-@@ -484,6 +484,7 @@ function doCreateUri(path: string, query
- 		});
- 	}
- 
-+	path = (window.location.pathname + "/" + path).replace(/\/\/+/g, "/")
- 	return URI.parse(window.location.href).with({ path, query });
- }
- 
-@@ -495,7 +496,7 @@ function doCreateUri(path: string, query
+@@ -398,7 +398,7 @@ class WorkspaceProvider implements IWork
  	if (!configElement || !configElementAttribute) {
  		throw new Error('Missing web configuration element');
  	}
--- a/patches/base-path.diff
+++ b/patches/base-path.diff
@@ -255,7 +255,17 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 +++ code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
-@@ -398,7 +398,7 @@ class WorkspaceProvider implements IWork
+@@ -144,7 +144,8 @@ class LocalStorageURLCallbackProvider ex
+ 			this.startListening();
+ 		}
+ 
+-		return URI.parse(window.location.href).with({ path: this._callbackRoute, query: queryParams.join('&') });
++		const path = (window.location.pathname + "/" + this._callbackRoute).replace(/\/\/+/g, "/");
++		return URI.parse(window.location.href).with({ path: path, query: queryParams.join('&') });
+ 	}
+ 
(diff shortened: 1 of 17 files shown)


benz0li:
Increase of build timeout is not required. It took something over 30 mins on my machine.

ℹ️ The pipeline for https://github.com/coder/code-server/pull/6369 failed due to an e2e test error.

patches/base-path.diff
```diff
@@ -255,15 +255,7 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/code/browser/workbench/workbench.ts
 +++ code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
-@@ -484,6 +484,7 @@ function doCreateUri(path: string, query
- 		});
- 	}
- 
-+	path = (window.location.pathname + "/" + path).replace(/\/\/+/g, "/")
```
code-asher, line 262: I am not sure the authentication callbacks are still used in code-server, but we might want to add this back.  Looks like it is found on line 147 now.

code-asher:
Weird, I am not sure why the e2e tests are failing.  I re-ran them a few times but it seems like code-server is not loading in time, and I am seeing similar issues locally, but the really weird thing is that it only happens if I have the browser dev tools open.  Maybe they added some kind of debug tool and it runs with dev tools open or in the e2e tests for some reason.

I will be able to investigate more next week.

patches/base-path.diff
benz0li: The following commits had a significant impact on patches:

* https://github.com/microsoft/vscode/commit/322bc2d7d86e8d0c2ccb1bfdabe594e2011d500f
  * File: `lib/vscode/src/vs/code/browser/workbench/workbench.ts`
  * Patches: `base-path.diff`, `proxy-uri.diff`
* https://github.com/microsoft/vscode/commit/ef42b49e806b4a4f1202ec2e31f441465a7047de
  * File: `lib/vscode/src/vs/server/node/serverServices.ts`
  * Patch: `telemetry.diff`

I am not sure I updated the patches correctly.

benz0li:
> I will be able to investigate more next week.

Thank you. It would be great to have a release by the end of this week.

benz0li:
Ping @code-asher

code-asher:
I have been trying to reproduce the hang but I cannot seem to get it to happen again.

I updated the callback path to account for the base path.

In addition to the above, I also tested display languages and the github token.

I get a lot (~23) of "throttler is disposed" errors when closing the browser but it does not seem to cause any issues so I think we can ignore it.  There is an open issue for it: https://github.com/microsoft/vscode/issues/188929

Assuming CI passes I think we will be good to go.

code-asher:
e2e tests still timing out unfortunately.

code-asher:
Sorted the test hang, but now there seems to be some syntax issue preventing WebKit browsers from working.

code-asher:
Thank you for experimenting with it.

I ran the CI build in Safari and the syntax error seems to be coming from some `static { }` blocks.  I guess Safari does not like these.

They are coming from here: https://github.com/microsoft/vscode/blob/6c3e3dba23e8fadc360aed75ce363ba185c49794/extensions/notebook-renderers/src/color.ts#L895-L902

It does not seem like this file has changed recently, so I wonder if the compilation target was changed and now it is outputting JavaScript that is not quite compatible with Safari.

code-asher:
Oh, to clarify the submodule, the e2e step downloads the build package from CI and uses that for the tests, so there is no need to populate lib/vscode.

code-asher:
In the 1.81.1 `static readonly white = new Color(new RGBA(255, 255, 255, 1))` gets transpiled into: 
```js
static {
  this.white = new I(new E(255, 255, 255, 1))
}
```

In the previous version we got:
```js
I.white = new I(new E(255, 255, 255, 1))
```

benz0li:
> I get the same error on vscode.dev so it appears to be an upstream issue.

No worries. We can also skip this version. If you feel the same, please close this issue.

I only prefer versions that work absolutely fine for all browsers.

benz0li:
Let us focus on the next release, then: https://github.com/microsoft/vscode/issues/189778

code-asher:
Sorry all, I was out last week so I was not able to get back to this.  I think skipping makes a lot of sense, we will need to see if we can get it patched upstream in the meantime if it has not already been.

PR #6397, opened 2023-08-25
feat(helm): Don't create Secret if existingSecret is provided

AntoineFr:
When I set `existingSecret` in values.yaml, a secret is still created even if it's not used. This PR removes it.

# Tests

```yaml
# secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-secret
type: Opaque
stringData:
  password: MySuperSecretPassword
```

```yaml
# values.yaml
existingSecret: my-secret
```

## Before

```yaml
# helm template ./ci/helm-chart -s templates/secrets.yaml
---
# Source: code-server/templates/secrets.yaml
apiVersion: v1
kind: Secret
metadata:
  name: release-name-code-server
  annotations:
    "helm.sh/hook": "pre-install"
  labels:
    app.kubernetes.io/name: code-server
    helm.sh/chart: code-server-3.12.1
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
type: Opaque
data:
# Here was a blank line I also removed in my PR
  password: "THh6aGdDYWZrUjMzNXlwdmxlbk95Z0g0"
```

## After

```
Error: could not find template templates/secrets.yaml in chart
```

### DIFF
--- a/ci/helm-chart/templates/secrets.yaml
+++ b/ci/helm-chart/templates/secrets.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.existingSecret }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Thank you!

PR #6419, opened 2023-09-03
🔧 Change config directory from `.config` to `.local`

bdsoha:
Updated `docker` documentation to use correct path locations.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -281,6 +281,7 @@ brew services start code-server
 mkdir -p ~/.config
 docker run -it --name code-server -p 127.0.0.1:8080:8080 \
   -v "$HOME/.local:/home/coder/.local" \
+  -v "$HOME/.config:/home/coder/.config" \
   -v "$PWD:/home/coder/project" \
   -u "$(id -u):$(id -g)" \
   -e "DOCKER_USER=$USER" \
--- a/docs/install.md
+++ b/docs/install.md
@@ -280,7 +280,7 @@ brew services start code-server
 # outside the container.
 mkdir -p ~/.config
 docker run -it --name code-server -p 127.0.0.1:8080:8080 \
-  -v "$HOME/.config:/home/coder/.config" \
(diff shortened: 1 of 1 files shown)


code-asher:
I wonder if we want both?  `.config` contains a config.yaml file code-server reads on startup.

bdsoha:
> I wonder if we want both? `.config` contains a config.yaml file code-server reads on startup.

I personally moved the file to `/etc/coder/config.yaml` to better conform with Linux conventions.

code-asher:
Fair, but `/etc` configuration tends to be world-readable and it may contain a password.

bdsoha:
> Fair, but `/etc` configuration tends to be world-readable and the `config.yaml` may contain a password.

I agree with your comment.
For my specific orchestration, the password is not contained in the config file.

code-asher:
Makes sense!  For the installation docs Iwhat do you think about keeping the `-v "$HOME/.config:/home/coder/.config" \` but also adding `-v "$HOME/.local:/home/coder/.local" \` so both the config and data are persisted outside the container?

code-asher:
_ince there is a comment above about `.config` we may want to make mention of what `.local` is as well.

bdsoha:
> Makes sense! For the installation doc example what do you think about keeping the `-v "$HOME/.config:/home/coder/.config" \` but also adding `-v "$HOME/.local:/home/coder/.local" \` so both the config and data are persisted outside the container?

VSCode doesn't use both paths.
If your goal is to also externally store and mount *user-configs* and *user-packages*, then yes.
Otherwise, IMO, I would just stick with the relevant path where VSCode's `settings.json` is stored.

code-asher:
> VSCode doesn't use both paths.

Right, but code-server does.

Is there a situation where you would not want to mount that config when copy-pasting this command?  Not doing so means that every time you restart the container your config is reset and a new password is generated.

code-asher (approved):
Thank you!

PR #6437, opened 2023-09-21
Build Node for CentOS 7

code-asher:
Node dropped support and I am not sure we can raise our minimum requirements.


code-asher:
Not sure I can use the workflow dispatch from a fork so I am going to resubmit to the main repo.

PR #6438, opened 2023-09-21
Build Node for CentOS 7

code-asher:
Node dropped support and I am not sure we can raise our minimum requirements.


code-asher:
We decided to just up the minimum glibc, since the project is meant to follow VS Code.

PR #6450, opened 2023-09-26
Enable secret storage

code-asher:
It seems that at some point the secret storage went browser-only, but the browser only persists storage if it gets a key from the server to do encryption.  So this implements that endpoint.  Normally the process is kicked off by the presence of a cookie that contains the endpoint but I patched the endpoint in directly because making cookies work with a base path is a bit annoying since they cannot be relative (for the login cookie we send the full path from the client to the server to account for that).

Additionally, I tested `--github-auth` and it works without our patch so I removed it.

Fixes https://github.com/coder/code-server/issues/6408
Fixes https://github.com/coder/code-server/issues/5072
Fixes https://github.com/coder/code-server/issues/6395


benz0li:
@code-asher Thank you for investigating. This is great news.

Could you please release v4.17.1 incorporating this change?

Thank you.

code-asher:
Definitely will, there is one extension that still seems to be having trouble authenticating so I am trying to figure that out, but once I do I will put out a patch release.

code-asher:
Actually I need to test the build changes for arm anyway so I am going to make an RC for now to do that.

PR #6453, opened 2023-09-27
Fix incorrect argon2 target in arm builds

code-asher:
I removed building from source because the comment said it was due to CentOS 7, which we no longer use.

However, it seems that the real issue is that argon2 is pulling the wrong binary for cross-builds?  So we need to force building from source instead.  This should fix https://github.com/coder/code-server/issues/6446 and fix https://github.com/coder/code-server/issues/6448.

Unfortunately building from source is causing the new Kerberos module to fail on arm64.  To fix that we will need to install the arm64 version of the library.

Additionally, building from source is causing keytar to fail (a longstanding issue that previously was accidentally ignored when it failed in CI).  It seems unique to GitHub's VM; I cannot reproduce it in a Docker container.  So, switch to a Docker container.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Install Node.js v18
         uses: actions/setup-node@v3
         with:
-          node-version: "18"
+          node-version: "18.15.0"
 
       - name: Install development tools
         run: |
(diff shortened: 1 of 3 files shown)


code-asher:
I just published an RC with this commit.  If you get a chance to try it please let me know if it fixes the issue!

Lan-Hekary:
I updated it via `curl -fsSL https://code-server.dev/install.sh | sh -s -- --version 4.17.1-rc.1`
and restarted the service, It works now as intended.
Thank you

code-asher:
Awesome thank you for the confirmation!

PR #6468, opened 2023-10-01
Modify manifest for window control overlay support

### LINKED ISSUE #3884: Add support for Window Controls Overlay
Ciao,
it would be nice to have support for window controls overlay: https://web.dev/window-controls-overlay/

When installing the webapp with MS Edge, for example, it would be possible to hide the title bar above the menu bar, thus making almost indistinguishable from the real vscode installed locally on your computer.

DanielTimLee:
Since the release of code-server v4.17.0 (Code 1.82), "Command Center" has become a default option. However, the current code-server PWA app lacks support for the Windows control overlay, resulting in an untidy appearance of the title bar. This commit introduces modifications to the manifest file to enable support for the window control overlay.

Fixes #3884

[Prior to patch]
image

[After the patch]
image

### DIFF
--- a/src/node/routes/vscode.ts
+++ b/src/node/routes/vscode.ts
@@ -55,6 +55,7 @@ export class CodeServerRouteWrapper {
             short_name: appName,
             start_url: ".",
             display: "fullscreen",
+            display_override: ["window-controls-overlay"],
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Thank you!

PR #6529, opened 2023-11-13
Add option to disable Show Local button

### LINKED ISSUE #5118: [Feat]: Add option for disabling "Show Local" while saving
## What is your suggestion?
Reuse the option `--disable-file-downloads` recently implemented on #5055 to remove the `Show Local` button when using the `Save As` option. This option allows the users to save the current file on their local machine.

Example:
Screenshot 2022-04-19 at 14 53 51

One could argue that having the option to save the file locally makes the download as accessible as having the `Download` button when right clicking the file name. As mentioned in the comments on the issue #1386, some industries have concerns about downloading files from the IDE.

Not that I'm aware of.

## How this feature could be implemented?
We could extend the code implemented on #5055. The value of the option `--disable-file-downloads` could be used within a condition that will either add the "Show Local" button or not. As I understand, the button is added on [simpleFileDialog.ts L279](https://github.com/microsoft/vscode/blob/12f99e06d005e5d871a1b1b264719e12ae107353/src/vs/workbench/services/dialogs/browser/simpleFileDialog.ts#L279). _Disclaimer, I'm not familiar with the code base, neither TypeScript, so there might be more places we will need patch._

Yes, but I might need some guidance setting up the environment.

jeffmax:
Fixes #5118

The PR does two things: It hides the "Show Local" button for Save/"Save As"/"Save Workspace As" when the --disable-file-downloads flag is used and it also introduces a new --disable-show-local flag that disables the "Show Local" button for Save/"Save As"/"Save Workspace As" and Open File. I also had to modify the `navigateMenus` test code so that I could target "Save" specifically. Without the changes it was choosing "Save Workspace As" when asked to navigate to "Save".

### DIFF
--- a/patches/disable-show-local.diff
+++ b/patches/disable-show-local.diff
@@ -1,0 +1,137 @@
+Index: code-server/lib/vscode/src/vs/server/node/serverEnvironmentService.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/server/node/serverEnvironmentService.ts
++++ code-server/lib/vscode/src/vs/server/node/serverEnvironmentService.ts
+@@ -17,6 +17,7 @@ export const serverOptions: OptionDescri
+ 	'disable-update-check': { type: 'boolean' },
+ 	'auth': { type: 'string' },
+ 	'disable-file-downloads': { type: 'boolean' },
++	'disable-show-local': { type: 'boolean' },
+ 	'locale': { type: 'string' },
+ 	'disable-getting-started-override': { type: 'boolean' },
+ 
+@@ -101,6 +102,7 @@ export interface ServerParsedArgs {
+ 	'disable-update-check'?: boolean;
+ 	'auth'?: string
+ 	'disable-file-downloads'?: boolean;
++	'disable-show-local'?: boolean;
+ 	'locale'?: string
+ 	'disable-getting-started-override'?: boolean,
+ 
+Index: code-server/lib/vscode/src/vs/server/node/webClientServer.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/server/node/webClientServer.ts
++++ code-server/lib/vscode/src/vs/server/node/webClientServer.ts
+@@ -335,6 +335,7 @@ export class WebClientServer {
+ 			webviewEndpoint: vscodeBase + this._staticRoute + '/out/vs/workbench/contrib/webview/browser/pre',
+ 			userDataPath: this._environmentService.userDataPath,
+ 			isEnabledFileDownloads: !this._environmentService.args['disable-file-downloads'],
++			isEnabledShowLocal: !this._environmentService.args['disable-show-local'],
+ 			isEnabledCoderGettingStarted: !this._environmentService.args['disable-getting-started-override'],
+ 			_wrapWebWorkerExtHostInIframe,
+ 			developmentOptions: { enableSmokeTestDriver: this._environmentService.args['enable-smoke-test-driver'] ? true : undefined, logLevel: this._logService.getLevel() },
+Index: code-server/lib/vscode/src/vs/workbench/browser/web.api.ts
+===================================================================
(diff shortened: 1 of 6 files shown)


code-asher:
Thank you, this is awesome!  I should be able to review more thoroughly this week but at first glance it looks good.  You can ignore the error on the `Build code-server` step just fyi, there is some issue right now where in CI not all the dependencies are being installed which I can hopefully fix soon.

One thought in the meantime, do we need the `disable-show-local` flag?  Or to ask it another way, is there a reasonable case where someone would want to disable the "show local" button but still allow downloads?  My current perception is that they will always be disabled together or not at all.

Thinking out loud, I could see a use for `disable-local-filesystem` eventually, and that one would disable both uploads and downloads, any interaction with the local filesystem at all.

jeffmax:
@code-asher Thanks! I agree that the `--disable-show-local` use-case is a bit confusing. I added it for users that wanted to begin to disable uploads or just generally wanted to disable "Show Local" functionality (I found it generally confusing and unexpected). I was actually planning on tackling a full `--disable-file-uploads` feature next, but I had not dug into that yet. I do have a use case where folks might need to enable/disable uploads and downloads independently of one another, so keeping two separate flags would be preferable to me. I agree though, fewer, clearer flags is better.

Another, related feature that I was hoping to add was the option to log uploads and downloads. It would involve something like adding

    this.logService.info(`readFile ${resource.path}`)

to:

https://github.com/microsoft/vscode/blob/1a5daa3a0231a0fbba4f14db7ec463cf99d7768e/src/vs/platform/files/node/diskFileSystemProviderServer.ts#L95

Is that something you would consider accepting?

code-asher:
> I do have a use case where folks might need to enable/disable uploads and downloads independently of one another

Got it, so what we want in the end is `--disable-file-uploads` and `--disable-file-downloads`.  What do you think about removing `--disable-show-local` for now, and we hold off disabling the "show local" button for opening once we implement `--disable-file-uploads`?  That way we can avoid having to support that extra flag indefinitely.

> Another, related feature that I was hoping to add was the option to log uploads and downloads

To me this feels like something that should be logged at trace level, would that meet your needs?  Or, depending on your goals, maybe you could use the telemetry data.

Also, that will log more than just uploads/downloads, right?  Is that OK for your use case?

jeffmax:
@code-asher Sure, I can remove the `--disable-show-local` flag and standalone tests for now and follow this PR up with one for `--disable-file-uploads` (I am actually digging through the upload code for this now). If I do that, can I roll download related "Show Local" changes on this PR into disable-file-downloads.diff?

Is `trace` another log level like `info` and `debug`? If so, that makes sense to me. I know it would catch more than uploads and downloads but I was looking for the best server-side place to record those actions - is there another spot within the server code that would be more specific to uploads and downloads? I know the client has a logger but it seems to log to the browser console, I didn't see that information going to the server.

code-asher:
> can I roll download related "Show Local" changes on this PR into disable-file-downloads.diff?

Yup, that sounds perfect to me.  Feel free to use it for the uploads diff as well if you want (although we should probably rename it if we do).

> Is trace another log level like info and debug?

Exactly, it would be `this.logService.trace` I think.  It is the most verbose logging level in VS Code.

> I was looking for the best server-side place to record those actions 

Ahhh yeah that makes sense, I think only the browser-side code knows that it is an upload/download.  I thought the browser logs also got stored server-side in a file but it seems I am wrong about that, or at least when I checked just now I did not see any, so `readFile` is looking like the closest place to me as well right now.

jeffmax:
@code-asher Is it ok to roll this into one PR? I was going to do two separate ones for download/upload but if I put them in one diff file its simpler as one. Would it be ok to put the trace statement in there as well?

code-asher:
> Is it ok to roll this into one PR
> Would it be ok to put the trace statement in there as well

Yup both things sound good to me!

jeffmax:
I am closing this in favor of PR #6557

PR #6557, opened 2023-11-29
Disable "Show Local" button, add --disable-file-uploads option and add file operation trace level logs

### LINKED ISSUE #5118: [Feat]: Add option for disabling "Show Local" while saving

jeffmax:
This PR replaces PR #6529. It adds an option to disable file uploads (--disable-file-uploads) and removes the "Show Local" button on the dialog box for relevant file actions for both --disable-file-uploads and --disable-file-downloads options (fix #5118).  It also adds trace level log statements for files that are read from or saved to disk.


code-asher (commented):
This looks great!  Thank you again.

Some unrelated e2e tests are failing, maybe the switch from `text` to `text-is`?  Could be it was matching on partial strings before.

code-asher:
One quirk: seems I can edit local files by dragging them to the editor pane, do we want to block that as well?

jeffmax:
I was looking at that as well.. best I could tell that was actually editing them in place (on the client machine with the client needing to give the browser permissions on the directory) and not actually uploading them to the server. Does that match what you are seeing?

code-asher:
Yup, that is what I was seeing.  My reasoning was that since we remove the "show local" button for opening files (which seems to be doing the same thing) we would want to prevent this as well otherwise it is possible to bypass the button removal by dragging instead.

You bring up a good point that it is not actually uploading anything though so leaving it as-is seems fine to me.

jeffmax:
Is there anything else I need to do, is that prettier error real?
Re: other failing tests- what is failing? I did check other code for use of the navigateMenus and thought I fixed things but it may have been lost when I rebased off the old PR.

code-asher:
Oh wait I think `Command Palette` needs to be `Command Palette...` now.

jeffmax:
@code-asher I ran `yarn fmt` and updated the call to navigateMenus for 'Command Palette'.

code-asher (approved):
Thanks!!  Awesome to have this in.

PR #6574, opened 2023-12-10
⬆ Update Debain version of base image

### LINKED ISSUE #6540: [Feat]: Update base image to use Debians's `bookworm` release
## What is your suggestion?

Update base image to use Debians's `bookworm` release

- Newer versions of base tools, such as: `python`, `git`, etc.
- Stable release version.

Yup, 👍.

bdsoha:
Fixes #6540

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -1,13 +1,13 @@
 # syntax=docker/dockerfile:experimental
 
-ARG BASE=debian:11
+ARG BASE=debian:12
 FROM scratch AS packages
 COPY release-packages/code-server*.deb /tmp/
 
 FROM $BASE
 
 RUN apt-get update \
-  && apt-get install -y \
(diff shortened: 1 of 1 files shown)


code-asher (commented):
Thank you!  The update looks good to me.  Are we sure we will not cause regressions with `--no-install-recommends`?  I am wondering if software that used to be installed will now be missing and we will get issue reports about it.

bdsoha:
> Thank you! The update looks good to me. Are we sure we will not cause regressions with `--no-install-recommends`? I am wondering if software that used to be installed will now be missing and we will get issue reports about it.

I will prepare a separate PR with the `--no-install-recommends` to compare installed packages with and without the flag.

code-asher (approved):
> I will prepare a separate PR with the --no-install-recommends to compare installed packages with and without the flag.

Perfect, thank you!

PR #6581, opened 2023-12-13
⬆ Update and organize `apt` packages

bdsoha:
#### Description

- Use `man-db` instead of `man` *(as proposed by `apt`)*
- Use `vim-tiny` instead of `vim.tiny` *(as proposed by `apt`)*
- Sort requirements alphabetically
- Update `fixuid` version
- Add `wget` to the base package list as it is widely used and available by default on many Linux distros.

--- 

Regarding the suggestion to use `--no-install-recommends` from #6574, the difference between the two installs is minor:

```diff
- 1 upgraded, 74 newly installed, 0 to remove and 5 not upgraded.
+ 1 upgraded, 53 newly installed, 0 to remove and 5 not upgraded.
- After this operation, 200 MB of additional disk space will be used.
+ After this operation, 190 MB of additional disk space will be used.
```

As a conclusion, it is not worth the `~10MB` to possibly cause regressions.

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -10,18 +10,18 @@ RUN apt-get update \
   && apt-get install -y \
     curl \
     dumb-init \
-    zsh \
-    htop \
-    locales \
-    man \
-    nano \
     git \
     git-lfs \
-    procps \
-    openssh-client \
-    sudo \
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Thank you for the updates and the analysis!

PR #6595, opened 2023-12-22
Fix non-functional docker link

Merith-TK:
registry.hub.docker.com leads to an blank white page removing "registry' from the URL fixes it and takes the user to the correct page


matifali (approved):

code-asher (approved):
Thank you!

PR #6670, opened 2024-02-09
Use rpm to install in Fedora image

dsseng:
Dnf will pull caches if invoked for local file install and is also unnecessary

Fixes #


code-asher (approved):
Ahh makes sense, thank you!

PR #6671, opened 2024-02-11
Update FAQ: clarify about hashed pass and compose file

ckesc:
I noticed that it is not obvious how to make hashed passwords work in compose file. So I tried a fix from here https://github.com/coder/code-server/issues/5327#issuecomment-1208106752
And it works! Lets add it to the docs


code-asher (approved):
Thank you!

code-asher:
Just FYI, Prettier seems to dislike the `...` for some reason so I went ahead and removed them for now.

PR #6676, opened 2024-02-15
Update .node-version

puneethrai:
This is to fix security vulnerability mentioned in https://www.hkcert.org/security-bulletin/node-js-multiple-vulnerabilities_20231016

Fixes #
Update Node version to 18.19.1 to fix vulnerabilities in NodeJS


code-asher:
Thank you for the PR!  We keep our version is sync with upstream so we will need to wait for them to update.  https://github.com/microsoft/vscode/blob/main/remote/.yarnrc

puneethrai:
@code-asher any idea when can the update be possible ?

code-asher:
They follow Electron, and the next Electron update appears to be 28, scheduled for Februrary.  Electron 28 uses Node 18.18.2 so I think that will be the next version.  I am not sure about 18.19.1 though.

puneethrai:
Thanks for the update @code-asher . So this is my first time PR for this repo, If you could guide if the PR is fine 

I'll change the version to 18.18.2 if that is what we upgrade to and I believe changelog will be maintained by you

code-asher:
Yeah I will handle the changelog.  I believe we can update to 18.18.2 once VS Code 1.87.0 comes out.

puneethrai:
Thanks @code-asher . I've made the appropriate changes. Will keep an eye on VSCode 1.87.0 release. Looking forward for my first PR merge

code-asher:
Apparently I was wrong about the Node version, looks like they are going to update it in 1.88, not 1.87.

puneethrai:
That would add another month or so right ? We will be dealing with vulnerabilities much longer.

code-asher:
> That would add another month or so right ?
Yup, that is right.

code-asher:
The time has finally come!

PR #6734, opened 2024-03-30
Use password in URL to login similar as in jupyter notebook servers

### LINKED ISSUE #6733: Allow login with token similar to jupyter notebook server
## What is your suggestion?

Simplify the login procedure by allowing to pass the session password in the url to get similar behavior as clicking on a jupyter link of the form
```bash
Jupyter Server 2.12.5 is running at:
    http://localhost:8888/lab?token=fbf7cc58bbd90a9f172490f5c57201e03ca967efe424090c872b3218110f29fa
```

It streamlines the sign in procedure.

```--auth none``` is a workaround but is not secure if there are multiple people working on the same node

Yes

### PR DESCRIPTION
**Pull request by benjaminum**
Fixes #6733 

This change allows passing the password in the URL to code-server to allow automatic login similar to the token in jupyter notebook servers.

We can use the following snippet to create the URL and launch code-server.
```bash
local password=$(openssl rand -hex 32)
echo http://$(hostname -f):8080/login?session_password=$password
PASSWORD=$password /path/to/code-server/bin/code-server --bind-addr 0.0.0.0:8080
```

Clicking on this link printed to the terminal by the above code will login the user to code-server without any interaction.
```bash
http://mycomputer.com:8080/login?session_password=fbf7cc58bbd90a9f172490f5c57201e03ca967efe424090c872b3218110f29fa
```


code-asher (commented):
Thank you for the PR!

I think we should do this on the backend rather than relying on automatically filling the form on the frontend.

If the password is in the query parameters, then we check it, if it matches then we set the cookie.  We could also redirect to strip out the `session_password` query variable so it does not persist in the URL (otherwise someone could potentially read it).

code-asher:
Closing for inactivity, but we can re-open if you come back to this!

PR #6754, opened 2024-04-11
Update Code to 1.88.1

benz0li:
Ping @code-asher

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -229,6 +229,11 @@ jobs:
         with:
           node-version-file: .node-version
 
+      # node-gyp is missing in (at least) npm 9.8.1.
+      # TODO: Remove once we update to npm>=10?
+      - name: Install node-gyp
+        run: npm install -g node-gyp
+
       - name: Fetch dependencies from cache
         id: cache-node-modules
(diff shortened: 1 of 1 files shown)


benz0li:
@code-asher Please re-run. My local build was successful.

code-asher:
Weird, I ran it a couple times but it keeps getting `ENOENT`.

Oh...I think I know why.  Something about `node-gyp` being removed from npm.  We updated Node to 18.18.2, but it did not trigger a rebuild of VS Code because of the cache (which itself is a bug I will have to fix), so it is only now causing an issue.

Might work to install `node-gyp` globally, let me test it later today and see.

code-asher (approved):

benz0li:
@code-asher What about the failed e2e tests?: https://github.com/coder/code-server/actions/runs/8698237868

code-asher:
Ah yeah I was worried about that, looks like the Express update broke something.

Not to worry though, the commit that was actually released is https://github.com/coder/code-server/commit/9a28bc29dbddb6886dfe03dc1c31320249a901ce which did pass, that commit was just for updating the changelog and Helm chart afterward.

code-asher:
Fixed!

PR #6783, opened 2024-05-03
Update Code to 1.89.0

benz0li:
Fixes https://github.com/coder/code-server/issues/6782

### DIFF
--- a/patches/base-path.diff
+++ b/patches/base-path.diff
@@ -10,7 +10,7 @@ Index: code-server/lib/vscode/src/vs/base/common/network.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/base/common/network.ts
 +++ code-server/lib/vscode/src/vs/base/common/network.ts
-@@ -198,7 +198,9 @@ class RemoteAuthoritiesImpl {
+@@ -205,7 +205,9 @@ class RemoteAuthoritiesImpl {
  		return URI.from({
  			scheme: platform.isWeb ? this._preferredWebSchema : Schemas.vscodeRemoteResource,
  			authority: `${host}:${port}`,
--- a/patches/disable-builtin-ext-update.diff
+++ b/patches/disable-builtin-ext-update.diff
@@ -7,7 +7,7 @@ Index: code-server/lib/vscode/src/vs/workbench/contrib/extensions/browser/extens
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/workbench/contrib/extensions/browser/extensionsWorkbenchService.ts
 +++ code-server/lib/vscode/src/vs/workbench/contrib/extensions/browser/extensionsWorkbenchService.ts
-@@ -283,6 +283,10 @@ export class Extension implements IExten
+@@ -284,6 +284,10 @@ export class Extension implements IExten
  			if (this.type === ExtensionType.System && this.productService.quality === 'stable') {
  				return false;
  			}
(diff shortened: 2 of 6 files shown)


benz0li:
`code-server-4.24.0-rc.1-linux-amd64.tar.gz` with Code 1.89.0 is deployed at https://coder.jupyter.b-data.ch.

Functionality [modified by patches] tested and found to work:

* base-path
* cli-window-open
* local-storage
* marketplace
* proxy-url
* service-worker
* web view

Workspaces and Jupyter Notebooks also work fine:

*  `ms-toolsai.jupyter@2024.3.1`
*  `ms-python.python@2024.4.1`

Cross reference:

* https://github.com/open-vsx/publish-extensions/issues/740

benz0li:
@code-asher I have built as usual. Somehow `code-server --version` shows `... with Code 1.90.0`.

**JupyterLab Terminal**

```bash
code-server --version
```

```
4.24.0-rc.1 fb2afbd9d62532be3952118adafff3972c63f3bc with Code 1.90.0
```

**code-server Terminal**

```bash
code-server --version
```

```
1.90.0
fb2afbd9d62532be3952118adafff3972c63f3bc
x64
```

I have no idea why. Just wanted to let you know.

code-asher:
> Somehow code-server --version shows ... with Code 1.90.0

How strange!  I believe it just reads `lib/vscode/package.json`.  Maybe VS Code branched 1.89 off the wrong commit.

Yeah I checked and the `package.json` says `1.90.0`.  Oh maybe we should use the `release/1.89` branch.  I think maybe they put the `1.89.0` tag on the wrong commit.

benz0li:
> Yeah I checked and the `package.json` says `1.90.0`. Oh maybe we should use the `release/1.89` branch, it seems to be right there. I think maybe they put the `1.89.0` tag on the wrong commit. Or possibly they messed up the release branch and force pushed it.

VS Code v1.89.0 is built off commit https://github.com/microsoft/vscode/commit/b58957e67ee1e712cebf466b995adf4c5307b2bd.

I will update this merge request accordingly.

benz0li:
> VS Code v1.89.0 is built off commit [microsoft/vscode@b58957e](https://github.com/microsoft/vscode/commit/b58957e67ee1e712cebf466b995adf4c5307b2bd).
> 
> I will update this merge request accordingly.

Updated on their side, too:

* https://github.com/microsoft/vscode/issues/210887#issuecomment-2094038559 ff
* https://github.com/microsoft/vscode/releases/tag/1.89.0

code-asher (approved):
Awesome, thank you!

benz0li:
@code-asher With the "Alternative", the patch releases are in sync. And why not keep the minor releases in sync, too?  
👉 I.e. "Code - OSS" v1.89.1 = code-server v4.89.1

code-asher:
It does sound reasonable to keep them in sync, unfortunately sometimes we make fixes or additions to code-server other than updating VS Code, so as soon as that happens we would be permanently out of sync, I think.

benz0li:
> It does sound reasonable to keep them in sync, unfortunately sometimes we make fixes or additions to code-server other than updating VS Code, so as soon as that happens we would be permanently out of sync, I think.

IMHO updates other than "Code - OSS" are simply additional patch releases, i.e. `.3`, `.4`, etc.

Breaking changes in code-server require updating the major version – accoring to [semver](https://semver.org).

Other than that, minor and patch releases remain in sync with upstream "Code - OSS".

code-asher:
I would not be comfortable releasing a feature update as a patch, I think.  

But I agree that those updates can just wait until the next release.  I am open to experimenting with this, if we do come across a feature update that absolutely has to be released early for some reason then we can revisit.

code-asher:
I put out v4.89.0-rc.1!

smerschjohann:
An alternative would be to use semver builds v4.89.1+coder1 like k3s.io is using (e.g. v1.30.0+k3s1). I think going back to v1.89.1+coder1 is not an option ;)

PR #6789, opened 2024-05-08
fix: Update node version

alexandersperling:
* Update node version to solve CVE-2024-21892

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "@types/express": "^4.17.17",
     "@types/http-proxy": "1.17.7",
     "@types/js-yaml": "^4.0.6",
-    "@types/node": "^18.0.0",
+    "@types/node": "^18.19.1",
     "@types/pem": "^1.14.1",
     "@types/proxy-from-env": "^1.0.1",
     "@types/safe-compare": "^1.1.0",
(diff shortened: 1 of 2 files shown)


code-asher:
Thank you for the PR!  This project just wraps the version used in VS Code, since that is the one that is tested and supported by them: https://github.com/microsoft/vscode/blob/main/remote/.yarnrc

Once they have updated, we can as well.

alexandersperling:
they now updated the version to 20.9.0, but this wont fix the CVE either...

code-asher:
Oh weird I wonder why they chose to do that.  Oh it probably matches whatever Node version is being used in Electron...

alexandersperling:
I wonder if it should work with node `20.12.1` since this is the version which is defined in the `.nvmrc`

You know why there are differences with the .yarnrc you linked before and the general node version used in the repo?

code-asher:
> You know why there are differences with the .yarnrc you linked before and the general node version used in the repo?

I am not sure, that does seem weird.  :thinking:

alexandersperling:
@code-asher they now updated to 20.11.1 and so did I 😄

code-asher:
Awesome!  We update VS Code on each release so we will need to wait until 1.90.0 or maybe 1.91.0.

code-asher:
Thanks again for the PR!  We ended up doing this as part of the VS Code upgrade so I will close this.

Weirdly, 20.11.1 keeps failing in CI (core dump) so I am trying out 20.12.1 from the .nvmrc instead.  It looks like they build with .nvmrc (20.12.1) but then they ship with the .yarnrc (20.11.1).  I am not sure why there is a discrepancy.  I am going to try both building and shipping with 20.12.1 and hopefully that is fine.

PR #6803, opened 2024-05-21
Allow emptyDir type volume mounts

### LINKED ISSUE #6802: [helm chart] Support emptyDir type volume mounts
## What is your suggestion?
I'd like to create emptyDir type volumes and mount them to the main container. 
The need for it arises from some setup tasks that are done by initContainers. For example, the apt mirror setup(/etc/apt/sources.list.d/debian.sources) can be brought in by an initContainer and transferred to the main container through a shared emptyDir volume.
Host path type volumes can do, but they are too heavy for my use cases
Yes

onichandame:
Fixes #6802

### DIFF
--- a/ci/helm-chart/templates/deployment.yaml
+++ b/ci/helm-chart/templates/deployment.yaml
@@ -177,9 +177,12 @@ spec:
         {{- if .existingClaim }}
         persistentVolumeClaim:
           claimName: {{ .existingClaim }}
-        {{- else }}
+        {{- else if .hostPath }}
         hostPath:
           path: {{ .hostPath }}
           type: Directory
+        {{- else }}
+        emptyDir:
+          {{- toYaml .emptyDir | nindent 10 }}
         {{- end }}
       {{- end }}
--- a/ci/helm-chart/values.yaml
+++ b/ci/helm-chart/values.yaml
@@ -190,6 +190,7 @@ extraVolumeMounts: []
   #   readOnly: true
   #   existingClaim: volume-claim
   #   hostPath: ""
+  #   emptyDir: {}
 
 extraConfigmapMounts: []
   # - name: certs-configmap


ci/helm-chart/templates/deployment.yaml
```diff
@@ -177,9 +177,12 @@ spec:
...
         persistentVolumeClaim:
           claimName: {{ .existingClaim }}
-        {{- else }}
+        {{- else if .hostPath }}
         hostPath:
           path: {{ .hostPath }}
           type: Directory
+        {{- else }}
+        emptyDir:
+          {{- toYaml .emptyDir | nindent 10 }}
```
code-asher, line 186: I am not too familiar with Helm, but if `emptyDir` is not specified then will this error?  Should we do `{{- else if .emptyDir }}` or make `emptyDir` default to `{}`?

code-asher (commented):
Thank you!  It looks good to me, just have one question.

ci/helm-chart/templates/deployment.yaml
onichandame: When emptyDir and all other configuration fields(hostPath etc.) are not set, the generated yaml would be `emptyDir: null`. I just tested that if such yaml is applied it would be transformed to `emptyDir: {}` by k8s.

I am not sure if it is the intended behaviour tho. I'd imagine that one of `existingClaim`|`hostPath`|`emptyDir` has to be set. Now it effectively defaults to emptyDir if not set. Would it be more desirable if an error is thrown when none of the fields above is set?

code-asher: Ahh that makes sense, thank you!

I am not sure either; I could see arguments for either way.  Hmm I vote we keep it as is it currently is.

Issue #6709, opened 2024-03-04
update ubuntu base image to jammy 


WisniewskiP:
## What is your suggestion?
update ubuntu to jammy 22.04 or add it as second image 

## Why do you want this feature?
it contains new version for many packages eg. cmake

## Are there any workarounds to get this functionality today?
using debian based image could be possible but requires modification of some scripts that we are using.

## Are you interested in submitting a PR for this?
I could make PR for this (change docker-bake.hcl) but I don't know what impact it could have on the image.
I don't know what dependencies could fail in new ubuntu and what should be tested and how to test it.

code-asher:
The only way to test at the moment is to build it locally then make sure it runs, although this can be a pain.  If you want, you can add a new block for Jammy and we can try it out on the next release.  If everything seems fine, we can eventually switch the default `ubuntu` tag to point to Jammy.

everflux:
As Ubuntu 20.04/focal is nearing EOL I think this should be really moved forward.

PR #6812, opened 2024-05-24
added ubuntu noble as base image

WisniewskiP:
Adds new ubuntu release as base image #6709

### DIFF
--- a/ci/release-image/docker-bake.hcl
+++ b/ci/release-image/docker-bake.hcl
@@ -18,6 +18,7 @@ group "default" {
     targets = [
         "code-server-debian-12",
         "code-server-ubuntu-focal",
+        "code-server-ubuntu-noble",
         "code-server-fedora-39",
         "code-server-opensuse-tumbleweed",
     ]
@@ -65,6 +66,17 @@ target "code-server-ubuntu-focal" {
     )
     args = {
         BASE = "ubuntu:focal"
+    }
+    platforms = ["linux/amd64", "linux/arm64"]
+}
+
+target "code-server-ubuntu-noble" {
+    dockerfile = "ci/release-image/Dockerfile"
+    tags = concat(
+        gen_tags_for_docker_and_ghcr("noble"),
+    )
+    args = {
+        BASE = "ubuntu:noble"
     }
     platforms = ["linux/amd64", "linux/arm64"]
 }


code-asher:
I have not tried them, but looks like they were pushed successfully.

https://hub.docker.com/layers/codercom/code-server/4.89.1-noble/images/sha256-2654982be8ab16f9f9b8071722ac90a96ca4ec27072999fabe9dcc211adc3e63?context=explore
https://hub.docker.com/layers/codercom/code-server/noble/images/sha256-2654982be8ab16f9f9b8071722ac90a96ca4ec27072999fabe9dcc211adc3e63?context=explore

WisniewskiP:
@code-asher I just tried this image it starts but there are somethings that are different.

in the base ubuntu image there is ubuntu user with ID 1000:
```
root@fbce18fd9fc5:/# id ubuntu
uid=1000(ubuntu) gid=1000(ubuntu) groups=1000(ubuntu),4(adm),20(dialout),24(cdrom),25(floppy),27(sudo),29(audio),30(dip),44(video),46(plugdev)
```
unfortunately you are assuming that user `coder` has this ID and you switch to it at the end of Dockerfile:
https://github.com/coder/code-server/blob/main/ci/release-image/Dockerfile#L55

when server starts it uses ubuntu user instead of coder:
```
[2024-05-27T07:57:13.851Z] info  Using user-data-dir /home/ubuntu/.local/share/code-server
[2024-05-27T07:57:13.860Z] info  Using config file /home/ubuntu/.config/code-server/config.yaml
[2024-05-27T07:57:13.860Z] info  HTTP server listening on http://0.0.0.0:8080/
[2024-05-27T07:57:13.860Z] info    - Authentication is enabled
[2024-05-27T07:57:13.860Z] info      - Using password from /home/ubuntu/.config/code-server/config.yaml
[2024-05-27T07:57:13.860Z] info    - Not serving HTTPS
[2024-05-27T07:57:13.860Z] info  Session server listening on /home/ubuntu/.local/share/code-server/code-server-ipc.sock
```

additionally sudo expects password because of wrong user.

to fix it it would be possible to user user name instead of id in Dockerfile. I can make PR for this if you agree with this fix.

code-asher:
The comment seems to imply we need to use the ID to support `DOCKER_USER` but I am not 100% sure that is true.  If we can use the user name instead and it still works, then that seems fine to me!

If that does not work, maybe we will have to do something in the entrypoint, like check if `whoami` is something other than `coder`, and if so rename it.

WisniewskiP:
I have looked at this and for me it works when I replace `USER 1000` with `USER coder`. But in helm-chart/values.yaml you are using:
```yaml
securityContext:
  enabled: true
  fsGroup: 1000
  runAsUser: 1000
```

this change could break some of your CI scripts and probably other scripts that are expecting user with id 1000.

code-asher:
Yeah I am not sure, I have not actually used or tested the Helm chart myself.

PR #6833, opened 2024-06-10
Update Node to 20.12.1

code-asher:
Still trying to figure out a failure to connect the first time that reproduces consistently for me.


code-asher:
I think I finally narrowed the issue down.  Seems to have something to do with the display language patch.

PR #6889, opened 2024-07-09
Update upgrade.md

Bulla199606:
Fixes #


code-asher:
This should not normally be necessary, and the paths do not quite look right either.  Could you explain more?

PR #6941, opened 2024-08-08
Update VS Code to 1.92.1

code-asher:
Closes https://github.com/coder/code-server/issues/6920

~When building locally, bootstrap-node.js was missing...not sure what is going on there.~  Edit: Looks like the files are bundled now (into `server-main.js`).  We could unbundle them again but I opted to call into `server-main.js` (took a small amount patching) since it also seems to do some nls stuff which is maybe important.  Also it handles all that VS Code-specific environment variable stuff we were copying before.

The es2020 patch does not work.  It results in invalid code.  Not sure if this is a bug with whatever is doing the transpilation (tsc?) but unfortunately I had to remove it.

~Still need to figure out what to do about the language stuff.~  ~I am going to see if our patch is still necessary.~ ~Edit: It is, but something about it does not quite work.~  Edit 2: Figured it out, looks like the browser needs a file that sets translations into a global so I am now generating that file locally since we cannot use Microsoft's CDN to pull down translations like how they have it set up.

~Also need to look into the proxy URI test failing.~  A flake?  Works fine locally.


bilogic:
yay

benz0li:
`code-server-4.92.2-rc.1-linux-amd64.tar.gz` with Code 1.92.2 is deployed at https://coder.jupyter.b-data.ch.

Functionality [modified by patches] tested and found to work:

* base-path
* cli-window-open
* local-storage
* marketplace
* proxy-url
* service-worker
* web view

Workspaces and Jupyter Notebooks also work fine:

*  `ms-toolsai.jupyter@2024.7.0`
*  `ms-python.python@2024.12.3`

code-asher:
Huzzah thank you for testing!

PR #6958, opened 2024-08-25
feat: introduce --abs-proxy-base-path that allows app proxying while code-server is not server at the root

### LINKED ISSUE #6770: Add --base-path flag for use with absolute proxy
This is not for code-server itself (since it works relatively), it would just be prepended to `absproxy` requests for applications that need the full path.

rafaelnferreira:
## What it is
- Allows proxy traffic to be properly routed when `code-server` is served under a path.

## How
- Introduces a new switch `abs-proxy-base-path`.
- When specified, `absproxy` requests will be forwarded using the value as prefix

## Example
- Code server is hosted at: `my-codeserver.com/user/123`
- React App is started at port 8080 under `PUBLIC_PATH` `/user/123`
- `abs-proxy-base-path` set to `/user/123`
- A `GET` request to `my-codeserver.com/user/123/absproxy/8080/app` will properly reach the app.

Fixes #6770

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -432,3 +432,12 @@ const config = {
 3. Access app at `<code-server-root>/absproxy/5173/` e.g. `http://localhost:8080/absproxy/5173/
 
 For additional context, see [this Github Issue](https://github.com/sveltejs/kit/issues/2958)
+
+### Prefixing `/absproxy/<port>` with a path
+
+This is a case where you need to serve an application via `absproxy` as explained above while serving `codeserver` itself from a path other than the root in your domain.
+
+For example: `http://my-code-server.com/user/123/workspace/my-app`. To achieve this result:
+
+1. start code server with the switch `--abs-proxy-base-path=/user/123/workspace`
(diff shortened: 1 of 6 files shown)


code-asher (approved):
This is great thank you for implementing!

code-asher:
Looks like our formatter is being a bit weird, let me fix it real quick

yuzhuoheng:
abs-proxy-base-path   It didn't work. This is my startup command:
企业微信截图_17288752683468
I tried to open it with a browser, and the page prompt was as follows:
企业微信截图_1728875413984

code-asher:
This flag is just for code-server's built-in proxy (to proxy to other applications on the remote), not for code-server itself.

To serve code-server at a base path, you should use a reverse proxy like NGINX or Caddy.

shalldie:
I want to serve code-server on a subpath like `/code`, but the `ci/helm-chart` can't edit pathType in `values.yaml`, which resulting in inability to rewrite.

I think it would be great if it could natively support subpath.

PR #6968, opened 2024-08-30
Fix install.sh refusing to download macos-arm64 standalone.

yuchenshi:
Fixes part of https://github.com/coder/code-server/issues/6659#issuecomment-2318943559

Ever since [v4.89.0-rc.1](https://github.com/coder/code-server/releases/tag/v4.89.0-rc.1), there's only macos-arm64 but no more amd64. This fixes the script to align with the actual release artifacts.

In the long run, if we ever add back macos-amd64, the check can be removed entirely.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -461,9 +461,9 @@ npm_fallback() {
 # Determine if we have standalone releases on GitHub for the system's arch.
 has_standalone() {
   case $ARCH in
-    amd64) return 0 ;;
-    # We only have amd64 for macOS.
-    arm64)
+    arm64) return 0 ;;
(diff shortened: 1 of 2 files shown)


code-asher:
Thanks again!  I suppose we should probably check the version to make sure we are doing the right check but that seems like it would be tricky in bash so I think this makes sense.  Looks like we have some script tests that need to be updated, I should really look into making CI run automatically instead of requiring approval to run first

yuchenshi:
Tests updated (and another CI approval please)

code-asher (approved):
Thank you!

PR #7085, opened 2024-11-22
fix extraInitContainers conditional

TheodoreHerzfeld:
When using `extraInitContainers` but not a persistant volume, the `initContainers` section does not render properly. This fixes that


code-asher (approved):
Thank you!

PR #7111, opened 2024-12-12
chore: update and pinned prettier to version 3.4.2 and fix ci

qupig:
According to the `Prettier` [documentation](https://prettier.io/docs/en/install#summary), we should always install an **exact version**:

> Install an **exact version** of Prettier locally in your project. This makes sure that everyone in the project gets the exact same version of Prettier. Even a patch release of Prettier can result in slightly different formatting, so you wouldn’t want different team members using different versions and formatting each other’s changes back and forth.

That is, in `package.json` it should look like `"prettier": "3.0.3"` instead of `"prettier": "^3.0.3"`.

Don’t forget the `--save-exact` parameter when installing or updating the `prettier`:

```sh
npm install --save-dev --save-exact prettier@3.4.2
```

Otherwise, following the [Development workflow](https://github.com/coder/code-server/blob/main/docs/CONTRIBUTING.md#development-workflow), one can easily get different `prettier` versions and end up with `CI` failing, and this is often a source of confusion.

Therefore, this PR mainly **pinned** the version of `prettier` and updates it to the latest version v3.4.2 by the way.

It's worth noting that I added the `commit` hash of the actual formatted code in `.git-blame-ignore-revs` to ignore non-actual code changes for git blame.
- It is recommended that future `prettier` version updates take the same action
- So note that this PR should NOT be squash/rebase merge, otherwise the commit hash ​​will not match
  - Or you could revise the commit hash in `.git-blame-ignore-revs` after merging

---

The original CI process uses the third-party workflow `actionsx/prettier@v3` with a fixed version of `prettier` (currently seems to be 3.0), which is another cause of confusion.

We should always use the version specified in `package.json`, and always maintain version numbers in one place.

The speed difference is insignificant here, and it's not a shortcoming in the overall CI process.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Prettier 3.4.2
+9b0340a09276f93c054d705d1b9a5f24cc5dbc97
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,17 +23,25 @@ concurrency:
 
 jobs:
   prettier:
-    name: Format with Prettier
+    name: Run prettier check
     runs-on: ubuntu-22.04
     timeout-minutes: 5
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
 
-      - name: Run prettier with actionsx/prettier
-        uses: actionsx/prettier@v3
+      - name: Install Node.js
+        uses: actions/setup-node@v4
         with:
-          args: --check --log-level=warn .
+          node-version-file: .node-version
+          cache: npm
+          cache-dependency-path: |
+            package-lock.json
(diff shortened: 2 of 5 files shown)


code-asher (approved):
Ahhh that explains why that block kept getting formatted differently.  Thank you for fixing this!

code-asher:
Hey, sorry about that, yeah we have not updated Express yet because doing so causes it to completely fail, and as you noticed all you get is a blank screen (I think all of the http requests error or time out).  I think I need to make that audit check optional.

code-asher:
I will drop that last commit and force merge

code-asher:
Thank you again!

Issue #7077, opened 2024-11-12
Fix FOUC issues in dark mode


qupig:
Currently, after selecting the dark theme, every operation such as opening/closing a folder will produce a brief full-screen white screen flicker i.e. `FOUC (Flash of Unstyled Content)`, which is very unfriendly to the eyes.

An additional question is that when the system/browser is in dark mode, the light theme is always the default when entering `code-server`. I always have to choose it manually. Maybe the matching theme should also be automatically applied?

code-asher:
These are probably best solved upstream, we are just wrapping VS Code's HTML.

That said, having to select the theme multiple times sounds like a bug, and it could be a problem created by our patches.  We patch the theme to be stored on the remote server, so as long as the config is not deleted from there it should keep using the same theme.

code-asher:
Actually we may need to solve it here anyway because we should also style the login page with the current theme.

code-asher:
I modified https://github.com/coder/code-server/issues/3019 to include styling the VS Code page before it loads, since it will all be the same work, and will close this one as a duplicate.

code-asher:
Ahh I see.  This is expected behavior, in that case.  You could work around it by seeding each instance with a `settings.json` file that has the theme set.

Issue #3019, opened 2021-04-02
Feature request for default login page


nooobcoder:
Please update the background of the login screen background. It has been static for quite a long time. Try putting cool animations to add a dynamic feeling to it. Also, bring a feature to change the welcome message on the login screen instead of the default message. I think showing some animated shapes (colorful) on a light theme would certainly look nice and there would be a dark theme toggle on right top side.

![image](https://user-images.githubusercontent.com/50350828/113375328-c3865f80-938c-11eb-9a11-9d35a33e3b9e.png)

code-asher:
Maybe we could use the same theme VS Code is using? That could be cool.

nooobcoder:
@code-asher what about the shades of purple theme? 
https://marketplace.visualstudio.com/items?itemName=ahmadawais.shades-of-purple

The color scheme is so good, maybe code-server should use it by default. Just a recommendation.

nooobcoder:
@jsjoeio @code-asher can I expect this feature in the next release `3.9.4` I am skipping updating to `3.9.3` as I could not see any major changes in code-server.

code-asher:
We don't have this on the `3.9.4` milestone; I think it's unlikely we'll get to it soon.

jsjoeio:
Echoing Asher, this is a lower priority for us. We also have some changes happening across our internal/external projects, which would be related to this. cc @vapurrmaid

vapurrmaid:
> Echoing Asher, this is a lower priority for us. We also have some changes happening across our internal/external projects, which would be related to this. cc @vapurrmaid

Hey @jsjoeio - yea, I think this is more likely to land June or shortly thereafter, as our internal design system is being built out in our April and May iterations. This exact ticket and `code-server` is one of the first uses-cases I have in mind and am building for as an acceptance state for our design system.

KaKi87:
Hello, any news on this ? Thanks

jsjoeio:
@KaKi87 not yet. It's still lower priority, but feel free to chime in with any thoughts!

vapurrmaid:
I have an update: as of the September release of Coder, we've internally started on a design system and we've implemented theming in-product. While the design system is not fully fleshed out, it brings us one step closer to being able to achieve this ticket with the design system.

KaKi87:
> feel free to chime in with any thoughts around why this is important to you and what you'd like to see

[...]

PR #7118, opened 2024-12-14
style: add dark theme color scheme

qupig:
Partially fixes the issue mentioned in #7077.

Using vscode theme colors is proposed in #3019, that would be nice. But until it is implemented, it might still be helpful as a fallback for browser adaptation in the absence of user preferences.

The problem is that this PR only solves the color scheme of login page, error pages.
I haven't found out how to modify the vscode page, so the FOUC is still there.

It seems that vscode page needs to be modified in `lib/vscode`? I.e. is this also an upstream issue?

I tested https://vscode.dev and it seems to have the same FOUC issue.
- But it seems to only appear when opening first time or pressing Enter in the address bar.
- i.e. it doesn't seem to happen when refreshing the page. (But happens in this project)

Or should actually compare to `Codespace`(https://github.dev) , which actually has no FOUC issues, in my testing.

| Login page (Light)  | Login page (Dark) |
|---|---|
| login-light | login-dark |

| Error page (Light)  | Error page (Dark) |
|---|---|
| error-light | error-dark |

### DIFF
--- a/src/browser/pages/error.css
+++ b/src/browser/pages/error.css
@@ -10,6 +10,7 @@
 }
 
 .error-display > .body {
+  color: #444;
   color: light-dark(#444, #ccc);
   font-size: 1.2rem;
 }
--- a/src/browser/pages/error.css
+++ b/src/browser/pages/error.css
@@ -10,7 +10,7 @@
 }
 
 .error-display > .body {
-  color: #444;
+  color: light-dark(#444, #ccc);
   font-size: 1.2rem;
 }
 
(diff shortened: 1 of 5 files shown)


code-asher:
Interesting about Codespaces, they must add something specially to Codespaces to avoid the FOUC, instead of adding it to the OSS repository for some reason.

This is really cool!  I have not seen `light-dark` before.  Should we have fallback colors in case the browser does not support `light-dark`?  It seems fairly new.  Like:

```
color: #444
color: light-dark(#444, #ccc)
```

code-asher:
To expand a little on fixing it for VS Code, we serve an HTML file from VS Code so we have no control over it outside, so yeah it will have to be fixed in VS Code itself.

Well, I guess we could intercept it and modify it but it seems a bit janky to do it that way.

code-asher (approved):

code-asher:
Agreed with your comments!  Thanks again, this is great.

PR #7156, opened 2025-01-08
Update Alpine instructions

jpetazzo:
While trying to set up code-server on Alpine, I ran into build errors. To solve them, I had to add Kerberos development libs. Also, it looks like npm config doesn't work in recent version of Node and doesn't seem to be necessary. These instructions were tested in the Docker `node:20-alpine` image.

Fixes #


code-asher (approved):
Thank you!

PR #7165, opened 2025-01-14
doc: update upstream documentation urls

tuunit:
### Motivation

Outdated upstream documentation urls

### Description

Updating the url for the upstream documentation of Pomerium, OAuth2-Proxy and Cloudflare Access


code-asher (approved):
Sorry for the delay, and thank you for the fixes!  Just FYI, it looks like there is a more generic Cloudflare link without the `en-gb` so I figure I will use that one.

PR #7173, opened 2025-01-19
Support HTTP BasicAuth for authentication with auth-user argument, password or hashedPassword

### LINKED ISSUE #7142: Support HTTP BasicAuth as alternative authentication
## What is your suggestion?

Support HTTP BasicAuth for authenticating the user with a provided username and password

I have a reverse proxy setup where the authentication is handled at the proxy level, and a  generated or pre-created Basic Auth is used to communicate with the back-end server, which may be still over a network and so HTTPS + Basic Auth is preferred over no authentication.

None, no authentication does not pass security requirements and cookie-based (UI driven login) does not work in this context.

Yes

gaberudy:
Fixes #7142

Supersedes previous pull request #7143

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -12,6 +12,7 @@ export enum Feature {
 
 export enum AuthType {
   Password = "password",
+  HttpBasic = "http-basic",
   None = "none",
 }
 
@@ -65,6 +66,7 @@ export interface UserProvidedCodeArgs {
 export interface UserProvidedArgs extends UserProvidedCodeArgs {
   config?: string
   auth?: AuthType
+  "auth-user"?: string
   password?: string
   "hashed-password"?: string
   cert?: OptionalString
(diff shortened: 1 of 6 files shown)


Parkreiner (commented):
Thank you for the contribution! Ended up leaving two small comments, but I'm worried I don't have enough architectural understanding of the project to give more finer-grained feedback

@bcpeinhardt @code-asher Pinging you just so you're aware of this PR

code-asher:

tachibanayui:
I would like to continue this if @gaberudy is no longer interested. I want to intergrate external authentication without setting `auth=none`

code-asher:
@tachibanayui that would be awesome.

If you do tackle this, one thought I had: I wonder if instead of a separate `auth-user` flag that affects only basic auth, we can make it part of the password, with the same format as basic auth itself (so `username:password`).

tachibanayui:
Since this is intended for single user anyway, we can hard code the username `code-server` and use the password in the config file

code-asher:
Good point, I like the sound of that.

PR #7177, opened 2025-01-21
Add optional idle timeout shutdown 

### LINKED ISSUE #1636: Add `--idle-timeout` to shutdown code-server when there isn't an active connection
If there hasn't been an active connection to code-server in X minutes, we should exit with status 0.

This would replace the heartbeat file (#1115) as a more friendly option.

cc @code-asher @antofthy

gaberudy:
I did this for my needs but realized there is a closed issue (not implemented) that is requesting the same thing.

Through the idle-timeout arg or IDLE_TIMEOUT environment variable, after the specified number of minutes since the heartbeat was last refreshed is detected, the code-server process shuts down. When run as the single process of a docker container, this just stops the container.

Fixes #1636

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -83,6 +83,7 @@ export interface UserProvidedArgs extends UserProvidedCodeArgs {
   "socket-mode"?: string
   "trusted-origins"?: string[]
   version?: boolean
+  "idle-timeout"?: number
   "proxy-domain"?: string[]
   "reuse-window"?: boolean
   "new-window"?: boolean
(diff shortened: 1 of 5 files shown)


gskifstad:
hi! our team would love this feature as well. I see there is a stray merge conflict causing linter errors, but otherwise, is there anything blocking this PR?

code-asher:
Yeah happy to merge this in if the conflicts and below are addressed.  

I think the heart counting inactivity and directly exiting the process is more than it should be doing.  Maybe we could refactor so the heart has an `onChange` event that is called when the state changes to either `active` or `inactive`?  Then we can tie it all together in `entry.ts`.

```
// could propagate the event through `server` or maybe we could just pass the `heart` itself back
server.onChange((state) => {
  if (state === "active") {
    clearTimeout(timer)
  } else if (args["idle-timeout"]) {
    timer = setTimeout(() => {
      logger.warn(`Idle timeout of ${args["idle-timeout"]} minutes exceeded`)
      wrapper.exit();
    }, args["idle-timeout"] * 60 * 1000);
  }
})
```

Something like that.

I think if we have the flag as `idle-timeout` we might want to have it in seconds by default and allow `m` or `h` or `s` suffixes?  Or change the flag to `idle-timeout-minutes` or something.

One other thing, we should check the `parseInt` for `NaN`.

code-asher:

PR #7187, opened 2025-01-29
[Helm Chart] fix replicas and existingSecret. bump to 3.25.3

### LINKED ISSUE #7186: Helm Chart is not using the replicaCount value, and the values file doesn't mention the variable existingSecrets

### OS/Web Information

- N/A

### Steps to Reproduce

1. set replicaCount: 2
2. deploy the helm application
3. check the deployment, replicacount is 1

### Expected

replicaCount to be set as per the value

existingSecret to be better documented in the values file

### Actual

replicaCount is not honored
existingSecret is to be found in the helm templates instead of values.yaml

### Logs

```shell

```

### Screenshot/Video

_No response_

### Notes

_No response_

mickkael:
Fixes #7186


code-asher (approved):
Thank you!

PR #7192, opened 2025-02-01
Update serviceWorker.ts

samss084:
Fixes #


Parkreiner:
Hi, we appreciate you taking making a PR. Unfortunately, it seems to be incomplete, as there are syntax errors that would prevent the code from compiling.

If that was a mistake, please feel free to open a new PR with the full changes and a description of what they aim to fix.

PR #7196, opened 2025-02-06
Add back macOS amd64 release

SongXiaoXi:
As per the [Supported Runners and Hardware Resources](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) documentation, use the macos-13 workflow label to build the macOS amd64 package.


code-asher (approved):
Thank you!

PR #7235, opened 2025-02-27
Fix issue 7232

benz0li:
- Resolve https://github.com/coder/code-server/issues/7232

### DIFF
--- a/ci/release-image/Dockerfile
+++ b/ci/release-image/Dockerfile
@@ -31,7 +31,10 @@ RUN sed -i "s/# en_US.UTF-8/en_US.UTF-8/" /etc/locale.gen \
   && locale-gen
 ENV LANG=en_US.UTF-8
 
-RUN adduser --gecos '' --disabled-password coder \
+RUN if grep -q 1000 /etc/passwd; then \
+    userdel -r "$(id -un 1000)"; \
+  fi \
+  && adduser --gecos '' --disabled-password coder \
   && echo "coder ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers.d/nopasswd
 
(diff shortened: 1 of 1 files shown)


benz0li:
Ping @code-asher

code-asher (approved):
Thank you!  And sorry for the delay.

code-asher:
Oh wait I already merged but had a thought, since we are grepping for `1000`, is there any chance this matches extra things?  Like maybe there is `someuser:x:31000`.  Probably unlikely, but I wonder if we want to use `getent passwd 1000` or something instead.

Also I wonder, if we already have a user 1000, could we rename it instead of delete/add?

benz0li:
> Oh wait I already merged but had a thought, since we are grepping for `1000`, is there any chance this matches extra things? Like maybe there is `someuser:x:31000`. Probably unlikely, but I wonder if we want to use `getent passwd 1000` or something instead.
> 
> Also I wonder, if we already have a user 1000, could we rename it instead of delete/add?

Yes. But this works just fine for both the official [Debian](https://hub.docker.com/_/debian) and [Ubuntu](https://hub.docker.com/_/ubuntu) Docker Image.

benz0li:
> Another thing I have not checked, if someone is using a different UID (`-u 1002:1002` for example), maybe we should check for that instead of 1000. I forget exactly how the mapping works though, so maybe this is not necessary.

I have not checked that either.

code-asher:
> Yes. But this works just fine for both the official [Debian](https://hub.docker.com/_/debian) and [Ubuntu](https://hub.docker.com/_/ubuntu) Docker Image.

Fair.  We can always revisit if we need to.

PR #7284, opened 2025-03-27
Allow OPTIONS requests through domain proxy

### LINKED ISSUE #7283: Proxy should allow OPTIONS
### Steps to Reproduce

Send an OPTIONS request to your application through code-server's proxy.

### Expected

Should get a 200 or 204 back.

### Actual

Gets blocked with a 401 if you have `--auth=password`, because `OPTIONS` requests do not include credentials.

### Notes

We should return a 200 or 204 for OPTIONS requests without checking auth.

helgehatt:
Fixes #7283

I believe the fix is as simple as allowing OPTIONS requests to pass through the proxy without authentication. Then the service behind the proxy should respond with 200 or 204.

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -84,6 +84,7 @@ export interface UserProvidedArgs extends UserProvidedCodeArgs {
   "trusted-origins"?: string[]
   version?: boolean
   "proxy-domain"?: string[]
+  "skip-auth-preflight"?: boolean
   "reuse-window"?: boolean
   "new-window"?: boolean
   "ignore-last-opened"?: boolean
@@ -252,6 +253,7 @@ export const options: Options<Required<UserProvidedArgs>> = {
     description: "GitHub authentication token (can only be passed in via $GITHUB_TOKEN or the config file).",
   },
   "proxy-domain": { type: "string[]", description: "Domain used for proxying ports." },
+  "skip-auth-preflight": { type: 'boolean', description: "Allows preflight requests through proxy without authentication." },
   "ignore-last-opened": {
     type: "boolean",
     short: "e",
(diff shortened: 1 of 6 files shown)


code-asher:
This might be OK, but something about letting things through the proxy without auth is making me nervous.  Would it work if we just respond with a 204 here?

code-asher:
Also thank you for working on this!

code-asher:
And I suppose we will need to do the same for the path-based proxy.

code-asher:
Actually, I suppose we have to let it through because it could be a 404 or something else, no way to know for sure that particular preflight will actually be a 204.

Will approve once we add it to the path proxy as well!

helgehatt:
What do you think is the right direction here? A `skip-auth-preflight` flag is definitely an option, but hardcoding adds a layer of security by obscurity. However, hardcoding 204 kind of defeats the purpose of preflighting requests.

code-asher:
> hardcoding 204 kind of defeats the purpose of preflighting requests

Yeah, you are right, but the attack surface is large since we automatically forward every port, so it makes me nervous.  A hardcoded response would eliminate the attack vector, and would probably be good enough for development?

But, I think for now my vote would be to gate this behind a `skip-auth-preflight` flag set to `false` by default.

What do you think?

helgehatt:
We can start off with a boolean `skip-auth-preflight` flag which can later be extended with a list of ports or a hardcoded mode?

code-asher:
That sounds good to me!

helgehatt:
@code-asher Have a look and see if these changes make sense to you.

code-asher (approved):
Perfect thank you!

PR #7286, opened 2025-03-27
Create Dockerfile

Dhishu:
Fixes #


code-asher:
Thanks for the contribution!  Could you explain the problem this solves, or why we might want to add this in addition to our existing Dockerfiles?

PR #7327, opened 2025-04-26
Cookie Expiration Age Fix

### LINKED ISSUE #7301: Allow configuring login cookie with an expiry duration
## What is your suggestion?

Allow configuring an (optional) expiry age for the login cookie so that it persists beyond a browser session. This is handy for not re-entering the password every time.

**Possible change**:
https://github.com/coder/code-server/blob/main/src/node/http.ts#L328

```ts
...
return {
  domain: getCookieDomain(url.host, req.args["proxy-domain"]),
  path: normalize(url.pathname) || "/",
  sameSite: "lax",
  // Load, if provided, max age for login cookie. 0 means it becomes a Session cookie (according to Express docs)
  maxAge: getConfigCookieMaxAgeAsMilliseconds() || 0,
}
```

I want to maintain my login status even after closing the browser (as the Cookie is currently session-only). Say keep the login cookie for a week or a month. Adding a max-age changes the cookie from `Session` to persistent for the given duration using Express' `maxAge` parameter ([docs](https://expressjs.com/en/api.html#res.cookie)).

No, none directly. User can manually edit the cookie in Devtools

I'm not that familiar with how to make new configuration argument and also how the config args are read (there's some middleware for `args` IIRC, but not sure how to use that). But I can give it a go, with some help.

The above given change should be roughly all that's needed, sans the config reading.

jonathandunne:
Fixes #7301

### DIFF
--- a/src/node/http.ts
+++ b/src/node/http.ts
@@ -318,6 +318,15 @@ export const getCookieOptions = (req: express.Request): express.CookieOptions =>
   // When logging in or out the request must include the href (the full current
   // URL of that page) and the relative path to the root as given to it by the
   // backend.  Using these two we can determine the true absolute root.
+
+  
+  function getConfigCookieMaxAgeAsMilliseconds(req: express.Request): number {
+    // the CLI flag or YAML key should be defined as "auth.cookie-max-age"
+    const days = Number(req.args["auth.cookie-max-age"] || 0)
+    return days * 24 * 60 * 60 * 1000
+  }
+  
+
   const url = new URL(
     req.query.base || req.body?.base || "/",
     req.query.href || req.body?.href || "http://" + (req.headers.host || "localhost"),
@@ -326,6 +335,7 @@ export const getCookieOptions = (req: express.Request): express.CookieOptions =>
     domain: getCookieDomain(url.host, req.args["proxy-domain"]),
     path: normalize(url.pathname) || "/",
     sameSite: "lax",
+    maxAge: getConfigCookieMaxAgeAsMilliseconds(req) || 0,
   }
 }
 


code-asher (changes requested):
Thank you for the contributing!  Looks good to me so far.  We will have to add it to `src/node/cli.ts` as well.

The `.` syntax is intriguing, but I want to say we call it `auth-cookie-max-age` for now.

src/node/http.ts
```diff
@@ -326,6 +335,7 @@ export const getCookieOptions = (req: express.Request): express.CookieOptions =>
     domain: getCookieDomain(url.host, req.args["proxy-domain"]),
     path: normalize(url.pathname) || "/",
     sameSite: "lax",
+    maxAge: getConfigCookieMaxAgeAsMilliseconds(req) || 0,
```
code-asher, line 338: Oh wait is `0` the right default?  This will delete the cookie instantly, I think?  We may want `-1`.

code-asher: Hmm actually zero or negative immediately expires the cookie it seems: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#max-agenumber

Should it just be undefined?

nb-programmer: I'd read it in the Express JS docs setting it to 0 makes it session
https://expressjs.com/en/api.html#res.cookie
image

They gave this explanation to the `expires` parameter, but also said `maxAge` is a "Convenient option for setting the expiry time relative to the current time in milliseconds". They didn't specify what to put here to make it session, nor what the default is.

code-asher: Yeah I also thought maybe it was an Express thing, but when I checked the source it seems to just immediately expire.

If `maxAge` is zero, it passes `expires` set to now and `maxAge` set to zero into `cookie.serialize()`:

https://github.com/expressjs/express/blob/f9954dd317a1b534e62a5ae3aa7f52f3582b8881/lib/response.js#L764-L771

`cookie.serialize()` simply adds `Max-Age=0` to the cookie:

https://github.com/jshttp/cookie/blob/2a1a4d8b2679b208f354e848e711dc8471fb83af/src/index.ts#L268-L274

nb-programmer: > Yeah I also thought maybe it was an Express thing, but when I checked the source it seems to just immediately expire.
> 
> If `maxAge` is zero, it passes `expires` set to now and `maxAge` set to zero into `cookie.serialize()`:
> 
> https://github.com/expressjs/express/blob/f9954dd317a1b534e62a5ae3aa7f52f3582b8881/lib/response.js#L764-L771
> 
> `cookie.serialize()` simply adds `Max-Age=0` and `Expires=<the now date>` to the cookie:
> 
> https://github.com/jshttp/cookie/blob/2a1a4d8b2679b208f354e848e711dc8471fb83af/src/index.ts#L268-L274
> 
> Now, if `expires` is zero (although this is actually a type error), and there is no max age it does indeed skip setting `Expires` in that case, which would result in a session cookie. The difference is they do `if (options.maxAge !== undefined)` versus `if (options.expires)`, so zero is not ignored in the max age case.

Hmm, what would be a good way to ensure it is properly a session cookie? Sending `null` seems illogical, but would work, or we can just skip the field if there is no config for it.

code-asher: Yeah I think something like `maxAge: getConfigCookieMaxAgeAsMilliseconds(req) ?? undefined` would do the trick.

code-asher: Ah wait `getConfigCookieMaxAgeAsMilliseconds` always returns a number, so we would have to do it in there.  We could return `number | undefined`.

code-asher:

PR #7328, opened 2025-04-26
Add suggestion for using tailscale to guide.md

jrmlhermitte:
Add a suggestion for using tailscale for code-server. This alternative is useful for those that want more options for connecting their IPads to their code server.
I'm not sure if it's acceptable to refer to a commercial option, but figured I would send this suggestion anyway in case it helps (feel free to decline).

Given this is just a documentation suggestion, I am not going to create an issue.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,19 @@ Code v99.99.999
 
 ## Unreleased
 
+## [4.99.3](https://github.com/coder/code-server/releases/tag/v4.99.3) - 2025-04-17
+
+Code v1.99.3
+
+### Added
+
+- Added `--skip-auth-preflight` flag to let preflight requests through the
+  proxy.
+
+### Changed
+
+- Update to Code 1.99.3.
+
 ## [4.99.2](https://github.com/coder/code-server/releases/tag/v4.99.2) - 2025-04-10
 
 Code v1.99.2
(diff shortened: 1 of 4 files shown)


code-asher (commented):
Thank you for the contribution!  Some thoughts:

1. Could we move it to the bottom of the section?  I would prefer to prioritize the other headings.
2. Does it need to be iPad-specific?  It should work for any client with Tailscale on it, right?  The advantage to me seems more about the server admin not having to register their own domain.
3. We should mention the tunnel uses HTTP by default and some features may not work, we already get a fair number of issues from folks not using HTTPS so it would be unfortunate if we had official documentation that suggests a partially broken setup without mentioning it.
4. Maybe we generalize this to something about using third party tunnels?  I know some people have used ngrok.

What about something like this (feel free to reword):

```
### Using third-party tunnels

Using a third-party tunnel service like [Tailscale](http://tailscale.com) or
[ngrok](https://ngrok.com/) allows you to connect to code-server securely
without needing to set up an external domain.

Note that depending on the service, the tunnel may use HTTP by default and some
code-server features will not work as they require a secure context (meaning
HTTPS or localhost), in which case you may still need to set up a certificate
and reverse proxy.  
```

If this feels too general, we can add more links to Tailscale's documentation, this is mostly a starting point.

Also tangentially related but I just found https://github.com/juanfont/headscale.  Never used it, but cool that it can be self-hosted.

jrmlhermitte:
Sorry I have not had time to get around to this. I think given tailscale has good updated docs [here](https://tailscale.com/kb/1166/vscode-ipad), we can possibly close this. The different solutions will always change over time and should be easily accessible so perhaps less is more. Thanks for the feedback, and https://github.com/juanfont/headscale is pretty interesting thanks!

Closing

PR #7337, opened 2025-05-02
fix: parse port in path proxy

kylecarbs:


sreya (approved):

code-asher (approved):

code-asher:
Ah wait does `parseInt` throw?  I think it just returns `NaN`.  Still technically fixes the original issue though :laughing: so maybe I can just update it in the next release.

Parkreiner:
@code-asher Yeah, you're right – a lot of the JS math functions just result in `NaN`

PR #7339, opened 2025-05-05
Allow configuring trusted domains at run-time

code-asher:
Also fixed a long-standing issue with repeatable flags in the config.


src/node/cli.ts
```diff
@@ -707,12 +713,16 @@ export function parseConfigFile(configFile: string, configPath: string): ConfigA
...
-    return `--${optName}=${opt}`
-  })
+  const configFileArgv = Object.entries(config)
+    .map(([optName, opt]) => {
+      if (opt === true) {
+        return `--${optName}`
+      } else if (Array.isArray(opt)) {
+        return opt.map((o) => `--${optName}=${o}`)
+      }
+      return `--${optName}=${opt}`
```
bdsoha, line 723:
```suggestion
      if (opt === true) {
        return `--${optName}`
      } 
      if (Array.isArray(opt)) {
        return opt.map((o) => `--${optName}=${o}`)
      }
      return `--${optName}=${opt}`
```

code-asher: Technically equivalent, true, but stylistically I like having one less line :stuck_out_tongue:

PR #7342, opened 2025-05-06
Adding a mention of code-server devcontainer feature

bartekgatzcoder:
The change is in "getting started" section and lists using the devcontainer feature of code-server (a separate GH project) as a potential method of using code-server in one's projects.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,7 +24,7 @@ on how to set up a Google VM on which you can install code-server.
 
 ## Getting started
 
-There are four ways to get started:
+There are five ways to get started:
 
 1. Using the [install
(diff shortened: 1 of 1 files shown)


code-asher:
Nice!  It looks like the `#code-server` header was removed though, should we just change the link `https://github.com/coder/devcontainer-features`?  Or maybe we should add the heading back.

code-asher:
Or we could link to `https://github.com/coder/devcontainer-features/blob/main/src/code-server/README.md`?

bartekgatzcoder:
@code-asher good catch. Please check now.

code-asher (approved):

PR #7366, opened 2025-06-03
Fix clipboard access issue for VIM extension in safari

li-yechao:
I typically use code-server in Safari with the VIM extension, but I've encountered issues with clipboard functionality due to the underlying VSCode architecture. After some exploration, I found a way to improve the clipboard experience in Safari: when certain events that can read the clipboard (such as pressing 'p' or 'ctrl+v') are triggered, the clipboard content is read into memory, and subsequent read requests within a short time window will return the cached content instead of reading from the clipboard

### DIFF
--- a/patches/clipboard.diff
+++ b/patches/clipboard.diff
@@ -137,18 +137,20 @@ Index: code-server/lib/vscode/src/vs/workbench/services/clipboard/browser/clipbo
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/workbench/services/clipboard/browser/clipboardService.ts
 +++ code-server/lib/vscode/src/vs/workbench/services/clipboard/browser/clipboardService.ts
-@@ -15,19 +15,36 @@ import { IWorkbenchEnvironmentService }
+@@ -15,19 +15,38 @@ import { IWorkbenchEnvironmentService }
  import { ILogService } from '../../../../platform/log/common/log.js';
  import { ILayoutService } from '../../../../platform/layout/browser/layoutService.js';
  import { getActiveWindow } from '../../../../base/browser/dom.js';
 +import { isSafari } from '../../../../base/browser/browser.js';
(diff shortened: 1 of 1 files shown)


li-yechao:
Although this patch might seem a bit like a hack, it works surprisingly well in Safari.

code-asher:
Thank you for the patch!  Could this fix be sent upstream to https://github.com/microsoft/vscode?  At a glance it seems like something that should be fixed upstream to me.

The VS Code folks can probably give a better review than I can, but my gut says that specifically reacting to Vim keybindings in the core might not be the right implementation.

li-yechao:
@code-asher As you mentioned, I don’t think VSCode would accept this patch. Given that WebKit (the core of Safari) has no plans to open the clipboard API, and considering the limitations of both VSCode's architecture and WebKit, supporting this in a VSCode extension isn't feasible. For now, the only viable solution is to modify the VSCode core via a patch.

li-yechao:
Even if this PR doesn’t get merged, that’s perfectly fine—I totally understand. While this patch works, it is indeed quite a hack. My intention is simply to provide this solution for anyone who might find it useful.

code-asher:
That makes sense!  Yeah I think it is not something we should merge so I will close but we can use this as an example if anyone else has issues.

Speaking of, what is the issue exactly?  Is it that sometimes pasting does not work, specifically when pasting in rapid succession?

li-yechao:
> Speaking of, what is the issue exactly? Is it that sometimes pasting does not work, specifically when pasting in rapid succession?

When you install VIM extension and enable ` "vim.useSystemClipboard": true`, `p` (paste function) in vim mode will be unavailable (in safari), because vscode's clipboard API calls `navigator.clipboard.readText()`, which must be called in events directly triggered by the user in webkit-based browsers, and vscode calls this API in the request of VIM Extension.

PR #7420, opened 2025-07-14
chore(release) Update build-standalone-release.sh

lemanschik:
Remove shell hacks use native node -p flag

Fixes #


code-asher (approved):
Thank you, that is much better!

PR #7479, opened 2025-09-06
Fix Open VSX marketplace URL gallery path issue

jinvien:
Description
This PR fixes a URL path issue in the Open VSX marketplace integration that was causing incorrect redirects.

Problem
vscode extension gallery was generating incorrect URLs with an extra /vscode segment:

❌ Incorrect: https://open-vsx.org/vscode/gallery/vscode/eamodio/gitlens/latest
✅ Correct: https://open-vsx.org/vscode/gallery/eamodio/gitlens/latest
The incorrect URL format was causing redirects to http://open-vsx.org/error instead of properly resolving extension resources.

Changes
Fixed the extension latest version URI template in extensionGalleryManifestService.ts by removing the redundant /vscode segment

### DIFF
--- a/patches/marketplace.diff
+++ b/patches/marketplace.diff
@@ -19,7 +19,7 @@ Index: code-server/lib/vscode/src/vs/platform/product/common/product.ts
 ===================================================================
 --- code-server.orig/lib/vscode/src/vs/platform/product/common/product.ts
 +++ code-server/lib/vscode/src/vs/platform/product/common/product.ts
-@@ -49,6 +49,16 @@ else if (globalThis._VSCODE_PRODUCT_JSON
+@@ -49,6 +49,17 @@ else if (globalThis._VSCODE_PRODUCT_JSON
  			version: pkg.version
  		});
  	}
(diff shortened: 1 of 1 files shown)


code-asher:
Ohhhhhhh interesting is this why extension installations sometimes fail??

But, I am not sure if we can merge this because there are other marketplaces (such as `code-marketplace`) that *do* include the `/vscode` so this would break those.  

Also, what about other VS Code-based alternatives like VSCodium?  I imagine they must have the same bug.  Maybe this needs to be fixed on Open VSX's side?

If not, then maybe we will need to pull this off `extensionsGallery` instead so it can be customized.

code-asher:
Actually, https://github.com/eclipse/openvsx/issues/1283 suggests we can set `extensionUrlTemplate` on `extensionsGallery` to fix this.  Looks like VSCodium works because it sets this, so we can do the same (in this same patch).

jinvien:
Yes, this is clearly a more elegant solution. I've updated the commit accordingly—please take a look when you have a moment.

Also, I’m not entirely sure whether my commit should include changes to changelog.md. I noticed in your contribution guidelines that updates related to the marketplace should be reflected in the changelog. Could you please advise on whether that’s needed here?

Thanks!

code-asher:
Awesome thank you!

> I’m not entirely sure whether my commit should include changes to changelog.md

Yup, I think it would be good to call out this fix in the change log.  No worries about it generally though, before releasing I always go through the commits and add anything that is missing to the change log.

code-asher (approved):

PR #7596, opened 2025-12-10
Adding possibility to include Authorization header when using externa…

### LINKED ISSUE #7595: Allow Authorization header to be set when working with a custom extension marketplace
## What is your suggestion?
It would be great to have an option to set the Authorization header when working with the custom extension marketplaces. The token could be defined in the ENV EXTENSIONS_GALLERY where we set the custom marketplace URL.
That would allow us to add a security layer to the custom marketplace
No
Yes

vmladenovic:
…l extension marketplace

Fixes #7595

### DIFF
--- a/patches/marketplace.diff
+++ b/patches/marketplace.diff
@@ -90,3 +90,18 @@ Index: code-server/lib/vscode/src/vs/platform/extensionResourceLoader/common/ext
  	}
  
  }
+Index: code-server/lib/vscode/src/vs/platform/externalServices/common/marketplace.ts
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/platform/externalServices/common/marketplace.ts
++++ code-server/lib/vscode/src/vs/platform/externalServices/common/marketplace.ts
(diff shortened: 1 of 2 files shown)


vmladenovic:
> Nice I like this! I think there might be some types that also need to be updated, but CI should tell us more once it runs.
> 
> I wonder if we can be more descriptive than `token`. Maybe `authorizationHeaderToken`? Or...hmm should we make it a general `headers` property maybe?? Then users could add whatever headers they need.

I guess we could be more descriptive and use `authorizationHeaderToken`. My idea was to keep it really simple so I wouldn't go with the general "header" property for now if it's OK with you?

code-asher (approved):
I have this suspicion that this is one of those things that ends up being replaced by a generic solution and then we are stuck maintaining the old thing forever hahaha but this is tiny enough that it should be no problem to deprecate in the future if necessary.  We have no idea if we really need a generic solution anyway, maybe this will be the only header we need.

Thanks again!

PR #7626, opened 2026-01-07
Create Dockerfile

AliVSCode:
Fixes #


code-asher:
Could you expand on what problem this is solving?  We already have docker files and I am not seeing how health-server.js is meant to be used.

PR #7646, opened 2026-01-24
docs: clarify error handlers are final and don't call next()

taosin:
## Summary
- Add doc comments to `errorHandler` and `wsErrorHandler` clarifying they are final handlers and intentionally do not call `next()`.

## Rationale
- Keeps behavior documentation consistent between HTTP and WebSocket error handlers.
- Makes intent visible at the function level (more discoverable than an inline comment).

## Test plan
- Not applicable (comment-only change).


taosin:
> Thanks for the PR! I am not convinced we need the comment, to me the `end` is self-explanatory.
> 
> We also make no such comments for why the non-websocket error handler also does not call `next()`, so it seems inconsistent.
> 
> I think it would make more sense as a comment on `wsErrorHandler` and `errorHandler` themselves though, because those would show up in the documentation when calling the functions, where `end` would not be visible and thus less self-explanatory. Something like:
> 
> > This handler does not call `next` even though it accepts it as an argument.  It should only be used as the final handler for a route.

Thanks for the review!

Good point — I agree the inline comment near `end()` is a bit noisy and also inconsistent with `errorHandler`.
I’ve moved the explanation to a doc comment on both `errorHandler` and `wsErrorHandler` to keep them consistent and make it visible where the functions are referenced, and removed the inline comment.

Issue #7688, opened 2026-03-02
--app-name should set the page title


code-asher:
Will need to add a variable to the HTML files to set the title to the app name (`login.html`, `error.html`).

And will need to patch for the VS Code HTML (`workbench.html`, `workbench-dev.html`) although not entirely sure how the titles there are set.

MicahZoltu:
Would this override the whole page title (my personal preference) or just the suffix (I suspect the path of least resistance)?  This is what my browser tab bar looks like with 3 different projects open:

<img width="437" height="46" alt="Image" src="https://github.com/user-attachments/assets/3201a58e-a90b-45f0-baba-2ccb2dd71845" />

The actual titles of each are:
```
README.md - workspace - code-server
README.md - workspace - code-server
README.md - workspace - code-server
```

Due to limited horizontal space, the `code-server` portion of the name (presumably what would be changed with `--app-name` if something more isn't done) isn't actually visible.

What I would like to see by setting `--app-name` environment variable on each instance would be:

<img width="429" height="43" alt="Image" src="https://github.com/user-attachments/assets/27cf0fe3-2c88-41df-beab-52174c699030" />

Alternatively, perhaps just naming like the following would still leave the extra info but make `--app-name` more meaningful:
```
<--app-name> - workspace - README.md
```
This way, when truncated the `--app-name` takes priority but you can still see the filename and opened directory if you have vertical tabs or mouse-over.

code-asher:
Good point, if we do just the suffix we might get away with not even having to patch anything, but if we want to change the order we will have to patch for sure.  

I am not sure if we should make that change here though, it sounds useful enough to go into VS Code core, if they are willing to accept such a patch.  Generally we want to avoid drifting from VS Code too far, especially for things that are not bugfixes.

MicahZoltu:
While the ordering isn't ideal in VSCode, it appears that github.dev already has "fixed" this problem to some extent.  It seems they somehow name the workspace root, which then is reflected in the title bar.  Is this something that can be done with code-server?

Since I'm using a different dockerized instance for each repository, my workspace root name is the same for each (`workspace`) because that is where I mount the volume.  I assumed that meant I was stuck with a shared name for every workspace but I see now that github.dev has a name that I don't think is a valid path, suggesting there may be some way to control the workspace name.  `MyRepository [Codespaces: Fantastic Space Broccoli]` is my workspace name which does show up in the title bar, before the `--app-name` would be.

[...]

PR #7693, opened 2026-03-04
feat: use --app-name in error page title

lustsazeus-lab:
This PR fixes partially addresses #7688 - when using --app-name flag, the error pages should display the custom app name instead of hardcoded "code-server".

## Changes
- Updated error.html to use {{APP_NAME}} template variable instead of hardcoded "code-server"
- Updated errors.ts to replace {{APP_NAME}} with the value from --app-name flag (defaults to "code-server")
- Added unit tests for the app-name feature

## Testing
- TypeScript compiles without errors
- Unit tests added for custom app-name and default value

Partially addresses #7688

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -519,6 +519,7 @@ export interface DefaultedArgs extends ConfigArgs {
   "extensions-dir": string
   "user-data-dir": string
   "session-socket": string
+  "app-name": string
   /* Positional arguments. */
   _: string[]
 }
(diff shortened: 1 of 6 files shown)


code-asher (approved):
Thank you for tackling this!  We might want to use the i18n strings for this eventually (like the login title), but this is definitely a step in the right direction.

FYI I modified your PR description from "fixes" and "closes" to "partially addresses" because to fully fix that issue we also need to set the VS Code titles (I think?  I need to test actually.  **Edit** yeah app name does not currently affect VS Code titles.)

PR #7725, opened 2026-03-29
fix: apply --app-name to web page titles

jonathandunne:
## Summary
- apply `--app-name` to VS Code web bootstrap product configuration
- set `nameShort`/`nameLong` so `${appName}` in tab titles uses the configured app name
- add a focused route test covering the injected configuration

## Testing
- attempted `npm install` (partially completed but failed in `lib/vscode` on missing `gssapi/gssapi.h` for `kerberos` under Node 24)
- attempted `npm run test:unit -- --runInBand --runTestsByPath test/unit/node/routes/vscode.test.ts test/unit/node/routes/login.test.ts`
- test run was blocked by a pre-existing TypeScript error in `src/node/http.ts` (`auth.cookie-max-age` missing on `DefaultedArgs`) before these route tests executed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,19 @@ Code v99.99.999
 
 ## Unreleased
 
+## [4.99.3](https://github.com/coder/code-server/releases/tag/v4.99.3) - 2025-04-17
+
+Code v1.99.3
+
+### Added
+
+- Added `--skip-auth-preflight` flag to let preflight requests through the
+  proxy.
+
+### Changed
+
+- Update to Code 1.99.3.
+
 ## [4.99.2](https://github.com/coder/code-server/releases/tag/v4.99.2) - 2025-04-10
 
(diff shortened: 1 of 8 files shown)


code-asher:
Thank you for figuring this out!  Sorry it has taken me a while to get to this, I should be able to properly test and review this week but some initial comments:

- Could we remove the Helm, doc, and cookie changes?  They seem unrelated to the core fix.  (Although maybe some of this is just the result of a conflict and a rebase will solve it.)  I do see the comment about the test run being blocked, let me look into that soon.
- I feel like it would make a lot of sense to add this to the `integration.diff` patch instead of a separate one.  What do you think?
- More of a note to myself, does this affect the patch we make to `aboutDetail`?  Wonder if it will show code-server twice now or something.

code-asher:
Superseded by https://github.com/coder/code-server/pull/7794

PR #7787, opened 2026-05-07
Update Code to 1.119.0

benz0li:


code-asher (approved):
Thank you!

code-asher:
Just FYI I pushed up a workflow that checks for VS Code updates every hour (unfortunately there seem to be no cross-repo release events so I have to poll) so this part should be automated now!

PR #7792, opened 2026-05-11
Add comments explaining what entrypoint is supposed to do

stuartpb:
This way the `DOCKER_USER` variable won't accidentally retain sudoer permissions for any user not named `coder` (almost certainly a security hole).

### DIFF
--- a/ci/release-image/entrypoint.sh
+++ b/ci/release-image/entrypoint.sh
@@ -5,15 +5,23 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ]; then
+# Rename the `coder` user if a new name is specified by the `DOCKER_USER` environment variable
+if [ "${DOCKER_USER-}" != "coder" ]; then
   USER="$DOCKER_USER"
+
+  # If the named user is not already present in the system
   if [ -z "$(id -u "$DOCKER_USER" 2>/dev/null)" ]; then
+
+    # Add a line adding the new user to sudoers without password prompting
     echo "$DOCKER_USER ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/nopasswd > /dev/null
+
+    # Rename the `coder` user and group
     # Unfortunately we cannot change $HOME as we cannot move any bind mounts
     # nor can we bind mount $HOME into a new home as that requires a privileged container.
     sudo usermod --login "$DOCKER_USER" coder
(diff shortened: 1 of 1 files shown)


code-asher:
How would this be reproduced?  We create the `coder` user as part of the image so it will always be the right user to remove: https://github.com/coder/code-server/blob/53d981a724a55e90f980dd8b147ba0028a956c24/ci/release-image/Dockerfile#L37-L38

If the user is not `coder` somehow, then these previous lines trying to rename it will fail anyway.

https://github.com/coder/code-server/blob/53d981a724a55e90f980dd8b147ba0028a956c24/ci/release-image/entrypoint.sh#L14-L15

stuartpb:
> We create the `coder` user as part of the image so it will always be the right user to remove

Okay, so now imagine the crazy hypothetical scenario where *literally any part of that very specific scaffolding for this specific fork of the codebase provided by the outside environment changes, for any reason*, like if, I dunno, *the name of the company changed and you missed one or two extant references to it*.

stuartpb:
Or rather, sorry, I guess I just don't understand how any of these variables are expected to work because there are *zero usable comments* here. The idea is that the line we're deleting *isn't* supposed to be deleting the user named by `DOCKER_USER`? How am I supposed to understand what this code's supposed to be doing?

stuartpb:
Here, I've updated the PR, here's a version where you can tell what cases would and wouldn't have bugs because the code *actually describes what it's trying to do*.

ci/release-image/entrypoint.sh
```diff
@@ -5,15 +5,23 @@ set -eu
 # Otherwise the current container UID may not exist in the passwd database.
 eval "$(fixuid -q)"
 
-if [ "${DOCKER_USER-}" ]; then
+# Rename the `coder` user if a new name is specified by the `DOCKER_USER` environment variable
+if [ "${DOCKER_USER-}" != "coder" ]; then
```
code-asher, line 9: This will now run when `DOCKER_USER` is blank where before it would not, should probably be something like 

```bash
if [ "${DOCKER_USER-}" ] && [ "$DOCKER_USER" != coder ]; then
```

Although, since we check if the user already exists below, it is probably not necessary to add this guard, but I suppose no harm in it either.

code-asher (commented):
Thanks!

> Okay, so now imagine the crazy hypothetical scenario

Yeah it would probably make sense for us to use the `USER` env var instead of hardcoding `coder` in case we ever change the user in the docker image and miss updating the entrypoint with the new name.

PR #7815, opened 2026-05-23
fix: install git extension fs-copyfile dependency

### LINKED ISSUE #7806: Source Control is stuck in 'scanning folder for git repository', fix found

### OS/Web Information

- Web Browser: Chrome
- Local OS: Macos
- Remote OS: Debian 12
- Remote Architecture: x86
- `code-server --version`: 4.117.0 and more

### Steps to Reproduce

I have been using version 4.100.3 all along, and today I tried to update to version 4.117.0 (npm install) code-server@4.117.0  --unsafe-perm ）， After the update, it was found that the Source Control has been stuck in "scanning folder for git repository".

### Expected

Source Control should show something

### Actual

I searched for @ builtin git in the extension and found that the git extension is not active.

I found the relevant logs in the 'Issues' window:
```
2026-05-21 01:11:21.601 [error] Activating extension vscode.git failed due to an error:
2026-05-21 01:11:21.601 [error] Error: Cannot find module '@vscode/fs-copyfile'
```

I also tried switching between 4.116.0 and 4.117.0.
I am trying to install `npm install @vscode/fs-copyfile` and reload window, and in the end it was fixed.

So was `@vscode/fs-copyfile` lost in the default installation?

### Logs

```shell

```

### Screenshot/Video

_No response_

Yes, this is also broken in VS Code web

### Notes

_No response_

cyphercodes:
Fixes #7806

## Summary
- Hoist the VS Code git extension's `@vscode/fs-copyfile` runtime dependency into the shared `lib/vscode/extensions` install manifest used by code-server's npm postinstall.
- Generate the extensions shrinkwrap from a temporary patched manifest so the released npm package installs `@vscode/fs-copyfile` deterministically without mutating the VS Code submodule package files.
- Add script coverage for the dependency-hoisting helper.

## Test plan
- `./ci/dev/lint-scripts.sh`
- `npm run test:scripts`
- Temporary install/resolve smoke test: generated the extensions package lock with `@vscode/fs-copyfile`, installed production dependencies, and verified `require.resolve("@vscode/fs-copyfile", { paths: ["<tmp>/git/dist"] })` resolves from `extensions/node_modules`.
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/.github/workflows/scripts.yaml
+++ b/.github/workflows/scripts.yaml
@@ -44,7 +44,7 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
       - name: Install test utilities
-        run: apk add bats checkbashisms
+        run: apk add bats checkbashisms jq
 
       - name: Check Bashisms
         run: checkbashisms ./install.sh
--- a/ci/build/add-vscode-extension-dependencies.sh
+++ b/ci/build/add-vscode-extension-dependencies.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+main() {
+  if [[ $# -ne 2 ]]; then
+    echo "Usage: $0 <extensions package.json> <VS Code extensions source dir>" >&2
+    exit 1
+  fi
+
+  local package_json="$1"
+  local vscode_extensions_src_path="$2"
+
+  # VS Code's optimizer keeps a runtime require("@vscode/fs-copyfile") in the
+  # built git extension but strips dependencies from the built extension
+  # package.json. code-server's npm postinstall installs from lib/vscode/extensions,
+  # so hoist this runtime dependency into the shared extension manifest.
+  add_dependency_from_extension_package \
+    "$package_json" \
+    "$vscode_extensions_src_path/git/package.json" \
+    "@vscode/fs-copyfile"
+}
+
+add_dependency_from_extension_package() {
+  local package_json="$1"
+  local source_package_json="$2"
+  local dependency="$3"
+
+  if [[ ! -f $package_json ]]; then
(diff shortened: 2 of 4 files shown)


code-asher (commented):
Awesome thank you for the PR!

I have a few worries, mainly that if the VS Code team ever adds new dependencies, or removes this one, or adds extensions with conflicting dependencies, we will probably not notice those until the npm module breaks again.  And also since we are not referencing the extension's lockfile we might install a different version leading to subtly different behavior between the GitHub and npm releases.

There are a few ways we could resolve these issues, but what do you think about this strategy:

1. Patch `lib/vscode/build/lib/extensions.ts` to emit the external modules in the package.json's dependencies.  I think probably the easiest way to do this is to return `packagedDependencies` from `fromLocalEsbuild` (in addition to the stream), and then in `fromLocal` we set `data.dependencies = packagedDependencies`.
2. Before we rsync the build, walk through each extension and generate a shrinkwrap for every extension with `node_modules` (or we could also do it as part of the same patch to `extensions.ts`, just would need to figure out the gulp syntax to do that).
3. Modify our postinstall to walk through each extension and install if necessary (I think the VS Code postinstall used to do this, if it still does we can skip this step).

So each extension with external dependencies will have their own shrinkwrap with its unique dependencies which should make the npm install match the GitHub release perfectly.

cyphercodes:
Pushed a small CI follow-up for the script-unit failure: the new script/tests call `jq`, but the Alpine script-test job only installed `bats` and `checkbashisms`. The workflow now installs `jq` too.

Local verification:
- `docker run --rm -v "$PWD:/work" -w /work alpine:3.17 sh -lc 'apk add bats checkbashisms jq >/dev/null && checkbashisms ./install.sh && ./ci/dev/test-scripts.sh'` ✅
- `git diff --check` ✅

PR #7818, opened 2026-05-25
Make app-name configurable via environment variable.

MicahZoltu:
Related to #7794 and continues fix to #7688.

I realized that CLI option is really inconvenient for my current setup, which is a general purpose code-server docker container that I can launch from any source code folder on my computer such that it is fully isolated via docker, but also retains my configuration settings.  It has an entrypoint that sets the workspace, but the CLI doesn't let you pass additional arguments after the workspace so I can't just add `--app-name $APP_NAME`.

By making it an environment variable, I can simply provide the environment variable to docker run, and it will use that if `--app-name` isn't provided.

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -677,9 +677,7 @@ export async function setDefaults(cliArgs: UserProvidedArgs, configArgs?: Config
   }
   args["proxy-domain"] = finalProxies
 
-  if (!args["app-name"]) {
-    args["app-name"] = "code-server"
-  }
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Ooo love to see `??=`.

MicahZoltu:
~~Unfortunately it appears I did something wrong here.  This results in the `${appName}` variable containing the literal string `$APP_NAME`.~~  😖

Nevermind, this was just me having some cruft lying around from back when I was testing things.

PR #7820, opened 2026-05-26
docs: fix 7 broken relative links in CONTRIBUTING.md

nickvigilante:
Relative paths like \`../test/unit\` work correctly on GitHub but produce 404s when the page is rendered at \`coder.com/docs/code-server/CONTRIBUTING\`. The docs site treats the page URL as a directory, so relative links resolve into \`/docs/code-server/...\` paths that don't exist.

Replace all 7 affected links with absolute \`github.com/coder/code-server/tree/main/...\` URLs.

Closes https://linear.app/codercom/issue/DOCS-244

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -201,7 +201,7 @@ There are four kinds of tests in code-server:
 Our unit tests are written in TypeScript and run using
 [Jest](https://jestjs.io/), the testing framework].
 
-These live under [test/unit](../test/unit).
+These live under [test/unit](https://github.com/coder/code-server/tree/main/test/unit).
 
 We use unit tests for functions and things that can be tested in isolation. The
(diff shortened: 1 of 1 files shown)


code-asher (commented):
Do you know if there is any way we can fix this on the docs site?  Something like if the link is relative, resolve it against the repo.

Just because now it will be more difficult to jump to the other files from within an IDE, especially since they are contributing docs (it is nice to reference them within the IDE while editing without getting kicked out to a browser).

nickvigilante:
Ah, I see your point. The tricky bit will be making sure we strictly only do relative links when linking to other Markdown files, or we use absolute links for docs. I'll do some investigation on this and get back to you!

nickvigilante:
@code-asher Closing this, since I'm working on a way to resolve this at the website level.

nickvigilante:
Fix in place and working. Closing.

PR #7842, opened 2026-06-10
fix: preserve path and query in openExternal URL rewrite

### LINKED ISSUE #7668: localhost openExternal rewrite loses URL /path

### OS/Web Information

- Web Browser: chrome
- Local OS: macOS
- Remote OS: Linux/Ubuntu
- Remote Architecture: amd64
- `code-server --version`: 1.107.0

### Steps to Reproduce

1.  `/usr/lib/code-server/lib/vscode/bin/helpers/browser.sh http://127.0.0.1:1234/my/path`
2. Opens URL `https://coder.<domain>/@<user>/<instance>/apps/code-server/proxy/1234/`
3. Be saddened by the observation that `/my/path` was dropped during the rewrite.

### Expected

Should have rewritten to URL `https://coder.<domain>/@<user>/<instance>/apps/code-server/proxy/1234/my/path` (with trailing `/my/path`).  When I update the URL in my browser to do this, this works - so the proxy itself is doing all the right things, it's just the rewrite that is incorrect afaics.

I have not explored with query params or anchors.  Given what I see in the code, I suspect these are lost too.

### Actual

Opened URL `https://coder.<domain>/@<user>/<instance>/apps/code-server/proxy/1234/`

### Logs

```shell

```

### Screenshot/Video

_No response_

This cannot be tested in native VS Code

I did not test VS Code web

I did not test GitHub Codespaces

- [x] I am using a secure context.

### Notes

I believe the relevant code is part of this code-server patch, and not vscode:
https://github.com/coder/code-server/blob/main/patches/proxy-uri.diff

From a squint at this patch, I think the `new URL(renderedTemplate, window.location.href)` call in `resolveExternalUri` is suspect.  I don't see anything here that _would_ copy through anything after the host portion of the URL.

1795771535y-cell:
## Description

Fixes #7668: The 
esolveExternalUri function in the proxy-uri patch was discarding the original URI path and query parameters when rewriting localhost URLs to go through the code-server proxy.

### Before
`http://127.0.0.1:1234/my/path?q=1` → `/proxy/1234/`

### After  
`http://127.0.0.1:1234/my/path?q=1` → `/proxy/1234/my/path?q=1`

## Changes
- Store the constructed proxy URL in a variable first
- Append the original URI path (with leading `/` stripped) to the proxy URL pathname
- Preserve the original URI query string by assigning it to the proxy URL search property

## Testing
The fix preserves the path and query parameters from localhost URLs when they are rewritten through the code-server proxy endpoint template.

### DIFF
--- a/patches/proxy-uri.diff
+++ b/patches/proxy-uri.diff
@@ -27,7 +27,7 @@ http.server` and it should show a notification and show up in the ports panel
 using the /proxy/port.
 
 Index: code-server/lib/vscode/src/vs/base/common/product.ts
-===================================================================
+====================================================================
 --- code-server.orig/lib/vscode/src/vs/base/common/product.ts
 +++ code-server/lib/vscode/src/vs/base/common/product.ts
 @@ -69,6 +69,7 @@ export interface IProductConfiguration {
@@ -54,7 +54,7 @@ Index: code-server/lib/vscode/src/vs/platform/remote/browser/remoteAuthorityReso
 @@ -86,9 +86,14 @@ export class RemoteAuthorityResolverServ
  		const connectionToken = await Promise.resolve(this._connectionTokens.get(authority) || this._connectionToken);
  		performance.mark(`code/didResolveConnectionToken/${authorityPrefix}`);
- 		this._logService.info(`Resolved connection token (${authorityPrefix}) after ${sw.elapsed()} ms`);
+ 		this._blogService.info(`Resolved connection token (${authorityPrefix}) after ${sw.elapsed()} ms`);
 +		let options: ResolvedOptions | undefined;
 +		if (this.productService.proxyEndpointTemplate) {
 +			const proxyUrl = new URL(this.productService.proxyEndpointTemplate, mainWindow.location.href);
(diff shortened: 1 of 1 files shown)


patches/proxy-uri.diff
```diff
@@ -116,46 +116,54 @@ Index: code-server/lib/vscode/src/vs/code/browser/workbench/workbench.ts
...
++					if (resolvedUri.path && resolvedUri.path !== '/') {
++						proxyUrl.pathname += resolvedUri.path.substring(1)
++					}
++					if (resolvedUri.query) {
++						proxyUrl.search = resolvedUri.query
++					}
++					resolvedUri = URI.parse(proxyUrl.toString())
 +				} else {
-+					throw new Error(`Failed to resolve external URI: ${uri.toString()}. Could not determine base url because productConfiguration missing.`)
++					throw new Error(Failed to resolve external URI: ${uri.toString()}. Could not determine base url because productConfiguration missing.)
```
code-asher: Looks like the backticks may have been accidentally dropped.  The argument here to `Error` should be a string.

code-asher (commented):
Thank you for fixing this!

1795771535y-cell:
@CodeRabbit Thank you for catching this! The backtick was accidentally dropped when I rewrote the error message. Fixed in the latest commit - the template string syntax is now correct: `throw new Error(Failed to resolve...) `

PR #7845, opened 2026-06-11
fix: re-enable skipped Heart tests and make non-asserting checks assert

voidmatcha:
This PR fixes test checks that cannot fail:

- **`it.only` committed in #7539** (`test/unit/node/heart.test.ts`): silently skipped the other 8 Heart tests in CI for the past 7 months. One of them stopped passing while disabled (its rejection handler runs on the microtask queue, after the assertion), re-enabled and repaired with a microtask drain.
- **`expect()` with no matcher** (`login.test.ts`, 4×): asserted nothing; the error-message tests passed whether or not the message rendered.
- **Dangling locator** (`extensions.test.ts`): `await page.getByText(...)` with no assertion.
- **One-shot `page.isVisible()` reads** (`downloads`/`uploads`/`github`/`codeServer`, 16×): [returns immediately with no auto-retry](https://playwright.dev/docs/api/class-page#page-is-visible); converted to web-first assertions:

```diff
-    expect(await codeServerPage.page.isVisible("text=Show Local")).toBe(false)
+    await expect(codeServerPage.page.locator("text=Show Local")).not.toBeVisible()
```

No assertions were removed and no behavior added. Every change strengthens an existing check.

**Verification:** `./ci/dev/test-unit.sh test/unit/node/heart.test.ts` - 9/9 pass; `tsc --noEmit` + repo ESLint clean on changed files. The Playwright e2e suite was not run locally (requires a full code-server build), the e2e changes are mechanical assertion-form conversions. Happy to iterate if CI surfaces anything.

---

Found while reviewing the test suite with [`e2e-skills/e2e-reviewer`](https://github.com/voidmatcha/e2e-skills/tree/main/skills/e2e-reviewer), which flags committed `.only` leaks, one-shot visibility reads, and matcher-less `expect()` calls.

### DIFF
--- a/test/e2e/codeServer.test.ts
+++ b/test/e2e/codeServer.test.ts
@@ -32,7 +32,7 @@ describe("code-server", ["--disable-workspace-trust"], {}, () => {
 
   test("should show the Integrated Terminal", async ({ codeServerPage }) => {
     await codeServerPage.focusTerminal()
-    expect(await codeServerPage.page.isVisible("#terminal")).toBe(true)
+    await expect(codeServerPage.page.locator("#terminal")).toBeVisible()
   })
 
   test("should open a file", async ({ codeServerPage }) => {
--- a/test/e2e/downloads.test.ts
+++ b/test/e2e/downloads.test.ts
@@ -18,7 +18,7 @@ describe("Downloads (enabled)", ["--disable-workspace-trust"], {}, async () => {
     // Action
     await codeServerPage.openContextMenu("text=unique-file.txt")
 
-    expect(await codeServerPage.page.isVisible("text=Download...")).toBe(true)
+    await expect(codeServerPage.page.locator("text=Download...")).toBeVisible()
   })
 
   test("should see the 'Show Local' button on Save As", async ({ codeServerPage }) => {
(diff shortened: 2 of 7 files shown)


code-asher (commented):
Thank you for catching these!!

~I think the `.not.toBeVisible()` ones will fail because `locator` will time out (since the element is not present) but otherwise it looks good to me.~ nvm, `locator` is the one that returns immediately, I was thinking of the `getBy` ones.

I feel like there has got to be a less janky way to wait for the microtask queue haha but I can look into that later.

**Edit**: hmm actually the ones that are visible are also failing.  Not really sure why.
**Edit 2**: oh just the login test ones are failing.

voidmatcha:
The password input is `required`, so submits with an empty field are blocked client-side and never POST. The failure videos show the native validation bubble at the moment of the assertion, so both login tests were never passable and the matcher-less `expect()` was hiding it.

image

- missing-password: assert the guard the user actually gets (`input.password:invalid`) instead of the unreachable server error
- rate limiter: refill the password before each attempt so all 15 actually POST (the re-render was clearing the required field), and `test.slow()` for the argon2 cost
- microtask drain: swapped for a single deterministic yield via `jest.requireActual("timers").setImmediate`

code-asher (approved):
Awesome thank you!

PR #7881, opened 2026-07-02
Backport installer workflow permissions hardening to v4.1.0

Edd88-pixel:
Backports the minimal GitHub Actions hardening from `main` to `v4.1.0` for `.github/workflows/installer.yml` only.

Related issue: https://github.com/coder/code-server/issues/7865
Reference workflow: https://github.com/coder/code-server/blob/v4.1.0/.github/workflows/installer.yml

Validation:
- `git diff --check`
- `actionlint` on `.github/workflows/installer.yml`
- `zizmor` on `.github/workflows/installer.yml`

Notes:
- The change is limited to `permissions: contents: read` at workflow level.
- The commit is signed.

### DIFF
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -19,6 +19,9 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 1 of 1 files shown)


code-asher:
Thank you for the PR!  If I recall correctly, I think the release workflow used to rely on the branch existing so I had to restore it after merging the PR but then I never deleted it again afterward.

I will go ahead and delete it now.

PR #7895, opened 2026-07-16
feat: hide debug view from activity bar

wdoxshzt:
Fixes #


code-asher:
Was this pull request opened on accident?  It does not look like something we would want to maintain here.  Closing for now but let me know if this was meant to fix some bug in code-server.

PR #7909, opened 2026-07-28
docs: document opening a file at a line via the payload query parameter

Grynn:
Relates to #1964.

## What

Adds an FAQ entry documenting how to open a specific file — optionally at a line and column — directly from a URL, using the `payload` query parameter (`openFile` + `gotoLineMode`) alongside the existing `folder` parameter.

## Why

#1964 ("Ability to open a file directly using a URL") has been open since 2020 and is in the backlog, but the capability already exists: code-server inherits upstream VS Code web's `payload` query parameter, so no new patch is needed — it just isn't documented anywhere. The FAQ currently only mentions `workspace` and `folder`.

Verified working on code-server 4.122.1 behind a reverse proxy:

```
https://<host>/?folder=/tmp&payload=%5B%5B%22gotoLineMode%22%2C%22true%22%5D%2C%5B%22openFile%22%2C%22vscode-remote%3A%2F%2F%3Chost%3E%2Ftmp%2Ftest.txt%3A6%3A3%22%5D%5D
```

opens `/tmp/test.txt` with the cursor at line 6, column 3.

The new entry includes a small shell snippet for generating such links from a CLI (e.g. an `code <file> --line N` style wrapper that emits a browser link), which is the use case that keeps coming up in #1964.

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -17,6 +17,7 @@
 - [Where is VS Code configuration stored?](#where-is-vs-code-configuration-stored)
 - [How can I reuse my VS Code configuration?](#how-can-i-reuse-my-vs-code-configuration)
 - [How does code-server decide what workspace or folder to open?](#how-does-code-server-decide-what-workspace-or-folder-to-open)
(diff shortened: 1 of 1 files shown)


code-asher (approved):
Very cool thank you!

PR #7914, opened 2026-07-30
Store state on remote instead of browser

### LINKED ISSUE #4212: Store state on remote instead of browser
**Edit by @code-asher:** Storing the state in the browser has (at least) the two following problems:

1.  If you switch to another browser you have to open and configure everything all over again (what this issue was originally about).
2. It is impossible to seed the initial state like you could in native VS Code by writing to the JSON state file.

Original issue content follows.

## OS/Web Information

- Web Browser: Chrome v91
- Local OS: Windows 10
- Remote OS: Ubuntu 20.04
- Remote Architecture: Debian
- `code-server --version`: 3.12.0

## Steps to Reproduce

1. Open a new code server instance in browser and make some changes (eg : hide remote host).
3. Open the same instance in incognito tab.
4. The changed made above will not be reflected, and you have to make those changes again.

## Expected
Till v3.11 the state of code-server was maintained across browsers. 

## Actual
The state is saved local storage instead of server and hence the changes does not reflect on browser changes.

## Notes
`scope === StorageScope.GLOBAL ? 'global' : this.payload.id;` it is in workbench.web.api.js.map
Perhaps it's related to where the state is saved?
From discussion : https://github.com/cdr/code-server/discussions/4185

Daywin67:
Fixes #4212

## Problem

Workbench state was stored in browser local storage, so it was not shared between browsers or incognito sessions.

## Motivation

The goal is to provide the same persistent development experience as using VS Code on a personal computer, with an identical workspace available from anywhere and from any browser.

Persisting the workbench state on the remote server is important to achieve that experience: settings, authenticated accounts, and other state should follow the remote workspace rather than the local browser.

## Solution

Store the relevant state on the remote, restoring the behavior expected by code-server users.

Based on @hefanbo fix proposition. Thanks to him. [Link](https://github.com/hefanbo/code-server/tree/feat/remote-storage)

## Test

Testing steps: 
- Flush code-server configuration
- Open blank vscode server on a first browser. Connect Github Account.
- Close this browser
- Open a second vsocde on another computer (or just incognito session)
- Enjoy the Github account already connected 

_The same behavior also works with a BYOM provider API key configured in GitHub Copilot._

Tested & deployed on my personal vscode-server self hosted instance.

## AI assistance and review

This change was developed with the assistance of an AI coding assistant. I reviewed and tested the implementation on my self-hosted code-server instance running on a NAS.

However, I cannot fully assess all possible side effects or interactions with other code-server features. Community review, testing, and criticism are very welcome, especially regarding compatibility, **security**, and backward compatibility.


code-asher:
Thank you for the PR!

Sorry I have not reviewed this, it is fairly large and I think it will take me a while to get all the context to properly review it so I am not sure when I can get to it.

Is there a way to instead use, say, the Electron version of the storage service?  It might be as simple as providing the `storage` channel (similar to what we did for display languages).

This change might be the correct way to do it in VS Code core but here we want to patch as minimally as possible to lower the maintenance burden and avoid drift (with new classes, every change to the storage service will likely require us to port it over, assuming we even catch those changes).

NevermoreN:
非常需要！

Daywin67:
Hello,

Thank you for the feedback @code-asher. I agree that keeping the patch as small as possible is important to reduce maintenance and avoid drifting from VS Code core.

I’ve revised the implementation to reuse VS Code’s existing Electron storage service instead of introducing new storage-service classes. The remote server exposes the storage service through a channel, similar to the existing display-language approach, and the browser-side service uses that channel for persistence.

I also agree with @nevermoreN that this functionality is indispensable. In my view, without persistent cross-device sync, code-server loses a significant part of its value, particularly for users accessing the same environment from multiple browsers or devices.

The GitHub build is passing with no issues on my fork (https://github.com/Daywin67/code-server/actions/runs/34258247556), and the updated code is currently being tested on my NAS.

What do you think of this new version?

patches/remote-secret-storage.diff
```diff
@@ -0,0 +1,86 @@
+AI: store secrets on remote for VS Code Web
```
code-asher: What is this `AI:` prefix for?

patches/remote-secret-storage.diff
```diff
@@ -0,0 +1,86 @@
...
+===================================================================
+--- code-server.orig/lib/vscode/src/vs/platform/secrets/common/secrets.ts
++++ code-server/lib/vscode/src/vs/platform/secrets/common/secrets.ts
+@@ -139,8 +139,13 @@ export class BaseSecretStorageService ex
+ 				return await readEncryptedSecret(
+ 					key,
+					(fullKey) => this.getValueFromStorage(key, fullKey, storageService),
+-					// If the storage service is in-memory, we don't need to decrypt
+-					this._type === 'in-memory' ? (v) => Promise.resolve(v) : (v) => this._encryptionService.decrypt(v),
++					// Don't decrypt if storage is in-memory or if encryption service is not available
```
code-asher: Instead of bypassing encryption, could we add the `encryption` channel?  I think that might make it just work.

code-asher:
Awesome thank you!  I will take a look as soon as I am able.

PR #7940, opened 2026-08-10
feat: add --socket-fd CLI option

### LINKED ISSUE #7939: Support socket fd inheritance for systemd socket activation
## What is your suggestion?

Implement a CLI argument to specify a file descriptor to bind to instead of directly binding to a inet or unix socket.

systemd allows services to be activated by sockets managed by itself keeping resource usage as none until needed. For this to work properly, the service has to accept a file descriptor specified by `LISTEN_FDS` environment variable (but predictibly starting at 3) instead of creating it's own listen socket.

[`systemd-socket-proxyd`](https://www.freedesktop.org/software/systemd/man/latest/systemd-socket-proxyd.html) should work but it creates a requirement of 2 new systemd units and an extra tcp port. Implementing a socket-fd parameter makes this a lot simpler.

yes.

runleveldev:
Fixes #7939

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -41,6 +41,7 @@
 - [How do I disable the proxy?](#how-do-i-disable-the-proxy)
 - [How do I disable file download?](#how-do-i-disable-file-download)
 - [Why do web views not work?](#why-do-web-views-not-work)
+- [Can I run code-server with systemd socket activation?](#can-i-run-code-server-with-systemd-socket-activation)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 <!-- prettier-ignore-end -->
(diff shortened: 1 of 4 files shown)


code-asher (approved):
Very cool thank you!

code-asher (commented):
Actually, had one thought.  Would it make sense to combine the flags?  They seem so similar apart from the unlink behavior.

What do you think about reusing `--socket`, but if it is a number (`parseInt()` returns non-NaN) then we assume it is an fd, and we skip the unlink and mode change?  Or does that overcomplicate the flag without any real benefit?

Do you know of any precedent from other existing applications?

runleveldev:
> Actually, had one thought. Would it make sense to combine the flags? They seem so similar apart from the unlink behavior.
> 
> What do you think about reusing `--socket`, but if it is a number (`parseInt()` returns non-NaN) then we assume it is an fd, and we skip the unlink and mode change? Or does that overcomplicate the flag without any real benefit?
> 
> Do you know of any precedent from other existing applications?

I was originally referencing `dockerd` but it uses a single "host" flag with schemes i.e. tcp://127.0.0.1:1234, unix:///path/to/sock, or fd://3. Obviously that would have been a more-jarring rewrite here. I did toy with combining this with the existing socket option, but I prefer to keep them explicit to avoid shadowing unlikely, but technically valid, paths (like `3` which currently would make a socket file in `$PWD/3`).

code-asher:
Fair point, yeah.  Thanks!

PR #7953, opened 2026-08-18
fix(npm): remove --unsafe-perm flag from install scripts

### LINKED ISSUE #7950: npm 12: installer passes removed --unsafe-perm flag, install fails with EUNKNOWNCONFIG
## Bug: `npm install -g code-server` fails on npm 12 with "Unknown cli flag: --unsafe-perm"

code-server's installer passes `--unsafe-perm` to npm, but **npm 12 removed that flag** and now errors on any unknown flag instead of ignoring it. Result: the install dies before it even starts.

## How to reproduce (2 commands, ~45 seconds)

```bash
npm --version   # must be 12.x (Node 22+ ships npm 11+; install npm@12 first if needed)
npm install -g code-server@4.117.0 --unsafe-perm
```

Expected: code-server installs.
Actual:

```text
npm error code EUNKNOWNCONFIG
npm error Unknown cli flag:
npm error   - --unsafe-perm
npm error Run `npm help config` for supported options.
```

That's it. No code-server source needed — the failure is npm rejecting the flag that `install.sh` passes.

## Where the flag lives

Three active places pass or require `--unsafe-perm`:

- `install.sh` line 439 — `npm install ... --unsafe-perm`
- `ci/build/npm-postinstall.sh` lines 79–92 — root-only `npm_config_unsafe_perm` guard
- `ci/build/npm-postinstall.sh` lines 111–116 — nested `npm install --unsafe-perm`

## Why it's safe to remove

npm stopped needing `--unsafe-perm` in npm 7 — when npm runs as root, lifecycle scripts already run with the effective UID/GID of the package directory owner. The flag has been a no-op for years; npm 12 just finally made passing it an error. See npm's changelog: https://docs.npmjs.com/cli/v7/using-npm/changelog/#root-user-does-not-drop-privileges

## Environment

- Node.js 22.22.2, npm 12.0.2
- code-server 4.117.0 (install fails before running)
- Debian 13

## Bonus caveat (read before saying "just remove it")

With `--unsafe-perm` gone, npm 12's **new lifecycle-script policy** also blocks code-server's postinstall by default:

```text
npm warn install-scripts 2 packages had install scripts blocked...
npm warn install-scripts   code-server@4.117.0 (postinstall: sh ./postinstall.sh)
npm warn install-scripts   argon2@0.44.0 (install: ...)
```

So removing the flag fixes the crash, but full npm 12 support also needs the `--allow-scripts=code-server,argon2` (or equivalent) handling. I have a test-backed patch for the `--unsafe-perm` part (36/36 script tests pass) — happy to open a PR if this is accepted.

aholten:
## Summary
Fixes #7950. Removes the --unsafe-perm flag from the three sites that pass or require it: npm 12 rejects unknown CLI flags (EUNKNOWNCONFIG) so the install dies before starting. The flag has been behaviorally irrelevant since npm v7.0.0 (2020): the changelog says the user/group/uid/gid/unsafe-perms configs are 'no longer relevant.' Scripts run as the working-dir owner regardless. 

## Changes
2 files: install.sh (drop flag from global install) + ci/build/npm-postinstall.sh (remove root guard that errors unless unsafe_perm=true, drop flag from dev-install). Validation: grep zero unsafe-perm refs, bash -n syntax OK, npm run test:scripts 34/34 pass. 

## Repro
 before: npm 12 + install -g code-server --unsafe-perm -> EUNKNOWNCONFIG. Follow-up tracked separately: npm 12 blocks postinstall scripts unless --allow-scripts is configured.

### DIFF
--- a/ci/build/npm-postinstall.sh
+++ b/ci/build/npm-postinstall.sh
@@ -76,20 +76,6 @@ main() {
     exit 1
   fi
 
-  # Under npm, if we are running as root, we need --unsafe-perm otherwise
-  # post-install scripts will not have sufficient permissions to do their thing.
-  if is_root; then
-    case "${npm_config_user_agent-}" in npm*)
-      if [ "${npm_config_unsafe_perm-}" != "true" ]; then
-        echo "Please pass --unsafe-perm to npm to install code-server"
-        echo "Otherwise post-install scripts will not have permissions to run"
-        echo "See https://docs.npmjs.com/misc/config#unsafe-perm"
-        echo "See https://stackoverflow.com/questions/49084929/npm-sudo-global-installation-unsafe-perm"
-        exit 1
-      fi
-      ;;
-    esac
-  fi
 
   if ! vscode_install; then
     echo "You may not have the required dependencies to build the native modules."
@@ -110,7 +96,7 @@ install_with_yarn_or_npm() {
   # end-user we want to keep using whatever package manager is in use.
   case "${npm_config_user_agent-}" in
     npm*)
-      if ! npm install --unsafe-perm --omit=dev; then
+      if ! npm install --omit=dev; then
         return 1
       fi
       ;;
--- a/install.sh
+++ b/install.sh
@@ -436,7 +436,7 @@ install_npm() {
     fi
     echoh "Installing with npm."
     echoh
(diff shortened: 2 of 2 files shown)


code-asher (approved):
Fabulous, thank you!

PR #7678, opened 2026-02-24
feat: expose --reconnection-grace-time CLI flag

tianyicui:
## Summary

- Expose VS Code Server's `--reconnection-grace-time` CLI argument through code-server, allowing users to configure how long the server waits for a disconnected client to reconnect before cleaning up the session
- Add support for `$CS_RECONNECTION_GRACE_TIME` environment variable and `config.yaml` setting
- The underlying VS Code Server (1.107+) already implements this flag in `serverEnvironmentService.ts`; code-server just wasn't passing it through

## Motivation

When a client machine sleeps (e.g., overnight), the WebSocket connection drops. VS Code Server's default 3-hour reconnection grace time expires, the server discards the reconnection token, and the user is forced to "Reload Window" on wake. Users who want longer grace periods currently have no way to configure this.

Fixes #7665

## Changes

- **`src/node/cli.ts`**: Added `reconnection-grace-time` to `UserProvidedCodeArgs` interface, `options` object, and `setDefaults()` (env var support via `$CS_RECONNECTION_GRACE_TIME`)
- **`test/unit/node/cli.test.ts`**: Added test for the new flag in the "parse all options" test and a dedicated env var test

## Usage

```bash
# CLI flag (value in seconds, e.g., 30 days)
code-server --reconnection-grace-time 2592000

# Environment variable
CS_RECONNECTION_GRACE_TIME=2592000 code-server

# config.yaml
reconnection-grace-time: 2592000
```

## Test plan

- [x] Existing CLI parser tests pass (68/68)
- [x] New test: `--reconnection-grace-time` is parsed correctly from CLI
- [x] New test: `$CS_RECONNECTION_GRACE_TIME` env var is picked up in `setDefaults()`
- [x] Manual: verify the flag is passed through to VS Code Server and sessions survive long disconnects

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7956, opened 2026-08-19
fix: honour --reconnection-grace-time when the browser closes

### LINKED ISSUE #7955: --reconnection-grace-time does not survive closing the browser tab
## Problem

`--reconnection-grace-time` (added in #7678) is documented as *"Override the reconnection grace time in seconds. Clients who disconnect for longer than this duration will need to reload the window."*

In practice the setting cannot deliver that, for two independent reasons.

**1. Closing the tab bypasses the grace time entirely.**

`BrowserLifecycleService` listens on `pagehide`/`beforeunload` and runs the workbench shutdown, which disposes the remote connection *gracefully*. On the server, `ManagementConnection` treats a graceful dispose as a finished client:

```
Event.once(this.protocol.onDidDispose)(() => {
    this._log(`The client has disconnected gracefully, so the connection will be disposed.`);
    this._cleanResources();
});
```

`_cleanResources()` runs immediately — the grace time is never consulted. So closing the tab kills the session regardless of the configured value.

**2. A second connection cuts the grace time to 5 minutes.**

`RemoteExtensionHostAgentServer` shortens every disconnected connection on each new connection:

```
for (const key in this._managementConnections) {
    this._managementConnections[key].shortenReconnectionGraceTimeIfNecessary();
}
```

which schedules `ProtocolConstants.ReconnectionShortGraceTime` (5 min). Opening a second tab is enough to cut a deliberately long grace time back down.

## Impact

The use case the flag exists for — leave a long-running task (a build, a CLI, an agent session) running and come back to it later — does not work. The session is gone as soon as the browser is closed.

## Expected

When an operator raises `--reconnection-grace-time` above the default, a disconnected session should survive for the configured duration, including across a closed browser. Installations that never set the flag should keep Code's stock behaviour.

## Environment

Reproduced on code-server 4.128.0 and confirmed present in `main` (4.133.0 / Code 1.133.0).

hugodeco:
Fixes #7955

## Summary

`--reconnection-grace-time` (#7678) cannot currently deliver what it documents, for two independent reasons:

1. **Closing the tab bypasses the grace time.** `BrowserLifecycleService` runs the workbench shutdown on `pagehide`/`beforeunload`, disposing the remote connection *gracefully*. `ManagementConnection` treats a graceful dispose as a finished client and calls `_cleanResources()` immediately — the grace time is never consulted, so the session dies with the tab however the flag is set.

2. **A second connection cuts it to 5 minutes.** Every new connection calls `shortenReconnectionGraceTimeIfNecessary()` on all disconnected connections, scheduling `ProtocolConstants.ReconnectionShortGraceTime`. Opening a second tab is enough to shorten a deliberately long grace time.

This adds `patches/session-preservation.diff`, which makes the configured grace time authoritative.

## Changes

`lib/vscode/src/vs/workbench/services/lifecycle/browser/lifecycleService.ts`

- `pagehide` persists UI state (`storageService.flush(WillSaveStateReason.SHUTDOWN)`) instead of unloading.
- A browser-driven unload no longer tears the workbench down. The veto handler is only passed on the browser path, so that is the discriminator: a `beforeunload` veto is still honoured, and an explicit in-product shutdown (`shutdown()`, which passes no veto handler) still unloads normally.

`lib/vscode/src/vs/server/node/remoteExtensionHostAgentServer.ts`

- The grace time is only shortened when it was left at or below `ProtocolConstants.ReconnectionGraceTime`. **Installations that never set the flag keep Code's stock behaviour**; only an operator who deliberately raised it opts into the longer wait.

Also updates the unreleased changelog section.

## Test plan

- [x] `quilt pop -a && quilt push -a` — full series applies cleanly with the new patch
- [x] `npm run typecheck-client` in `lib/vscode` — passes
- [x] `npm run lint:ts` — passes
- [x] `npm run test:unit` — 331/332; the one failure (`should unlink a socket before listening on the socket`, expecting `EACCES`) is environmental, from running as root, and exercises `src/node/app.ts`, which this PR does not touch
- [x] `npm run valid-layers-check` fails identically with and without this patch (pre-existing, in `network.ts` / `extensionResourceLoader.ts`)
- [x] Behaviour verified in production on 4.128.0: with `--reconnection-grace-time 2592000`, sessions survive closing the browser and reconnect on return

I was not able to run a full product build on the machine I had available (disk), so a CI build is the remaining check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,17 @@ Code v99.99.999
 
 ## Unreleased
 
+### Fixed
+
+- `--reconnection-grace-time` is now honoured when the browser goes away.
+  Closing the tab used to dispose the connection gracefully, which the server
+  treats as a finished client and cleans up at once, so the grace time was never
+  consulted; and any new connection shortened every disconnected session to the
+  5-minute short grace, so opening a second tab cut a deliberately long grace
+  time back down. Sessions now survive a closed browser for as long as the
+  configured grace time. Installations that never set the flag keep the previous
+  behaviour.
+
 ## [4.133.0](https://github.com/coder/code-server/releases/tag/v4.133.0) - 2026-08-17
 
 Code v1.133.0
--- a/patches/series
+++ b/patches/series
@@ -25,3 +25,4 @@ signature-verification.diff
 copilot.diff
 app-name.diff
 csp-hashes.diff
+session-preservation.diff
--- a/patches/session-preservation.diff
+++ b/patches/session-preservation.diff
@@ -1,0 +1,113 @@
+Preserve the remote session when the browser goes away.
+
+--reconnection-grace-time lets an operator say how long a disconnected session
+should be kept. Two behaviours inherited from Code make that setting unable to
+deliver on its promise:
+
+1. Closing the tab runs the browser workbench's shutdown, which disposes the
+   remote connection *gracefully*. The server reads a graceful dispose as "the
+   client is finished" and cleans up immediately, so the grace time is never
+   consulted at all -- the session dies with the tab no matter how the flag is
+   set.
+
+2. Any new connection shortens every disconnected session's grace time to
+   ProtocolConstants.ReconnectionShortGraceTime (5 minutes). Opening a second
+   tab is enough to cut a deliberately long grace time down to five minutes.
+
+This patch makes the configured grace time authoritative:
+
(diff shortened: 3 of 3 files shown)


code-asher:
Thank you for the PR!

This feels like something we should commit directly to VS Code rather than try to maintain as a patch here.  Could we try submitting it upstream?

PR #7983, opened 2026-09-04
fix: guard FORCE_NODE_VERSION in the version error path

denusklo:
Installing on an unsupported node version dies with `FORCE_NODE_VERSION: unbound variable` instead of the intended error message. npm-postinstall.sh runs under `set -u`, and this is the one place the variable is read bare; the other reads all default it with `${FORCE_NODE_VERSION:-}`.

The lines the script never reaches are exactly the ones that tell the user the override exists and how to set it, so the people who need the message are the only ones who cannot see it.

Verified on Windows 11 with node 22 against 4.135.0: before, the install ends at the unbound variable; after, it prints the version error plus the override instructions and exits 1 as intended. Related to the install paths discussed in #1397.


code-asher (approved):
Good find, thank you!

Issue #1537, opened 2020-04-17
readlink error running code-server macOS


nhooyr:
```
/tmp/code-server-3.1.1-darwin-x86_64$ ./code-server
readlink: illegal option -- f
usage: readlink [-n] [file ...]
info  code-server 3.1.1 28e91ba70cd70fa9adf3f2e3e3b87631b5667ecf
info  HTTP server listening on http://127.0.0.1:8080
info    - Password is 4afa6b3d9cb0df0935c1540d
info      - To use your own password set the PASSWORD environment variable
info      - To disable use `--auth none`
info    - Not serving HTTPS
info  Automatic updates are enabled
info  SSH server listening on localhost:62276
info    - To disable use `--disable-ssh`
```

nhooyr:
Fixed in #1546

PR #7984, opened 2026-09-05
Add a windows launcher to the standalone release

denusklo:
The standalone release ships `bin/code-server`, a shell script that resolves its own location and execs the bundled node against the release root. Windows cannot run it, so a windows release arrives with no way to start it short of knowing where node ended up and typing both paths.

This adds the same launcher as a `.cmd`, copied into `bin/` only when `OS` is windows, so no other release changes at all. It is deliberately the shortest thing that works: `%~dp0` gives the bin directory, the root is one above it, and the bundled node is handed the root and every argument.

Verified against a real windows release tree: `--version` answers, it works from any working directory, and a bad flag comes back as exit code 1.

Two things a reviewer should know. The shell launcher resolves symlinks before computing the root, because of #1537. The `.cmd` does not, since `%~dp0` reports the directory of the link rather than of the target and batch has no readlink. A copy of the file works anywhere; a symlink to it does not.

The file is checked in with CRLF, which is what a `.cmd` needs on the platform it runs on and what a windows checkout would produce for it anyway.

### DIFF
--- a/ci/build/build-release.sh
+++ b/ci/build/build-release.sh
@@ -41,6 +41,11 @@ main() {
     rsync ./ci/build/code-server.sh "$RELEASE_PATH/bin/code-server"
     chmod 755 "$RELEASE_PATH/bin/code-server"
 
+    # Windows cannot run the shell launcher, so it gets one of its own.
+    if [ "$OS" = windows ]; then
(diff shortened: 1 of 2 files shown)


code-asher (approved):
Makes sense!

PR #7985, opened 2026-09-05
Default the session socket to a named pipe on windows

denusklo:
The session socket defaults to a path inside the user data directory, which windows cannot bind: it has no Unix sockets. The bind failure is caught and warned about, and the server is returned anyway, so startup prints

```
warn  Could not create socket at ...\code-server-ipc.sock
info  Session server listening on ...\code-server-ipc.sock
```

about a socket that does not exist, and nothing listens on it. That socket is what `shouldOpenInExistingInstance` consults, so on windows `code-server <file>` run from a shell can never find a running instance and starts its own server instead.

Windows has the same facility under a different name, so this defaults to a named pipe there. The name is derived from the user data directory, so instances with separate data directories do not collide and a later invocation with the same one finds the first. It is case-folded before hashing because windows compares paths that way, and the same directory typed two ways should not produce two pipes. An explicitly passed `--session-socket` is untouched on every platform.

Nothing downstream needed changing: the paths are only ever handed to `net.connect` and `http.request`, both of which take a pipe name on windows, and `listen()`'s unlink-first only has to tolerate ENOENT, which it already does.

Measured against a real windows build, one flag apart:

| invocation | result |
| --- | --- |
| `--session-socket '\\.\pipe\...'` | no warning; "Session server listening on" that pipe; `canConnect` true; `GET /session` answers 200 |
| default | the warning above, the phantom "listening" line, and no such file on disk |

The helper takes the platform as a parameter rather than reading `process.platform` directly, following `getEnvPaths` and `constructOpenOptions`, so both branches are tested where the tests run.

### DIFF
--- a/src/node/cli.ts
+++ b/src/node/cli.ts
@@ -1,4 +1,5 @@
 import { field, Level, logger } from "@coder/logger"
+import * as crypto from "crypto"
 import { promises as fs } from "fs"
 import { load } from "js-yaml"
 import * as path from "path"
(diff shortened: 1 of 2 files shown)


code-asher (approved):
Looks good to me, thank you!

PR #8012, opened 2026-09-17
fix: preserve path, query, and fragment in proxy URLs (#7668)

### LINKED ISSUE #7573: URL translation via VSCODE_PROXY_URI drops query parameters

### OS/Web Information

- Web Browser: Chrome
- Local OS: MacOS
- Remote OS: Ubuntu
- Remote Architecture: amd64
- `code-server --version`: 1.102.2

### Steps to Reproduce

1. Run code-server with a configured `VSCODE_PROXY_URI`
2. Attempt to redirect from a local URL to the proxy, where the URL contains a query parameter.
    (Example: http://127.0.0.1:8888/lab?token=2212f0697bfed279e9a6fa898291a0f44f313b3510965815)
3. Observe that code-server will correctly translate the base url to `8888-{my-proxy}/lab`, but that the query param `token=...` is not passed along. Navigating to the correct URL directly (`8888-{my-proxy}/lab?token=...`) works correctly.

The example I'm using here is based on attempting to launch jupyterlab via a terminal in code-server, 
but this bug is not specific to jupyterlab, the problem presents itself for any URL that contains a query parameter part.

### Expected

The query parameters are copied over to the proxied URL, and translation works seamlessly.

### Actual

The query parameters are dropped from the proxied URL, and need to be manually supplied in order to provide the correct data to the server.

### Logs

```shell

```

### Screenshot/Video

_No response_

This cannot be tested in native VS Code

I did not test GitHub Codespaces

- [x] I am using a secure context.

### Notes

I suspect that [this part](https://github.com/coder/code-server/blob/v4.106.2/patches/proxy-uri.diff#L59-L62) of the proxy_uri patch does not function correctly, but I know too little of the workings of vscode to be sure.

### LINKED ISSUE #7668: localhost openExternal rewrite loses URL /path

yykaue:
Fixes #7668
Fixes #7573

Localhost URLs opened through `browser.sh` lose their original path when rewritten through the port proxy. Query parameters and fragments are also discarded.

This change appends the original path to the proxy base and preserves the query and fragment using URI components. It supports proxy templates with or without a trailing slash and avoids double encoding.

For example:
- Before: `http://127.0.0.1:1234/my/path?q=1#section` → `/proxy/1234/`
- After: `http://127.0.0.1:1234/my/path?q=1#section` → `/proxy/1234/my/path?q=1#section`

Added end-to-end regression coverage for default, relative, and subdomain proxy templates, including encoded characters.

Validation:
- All 13 isolated reproduction cases passed.
- Patch application, Prettier, ESLint, and changed-file TypeScript checks passed.
- Full test-project TypeScript diagnostics are unchanged from `main`.
- Full end-to-end tests were not run because a VS Code build was unavailable.


code-asher (approved):
Awesome thank you!

PR #8015, opened 2026-09-21
fix(webview): load Markdown custom styles in Firefox (#7913)

### LINKED ISSUE #7913: [Firefox] Markdown Preview fails to load workspace CSS via markdown.styles (NS_ERROR_UNKNOWN_HOST for vscode-resource.vscode-cdn.net)

### OS/Web Information

* Web Browser: Firefox
  * `firefox-151.0-2.fc42.x86_64`
* Local OS: Fedora 42
* Remote OS: Fedora 42
* Remote Architecture: amd64
* `code-server --version`: 4.105.1 811ec6c1d60add2eb92446161ca812828fdbaa7f with Code 1.105.1

### Bug Description

When opening a Markdown preview in Firefox, local CSS files referenced in `markdown.styles` fail to load. The browser attempts to resolve the virtual `vscode-resource.vscode-cdn.net` domain via DNS instead of intercepting it via the Service Worker, resulting in `NS_ERROR_UNKNOWN_HOST`.

This issue is specific to Firefox and its implementation of Service Worker / iframe isolation (dFPI).

Looks close to coder/code-server#7892 … BUT!!! The same setup works for me perfectly in Chromium-based browsers.  <br>So I hope some workaround (custom settings, hacks) exists that can make work it with Firefox.

### Steps to Reproduce

1. Run `code-server` and access it via Firefox.
2. In a workspace, create a file `test.css` in the root directory.
3. Add the following to `.vscode/settings.json`:

   ```json
   {
       "markdown.styles": ["test.css"]
   }
   ```
4. Open any `.md` file and open the Markdown Preview.
5. Check the developer tools Network and Console panels.

### Expected

The Service Worker intercepts the request for `test.css` and serves the local workspace file, applying the custom styles to the preview (as it does in Chromium).

### Actual

The preview renders without the custom CSS. In the Developer Tools Console:

```text
Could not load 'markdown.styles': test.css
```

In the Network panel, the request for `test.css` fails:

* **Host**: `vscode-remote+<punycode>.vscode-resource.vscode-cdn.net`
* **Error**: `NS_ERROR_UNKNOWN_HOST`
* **DNS Resolution**: System

### Investigation Details

Based on HAR exports and console logs:

1. The Service Worker **is registered and active**.
2. Extension resources inside the webview (e.g., `index.bundle.js` from extensions) are successfully intercepted and return `HTTP/2 200`.
3. However, the dynamically injected `<link rel="stylesheet">` for `markdown.styles` bypasses the Service Worker and goes directly to the network.
4. Setting `privacy.partition.serviceWorkers` to `false` in `about:config` completely breaks `code-server` initialization (`Error: IndexedDB database 'vscode-web-db' is closed`), so it cannot be used as a workaround.

It appears to be a race condition or a cross-origin iframe isolation bug specific to Firefox, where the Service Worker fails to control the CSS fetch request initiated by the Markdown preview webview.

### Attachments

Please find the attached `console.log` and `.har` (`har.zip`) network export demonstrating the bypassed Service Worker.

[console-export-2026-7-30_12-31-17.log](<https://github.com/user-attachments/files/30540719/console-export-2026-7-30_12-31-17.log>)

[xn--80ach2ae.xn--80apqgfe.xn--p1ai_Archive [26-07-30 12-32-00].har.zip](<https://github.com/user-attachments/files/30540774/xn--80ach2ae.xn--80apqgfe.xn--p1ai_Archive.26-07-30.12-32-00.har.zip>)

### Logs

```shell
```

### Screenshot/Video

*No response*

Yes, this is also broken in VS Code web

### Notes

*No response*

yykaue:
## Summary

- Fetch local webview stylesheets through the service worker in Firefox 149 and later.
- Apply the fetched CSS as constructed stylesheets while preserving the original base URL, media, and disabled state.
- Add regression coverage for custom Markdown preview styles.
- Update the changelog.

## Validation

- `npm run build`
- `npm run gulp compile-web`
- `npm run compile-client`
- Manually verified with Firefox 151 that `markdown.styles` is applied correctly.

The targeted Playwright E2E test is currently blocked before the workbench loads by an unrelated CSS module MIME-type issue in the development build.

Fixes #7913


code-asher:
Thank you for the PR!

This looks like something we ought to contribute to VS Code itself rather than maintaining as a patch here.  Could we send this up to https://github.com/microsoft/vscode?

yykaue:
> Thank you for the PR!
> 
> This looks like something we ought to contribute to VS Code itself rather than maintaining as a patch here. Could we send this up to https://github.com/microsoft/vscode?

Thanks for the feedback! I completely agree — this is better suited to VS Code itself rather than being maintained as a patch here. I'm happy to send it upstream to https://github.com/microsoft/vscode.

I'll open a PR/issue there and link back to this discussion. Do you want me to close this PR in the meantime, or keep it open until the upstream change is merged?

code-asher:
Awesome thank you!  I think we go ahead and close for now, and we can reopen if there ends up being something code-server-specific we have to fix.

